PR #1, opened 2020-07-28
Fix the Missing a Javadoc comment problem during compilation

OpenOpened:
fix problem during compilation :

`Error: java: JavadocType: Missing a Javadoc comment.`

The solution is to add comment to the class


wuchong:
Thanks. I have addressed this.

Issue #14, opened 2020-08-13
No entry to modify the DBZ properties


NextMark:
Mysql cdc connector need add the entry to modify the DBZ properties.

wuchong:
Yes, I think this is a reasonable feature. Passing through debezium's properties will be very helpful. For example, we can support 'debezium.*' to pass throught the properties.

wuchong:
Fixed in c89b46f14a26ed9f00c0c357c3c12420e9ae68f0

PR #15, opened 2020-08-14
Fix add entry to modify DBZ properties.

NextMark:


flink-connector-mysql-cdc/src/main/java/com/alibaba/ververica/cdc/connectors/mysql/table/MySQLTableSourceFactory.java
```diff
@@ -75,6 +76,11 @@
 			"MySQL database cluster as another server (with this unique ID) so it can read the binlog. " +
 			"By default, a random number is generated between 5400 and 6400, though we recommend setting an explicit value.");
 
+	private static final ConfigOption<Map<String, String>> PROPERTIES = ConfigOptions.key("properties")
+		.mapType()
+		.noDefaultValue()
+		.withDescription("The Debezium MySQL connector properties.");
+
```
wuchong, line 83: Currently, `ConfigOption` doesn't support Map type, and compress all the properties in one entry is not user-friendly. Could you take the implementation of Kafka `properties.*` as an example? see https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaOptions.java#L292

NextMark: OK, i will modify it, and add the support for PG CDC .

wuchong:
Thanks for the updating. I will modify and merge it.

This fixes #14

Issue #10, opened 2020-08-12
Remove the Unsafe.monitorEnter usage in code to support JDK11


wuchong:
Currently, it will throw the following exception if using JDK11

```
Caused by: java.lang.NoSuchMethodError: sun.misc.Unsafe.monitorEnter(Ljava/lang/Object;)V
    at com.alibaba.ververica.cdc.debezium.internal.DebeziumChangeConsumer.handleBatch(DebeziumChangeConsumer.java:101)
    at io.debezium.embedded.ConvertingEngineBuilder.lambda$notifying$2(ConvertingEngineBuilder.java:81)
    at io.debezium.embedded.EmbeddedEngine.run(EmbeddedEngine.java:810)
    at io.debezium.embedded.ConvertingEngineBuilder$2.run(ConvertingEngineBuilder.java:170)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
    at java.base/java.lang.Thread.run(Thread.java:834)
```

The JDK version
```
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)
```

The reason is that we are using `Unsafe.monitorEnter` to manually hold the checkpoint lock during database snapshot phase. If we want to remove the `Unsafe.monitorEnter`, an approach is that, we have to using `synchronize` keyword to hold the checkpoint lock in `SourceFunction.run` method.

balint133:
Hey @wuchong I really like the library, removing lots of moving parts. 
I took the initiative and opened a PR removing the UNSAFE calls to make the library compatible with Java 11. It'd be nice if you could give some feedback and hopefully get it merged.

wuchong:
Hi @balint133 ! Glad to hear that you are inteseted on this issue. The introducing of `Unsafe.monitorEnte` is that we want to dynamically hold and release the checkpoint lock for database snapshot phase. The reason is that we can't perform checkpoint when scanning database table as we don't have an offset.

wuchong:
[...]

PR #19, opened 2020-08-20
#10 Remove the use of UNSAFE.monitorEnter() and monitorExit() calls

balint133:


balint133:
Closing due to being incomplete. Change only works if Snapshot and binlog boundaries are within the same batch, produces inconsistent results in case snapshot spans across multiple batches.

PR #70, opened 2020-12-18
[ISSUE-69] DebeziumSourceFunction only stores monitored table ddls wh…

docete:
…ich cause error in some situation

The root cause is: DebeziumSourceFunction only stores monitored table ddls. If there are some renaming operations, debezium can't rebuild table schema from the monitored table ddls. This PR fix this issue. See [issue-69](https://github.com/ververica/flink-cdc-connectors/issues/69)


wuchong (changes requested):
The `MySQLSourceTest#testCheckpointAndRestore#assertHistoryState` is failed because we store all DDLs now.

docete:
> The `MySQLSourceTest#testCheckpointAndRestore#assertHistoryState` is failed because we store all DDLs now.

fixed!

wuchong (approved):
LGTM.

PR #85, opened 2021-01-14
[fix bug] offset state missing if no binlog fetched tile checkpoint after restore.

dongdongking008:


wuchong (approved):
Thanks for the contribution @dongdongking008 . LGTM.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -222,7 +222,15 @@ private void snapshotOffsetState() throws Exception {
 				serializedOffset = restoredOffsetState.getBytes(StandardCharsets.UTF_8);
 			}
 		} else {
-			serializedOffset = consumer.snapshotCurrentState();
+			byte[] currentState = consumer.snapshotCurrentState();
+			if (currentState == null) {
+				// the consumer has been initialized, but has not yet received any data,
+				// which means we need to return the originally restored offsets
+				serializedOffset = restoredOffsetState.getBytes(StandardCharsets.UTF_8);
```
MichaelXucf, line 229: An NP will be thrown here when the Application started for the first time and restoredOffsetState is null .

dongdongking008: Hi MichaelXucf,  could you please give a test case?
In my opinion, the case you mentioned has been covered by test cases `testCheckpointAndRestore`(Step-1) both in MySQLSourceTest and PostgreSQLSourceTest. 
When the Application started for the first time and restoredOffsetState is null, debezium will mark current offset and take a snapshot. In this case, currentState won't be null.

MichaelXucf: when try to capture a table with no data in it, it will throw the Exception.  The test case ```testCheckpointAndRestore(Step-1)``` doesn't cover this case.
You can try it with the followiing code
```java
public class MyTest {

    @Test
    public void testSource() throws Exception {
        DebeziumSourceFunction<String> sourceFunction = MySQLSource.<String>builder()
                .hostname("localhost")
                .port(3306)
                .databaseList("inventory")
                .username("root")
                .password("123456")
                .tableList(new String[]{"inventory.test"})
                .build();

        env.enableCheckpointing(60000);
        env.getCheckpointConfig().setCheckpointingMode(CheckpointingMode.EXACTLY_ONCE);

        DataStreamSource<String> streamSource = env.addSource(sourceFunction);
        streamSource.print("mysql server");

        env.execute("mysql server cdc");
    }
}
```

create a test table in database inventory and keep it empty, then run the above test case:
```sql
CREATE TABLE `test` (
  `id` int NOT NULL,
  `name` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

state-001
state-002

state-003

dongdongking008: MichaelXucf, you are right.
I think that is the point: capture a table that never had data.
We get offset from SourceRecord.sourceOffset() but it returns null, since there are no any records found both in snapshot phase and streaming change phase.
@wuchong would you please reopen this PR or just fix this problem in another PR?

wuchong: Please feel free to open a new pull request to fix this @MichaelXucf @dongdongking008

MichaelXucf: Hi @wuchong  @dongdongking008, I have committed a [pull request #119](https://github.com/ververica/flink-cdc-connectors/pull/119) to fix this.

PR #91, opened 2021-01-26
feat: add support to jdk 9+

wallezhang:
#10 I used a queue to store the change event and another consumption thread to consume it, eliminating the use of the `Unsafe` class

### DIFF
--- a/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
+++ b/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
@@ -43,10 +43,10 @@ import com.alibaba.ververica.cdc.debezium.internal.DebeziumOffset;
 import com.alibaba.ververica.cdc.debezium.internal.DebeziumOffsetSerializer;
 import com.alibaba.ververica.cdc.debezium.internal.FlinkDatabaseHistory;
 import com.alibaba.ververica.cdc.debezium.internal.FlinkOffsetBackingStore;
+import com.alibaba.ververica.cdc.debezium.internal.Handover;
 import io.debezium.document.DocumentReader;
 import io.debezium.document.DocumentWriter;
 import io.debezium.embedded.Connect;
-import io.debezium.engine.ChangeEvent;
 import io.debezium.engine.DebeziumEngine;
 import io.debezium.engine.spi.OffsetCommitPolicy;
 import io.debezium.heartbeat.Heartbeat;
@@ -62,11 +62,9 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Properties;
 import java.util.UUID;
-import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
@@ -117,14 +115,8 @@ public class DebeziumSourceFunction<T> extends RichSourceFunction<T>
     private ExecutorService executor;
     private DebeziumEngine<?> engine;
 
-    /** The error from {@link #engine} thread. */
-    private transient volatile Throwable error;
-
-    /** Flag indicating whether the consumer is still running. */
-    private volatile boolean running = true;
-
-    /** The consumer to fetch records from {@link DebeziumEngine}. */
-    private transient volatile DebeziumChangeConsumer<T> debeziumConsumer;
+	/** The consumer to fetch records from {@link DebeziumEngine}. */
+	private transient volatile DebeziumChangeConsumer<T> debeziumConsumer;
 
     /**
      * The offsets to restore to, if the consumer restores state from a checkpoint.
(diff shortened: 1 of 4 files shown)


flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -138,7 +145,8 @@ public void open(Configuration parameters) throws Exception {
 		ThreadFactory threadFactory = new ThreadFactoryBuilder()
 			.setNameFormat("debezium-engine")
 			.build();
-		this.executor = Executors.newSingleThreadExecutor(threadFactory);
+		this.executor = Executors.newFixedThreadPool(2, threadFactory);
+		this.changeEventQueue = new LinkedBlockingQueue<>();
```
wuchong, line 149: We should set a max size for the queue to avoid OOM when the main thread is blocked.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -303,6 +311,9 @@ public void run(SourceContext<T> sourceContext) throws Exception {
 		// run the engine asynchronously
 		executor.execute(engine);
 
+		// run the real debezium consumer
+		executor.execute(debeziumConsumer);
```
wuchong, line 315: We don't need to start a new thread to poll elements from the queue. We can poll in the following while loop.

pom.xml
```diff
@@ -65,7 +65,7 @@ under the License.
         <flink.version>1.11.1</flink.version>
         <debezium.version>1.2.1.Final</debezium.version>
         <testcontainers.version>1.14.3</testcontainers.version>
-        <java.version>1.8</java.version>
+        <java.version>11</java.version>
```
wuchong, line 68: I think we don't need to upgrade java version. Compiling in java 8 should be compatible with java 11.

wuchong (commented):
Sorry for the late reviewing. I left some comments.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
wallezhang: I wonder if it is necessary to use the `configuration` parameter here, just like: `this.changeEventQueue = new LinkedBlockingQueue<>(parameters.getInteger(ConfigOptions.key("flink.cdc.connector.change.event.queue.size").intType().defaultValue(1024)))`
Or use a constant directly in code

pom.xml
wallezhang: Ok, I will rollback this change.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
wuchong: Size one blocking queue is enough. This blocking queue is only used for handover data. However, I just find that blocking queue may not be sufficient here, because we need extra exception reporting. I'm wondering we may need something similar to the [`Handover`](https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/internals/Handover.java#L50) class in kafka connector.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
wallezhang: OK, I will modify this.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
wallezhang: Yes, it's really enough to use size one blocking queue.😎Actually the event is handled one by one, and in order to maintain the semantic of handover class, so I think we should use `ChangeEvent<SourceRecord, SourceRecord>` as the generic type. The producer put the events into the handover structure in a loop.

wuchong: I think it would be better to use `List<ChangeEvent>`, this can reduce the frequent communication between `DebeziumChangeConsumer` and `Handover`. `ConsumerRecords` is also a List of `ConsumerRecord`.

wallezhang: OK, I will submit a revision later.

wallezhang:
Sorry for the late revision.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
wallezhang: done

pom.xml
wallezhang: done

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
wallezhang: done

wuchong:
Hi @wallezhang , could you rebase and fix the conflicts?

wallezhang:
> Hi @wallezhang , could you rebase and fix the conflicts?

Yes, I'm working on it.

wuchong:
Thanks for the contribution @wallezhang @fsk119 . This feature has been merged.

PR #95, opened 2021-01-29
upgrade flink version to 1.12.0

Jiabao-Sun:
upgrade flink version to 1.12.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This README is meant as a brief walkthrough on the core features with Flink CDC 
 
 We need several steps to setup a Flink cluster with the provided connector.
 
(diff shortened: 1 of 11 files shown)


Sdu0:
太棒了，期待已久

wuchong:
Thanks for the contribution @Jiabao-Sun . I upgrade the Flink version to 1.12.1 and revert changes of changelog-json. Previously it added 2 new options which is not needed, because we can pass the default value to the paranet json deserializer. If we want to support these 2 new options, we should do that in another PR. 

I will merge this PR once CI is passed.

wuchong:
Merged in 2dfb59f3c783d8bc56b25a72517aa5589451cb85

PR #105, opened 2021-02-18
[debezium] Fix bug with inefficient batch event handling

SmirAlex:
During studying DebeziumChangeConsumer source code I found out, that methods that supposed to process batch of events process them one by one, which is not efficient (there are always only 1 element during processing in DebeziumCollector records queue). Fixed this bug in this commit.

### DIFF
--- a/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumChangeConsumer.java
+++ b/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumChangeConsumer.java
@@ -92,10 +92,11 @@ public class DebeziumChangeConsumer<T> implements DebeziumEngine.ChangeConsumer<
 			List<ChangeEvent<SourceRecord, SourceRecord>> changeEvents,
 			DebeziumEngine.RecordCommitter<ChangeEvent<SourceRecord, SourceRecord>> committer) throws InterruptedException {
 		try {
+			Map<String, ?> lastSourcePartition = null;
+			Map<String, ?> lastSourceOffset = null;
 			for (ChangeEvent<SourceRecord, SourceRecord> event : changeEvents) {
 				SourceRecord record = event.value();
 				deserialization.deserialize(record, debeziumCollector);
-
 				if (isInDbSnapshotPhase) {
 					if (!lockHold) {
 						MemoryUtils.UNSAFE.monitorEnter(checkpointLock);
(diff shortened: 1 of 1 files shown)


wuchong:
We can't assume there is only one element in the `List<SourceRecord>`, otherwise we may lose data. According to the javadoc of `ChangeConsumer` and the implementation, there is possible the size is larger than 1.

SmirAlex:
Maybe you misunderstood me. I don't assume there is only one element in `List<SourceRecord>`. In this code I accumulate records from this list in queue of [DebeziumCollector](https://github.com/ververica/flink-cdc-connectors/blob/e2d4ff735c7a94fa5073622260a78e5c0229ee97/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumChangeConsumer.java#L182) and pass these accumulated records to the method [emitRecordsUnderCheckpointLock](https://github.com/ververica/flink-cdc-connectors/blob/e2d4ff735c7a94fa5073622260a78e5c0229ee97/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumChangeConsumer.java#L134). I take last record offset as an offset of the whole batch of records (i.e. `List<SourceRecord>`). Currently in master branch all records are being passed to method [emitRecordsUnderCheckpointLock](https://github.com/ververica/flink-cdc-connectors/blob/e2d4ff735c7a94fa5073622260a78e5c0229ee97/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumChangeConsumer.java#L134) one by one (because as I said there are only 1 element in [queue](https://github.com/ververica/flink-cdc-connectors/blob/e2d4ff735c7a94fa5073622260a78e5c0229ee97/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumChangeConsumer.java#L182) at any given time), which is not efficient especially considering that locks are taken there.

wuchong:
Thanks @SmirAlex . I got what you mean. However, we are going to refactor `DebeziumChangeConsumer` to avoid using `UNSAFE`, see #91. After the refactoring, the elements consumed from the `ChangeConsumer#handleBatch` will be delivered to a queue instead of emitting to downstream. I think we can do the improvement after this change.

SmirAlex:
Thanks for the answer! I'd like to note that after refactoring of this code problem will still exist, so as you said, it need to be fixed in the future

wuchong:
Yes. You can create an issue to track this problem.

SmirAlex:
Ok. There it is - https://github.com/ververica/flink-cdc-connectors/issues/109

PR #119, opened 2021-03-09
[fix bugs] fixed NPE issues when take a snapshot for an empty table.

MichaelXucf:
when take a snapshot for an empty table, both currentState and restoredOffsetState will be null and will throw an NullPointerException.

### DIFF
--- a/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
+++ b/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
@@ -265,7 +265,7 @@ public class DebeziumSourceFunction<T> extends RichSourceFunction<T>
             }
         } else {
             byte[] currentState = consumer.snapshotCurrentState();
-            if (currentState == null) {
+            if (currentState == null && restoredOffsetState != null) {
                 // the consumer has been initialized, but has not yet received any data,
                 // which means we need to return the originally restored offsets
                 serializedOffset = restoredOffsetState.getBytes(StandardCharsets.UTF_8);
--- a/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumOffsetSerializer.java
+++ b/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumOffsetSerializer.java
@@ -20,6 +20,7 @@ package com.alibaba.ververica.cdc.debezium.internal;
 
 import org.apache.flink.annotation.Internal;
 
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.DeserializationFeature;
 import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.io.IOException;
@@ -38,6 +39,7 @@ public class DebeziumOffsetSerializer {
 
     public DebeziumOffset deserialize(byte[] bytes) throws IOException {
         ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.enable(DeserializationFeature.USE_LONG_FOR_INTS);
         return objectMapper.readValue(bytes, DebeziumOffset.class);
     }
 }


flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumOffsetSerializer.java
```diff
@@ -38,6 +39,7 @@
 
     public DebeziumOffset deserialize(byte[] bytes) throws IOException {
         ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.enable(DeserializationFeature.USE_LONG_FOR_INTS);
```
wuchong, line 42: Why need this?

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -265,7 +265,7 @@ private void snapshotOffsetState(long checkpointId) throws Exception {
             }
         } else {
             byte[] currentState = consumer.snapshotCurrentState();
-            if (currentState == null) {
+            if (currentState == null && restoredOffsetState != null) {
```
wuchong, line 268: Could you add a test which reproduces this NPE?

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumOffsetSerializer.java
MichaelXucf: I have added  an issue #121 to explain this. You can take a look.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
MichaelXucf: This is the test case:

prepare database and two table:
```sql
CREATE DATABASE IF NOT EXISTS inventory;

use inventory;

CREATE TABLE products (
  id INTEGER NOT NULL AUTO_INCREMENT PRIMARY KEY,
  name VARCHAR(255) NOT NULL,
  description VARCHAR(512),
  weight FLOAT
);
ALTER TABLE products AUTO_INCREMENT = 101;

INSERT INTO products 
VALUES (default,"scooter","Small 2-wheel scooter",3.14),
       (default,"car battery","12V car battery",8.1),
       (default,"12-pack drill bits","12-pack of drill bits with sizes ranging from #40 to #3",0.8),
       (default,"hammer","12oz carpenter's hammer",0.75),
       (default,"hammer","14oz carpenter's hammer",0.875),
       (default,"hammer","16oz carpenter's hammer",1.0),
       (default,"rocks","box of assorted rocks",5.3),
       (default,"jacket","water resistent black wind breaker",0.1),
       (default,"spare tire","24 inch spare tire",22.2);

CREATE TABLE `test` (
  `id` int NOT NULL,
  `name` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

```java
package com.alibaba.ververica.cdc.connectors.mysql;

import com.alibaba.ververica.cdc.debezium.DebeziumSourceFunction;
import com.alibaba.ververica.cdc.debezium.StringDebeziumDeserializationSchema;
import org.apache.flink.streaming.api.CheckpointingMode;
import org.apache.flink.streaming.api.datastream.DataStreamSource;
import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
import org.junit.Test;

public class MyTest {
    @Test
    public void testSource() throws Exception {
        DebeziumSourceFunction<String> sourceFunction = MySQLSource.<String>builder()
                .hostname("localhost")
                .port(3306)
                .databaseList("inventory")
                .username("root")
                .password("123456")
                .tableList(new String[]{"inventory.products"})// OK, table has data in it.
                .tableList(new String[]{"inventory.test"})// emtpy table, throw Exception
                .tableList(new String[]{"inventory.null"})// not exist table, throw Exception
                .build();

        env.enableCheckpointing(10000);
        env.getCheckpointConfig().setCheckpointingMode(CheckpointingMode.EXACTLY_ONCE);

        DataStreamSource<String> streamSource = env.addSource(sourceFunction);
        streamSource.print("mysql server");

        env.execute("mysql server cdc");
    }
}
```

if we don't change the if condition to ```if (currentState == null && restoredOffsetState != null) ```, it will throw exception when capature a empty table or not exist table.
This problem has been mentioned in the pull quest #85 .

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/DebeziumOffsetSerializer.java
MichaelXucf: This commit may need to be put in another pull request since it is for another issue. when I added this pull request, I just had one commit ```a27b810``` for it. I don't why this commit ```66a5a92``` appears here.

wuchong:
Fixed in f99887755a2d13bd173402512cca46cbbf7410b0

PR #176, opened 2021-05-08
[debezium] Bump Flink version to 1.13.0

leonardBang:
Flink has released its major version to 1.13.0, we can bump to the latest version now.


wuchong (approved):
LGTM

PR #194, opened 2021-05-28
Add support to jdk 9+

fsk119:
This is based on #91 with some minor fix and resolve conflicts.


flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/Handover.java
```diff
@@ -0,0 +1,223 @@
...
+public class Handover implements Closeable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(Handover.class);
+    private final Object lock = new Object();
+
+    private Pair<
+                    RecordCommitter<ChangeEvent<SourceRecord, SourceRecord>>,
+                    List<ChangeEvent<SourceRecord, SourceRecord>>>
+            next;
+    private volatile Throwable error;
```
wuchong, line 62: Do not need to declare `volatile`, because all the reads and writes on `error` are already protected by the lock.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -345,73 +376,60 @@ public void run(SourceContext<T> sourceContext) throws Exception {
...
                 DebeziumEngine.create(Connect.class)
                         .using(properties)
-                        .notifying(debeziumConsumer)
+                        .notifying(changeConsumer)
                         .using(OffsetCommitPolicy.always())
                         .using(
                                 (success, message, error) -> {
-                                    if (!success && error != null) {
-                                        this.reportError(error);
+                                    debeziumChangeFetcher.close();
```
wuchong: Don't need to close fetcher here, just need to report error and the fetcher will be closed .

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -145,12 +157,29 @@
...
      * Unique name of this Debezium Engine instance across all the jobs. Currently we randomly
      * generate a UUID for it. This is used for {@link FlinkDatabaseHistory}.
      */
     private transient String engineInstanceName;
 
+    /** Consume the events from the engine and commit the offset to the engine. */
+    private transient DebeziumChangeConsumer changeConsumer;
+
+    /** The consumer to fetch records from {@link Handover}. */
+    private transient volatile DebeziumChangeFetcher<T> debeziumChangeFetcher;
```
wuchong: Don't need `volatile` because it will not be accessed concurrently.

wuchong:
I helped to improve some code.

fsk119:
Thanks for @wuchong help. LGTM.

PR #207, opened 2021-06-08
[common] Fix schema history run out of memory

fsk119:
*This PR aims to solve the out of memory poblem caused by the frequent schema changes. In this PR, we propose to use `FlinkDatabaseSchemaHistory`, which only stores the current snapshot of the table schema. If the pipeline fails and recovers from the state, it doesn't need to replay all the ddl history any more but only applies the maintained tables.*

*When starting from a state, the source funciton will automatically determine which history should use according to the state structure.  Users don't need to pay attention to the internal implementation.*

**Break Changes**
Because the new database history has a different state structure comparing to before. Therefore, it's not avaliable to start the pipeline with a state with different structure.

### DIFF
--- a/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
+++ b/flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
@@ -47,6 +47,7 @@ import com.alibaba.ververica.cdc.debezium.internal.FlinkDatabaseHistory;
 import com.alibaba.ververica.cdc.debezium.internal.FlinkDatabaseSchemaHistory;
 import com.alibaba.ververica.cdc.debezium.internal.FlinkOffsetBackingStore;
 import com.alibaba.ververica.cdc.debezium.internal.Handover;
+import com.alibaba.ververica.cdc.debezium.internal.SchemaRecord;
 import io.debezium.document.DocumentReader;
 import io.debezium.document.DocumentWriter;
 import io.debezium.embedded.Connect;
@@ -54,7 +55,6 @@ import io.debezium.engine.DebeziumEngine;
 import io.debezium.engine.spi.OffsetCommitPolicy;
 import io.debezium.heartbeat.Heartbeat;
 import io.debezium.relational.history.DatabaseHistory;
-import io.debezium.relational.history.HistoryRecord;
 import org.apache.commons.collections.map.LinkedMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -121,8 +121,14 @@ public class DebeziumSourceFunction<T> extends RichSourceFunction<T>
     /** The maximum number of pending non-committed checkpoints to track, to avoid memory leaks. */
     public static final int MAX_NUM_PENDING_CHECKPOINTS = 100;
 
-    private static final String LEGACY_IMPLEMENTATION_KEY = "internal.implementation";
-    private static final String LEGACY_IMPLEMENTATION_VALUE = "legacy";
+    /**
+     * The configuration represents the Debezium MySQL Connector uses the legacy implementation or
+     * not.
+     */
+    public static final String LEGACY_IMPLEMENTATION_KEY = "internal.implementation";
+
+    /** The configuration value represents legacy implementation. */
+    public static final String LEGACY_IMPLEMENTATION_VALUE = "legacy";
 
     // ---------------------------------------------------------------------------------------
     // Properties
@@ -148,10 +154,10 @@ public class DebeziumSourceFunction<T> extends RichSourceFunction<T>
     // ---------------------------------------------------------------------------------------
 
     /**
-     * Structure to maintain the current schema history. The content in {@link HistoryRecord} is up
+     * Structure to maintain the current schema history. The content in {@link SchemaRecord} is up
      * to the implementation of the {@link DatabaseHistory}.
      */
-    private static final ConcurrentMap<String, Collection<HistoryRecord>> HISTORY =
+    private static final ConcurrentMap<String, Collection<SchemaRecord>> HISTORY =
             new ConcurrentHashMap<>();
 
     /**
@@ -171,8 +177,9 @@ public class DebeziumSourceFunction<T> extends RichSourceFunction<T>
      * State to store the history records, i.e. schema changes.
      *
      * @see FlinkDatabaseHistory
+     * @see FlinkDatabaseSchemaHistory
      */
-    private transient ListState<String> historyRecordsState;
+    private transient ListState<String> schemaRecordsState;
 
     // ---------------------------------------------------------------------------------------
     // Worker
@@ -228,7 +235,7 @@ public class DebeziumSourceFunction<T> extends RichSourceFunction<T>
                         new ListStateDescriptor<>(
                                 OFFSETS_STATE_NAME,
                                 PrimitiveArrayTypeInfo.BYTE_PRIMITIVE_ARRAY_TYPE_INFO));
-        this.historyRecordsState =
+        this.schemaRecordsState =
                 stateStore.getUnionListState(
                         new ListStateDescriptor<>(
                                 HISTORY_RECORDS_STATE_NAME, BasicTypeInfo.STRING_TYPE_INFO));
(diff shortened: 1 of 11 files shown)


flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -519,4 +535,46 @@ public LinkedMap getPendingOffsetsToCommit() {
...
+    private Class<?> determineDatabase() {
+        boolean isCompatibleWithLegacy =
+                FlinkDatabaseHistory.isCompatible(StateUtils.retrieveHistory(engineInstanceName));
+        if (LEGACY_IMPLEMENTATION_VALUE.equals(properties.get(LEGACY_IMPLEMENTATION_KEY))) {
+            // specifies the legacy implementation but the state may be incompatible
+            if (isCompatibleWithLegacy) {
+                return FlinkDatabaseHistory.class;
+            } else {
+                throw new IllegalStateException(
+                        "Use the legacy implementation of the database history but FlnkDatabaseHistory can't load the state.");
```
leonardBang, line 548:
```suggestion
                         "The configured option 'debezium.internal.implementation' is 'legacy', but the state of source is incompatible with this implementation, you should remove the the option.");
```
```suggestion
                        "Use the legacy implementation of the database history but FlnkDatabaseHistory can't load the state.");
```

flink-connector-mysql-cdc/src/test/java/com/alibaba/ververica/cdc/connectors/mysql/table/MySQLConnectorITCase.java
```diff
@@ -144,9 +144,11 @@ public void testConsumingAllEvents()
                     "UPDATE products SET description='new water resistent white wind breaker', weight='0.5' WHERE id=110;");
             statement.execute("UPDATE products SET weight='5.17' WHERE id=111;");
             statement.execute("DELETE FROM products WHERE id=111;");
+            statement.execute(
+                    "INSERT INTO products(id, description, weight) VALUES (default, 'Go go go', 0.1);");
```
leonardBang, line 148: The statement is meaningless?

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -519,4 +535,46 @@ public LinkedMap getPendingOffsetsToCommit() {
...
+            throw new IllegalStateException("Can't determine which DatabaseHistory to use.");
+        }
+    }
+
+    // ---------------------------------------------------------------------------------------
+
+    /** Utils to get/put/remove the history. */
+    public static final class StateUtils {
+        public static void registerHistory(
+                String engineName, Collection<HistoryRecord> engineHistory) {
```
leonardBang, line 568: we should `synchronize` visit the data to ensure the data won't change when checkpoint ?

flink-connector-postgres-cdc/src/test/resources/ddl/inventory.sql
```diff
@@ -21,7 +21,7 @@ SET search_path TO inventory;
 -- Create and populate our products using a single insert with many rows
 CREATE TABLE products (
   id SERIAL NOT NULL PRIMARY KEY,
-  name VARCHAR(255) NOT NULL,
+  name VARCHAR(255) NOT NULL DEFAULT 'flink',
```
leonardBang, line 24: ditto

leonardBang (commented):
Thanks @fsk119 for the contribution, I left some comments

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/FlinkDatabaseSchemaHistory.java
```diff
@@ -0,0 +1,205 @@
...
+            Map<String, ?> source,
+            Map<String, ?> position,
+            String databaseName,
+            String schemaName,
+            String ddl,
+            TableChanges changes)
+            throws DatabaseHistoryException {
+        final HistoryRecord record =
+                new HistoryRecord(source, position, databaseName, schemaName, ddl, changes);
+        for (TableChanges.TableChange change : changes) {
```
wuchong, line 116: Is it possible the `changes` is null? What happens when it is null?

flink-connector-mysql-cdc/src/test/java/com/alibaba/ververica/cdc/connectors/mysql/MySQLSourceTest.java
```diff
@@ -744,10 +909,72 @@ public void go() throws Exception {
...
 
+    private static Properties createDebeziumProperties(boolean useLegacyImplementation) {
+        Properties debeziumProps = new Properties();
+        if (useLegacyImplementation) {
+            debeziumProps.put("internal.implementation", "legacy");
+            // check legacy mysql record type
+            debeziumProps.put("transforms", "snapshotasinsert");
+            debeziumProps.put(
+                    "transforms.snapshotasinsert.type",
+                    "io.debezium.connector.mysql.transforms.ReadToInsertEvent");
```
wuchong, line 920: Are the configurations also required when using legacy mode? I mean do users also need to set them when using legacy mode.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/internal/FlinkDatabaseSchemaHistory.java
```diff
@@ -0,0 +1,205 @@
...
+            switch (change.getType()) {
+                case CREATE:
+                case ALTER:
+                    tables.put(
+                            change.getId(),
+                            new HistoryRecord(tableChangesSerializer.toDocument(change)));
```
wuchong, line 122: I would like to use `HistoryRecord(Map<String, ?> source, Map<String, ?> position, String databaseName, String schemaName, String ddl, TableChanges changes)` to construct a complete `HistoryRecord`. Your way may have some problems because it's not a valid `HistoryRecord`, because TableChanges is not in "tableChanges" field.

flink-connector-mysql-cdc/src/test/java/com/alibaba/ververica/cdc/connectors/mysql/MySQLSourceTest.java
```diff
@@ -401,6 +414,89 @@ public void go() throws Exception {
...
+                    };
+            runThread5.start();
+
+            try (Connection connection = database.getJdbcConnection();
+                    Statement statement = connection.createStatement()) {
+
+                statement.execute(
+                        "INSERT INTO products(id, description, weight) VALUES (default, 'Go go go', 111.1)");
+                statement.execute(
+                        "ALTER TABLE products ADD comment_col VARCHAR(100) DEFAULT 'cdc'");
```
wuchong, line 442: Could you also add the partial insert and alter column (before checkpoint, and after restore) in IT case? I think we didn't cover record deserialization with the new `DatabaseHistory`.

flink-connector-mysql-cdc/src/test/java/com/alibaba/ververica/cdc/connectors/mysql/MySQLSourceTest.java
leonardBang: yes

flink-connector-mysql-cdc/src/test/java/com/alibaba/ververica/cdc/connectors/mysql/MySQLSourceTest.java
leonardBang: I didn't add this for IT case for now, this PR’s change won't affect the IT case.
And in my  [mysql-cdc-2.0 branch](https://github.com/leonardBang/flink-cdc-connectors/blob/mysql_parallel_nolock/flink-connector-mysql-cdc/src/test/java/com/alibaba/ververica/cdc/connectors/mysql/debezium/reader/BinlogSplitReaderTest.java#L267), I've reconstructed  the  ITCase.

fsk119:
Thanks for @leonardBang 's improvements. LGTM. 

Look forward for your meetup. :)

wuchong (approved):
LGTM.

PR #221, opened 2021-06-22
provide inteceptor

lqjack:


leonardBang:
@lqjack Thanks for the contribution, could you add description for this PR? 
I suggest we discuss the new feature first and develop it after we have reached a consensus

lqjack:
@leonardBang  Thanks the response.  The reason why we need to modify the code as below : 
1,  background : we need custom logic to inject the handle SourceRecord .  ​So I suggest that we can provide the interceptor interface for developer to inject custom logical for themselves. 
2, For the extension for Apache Flink in inputformat/outputformat/SourceFunction/SinkFounction JIRA: https://issues.apache.org/jira/browse/FLINK-23103.

fsk119:
I just curious what kind of `Inteceptor` to you need? Flink has already provide us with various push-down interfaces, e.g. `SupportsProjectionPushDown`. Could you share with us with more details?

[1] https://ci.apache.org/projects/flink/flink-docs-master/docs/dev/table/sourcessinks/#source-abilities

Issue #72, opened 2020-12-18
数据传输过程中，job无法cancel掉


laifulin:
2020-12-18 22:38:00,329 WARN  org.apache.flink.runtime.taskmanager.Task                    [] - Task 'Source: TableSourceScan(table=[[default_catalog, default_database, 表名]], fields=[字段名]) -> MiniBatchAssigner(interval=[2000ms], mode=[ProcTime]) -> Calc(select=[sql语句]) (1/1)' did not react to cancelling signal for 30 seconds, but is stuck in method:
 org.apache.flink.streaming.runtime.tasks.StreamTaskActionExecutor$SynchronizedStreamTaskActionExecutor.runThrowing(StreamTaskActionExecutor.java:91)
org.apache.flink.streaming.runtime.tasks.mailbox.Mail.run(Mail.java:78)
org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.processMail(MailboxProcessor.java:282)
org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.runMailboxStep(MailboxProcessor.java:190)
org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.runMailboxLoop(MailboxProcessor.java:181)
org.apache.flink.streaming.runtime.tasks.StreamTask.runMailboxLoop(StreamTask.java:566)
org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:536)
org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:721)
org.apache.flink.runtime.taskmanager.Task.run(Task.java:546)
java.lang.Thread.run(Thread.java:748)

当表的数据量很多的时候，任务运行中，在web ui上cancel任务时，发现无法关闭任务，一直显示CANCELING，暂用slot也没有释放掉   ---   standalone集群模式下出现

jasonzhangz:
目前也遇到这样的问题，pg cdc , 使用 flink cancel 命令方式，web ui 方式， flink rest api方式尝试取消任务，都一直处于CANCELING。

yaoyi:
@jasonzhangz 请问还有出现这种情况吗，我这边使用最新2.2.1版本，pg cdc还是会出现无法取消的情况

PatrickRen:
[...]

PR #225, opened 2021-06-29
[ISSUE-11] MongoDB cdc connector of Flink 1.13.0

Jiabao-Sun:
### Summary

Use Debezium EmbeddedEngine to drive a SourceConnector of  [MongoDB Kafka Connector](https://github.com/mongodb/mongo-kafka).
MongoDB Kafka Connector's  CDC mechanism is [MongoDB ChangeSteams](https://docs.mongodb.com/manual/changeStreams/) which is available for replica sets and sharded clusters.

For a fully detailed documentation of MongoDB Kafka Connector , please see [Documentation](https://docs.mongodb.com/kafka-connector/master/kafka-source/).
___________________

#### 前置条件
- MongoDB Version > 3.6 .
- MongoDB run in replica sets or sharded clusters mode.
- changeStream and read privileges required. [Database-User-Roles](https://docs.mongodb.com/v4.0/reference/built-in-roles/#database-user-roles)

Create a user:
```javascript
use admin;
db.createUser(
 {
   user: "flinkuser",
   pwd: "flinkpw",
   roles: [
      { role: "read", db: "admin" },
      { role: "readAnyDatabase", db: "admin" }
   ]
 }
);
```

Flink SQL Example：
```sql
CREATE TABLE mongodb_source (
   _id STRING PRIMARY KEY, #upsert stream, primary key required
   name STRING,
   description STRING,
   weight DECIMAL(10,3)
) WITH (
   'connector' = 'mongodb-cdc',
   'hosts' = '127.0.0.1:27017',
   'username' = 'flinkuser',
   'password' = 'flinkpw',
   'database' = 'inventory',
   'collection' = 'products'
);

SELECT _id, name, description, weight FROM mongodb_source;
```

DataStream Example：
```java
import com.alibaba.ververica.cdc.debezium.StringDebeziumDeserializationSchema;
import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
import org.apache.flink.streaming.api.functions.source.SourceFunction;

public class MongoDBSourceExample {
    public static void main(String[] args) throws Exception {
        SourceFunction<String> sourceFunction = MongoDBSource.<String>builder()
                .hosts("127.0.0.1:27017")
                .username("flinkuser")
                .password("flinkpw")
                .database("inventory")
                .collection("products")
                .build();

        env.addSource(sourceFunction)
                .print().setParallelism(1); // use parallelism 1 for sink to keep message ordering

        env.execute();
    }
}
```
___________________

### Connector Options
| Name | Description | Required | Default | 
| --- | --- | --- | --- |
| hosts | The comma-separated list of hostname and port pairs of the MongoDB servers.<br> eg. localhost:27017,localhost:27018 | true | - |
| username | Name of the database user to be used when connecting to MongoDB.<br> This is required only when MongoDB is configured to use authentication. | false | - |
| password | Password to be used when connecting to MongoDB.<br> This is required only when MongoDB is configured to use authentication. | false | - |
| database | Name of the database to watch for changes. | true | - |
| collection | Name of the collection in the database to watch for changes. | true | - |
| connection.options | The ampersand-separated [connection options](https://docs.mongodb.com/manual/reference/connection-string/#std-label-connections-connection-options) of MongoDB. eg. `replicaSet=test&connectTimeoutMS=300000`. | false | - |
| errors.tolerance| Whether to continue processing messages if an error is encountered.  | false | none |
| errors.log.enable | Whether details of failed operations should be written to the log file.  | false | true |
| copy.existing | Whether copy existing data from source collections | false | true |
| copy.existing.pipeline | An array of JSON objects describing the pipeline operations to run when copying existing data. This can improve the use of indexes by the copying manager and make copying more efficient. eg. `[{"$match": {"closed": "false"}}]` ensures that only documents in which the closed field is set to false are copied. | false | - |
| copy.existing.max.threads | The number of threads to use when performing the data copy. | false | Processors Count |
| copy.existing.queue.size |  The max size of the queue to use when copying data. | false |16000|
| poll.max.batch.size | Maximum number of change stream documents to include in a single batch when polling for new data. |  false | 1000 |
| poll.await.time.ms | The amount of time to wait before checking for new results on the change stream. | false | 1500 |
| heartbeat.interval.ms | The length of time in milliseconds between sending heartbeat messages. Use 0 to disable. | false | 0 |

___________________

### Reference
[MongoDB Kafaka Connector](https://docs.mongodb.com/kafka-connector/master/kafka-source/)
[MongoDB Change Streams](https://docs.mongodb.com/manual/changeStreams/)

___________________


Jiabao-Sun:
Hi, Jark. @wuchong 
Do you have any suggestions?

wuchong:
@Jiabao-Sun , I went through the API and examples, and it looks very well. I think it's already in a good shape. 
I will find time to review it, but quite busy this week.

carloscbl:
I've compiled and tested in my env and seems to be working good! flink 1.13

Jiabao-Sun:
> I've compiled and tested in my env and seems to be working good! flink 1.13

Thanks for your reply.

carloscbl:
org.apache.flink.table.api.ValidationException: Table 'MongoDB-CDC' declares metadata columns, but the underlying DynamicTableSource doesn't implement the SupportsReadingMetadata interface. Therefore, metadata cannot be read from the given source.

Is this intended?

Jiabao-Sun:
> org.apache.flink.table.api.ValidationException: Table 'MongoDB-CDC' declares metadata columns, but the underlying DynamicTableSource doesn't implement the SupportsReadingMetadata interface. Therefore, metadata cannot be read from the given source.
> 
> Is this intended?



> org.apache.flink.table.api.ValidationException: Table 'MongoDB-CDC' declares metadata columns, but the underlying DynamicTableSource doesn't implement the SupportsReadingMetadata interface. Therefore, metadata cannot be read from the given source.
> 
> Is this intended?

@carloscbl Currently this ability has not been realized, but I think it's a helpful suggestion.

We can provide some metadata such as [Debezium Format](https://ci.apache.org/projects/flink/flink-docs-release-1.13/zh/docs/connectors/table/formats/debezium/) describes.

Key | Data Type | Description
-- | -- | --
schema | STRING NULL | JSON string describing the schema of the payload. Null if the schema is not included in the Debezium record.
ingestion-timestamp | TIMESTAMP_LTZ(3) NULL | The timestamp at which the connector processed the event. Corresponds to the ts_ms field in the Debezium record.
source.timestamp | TIMESTAMP_LTZ(3) NULL | The timestamp at which the source system created the event. Corresponds to the source.ts_ms field in the Debezium record.
source.database | STRING NULL | The originating database. Corresponds to the source.db field in the Debezium record if available.
source.schema | STRING NULL | The originating database schema. Corresponds to the source.schema field in the Debezium record if available.
source.table | STRING NULL | The originating database table. Corresponds to the source.table or source.collection field in the Debezium record if available.

CC @wuchong @leonardBang
If you think this is a necessary feature, I'm happy to open a new PR to deal with it.

yangxusun9:
when i cancle a  job which is  mongocdc stream join kafka stream , i found mongocdc stream can't be closed .There are some pictures to show:
image
TM log
image
then i check the pom.xml ,i found you used mongo-kafka-connect ,but not debezium-connector-mongodb ,Can you tell me why . i think this is the reason to my question

Jiabao-Sun:
@yangxusun9 

Q1.

Maybe relate to #72. 

Please help check: 
- Snapshot phase is done.
- Whether the source table is changing rapidly.
- Is the `execution.checkpointing.interval` set too short causing the checkpoint to be failed all the time.


Q2. 

Debezium mongodb connector cannot extract full document of a changed record.
So it's difficult to transform to flink's upsert stream.
Please see the previous PR for detail [PR-123](https://github.com/ververica/flink-cdc-connectors/pull/123).

yangxusun9:
@Jiabao-Sun Thanks for your reply! 
to Q1,I'm sure that Snapshot phase is done but  **the source table is changing rapidly** .you means it can't be closed if the source table is changing rapidly

Jiabao-Sun:
> @Jiabao-Sun Thanks for your reply!
> to Q1,I'm sure that Snapshot phase is done but **the source table is changing rapidly** .you means it can't be closed if the source table is changing rapidly

@yangxusun9 
If you cancel the flink job with a savepoint, must acquire the checkpoint lock first.
Rapidly changing or too short checkpointing interval may reduce the possibility of acquiring the checkpoint lock.
In addition, long time savepoint will also increase the time to cancel the task.

However, there may be other reasons that cause this problem.
For example, the operators were not interrupted or destroyed as expected.

If you can reproduce this problem, is it convenient to provide a detailed log for troubleshooting?

Jiabao-Sun:
Upgrade mongo-kafka-connect from 1.4.0 to 1.6.1 and mongo-driver from 4.2.1 to 4.3.1 to support MongoDB 5.0.

yangxusun9:
> > @Jiabao-Sun Thanks for your reply!
> > to Q1,I'm sure that Snapshot phase is done but **the source table is changing rapidly** .you means it can't be closed if the source table is changing rapidly
> 
> @yangxusun9
> If you cancel the flink job with a savepoint, must acquire the checkpoint lock first.
> Rapidly changing or too short checkpointing interval may reduce the possibility of acquiring the checkpoint lock.
> In addition, long time savepoint will also increase the time to cancel the task.
> 
> However, there may be other reasons that cause this problem.
> For example, the operators were not interrupted or destroyed as expected.
> 
> If you can reproduce this problem, is it convenient to provide a detailed log for troubleshooting?

It is solved because of the problem of using the platform. If you directly call cancel command, it can be cancelled normally
Thanks for your help!

Jiabao-Sun:
> > > @Jiabao-Sun Thanks for your reply!
> > > to Q1,I'm sure that Snapshot phase is done but **the source table is changing rapidly** .you means it can't be closed if the source table is changing rapidly
> > 
> > 
> > @yangxusun9
> > If you cancel the flink job with a savepoint, must acquire the checkpoint lock first.
> > Rapidly changing or too short checkpointing interval may reduce the possibility of acquiring the checkpoint lock.
> > In addition, long time savepoint will also increase the time to cancel the task.
> > However, there may be other reasons that cause this problem.
> > For example, the operators were not interrupted or destroyed as expected.
> > If you can reproduce this problem, is it convenient to provide a detailed log for troubleshooting?
> 
> It is solved because of the problem of using the platform. If you directly call cancel command, it can be cancelled normally
> Thanks for your help!

Hi @yangxusun9
Thank you for your feedback on this issue.
In order to prevent others from encountering the same problem, is it convenient to tell what framework is used, zeppelin or others?

flink-connector-mongodb-cdc/src/test/java/com/ververica/cdc/connectors/mongodb/table/MongoDBConnectorITCase.java
```diff
@@ -0,0 +1,361 @@
...
+        String sourceDDL =
+                String.format(
+                        "CREATE TABLE full_types (\n"
+                                + "    _id STRING,\n"
+                                + "    stringField STRING,\n"
+                                + "    uuidField STRING,\n"
+                                + "    md5Field STRING,\n"
+                                + "    dateField DATE,\n"
+                                + "    dateBefore1970 DATE,\n"
+                                + "    timestampField TIMESTAMP,\n"
```
wuchong, line 199: Add test for `TIMESTAMP_LTZ` as well.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBConnectorDeserializationSchema.java
```diff
@@ -0,0 +1,727 @@
...
+                        + "' of type "
+                        + docObj.getBsonType());
+    }
+
+    private TimestampData convertToTimestamp(BsonValue docObj) {
+        if (docObj.isDateTime()) {
+            return TimestampData.fromEpochMillis(docObj.asDateTime().getValue());
+        }
+        if (docObj.isTimestamp()) {
+            return TimestampData.fromEpochMillis(docObj.asTimestamp().getTime() * 1000L);
```
wuchong, line 495: What does the underlying timestamp value means? Is there any documentation to reference? 

Usually, `TimestampData.fromEpochMillis` is used for `TIMESTAMP_LTZ` types where the epoch milliseconds represent Java `Instant`. We recommend to use `TimestampData.fromLocalDateTime` for `TIMESTAMP` types.

wuchong (commented):
Thanks @Jiabao-Sun for the great and high-quality work! The pull request looks good to me in general. I don't have much knowledge about MongoDB, so I only left some comments from Flink side. 

Cheers!

wuchong:
Besides, I created #393 for the mongodb-cdc connector, would like to take it as well ?

Jiabao-Sun:
> Besides, I created #393 for the mongodb-cdc connector, would like to take it as well ?

Thanks, be happy to do with it.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/internal/MongoDBConnectorSourceTask.java
```diff
@@ -0,0 +1,197 @@
...
+                        outSourceRecords.add(current);
+                    }
+                }
+            } else {
+                // Snapshot Phase Ended, Condition 2:
+                // No changing stream event comes and source task is finished copying,
+                // then exit the snapshot phase.
+                if (!isCopying()) {
+                    outSourceRecords =
+                            Collections.singletonList(markLastSnapshotRecord(lastSnapshotRecord));
```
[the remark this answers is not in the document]
Jiabao-Sun: Thanks, Jark
`lastSnapshotRecord` may be `null` when the source collection is empty and a short poll waiting time is set.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/internal/MongoDBConnectorSourceTask.java
```diff
@@ -0,0 +1,197 @@
...
+                            outSourceRecords.add(lastSnapshotRecord);
+                        }
+                        // Keep the current last snapshot record.
+                        // When exit snapshot phase, mark it as the last of all snapshot records.
+                        lastSnapshotRecord = current;
+                    } else {
+                        // Snapshot Phase Ended, Condition 1:
+                        // Received non-snapshot record, exit snapshot phase immediately.
+                        if (lastSnapshotRecord != null) {
+                            outSourceRecords.add(markLastSnapshotRecord(lastSnapshotRecord));
```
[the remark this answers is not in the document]
Jiabao-Sun: No, we don't.

The MongoDB Connector doesn't mark the last snapshot record as Debezium. 
So we should retain the current last snapshot record (do not emit it in current batch) util new snapshot batch come or the snapshot phase is done.
If the snapshot phase is done and there is a retained snapshot record, mark it as the last snapshot of all and emit it to trigger `DebeziumChangeFetcher` to quit the snapshot phase too.

The naming of `lastSnapshotRecord` may be somewhat ambiguous, I will change it for better readability.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/MongoDBSource.java
```diff
@@ -0,0 +1,383 @@
...
+         * https://docs.mongodb.com/kafka-connector/current/kafka-source
+         */
+        public DebeziumSourceFunction<T> build() {
+            Properties props = new Properties();
+
+            props.setProperty(
+                    "connector.class", MongoDBConnectorSourceConnector.class.getCanonicalName());
+            props.setProperty("name", "mongodb_binlog_source");
+
+            props.setProperty(MongoSourceConfig.CONNECTION_URI_CONFIG, checkNotNull(connectionUri));
```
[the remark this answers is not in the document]
Jiabao-Sun: Sorry Jark, we didn't choose Debezium's MongoDB Connector but MongoDB's Kafka Connector. 
https://docs.mongodb.com/kafka-connector/current/kafka-source/

The `mongodb.name` property is used for kafka topics prefix as `topic.prefix` in MongoDB's Kafka Connector and I think it's not required for our flink cdc connector.

The mechanisms of the two connectors are different:
- For Debezium's MongoDB Connector, it read the `oplog.rs` collection of each replica set master node.
- For MongoDB's Kafka Connector, it subscribe change stream of MongoDB.

MongoDB's `oplog.rs` collection didn't keep the changed record's `update before state`, so it's hard to extract the full document state by one `oplog.rs`'s record and convert it to Flink's upsert changelog. And MongoDB 5 (released in July) has changed the oplog format, so the current Debezium connector cannot be used with it. 

The MongoDB's change streams provided a `updateLookup` feature to return the most current majority-committed version of the updated document. And it's easy to convert change stream event to Flink's upsert changelog.
https://docs.mongodb.com/manual/changeStreams/
https://docs.mongodb.com/manual/reference/change-events/

By the way, Debezium's MongoDB change streams exploration is on roadmap.
If it's done, we can provide two engine for users to choose.
https://issues.redhat.com/browse/DBZ-435

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBConnectorDeserializationSchema.java
Jiabao-Sun: https://docs.mongodb.com/manual/reference/bson-types/

**Timestamp**

> BSON has a special timestamp type for internal MongoDB use and is not associated with the regular Date type. This internal timestamp type is a 64 bit value where:
>- the most significant 32 bits are a time_t value (seconds since the Unix epoch)
>- the least significant 32 bits are an incrementing ordinal for operations within a given second.

The BSON timestamp type is represented by a java long type. High 32 bit's represent unix epoch seconds and low 32 bits's represent an incrementing value to keep time increments and avoid duplication. 

____________

**DateTime**

>BSON Date is a 64-bit integer that represents the number of milliseconds since the Unix epoch (Jan 1, 1970). This results in a representable date range of about 290 million years into the past and future.

Jiabao-Sun: Thanks Jark to for the detailed explanation of `TIMESTAMP_LTZ`.
As the documentation mentioned above, they are semantically equivalent:
- BSON `DateTime`  - `TIMESTAMP_LTZ(3)`
- BSON `Timestamp` -`TIMESTAMP_LTZ(0)`

We treated BSON `Timestamp` as TIMESTAMP_LTZ(0) because it's minimum precision is seconds and the incrementing value does not represent the exact time interval.

So, I think the transformation in the code can work. 
If there is a deviation in my understanding, please help to correct me.

flink-connector-mongodb-cdc/src/test/java/com/ververica/cdc/connectors/mongodb/table/MongoDBConnectorITCase.java
Jiabao-Sun: Fixed

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/MongoDBSource.java
```diff
@@ -0,0 +1,383 @@
...
+import java.util.Properties;
+
+import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * A builder to build a SourceFunction which can read snapshot and continue to consume change stream
+ * events.
+ */
+public class MongoDBSource {
```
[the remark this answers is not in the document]
Jiabao-Sun: Fixed

Jiabao-Sun:
@wuchong 

Thanks a lot for the review and helpful suggestions.
I made some changes to the problem you pointed out, but there are still some issues that need to be confirmed.
If you are available, please help to review it again.

Best

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/internal/MongoDBConnectorSourceTask.java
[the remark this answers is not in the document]
Jiabao-Sun: Fixed.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/MongoDBSource.java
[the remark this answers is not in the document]
wuchong: Thanks for the detailed explanation.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBConnectorDeserializationSchema.java
wuchong: Thanks @Jiabao-Sun , I think it makes sense to map BSON `DateTime` and `Timestamp` to Flink `TIMESTAMP_LTZ` because they all represents the seconds since epoch. However, if we map them to Flink `TIMESTAMP` (i.e. TIMESTAMP WITHOUT TIME ZONE), then we need to do a conversion (epoch seconds to local clock time) just like Java `Instant` to Java `LocalDateTime`. This conversion requires a time zone which means what time zone to dispaly the instant in string. Usually, the time zone is configured by users and not hard code, e.g. [table.local-time-zone](https://ci.apache.org/projects/flink/flink-docs-master/docs/dev/table/config/#table-local-time-zone) in Flink SQL , and [server-time-zone](https://ververica.github.io/flink-cdc-connectors/release-2.0/content/connectors/mysql-cdc.html) in mysql-cdc.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/MongoDBSource.java
```diff
@@ -0,0 +1,386 @@
...
+         */
+        public DebeziumSourceFunction<T> build() {
+            Properties props = new Properties();
+
+            props.setProperty(
+                    "connector.class", MongoDBConnectorSourceConnector.class.getCanonicalName());
+            props.setProperty("name", "mongodb_binlog_source");
+
+            props.setProperty(MongoSourceConfig.CONNECTION_URI_CONFIG, checkNotNull(connectionUri));
+            props.setProperty(MongoSourceConfig.DATABASE_CONFIG, checkNotNull(database));
```
XuQianJin-Stars, line 301: hi @Jiabao-Sun Some mongodb needs to have an account and password to connect, here does not consider the setting of the account and password?

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBTableSource.java
```diff
@@ -0,0 +1,195 @@
...
+        RowType rowType = (RowType) physicalSchema.toRowDataType().getLogicalType();
+        TypeInformation<RowData> typeInfo =
+                scanContext.createTypeInformation(physicalSchema.toRowDataType());
+
+        DebeziumDeserializationSchema<RowData> deserializer =
+                new MongoDBConnectorDeserializationSchema(rowType, typeInfo);
+
+        MongoDBSource.Builder<RowData> builder =
+                MongoDBSource.<RowData>builder()
+                        .connectionUri(uri)
```
XuQianJin-Stars, line 111: @Jiabao-Sun There is also the url here. If Shard, there will be multiple ip:ports.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/MongoDBSource.java
Jiabao-Sun: Hi, @XuQianJin-Stars Please use this pattern `mongodb://username:password@localhost/`

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBTableSource.java
Jiabao-Sun: Hi. @XuQianJin-Stars 
The uri represents MongoDB's connecting string, you can see https://docs.mongodb.com/manual/reference/connection-string/#standard-connection-string-format for detail.

For replica set.
`mongodb://mongodb1.example.com:27317,mongodb2.example.com:27017/?replicaSet=mySet`

For sharded cluster.
`mongodb://mongos0.example.com:27017,mongos1.example.com:27017,mongos2.example.com:27017`

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBConnectorDeserializationSchema.java
Jiabao-Sun: Thanks @wuchong, I'll add a `server-time-zone` configuration for users to configure.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBTableSource.java
wuchong: This uri pattern looks a little complicate. I'm afraid most MongoDB newbies can't set a correct URI at first time. 
Can we decouple the different parts of the URI? e.g. `hosts`, `replica-set`, `username`, `password`. 

Decouple can also makes things flexible, e.g. we would like to mask secrets in logging and displaying.

yangxusun9:
> > > > @Jiabao-Sun Thanks for your reply!
> > > > to Q1,I'm sure that Snapshot phase is done but **the source table is changing rapidly** .you means it can't be closed if the source table is changing rapidly
> > > 
> > > 
> > > @yangxusun9
> > > If you cancel the flink job with a savepoint, must acquire the checkpoint lock first.
> > > Rapidly changing or too short checkpointing interval may reduce the possibility of acquiring the checkpoint lock.
> > > In addition, long time savepoint will also increase the time to cancel the task.
> > > However, there may be other reasons that cause this problem.
> > > For example, the operators were not interrupted or destroyed as expected.
> > > If you can reproduce this problem, is it convenient to provide a detailed log for troubleshooting?
> > 
> > 
> > It is solved because of the problem of using the platform. If you directly call cancel command, it can be cancelled normally
> > Thanks for your help!
> 
> Hi @yangxusun9
> Thank you for your feedback on this issue.
> In order to prevent others from encountering the same problem, is it convenient to tell what framework is used, zeppelin or others?

Flink-VVP
This's link to documentation https://help.aliyun.com/document_detail/169590.html

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBTableSource.java
XuQianJin-Stars: Hi @Jiabao-Sun I also think it is better to set the username and password separately.

Jiabao-Sun: Thanks, I will add added separate configurations for that.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBConnectorDeserializationSchema.java
dongdongking008: https://docs.mongodb.com/manual/reference/bson-types/#timestamps

Jiabao-Sun: Thanks @wuchong for the detailed explanation.

I added a new config option `local-time-zone` to let users decide which time zone to convert to `TIMESTAMP`.

I think this is semantically more similar to `table.local-time-zone` in Flink SQL rather than `server-time-zone` of MySQL, so I named it as `local-time-zone`.

By the way, is there a method provided by Flink to get the `table.local-time-zone` at runtime, so we may not need to add additional config option.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBTableSource.java
Jiabao-Sun: @wuchong @XuQianJin-Stars

New config options have been added to replace the `uri`.

- `hosts`:  The comma-separated list of hostname and port pairs of the MongoDB servers. 
- `user`:  Database user to be used when connecting to MongoDB.
- `password`: Password to be used when connecting to MongoDB.
- `mongodb.replicaset`: Specifies the name of the replica set. It is not necessary, but can speed up your connection times to explicitly state the servers configured by `hosts`. 
- `mongodb.authsource`: This is required only when MongoDB is configured to use authentication with another authentication database than admin. Defaults to `admin`.
- `mongodb.connect.timeout.ms`: The time in milliseconds to attempt a connection before timing out. Defaults to 10000 (10 seconds).
- `mongodb.socket.timeout.ms`: The time in milliseconds to attempt a send or receive on a socket before the attempt times out. Defaults to 0.
- `mongodb.ssl.enabled`: Connector will use SSL to connect to MongoDB instances.
- `mongodb.ssl.invalid.hostname.allowed`: When SSL is enabled this setting controls whether strict hostname checking is disabled during connection phase.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBConnectorDeserializationSchema.java
wuchong: Yes, you can get it by `org.apache.flink.table.factories.DynamicTableFactory.Context#getConfiguration`.

Jiabao-Sun: Thanks @wuchong.

Removed table option 'local-time-zone', use time zone of `table.local-time-zone` at runtime.

Jiabao-Sun:
@wuchong @XuQianJin-Stars 
The problems mentioned above has been solved. 
If you are available, please help review it again.

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBTableSourceFactory.java
```diff
@@ -0,0 +1,335 @@
...
+
+    private static final ConfigOption<String> COLLECTION =
+            ConfigOptions.key("collection")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Name of the collection in the database to watch for changes.");
+
+    private static final ConfigOption<String> REPLICA_SET =
+            ConfigOptions.key("mongodb.replicaset")
```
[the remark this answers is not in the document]
Jiabao-Sun: Sorry @wuchong, 

I think this new ampersand config option may cause confusion for those MongoDB expert users that they may be more used to use connection strings.

I have two proposals.
1.  Use `mongodb.connect` as config prefix to mark connection options.
```
For example:

CREATE TABLE WITH (
     hosts = 'localhost:27017,localhost:27018',
     username = 'user',
     password = 'pwd',
     mongodb.connect.authsource = 'nonadmin',
     mongodb.connect.replicaset = 'rs',
     mongodb.connect.sslenabled = 'false'
);
```

2.  Provide config option `uri` as an alternative of `hosts`.
```
We can directly use uri with connection parameters and append the `username` and `password` as `userInfo`  to the connection URI.

For example: 

CREATE TABLE WITH (
     uri = 'mongodb://localhost:27017,mongodb://localhost:27018/?authsource=nonadmin&replicaset=rs',
     username = 'user',
     password = 'pwd'
);

It s equivalent to the following but which has no connection option

CREATE TABLE WITH (
     hosts = 'localhost:27017,localhost:27018',
     username = 'user',
     password = 'pwd'
);
```

Do you think so?

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBTableSourceFactory.java
```diff
@@ -0,0 +1,335 @@
...
+    private static final ConfigOption<String> HOSTS =
+            ConfigOptions.key("hosts")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The comma-separated list of hostname and port pairs of the MongoDB servers. "
+                                    + "eg. localhost:27017,localhost:27018");
+
+    private static final ConfigOption<String> USER =
+            ConfigOptions.key("user")
```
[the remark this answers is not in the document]
Jiabao-Sun: Fixed

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/table/MongoDBTableSourceFactory.java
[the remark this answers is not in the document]
Jiabao-Sun: @wuchong 
Yes, It's case insensitive.

As [connection-string-options](https://docs.mongodb.com/manual/reference/connection-string/#connection-string-options) says:

> Connection options are pairs in the following form: name=value.
> The option name is case insensitive when using a driver.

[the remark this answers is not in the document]
Jiabao-Sun: @wuchong @leonardBang @XuQianJin-Stars

The `connection.options` has been added in the latest commit.
If you think `mongodb.connect.` prefix is better, I can make a change for it.

wuchong (approved):
The changes looks good to me.  I will merge it once build is passed. Thanks for the great work and long time patience @Jiabao-Sun .

Jiabao-Sun:
Thanks @wuchong @leonardBang @XuQianJin-Stars @carloscbl @yangxusun9 for your help.

It is a great honor to be able to complete it.
The documentation is ready too.

Cheers!

Issue #226, opened 2021-06-29
issue: java.lang.Integer cannot be cast to java.lang.Long


Jiabao-Sun:
ClassCastException raises when mapping a MYSQL Integer type to Flink SQL BigInt Type.

```
Caused by: java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.convertToLong(RowDataDebeziumDeserializeSchema.java:231) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.lambda$wrapIntoNullableConverter$7b91dc26$1(RowDataDebeziumDeserializeSchema.java:394) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.convertField(RowDataDebeziumDeserializeSchema.java:384) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.lambda$createRowConverter$508c5858$1(RowDataDebeziumDeserializeSchema.java:371) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.lambda$wrapIntoNullableConverter$7b91dc26$1(RowDataDebeziumDeserializeSchema.java:394) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.extractAfterRow(RowDataDebeziumDeserializeSchema.java:127) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.deserialize(RowDataDebeziumDeserializeSchema.java:102) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.internal.DebeziumChangeFetcher.handleBatch(DebeziumChangeFetcher.java:217) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.internal.DebeziumChangeFetcher.runFetchLoop(DebeziumChangeFetcher.java:151) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.DebeziumSourceFunction.run(DebeziumSourceFunction.java:420) ~[classes/:?]
[...]

PR #227, opened 2021-06-29
[ISSUE-226] Fix integer type convert problem

Jiabao-Sun:
[issue:226](#226)

ClassCastException raises when mapping a MYSQL Integer type to Flink SQL BigInt Type.

```
Caused by: java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.convertToLong(RowDataDebeziumDeserializeSchema.java:231) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.lambda$wrapIntoNullableConverter$7b91dc26$1(RowDataDebeziumDeserializeSchema.java:394) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.convertField(RowDataDebeziumDeserializeSchema.java:384) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.lambda$createRowConverter$508c5858$1(RowDataDebeziumDeserializeSchema.java:371) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.lambda$wrapIntoNullableConverter$7b91dc26$1(RowDataDebeziumDeserializeSchema.java:394) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.extractAfterRow(RowDataDebeziumDeserializeSchema.java:127) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.table.RowDataDebeziumDeserializeSchema.deserialize(RowDataDebeziumDeserializeSchema.java:102) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.internal.DebeziumChangeFetcher.handleBatch(DebeziumChangeFetcher.java:217) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.internal.DebeziumChangeFetcher.runFetchLoop(DebeziumChangeFetcher.java:151) ~[classes/:?]
    at com.alibaba.ververica.cdc.debezium.DebeziumSourceFunction.run(DebeziumSourceFunction.java:420) ~[classes/:?]
    at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:110) ~[flink-streaming-java_2.11-1.13.0.jar:1.13.0]
    at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:66) ~[flink-streaming-java_2.11-1.13.0.jar:1.13.0]
    at org.apache.flink.streaming.runtime.tasks.SourceStreamTask$LegacySourceFunctionThread.run(SourceStreamTask.java:269) ~[flink-streaming-java_2.11-1.13.0.jar:1.13.0]
```


flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
```diff
@@ -238,7 +238,7 @@ private long convertToLong(Object dbzObj, Schema schema) {
 
     private double convertToDouble(Object dbzObj, Schema schema) {
         if (dbzObj instanceof Float) {
-            return (double) dbzObj;
+            return ((Float) dbzObj).doubleValue();
```
leonardBang, line 241: could we add a test for this, this is pretty interesting, you may meet precision problem.

leonardBang, line 241: Could we add a test for this?  This is pretty interesting, you may meet precision problem.

flink-connector-mysql-cdc/src/test/resources/ddl/column_type_test.sql
```diff
@@ -25,6 +25,8 @@ CREATE TABLE full_types (
     small_un_c SMALLINT UNSIGNED,
     int_c INTEGER ,
     int_un_c INTEGER UNSIGNED,
+    int11_c INT(11) ,
+    int11_un_c INT(11) UNSIGNED,
```
leonardBang, line 29: One column to check `The type in db is INT, but defined BIGINT in Flink SQL` is enough,
Could you add `The type in db is FLOAT, but defined DOUBLE` in Flink SQL ?

leonardBang (commented):
@Jiabao-Sun Thanks for the contribution,  I left two comments.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
Jiabao-Sun: @leonardBang 

Thanks for the review.
Actually, debezium convert both MySQL `FLOAT` and `DOUBLE` type to `java.util.Double` with the schema `FLOAT64`. 
Please see [Debezium Documentation](https://debezium.io/documentation/reference/connectors/mysql.html).

The difference in precision we see is just because the float type is converted to the double type and It is no problem to restore it to float type again.
```java
float a = 123.102f;
double b = (double) a;

float c = (float) b;
```

So the code in condition one will not be executed.
```java
private double convertToDouble(Object dbzObj, Schema schema) {
    if (dbzObj instanceof Float) { //condition 1
         return (double) dbzObj;
    }
    ...
}
```

But purely semantically speaking, it will cause java.lang.ClassCastException: java.lang.Float cannot be cast to java.lang.Double.
```java
    public static void main(String[] args) {
        Float a = 123.102f;
        convertToDouble(a);
    }

    private static double convertToDouble(Object dbzObj) {
        if (dbzObj instanceof Float) {
            return (double) dbzObj;
        } else if (dbzObj instanceof Double) {
            return (double) dbzObj;
        } else {
            return Double.parseDouble(dbzObj.toString());
        }
    }
```

leonardBang: Thanks @Jiabao-Sun for the detail explanation, the change make sense to me. 
User should use correcting data type mapping so that they may meet precision problem although our code is correct.

flink-connector-mysql-cdc/src/test/resources/ddl/column_type_test.sql
leonardBang: I think only `int11_c INT(11) ,` column is enough?

Jiabao-Sun: Yes, I will remove it.

Jiabao-Sun: > I think only `int11_c INT(11) ,` column is enough?

fixed it.

leonardBang (approved):
LGTM, Thanks @Jiabao-Sun

PR #237, opened 2021-07-07
add `slot.name` to optional arguments

ivanaslamov:
currently it is not possible to specify `slot.name` in `CREATE TABLE () WITH(..., 'slot.name' = ...)`. Trying to do so results in an error: 
```
Caused by: org.apache.flink.table.api.ValidationException: Unsupported options found for 'postgres-cdc'.
[error] Unsupported options:
[error] slot.name
[error] Supported options:
```
this quick change fixes the problem


Jiabao-Sun:
Hi @ivanaslamov 
Try to use 'debezium.slot.name' = 'xxx'.
The cdc connector accepts additional properties  by `debezium` prefix.

Jiabao-Sun:
> Hi @ivanaslamov
> Try to use 'debezium.slot.name' = 'xxx'.
> The cdc connector accepts additional properties by `debezium` prefix.

Sorry, I just saw the source code of PostgreSQLTableFactory which provide `SLOT_NAME` option but not added into optional options. It's indeed a problem.

By the way, we can use 'debezium.slot.name' as a temporary solution.

ivanaslamov:
oh I see, sounds good, sorry I didn't know how `debezium` prefix works.

PR #275, opened 2021-07-28
[mysql] Enable single server id for MySqlParallelSource

leonardBang:
Enable single server id for MySqlParallelSource for compatibility consideration


wuchong (approved):
LGTM.

PR #282, opened 2021-08-01
[debezium] Fix uncompleted state of schema history when engine stop before checkpoint

leonardBang:


flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -353,12 +363,18 @@ private void snapshotHistoryRecordsState() throws Exception {
...
+                        String.format(
+                                "The schema records for engine %s has been removed, checkpoint failed.",
+                                engineInstanceName));
+            } else {
+                schemaRecordsState.add(engineInstanceName);
+                Collection<SchemaRecord> records = StateUtils.retrieveHistory(engineInstanceName);
+                DocumentWriter writer = DocumentWriter.defaultWriter();
+                for (SchemaRecord record : records) {
+                    schemaRecordsState.add(writer.write(record.toDocument()));
+                }
```
wuchong, line 377: I think this is still problematic, because the getting of status and records is not atomic, which means the status is ok but we still get a null records.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -354,11 +364,17 @@ private void snapshotHistoryRecordsState() throws Exception {
...
         if (engineInstanceName != null) {
             schemaRecordsState.add(engineInstanceName);
-
-            Collection<SchemaRecord> records = StateUtils.retrieveHistory(engineInstanceName);
-            DocumentWriter writer = DocumentWriter.defaultWriter();
-            for (SchemaRecord record : records) {
-                schemaRecordsState.add(writer.write(record.toDocument()));
+            Collection<SchemaRecord> records = checkStatusAndRetrieveHistory(engineInstanceName);
+            if (records.isEmpty()) {
+                throw new FlinkRuntimeException(
```
wuchong, line 369: We shoudn't throw exception here, because history might be empty when the job is just started.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/DebeziumSourceFunction.java
```diff
@@ -160,6 +161,13 @@
     private static final ConcurrentMap<String, Collection<SchemaRecord>> HISTORY =
             new ConcurrentHashMap<>();
 
+    /**
+     * The schema history will be clean up once {@link DatabaseHistory#stop()}, the checkpoint
+     * should fail when this happens.
+     */
+    private static final ConcurrentMap<String, Boolean> HISTORY_CLEANUP_STATUS =
```
wuchong, line 168: Could you move `HISTORY` and `HISTORY_CLEANUP_STATUS` and the accessing methods into another class, and just expose the getter/setters? In this way, we can avoid users/devs to access the underlying  `HISTORY` without lock protect.

wuchong (approved):
LGTM.

wuchong:
Please check the failed build.

flink-connector-mysql-cdc/src/test/java/com/alibaba/ververica/cdc/connectors/mysql/MySqlSourceTest.java
```diff
@@ -848,6 +850,75 @@ public void testLoadIllegalState() throws Exception {
...
+                        source.snapshotState(new StateSnapshotContextSynchronousImpl(102, 102));
+                    }
+                    fail("Should fail.");
+                } catch (Exception e) {
+                    assertTrue(e instanceof FlinkRuntimeException);
+                    assertTrue(
+                            e.getMessage()
+                                    .contains(
+                                            String.format(
+                                                    "The schema records for engine %s has been removed, checkpoint failed.",
```
wuchong, line 915: We may need to update this.

flink-connector-debezium/src/main/java/com/alibaba/ververica/cdc/debezium/utils/DatabaseHistoryUtil.java
```diff
@@ -0,0 +1,85 @@
...
+     * Retrieves history of schema safely, this method firstly checks the history status of specific
+     * engine, and then return the history of schema if the history exists(didn't clean up yet).
+     * Returns null when the history of schema has been clean up.
+     */
+    public static Collection<SchemaRecord> retrieveHistory(String engineName) {
+        synchronized (HISTORY) {
+            if (Boolean.TRUE.equals(HISTORY_CLEANUP_STATUS.get(engineName))) {
+                throw new IllegalStateException(
+                        String.format(
+                                "Retrieve schema history failed, the schema records for engine %s has been removed.",
```
wuchong, line 78:
```suggestion
                                "Retrieve schema history failed, the schema records for engine %s has been removed, this might because the debezium engine has been shutdown due to other errors.",
```

PR #335, opened 2021-08-19
add slot.name to supported options of postgres-cdc-connector

dongdongking008:


leonardBang:
Thanks @dongdongking008 for the contribution, the changes looks good, could you add a test in `PostgreSQLTableFactoryTest` ?

dongdongking008:
> Thanks @dongdongking008 for the contribution, the changes looks good, could you add a test in `PostgreSQLTableFactoryTest` ?

Test added, and also added in `MySqlTableSourceFactoryTest`.

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlTableSourceFactoryTest.java
```diff
@@ -190,6 +194,30 @@ public void testEnableParallelReadSourceLatestOffset() {
...
+    @Test
+    public void testOptionalOptions() {
+        Factory factory = new MySqlTableSourceFactory();
+        Set<ConfigOption<?>> optionalOptions = factory.optionalOptions();
+        for (Field field : MySqlSourceOptions.class.getDeclaredFields()) {
+            if (Modifier.isStatic(field.getModifiers()) && field.getType() == ConfigOption.class) {
+                try {
+                    field.setAccessible(true);
+                    ConfigOption<?> option = (ConfigOption<?>) field.get(factory);
+                    if (option.hasDefaultValue()) {
```
leonardBang, line 206: Although the test works well now,  IIUC there's no rule that the optional option must contains a default value, thus this test may be problematic after we add an optional option without default value.
I suggest we can simple add unit test in `testOptionalProperties`, WDYT ?

flink-connector-postgres-cdc/src/test/java/com/ververica/cdc/connectors/postgres/table/PostgreSQLTableFactoryTest.java
```diff
@@ -90,6 +93,30 @@ public void testCommonProperties() {
...
+    @Test
+    public void testOptionalOptions() {
+        Factory factory = new PostgreSQLTableFactory();
+        Set<ConfigOption<?>> optionalOptions = factory.optionalOptions();
+        for (Field field : factory.getClass().getDeclaredFields()) {
+            if (Modifier.isStatic(field.getModifiers()) && field.getType() == ConfigOption.class) {
+                try {
+                    field.setAccessible(true);
+                    ConfigOption<?> option = (ConfigOption<?>) field.get(factory);
+                    if (option.hasDefaultValue()) {
```
leonardBang, line 105: ditto

leonardBang (commented):
Thanks @dongdongking008 for the contribution, I left one comment for tests

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlTableSourceFactoryTest.java
dongdongking008: `there's no rule that the optional option must contains a default value, thus this test may be problematic after we add an optional option without default value.`
That's right, but what about the rule that options have a default value must be optional options?  

The test will just catch options have a default value but not in the `optionalOptions`. 

And I do agree to add unit test for `slot.name` in `testOptionalProperties` for consistent.

leonardBang: >That's right, but what about the rule that options have a default value must be optional options?

I think no this kind of rule in Flink or Flink CDC

>The test will just catch options have a default value but not in the optionalOptions.Yes, I agree that 
Yes, you're right. Thus we cannot check the new add optional option without default value.

leonardBang: >That's right, but what about the rule that options have a default value must be optional options?

I think no this kind of rule in Flink or Flink CDC
>The test will just catch options have a default value but not in the optionalOptions.Yes, I agree that 

Yes, you're right. Thus we cannot check the new add optional option without default value.

leonardBang: Could we delete the test `testOptionalOptions` as we discussed above? @dongdongking008

dongdongking008: ok.

leonardBang (approved):
Thanks @dongdongking008 for the nice work, LGTM
Although the final change is brief,  what we discussed is pretty interesting. `^_^`

leonardBang:
The failed test is not related to this PR, record in #358 
merging...

leonardBang:
@dongdongking008 Would you like to open a PR for release-2.0 branch ?

leonardBang:
I've cherry-picked, plz ignore my last comment

PR #337, opened 2021-08-19
[issue-336] Remove schemas enable parameter when construct DebeziumEngine in DebeziumSourceFunction

lsyldliu:


leonardBang (approved):
Thanks @lsyldliu for the contribution, LGTM.
The `DebeziumEngine` uses `Connect.class` in `DebeziumSourceFunction`  which never use these parameters, I have tested in local environment.

PR #349, opened 2021-08-23
[build] Run integration tests for each building.

leonardBang:


leonardBang:
The CI tests passed, and I manually checked all unit tests and ITCase has been checked

PR #357, opened 2021-08-24
[docs] Add checkpoint and primary key setting for example in tutorials

leonardBang:


leonardBang:
cherry-pick to 2.0 branch

PR #387, opened 2021-09-02
[ISSUE-371] Fix enum type cause DataException

Jiabao-Sun:
This problem happens when serialize and deserialize `MySqlSplit`.
Cause debezium `JsonTableChangeSerializer` ignored `defaultValue` and `enumValues` when serializing `Table.Column`.

Rewritten `JsonTableChangeSerializer` as a compromise and it works fine.

Relate to #371

### DIFF
--- a/flink-connector-debezium/src/main/java/io/debezium/relational/history/JsonTableChangeSerializer.java
+++ b/flink-connector-debezium/src/main/java/io/debezium/relational/history/JsonTableChangeSerializer.java
@@ -1,0 +1,201 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.debezium.relational.history;
+
+import io.debezium.document.Array;
+import io.debezium.document.Array.Entry;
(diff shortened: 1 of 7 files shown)


Jiabao-Sun:
Sorry, tests fixed.

Jiabao-Sun:
More tests is needed to check for side effects.

Jiabao-Sun:
Sorry @leonardBang , this PR will be closed for the reason as follows.

Simply serializing the default value will not solve this issue, it may cause other type conversion problems.
Before debezium fix it, we may need to use other ways to solve this problem.

leonardBang:
@Jiabao-Sun It makes sense to me, thanks for your effort.

luoyuxia:
@Jiabao-Sun Why simply serializing the default value will not solve this issue? I think it's fine. I wonder what's your concern.

Jiabao-Sun:
> @Jiabao-Sun Why simply serializing the default value will not solve this issue? I think it's fine. I wonder what's your concern.

@luoyuxia Hi, you can see this PR for detail.
https://github.com/debezium/debezium/pull/2671

luoyuxia:
@Jiabao-Sun Thanks~

PR #409, opened 2021-09-08
[build] Improve the module scope of maven-shade-plugin execution

ashulin:
Improve the module scope of maven-shade-plugin execution：
Make only the sql module execute shade dependencies.
And fixed the problem that SPI files were not relocated after Kafka relocation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -318,61 +318,6 @@ under the License.
 
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-shade-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>shade-flink</id>
-                        <phase>package</phase>
-                        <goals>
-                            <goal>shade</goal>
-                        </goals>
-                        <configuration>
-                            <shadeTestJar>true</shadeTestJar>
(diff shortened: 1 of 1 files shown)


wuchong:
Hi @ashulin , what do you mean "SPI files were not relocated after Kafka relocation" ?

ashulin:
> Hi @ashulin , what do you mean "SPI files were not relocated after Kafka relocation" ?
Example：

before fixing

```
#META-INF/services/org.apache.kafka.common.config.provider.ConfigProvider


org.apache.kafka.common.config.provider.FileConfigProvider
```

after fixing

```
#META-INF/services/com.ververica.cdc.connectors.shaded.org.apache.kafka.common.config.provider.ConfigProvider


com.ververica.cdc.connectors.shaded.org.apache.kafka.common.config.provider.FileConfigProvider
```

wuchong:
What's the problem if we do not relocate the SPI file?

ashulin:
> What's the problem if we do not relocate the SPI file?

I have not encountered any problems, but logically it is necessary to relocate the SPI file.

lsyldliu:
@ashulin After discuss with @wuchong  offline, we should follow flink project way, minimize dependence . I will open a new pr, tks.

PR #420, opened 2021-09-10
[mongodb] Remove no need config option prefix debezium.

Jiabao-Sun:
Remove no need config option prefix `debezium.`


wuchong (approved):
LGTM.

PR #437, opened 2021-09-17
[tests] Introduce assertEqualsInOrder/assertEqualsInAnyOrder for tests

### LINKED ISSUE #436: [tests] The test BinlogSplitReaderTest#testReadAllBinlogSplitsForOneTable is unstable
https://github.com/ververica/flink-cdc-connectors/runs/3628680617

leonardBang:
Introduce assertEqualsInOrder/assertEqualsInAnyOrder for tests,
This also fixes #436 which we should not use equals to compare the un-ordered result because the snapshot data comes from  different task and may arrives in different order.


leonardBang:
The CI passed, due to this is a test improvements, I'll merge it.

PR #438, opened 2021-09-17
support all mysql data types

luoyuxia:
Support all mysql data type and refine the document of datatype mapping.


luoyuxia:
Now, the ENUM/SET datatype  will cause failure while using MySQL CDC 2.0 shown in #371 .

As #387 shows, it's caused by JsonTableChangeSerializer ignored `defaultValue`, `hasDefaultValue`, `enumValues` of ColumnImpl in Debzium and there has been a PR debezium/debezium#2671 to fix it.

But actually, to simply fix it,  we only need to consider the fields `enumValues`, and it's ok to ignore the field `defaultValue` and `hasDefaultValue`.  So I think we can implement a custom JsonTableChangeSerializer that only serialize and deserialize the field `enumValues` to work around it before the PR debezium/debezium#2671 is merged into Debezium.  

WDYT @leonardBang @Jiabao-Sun

flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
```diff
@@ -223,9 +233,10 @@ public Object convert(Object dbzObj, Schema schema) {
                 return convertToBinary();
             case DECIMAL:
                 return createDecimalConverter((DecimalType) type);
+            case ARRAY:
+                return createArrayConverter((ArrayType) type);
             case ROW:
                 return createRowConverter((RowType) type);
-            case ARRAY:
```
leonardBang, line 228: hnit: no need to move the position

flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
```diff
@@ -413,13 +424,39 @@ public Object convert(Object dbzObj, Schema schema) {
...
+        ObjectMapper objectMapper = new ObjectMapper();
+        ObjectWriter objectWriter = objectMapper.writer();
         return new DeserializationRuntimeConverter() {
 
             private static final long serialVersionUID = 1L;
 
             @Override
             public Object convert(Object dbzObj, Schema schema) {
-                return StringData.fromString(dbzObj.toString());
+                // if the dbzObj is instance of Struct, assume it represents GEOMETRY
```
leonardBang, line 435: The assumption is weak, could we check it more strictly?

docs/content/connectors/mysql-cdc.md
```diff
@@ -463,36 +475,119 @@ Data Type Mapping
...
+**Note:**
+- The precision for DECIMAL data type is up to 65 in MySQL, but the precision for DECIMAL is limited to 38 in Flink.
+  So if you define a decimal column whose precision is greater than 38, you should map it to STRING to avoid precision loss.
+- The JSON data type in MySQL will be converted into STRING with JSON format in Flink.
+- Currently, for BLOB data type in MySQL, only the blob whose length isn't greater than 2,147,483,647(2 ** 31 - 1) is supported. 
+- The spatial data types in MySQL will be converted into STRING with a fixed format like:<br>
+  ```json
+  {"srid": 0, "type": "POINT", "coordinates": [1.0,1.0]}
+  ```
+
```
leonardBang, line 590: We can add a `note` column in above type mapping table, and then explain the note.

docs/content/connectors/mysql-cdc.md
```diff
@@ -463,36 +475,119 @@ Data Type Mapping
...
+        VARCHAR(n)
+      </td>
+      <td>VARCHAR(n)</td>
+    </tr>
+    <tr>
+      <td>
+        BIT(n)<br>
+        where n > 1
+      </td>
+      <td>BINARY(⌊n/8⌋)</td>
```
leonardBang, line 506: `⌊n/8⌋` is a special math signal ?

docs/content/connectors/mysql-cdc.md
leonardBang: The type mapping table is pretty long which is more than one page in the web site, and the mappings is not very clear. Could we organize it better?

flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
```diff
@@ -480,6 +517,39 @@ public Object convert(Object dbzObj, Schema schema) {
...
+            if (dbzObj instanceof List) {
+                List<?> elements = (List<?>) dbzObj;
+                final Object[] elementArray =
+                        (Object[]) Array.newInstance(elementClass, elements.size());
+                for (int i = 0; i < elements.size(); i++) {
+                    elementArray[i] = elementConverter.convert(elements.get(i), schema);
+                }
+                return new GenericArrayData(elementArray);
+            } else if (dbzObj instanceof String) {
+                // for SET datatype in mysql, debezium will always return a string split by comma
```
leonardBang, line 535: Does debezium object contain any type information that can use to infer the object data is SET?

flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
```diff
@@ -480,6 +517,39 @@ public Object convert(Object dbzObj, Schema schema) {
...
+                final Object[] elementArray =
+                        (Object[]) Array.newInstance(elementClass, elements.length);
+                for (int i = 0; i < elements.length; i++) {
+                    elementArray[i] = elementConverter.convert(elements[i], schema);
+                }
+                return new GenericArrayData(elementArray);
+            } else {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Unable to convert to arrayType from unexpected value '%s', only List and String type can be converted to arrayType",
```
leonardBang, line 547:
```suggestion
                                "Unable convert to Flink ARRAY type from unexpected value '%s', only MySQL LIST and SET type could convert to ARRAY type",
```

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/table/MySqlTableSource.java
```diff
@@ -165,6 +165,7 @@ private Configuration getParallelSourceConf() {
         properties.put("database.port", String.valueOf(port));
         properties.put("database.history.skip.unparseable.ddl", String.valueOf(true));
         properties.put("database.server.name", DATABASE_SERVER_NAME);
+        properties.put("bigint.unsigned.handling.mode", "precise");
```
leonardBang, line 168: add a note?

pom.xml
```diff
@@ -66,6 +66,7 @@ under the License.
     <properties>
         <flink.version>1.13.1</flink.version>
         <debezium.version>1.5.2.Final</debezium.version>
+        <geometry.version>2.0.0</geometry.version>
```
leonardBang, line 69:
```suggestion
        <geometry.version>2.2.0</geometry.version>
```
`flink-table-planner` also contains this dependency,  could we use same version ?

docs/content/connectors/mysql-cdc.md
luoyuxia: yes.
But I use the wrong math signal, we should use ⌈n/8⌉, which represents ceil(n/8)

flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
luoyuxia: There's no LIST type in MySQL.
For MySQL SET type, debezium will always return a string. 
And I also consider List type in here is  to make it can be used in PG.

luoyuxia:
@leonardBang

Jiabao-Sun:
@luoyuxia

I think it's a nice enhancement to serialize enumsValues first.
It can solve the problem that the currently known Enum type is not supported.

Currently, I have not tested the type unsupported problem caused by the missing defaultValue.
If there is indeed a problem caused by the missing defaultValue, we can also consider using serializing tools like `Kryo` to serialize `defaultValue` before [debezium-2671] is released. The only concern with custom serializers is that we may not cover all possible types.

I think we can accept this PR first. @leonardBang What do you think?

PR #458, opened 2021-09-23
[docs] Update the dependency version to 2.1.0 in documentation

leonardBang:
Update the dependency version to 2.1.0


leonardBang:
The change only contains documents, merge it after the compile success

PR #459, opened 2021-09-23
[docs] Update the dependency version to 2.0.2 in documentation

leonardBang:


leonardBang:
The change only contains documents, merge it after the compile success

PR #475, opened 2021-09-29
Add support for copy.existing.namespace.regex and pipeline config to mongo-cdc

bobby-richard:
Problem: I need to capture data from multiple tenant databases in the same mongodb cluster

The connector already allowed you to capture data from multiple databases/collections, but this was not very useful without the ability to filter. The copy.existing.namespace.regex setting already exists in the kafka source connector and does exactly this. So this PR just exposes the existing setting to the flink connector.

### DIFF
--- a/docs/content/connectors/mongodb-cdc.md
+++ b/docs/content/connectors/mongodb-cdc.md
@@ -152,14 +152,14 @@ Connector Options
       <td>required</td>
       <td style="word-wrap: break-word;">(none)</td>
       <td>String</td>
-      <td>Name of the database to watch for changes.</td>
+      <td>Name of the database to watch for changes. If set to empty string, then all databases will be captured (databases can be filtered using the copy.existing.namespace.regex setting).</td>
     </tr>
     <tr>
       <td>collection</td>
       <td>required</td>
       <td style="word-wrap: break-word;">(none)</td>
       <td>String</td>
-      <td>Name of the collection in the database to watch for changes.</td>
+      <td>Name of the collection in the database to watch for changes. If set to empty string, then all collections will be caputred (collections can be filtered using the copy.existing.namespace.regex setting)</td>
     </tr>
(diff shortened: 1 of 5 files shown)


leonardBang:
Thanks @bobby-richard for the contribution, @Jiabao-Sun Could you help review this PR?

Jiabao-Sun:
@leonardBang OK

Jiabao-Sun:
Thanks @bobby-richard, It's a nice improvement. This setting was not exposed before because multiple databases and collections are complicated for users.

The `copy.existing.namespace.regex` setting to provide a regular expression that matches specific collections by their namespace, but there is no subscription for these collection changes. 

If we need to improve the support of multiple databases and collections, I think we should also expose `pipeline` configuration for `MongoDBTableSource` according to the [multiple-source-example](https://docs.mongodb.com/kafka-connector/current/kafka-source/#multiple-source-example).

https://github.com/ververica/flink-cdc-connectors/blob/e19b78691ece0af61db42800fb97be64242a77e9/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/MongoDBSource.java#L181-L188

bobby-richard:
Sounds good @Jiabao-Sun, I will expose `pipeline` as well.

bobby-richard-sed:
@Jiabao-Sun Exposed the pipeline configuration as well. Nice catch!

bobby-richard:
@Jiabao-Sun Exposed pipeline configuration as well. Nice catch!

Jiabao-Sun:
> @leonardBang @Jiabao-Sun any updates?

Thanks @bobby-richard, it's a good enhencement that we can subscribe and filter multiple databases and collections.

@wuchong @leonardBang Do you have any suggestions?

Jiabao-Sun:
closed #940

PR #480, opened 2021-10-03
[docs] fix MySQLSource typo

mincwang:
When cdc version more than 2.0, MySQLSource should replace to MySqlSource. This is a bug left over from the last PR. (https://github.com/ververica/flink-cdc-connectors/pull/440)


leonardBang:
Thanks @mincwang for the contribution, LGTM,  I update a little as well

PR #496, opened 2021-10-11
Support metadata columns for mysql-cdc connector

wuchong:
We introduced following metadata columns in this PR:

| Key | DataType | Description |
| --- | --- | --- |
| database_name | STRING NOT NULL |  Name of the database that contain the row. |
| table_name | STRING NOT NULL |  <div>Name of the table that contain the row.</div> |
| op_ts | TIMESTAMP_LTZ(3) NOT NULL | It indicates the time that the change was made in the database. If the record is read from snapshot of the table instead of the binlog, the value is always 0. |


leonardBang:
The tests failed

ashulin:
I did a review of your code, and I think your implementation is better.

PR #499, opened 2021-10-12
[build] Refactor maven-shade-plugin style follow flink standard

lsyldliu:


lsyldliu:
image
As you see, it includes the factory service file, also add  the service transfomers in the parent shade plugin?

PR #512, opened 2021-10-20
[build] Add necessary shaded dependencies

Jiabao-Sun:
Add necessary shaded dependencies


lsyldliu:
> > IMO, we just need include `com.fasterxml.jackson.core:*` in three module, the job can work well.
> 
> `com.fasterxml.jackson.datatype:jackson-datatype-jsr310` for JSR-310 Date and Time serialization.
> 
> `com.google.guava:guava` was dependent by
> 
> * `io.debezium.connector.mysql.GtidSet`
> * `io.debezium.util.ObjectSizeCalculator`
> * `io.debezium.connector.base.ChangeEventQueue`

Ok, tks

PR #521, opened 2021-10-25
[mysql] if colunm defined default value and nullable,when column value is null. flink-cdc catch field value is default

GOODBOY008:
fix issue:https://github.com/ververica/flink-cdc-connectors/issues/516

### DIFF
--- a/flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
+++ b/flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
@@ -574,7 +574,10 @@ public final class RowDataDebeziumDeserializeSchema
                     if (field == null) {
                         row.setField(i, null);
                     } else {
-                        Object fieldValue = struct.get(field);
+                        Object fieldValue =
+                                field.schema().isOptional()
+                                        ? struct.getWithoutDefault(fieldName)
+                                        : struct.get(field);
                         Schema fieldSchema = schema.field(fieldName).schema();
                         Object convertedField =
                                 convertField(fieldConverters[i], fieldValue, fieldSchema);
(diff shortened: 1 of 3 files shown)


flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
```diff
@@ -574,7 +574,10 @@ public Object convert(Object dbzObj, Schema schema) throws Exception {
                     if (field == null) {
                         row.setField(i, null);
                     } else {
-                        Object fieldValue = struct.get(field);
+                        Object fieldValue =
+                                field.schema().isOptional()
+                                        ? struct.getWithoutDefault(fieldName)
+                                        : struct.get(field);
```
leonardBang, line 580: Should we directly use `Object fieldValue = struct.getWithoutDefault(fieldName);`, the CDC process should never change the original data,  but only capture and transfer it. The assignment of Default value should happen in DB server side.

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
```diff
@@ -1049,6 +1052,74 @@ public void testStartupFromTimestamp() throws Exception {
...
+                    "+I[rocks, box of assorted rocks]",
+                    "+I[jacket, water resistent black wind breaker]",
+                    "+I[spare tire, 24 inch spare tire]"
+                };
+
+        List<String> actual = TestValuesTableFactory.getResults("sink");
+        assertEqualsInAnyOrder(Arrays.asList(expected), actual);
+        result.getJobClient().get().cancel().get();
+    }
     // ------------------------------------------------------------------------------------
```
leonardBang, line 1123: Introduce an IT test is a little heavy, we can add a column named `nullable_col_with_default_value VARCHAR(255) DEFAULT 'flink',` in existed table `column_type_test.fulltypes` and cover the change by existed tests like `XXXConnectorITCase#testAllTypes`.

The above change will affect all connectors who use `RowDataDebeziumDeserializeSchema`, right? 
 In this way, we can cover all connectors well, and we can also reuse `column_type_test.sql` instead of introducing `column_default_value.sql` file.

leonardBang (commented):
Thanks @GOODBOY008 for the contribution, I left some comments

flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/table/RowDataDebeziumDeserializeSchema.java
GOODBOY008: I suddenly understood your means. The logic of my code looks stupid.haha....

PR #541, opened 2021-10-29
[tests] Reduce the total test time

wuchong:
The Azure test time is reaching to 1hour, the pull request is aiming to reduce the test time to <40min.


pom.xml
```diff
@@ -69,7 +69,7 @@ under the License.
         <flink.version>1.13.1</flink.version>
         <debezium.version>1.5.2.Final</debezium.version>
         <geometry.version>2.2.0</geometry.version>
-        <testcontainers.version>1.15.1</testcontainers.version>
+        <testcontainers.version>1.16.2</testcontainers.version>
```
leonardBang, line 72: The oracle tests failed, IIRC, the oracle Container is build image on the flight, but `1.16.2` doesn't offer such ability or we need make some change.

PR #543, opened 2021-11-01
[mysql] polish binlog offset compare semantic

zhjwpku:
the `isAtOrBefore & isBefore` was actually meaning
`isAtOrAfter & isAfter`, so let's do this right ;)

Signed-off-by: Zhao Junwang <zhjwpku@gmail.com>

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/offset/BinlogOffset.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/offset/BinlogOffset.java
@@ -227,10 +227,18 @@ public class BinlogOffset implements Comparable<BinlogOffset>, Serializable {
     }
 
     public boolean isAtOrBefore(BinlogOffset that) {
-        return this.compareTo(that) >= 0;
+        return this.compareTo(that) <= 0;
     }
 
     public boolean isBefore(BinlogOffset that) {
+        return this.compareTo(that) < 0;
+    }
+
+    public boolean isAtOrAfter(BinlogOffset that) {
+        return this.compareTo(that) >= 0;
+    }
+
+    public boolean isAfter(BinlogOffset that) {
         return this.compareTo(that) > 0;
     }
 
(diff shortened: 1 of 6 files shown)


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/offset/BinlogOffset.java
```diff
@@ -227,10 +227,18 @@ public int compareTo(BinlogOffset that) {
     }
 
     public boolean isAtOrBefore(BinlogOffset that) {
```
leonardBang, line 229: We can rename this method to `isAtOrAfter` in IDE to avoid miss as well as renaming `isBefore` to is `is`  After

leonardBang (commented):
Thanks @zhjwpku for the contribution, I left one comment

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/offset/BinlogOffset.java
zhjwpku: I could have done that, but I think left `isAtOrBefore` and `isBefore` as where it was is more flexible, since at some circumstances, `isBefore` is a more natural way to express, e.g.

```
for (MySqlSnapshotSplit split : assignedSnapshotSplit) {
  // find the min binlog offset
  BinlogOffset binlogOffset = splitFinishedOffsets.get(split.splitId());
  if (minBinlogOffset == null || binlogOffset.isBefore(minBinlogOffset)) {
      minBinlogOffset = binlogOffset;
  }
}
```

leonardBang: The method `isAtOrBefore` and `isBefore` is welcome, what I suggested  is only rename method firstly and then add introduce methods, in this way, we can avoid misuse.

zhjwpku: I kind of know why this patch make you worried 😟 , I use IntellJ to find usage of `isAtOrBefore` and `isBefore` and replace them one by one, and during the replacement I changed some logic, eg:

image

I thought HighWatermark itself should not be emitted so `isAfter` instead of `isAtOrAfter` here.

And this:

image

If  the highWatermark equals maxHighWatermark, tableIdBinlogPositionMap doesn't need to change, so again `isAfter` intead of `isAtOrAfter`.

I think these are what worries you the most, and I really need you to review these logic changes to make sure they are in good shape, thanks 😊

leonardBang: Yes, you got it. All your changes looks good to me, I just suggest a safe way for your changes.
I think you can even split your changes in two commits. The first one aims to fix existed method name typo, the second one contains several improvements.

zhjwpku: Yeah, greate idea 💡

PR #563, opened 2021-11-04
[mysql] polish binlog offset compare semantic #543

leonardBang:


leonardBang:
The PR used to build CI for #543

PR #564, opened 2021-11-04
[tests] Upgrade testcontainers dependency to 1.16.2

MartijnVisser:
* Upgrade the testcontainers dependency from 1.15.1 to 1.16.2. This newer version should be quicker and also includes support for Oracle XE

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@ under the License.
         <flink.version>1.13.1</flink.version>
         <debezium.version>1.5.4.Final</debezium.version>
         <geometry.version>2.2.0</geometry.version>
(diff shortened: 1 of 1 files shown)


leonardBang:
Thanks @MartijnVisser for the contribution, but this may lead to error that oracle container is building image on-flight which the newer version doesn't support IIRC.

MartijnVisser:
@leonardBang I believe that with the newer version it can pull existing containers, like explained on https://www.testcontainers.org/modules/databases/oraclexe/

I'll try it locally first :)

wuchong:
Thanks @MartijnVisser , I continued the work in #591

MartijnVisser:
Thanks for that @wuchong !

PR #569, opened 2021-11-05
[docs] Fix oracle DATE type mapping

ashulin:
image
Oracle's DATE type is actually a datetime type, which should be mapped to flink's TIMESTAMP type.


leonardBang:
Thanks @ashulin for the contribution, LGTM, I left a picture here for tracking
image

leonardBang (approved):

Issue #585, opened 2021-11-09
MongoDB cdc配置heartbeat.interval.ms后会报错


liuxiaolinlxl:
想从savepoint恢复任务，恢复的时候报错（停止任务时间3小时左右，相应oplog存在）：ResumeToken not found. Cannot create a change stream cursor
查文档说：Note: heartbeat.interval.ms is highly recommended to set a proper value larger than 0 if the collection changes slowly. The heartbeat event can pushing the resumeToken forward to avoid resumeToken being expired when we recover the Flink job from a checkpoint or savepoint.
然后将heartbeat.interval.ms设为10000，启动就报错：Caused by: java.lang.NullPointerException: null
	at com.ververica.cdc.connectors.mongodb.internal.MongoDBConnectorSourceTask.markRecordTimestamp(MongoDBConnectorSourceTask.java:161)
	at com.ververica.cdc.connectors.mongodb.internal.MongoDBConnectorSourceTask.poll(MongoDBConnectorSourceTask.java:107)
	at io.debezium.embedded.EmbeddedEngine.run(EmbeddedEngine.java:788)
	at io.debezium.embedded.ConvertingEngineBuilder$2.run(ConvertingEngineBuilder.java:188)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)

请问怎么解决，或者设置多少合适才能从savepoint恢复任务。

Issue #579, opened 2021-11-08
mongodb-cdc加了心跳就报空指针错


aokinba:
**Describe the bug**
mongodb-cdc加了心跳就报空指针错

**Environment :**
 - Flink version :  1.13.1
 - Flink CDC version: mongodb-cdc，最新master
 - Database and version: mongodb

![1636341697563_13E917EC-2CAC-46a3-AEE5-DA9C30319FB6](https://user-images.githubusercontent.com/6531769/140679512-1e16afe2-10d0-4f6e-85b7-66b9197d2ecb.png)
![1636341697589_5F151C52-FEA7-493a-A814-09244856402A](https://user-images.githubusercontent.com/6531769/140679523-20de375f-6ec4-4756-a185-86a6529014d7.png)

Jiabao-Sun:
Thanks @aokinba to report this problem. I'll fix it.

PR #586, opened 2021-11-09
[hotfix][mongodb] Fix NullPointException when heartbeat is enabled

Jiabao-Sun:
Fix NullPointException when heartbeat is enabled.

#579 
#585


flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/internal/MongoDBConnectorSourceTask.java
```diff
@@ -102,7 +108,13 @@ public void commitRecord(SourceRecord record, RecordMetadata metadata)
             if (sourceRecords != null && !sourceRecords.isEmpty()) {
                 outSourceRecords = new LinkedList<>();
                 for (SourceRecord current : sourceRecords) {
+                    if (isHeartbeatRecord(current)) {
+                        outSourceRecords.add(markHeartbeatRecord(current));
+                        continue;
+                    }
+
                     markRecordTimestamp(current);
```
leonardBang, line 116: how about rename method name to `markTimestampForHeartbeatRecord` and  `markTimestampForDataRecord` ?
It's better If we can abstract a new method `markTimestampForRecord` and do all mark works in that method will make the code more readable.

leonardBang:
The test failed

leonardBang (approved):
Thanks @Jiabao-Sun for the quick fix, LGTM

PR #588, opened 2021-11-09
[hotfix][mongodb] Fix problem of cannot look up fields on non-struct type

Jiabao-Sun:
Fix problem of cannot look up fields on non-struct type


Jiabao-Sun:
@leonardBang, If you are available, please help review it.

wuchong (approved):
Thanks @Jiabao-Sun for the fixing. I also encountered this problem when adding mongodb e2e tests. And this PR does fix the problem. However, it would be better to add a test in the PR which can reproduce the problem when fixing a bug.

PR #592, opened 2021-11-10
[mysql]: fix the normalizedSplitRecords logic

zhjwpku:
When nomarlize the snapshotRecords and binlogRecords, we should
remove from snapshotRecords by binlogRecords.keys().

Short version reason: it's the rule from DBLog paper

Long version reason. TLDR;

Let't see the followling illustration:

```
    snapshot readview(sr)
            |
            v
------lw---------hw-----    <- binlog stream
```

For RR isolation level, the readview is a point corresponding to
the binlog stream, let's name it sr.

1. For an update binlog(ub):

If lw <= ub <= sr, the normalizedBinlogRecord first add the
new value(i.e the updateAfter), and then add the snapshotRecords,
so there will be two record for this key, both contain the new value,
there is no problem since the downstream use the upsert logic.

But if sr < ub <= hw, the snapshotRecords contains the old
value, and the normalizedBinlogRecord first add the new value,
and then add the old value from the snapshotRecords, now there
is problem, since the upsert logic will apply the old value
to the downstream, data inconsistency ;(

2. For an delete binlog(db)

If lw <= db <= sr, the snapshotRecords does not contain the
corresponding key, and if sr < db <= hw, the snapshotRecords
contains the key, so the logic `if contain then delete` is
just ok, no need to throw Exception if snapshotRecords does
not contains the key.

3. For an create binlog(cb)

If lw <= db <= sr, there will be two create binlog in the
normalizedBinlogRecords with the same value, no problem
since downstream apply the upsert logic.

So, a simple solution to update and delete problem is to
delete the corresponding key from snapshotRecords.
For create, delete from snapshotRecords will eliminate
the duplication.

For RC isolation level, delete from snapshotRecords first
will also work, I will leave the analysis for those who
will see this patch and interested.

[v2]

After some discussion with @wuchong , the upsert logic should be the
following logic:

0. add `lowWatermarkEvent` into `normalizedBinlogRecords`
1. iterate all the records in `binlogRecords`
    1.0. for DELETE records, we get the key and `Map.remove` it from `snapshotRecords`
    1.1. for all other kind records, we get the key and `Map.put` it into `snapshotRecords` (we still need to decorate the UPDATE event into READ event).
2. add `snapshotRecords.values()` into `normalizedBinlogRecords`
3. add `highWatermarkEvent` into `normalizedBinlogRecords`
4. return `normalizedBinlogRecords`

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/RecordUtils.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/RecordUtils.java
@@ -153,9 +153,6 @@ public class RecordUtils {
                     Envelope.Operation operation =
                             Envelope.Operation.forCode(
                                     value.getString(Envelope.FieldName.OPERATION));
-                    if (snapshotRecords.containsKey(key)) {
-                        snapshotRecords.remove(key);
-                    }
                     switch (operation) {
                         case UPDATE:
                             Envelope envelope = Envelope.fromSchema(binlog.valueSchema());
@@ -174,13 +171,13 @@ public class RecordUtils {
                                             binlog.key(),
                                             binlog.valueSchema(),
                                             envelope.read(updateAfter, source, ts));
-                            normalizedBinlogRecords.add(record);
+                            snapshotRecords.put(key, record);
                             break;
                         case DELETE:
-                            // The deleting from snapshotRecords has been promoted to outer scope.
+                            snapshotRecords.remove(key);
                             break;
                         case CREATE:
-                            normalizedBinlogRecords.add(binlog);
+                            snapshotRecords.put(key, binlog);
                             break;
                         case READ:
                             throw new IllegalStateException(
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/RecordUtils.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/RecordUtils.java
@@ -153,6 +153,9 @@ public class RecordUtils {
                     Envelope.Operation operation =
                             Envelope.Operation.forCode(
                                     value.getString(Envelope.FieldName.OPERATION));
+                    if (snapshotRecords.containsKey(key)) {
+                        snapshotRecords.remove(key);
+                    }
                     switch (operation) {
                         case UPDATE:
                             Envelope envelope = Envelope.fromSchema(binlog.valueSchema());
@@ -174,14 +177,7 @@ public class RecordUtils {
                             normalizedBinlogRecords.add(record);
                             break;
                         case DELETE:
-                            if (snapshotRecords.containsKey(key)) {
-                                snapshotRecords.remove(key);
-                            } else {
-                                throw new IllegalStateException(
-                                        String.format(
-                                                "The delete record %s doesn't exists in table split %s.",
-                                                binlog, split));
-                            }
+                            // The deleting from snapshotRecords has been promoted to outer scope.
                             break;
                         case CREATE:
                             normalizedBinlogRecords.add(binlog);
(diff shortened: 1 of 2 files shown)


zhjwpku:
How to reproduce the problem:

Set breakpoint in the following positions:

[1]: https://github.com/ververica/flink-cdc-connectors/blob/release-2.0/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java#L149
[2]: https://github.com/ververica/flink-cdc-connectors/blob/release-2.0/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java#L152
[3]: https://github.com/ververica/flink-cdc-connectors/blob/release-2.0/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/RecordUtils.java#L144

When the debug program hit [1], do some stuff on the source mysql instance, and resume the process, when it hit [2], do some stuff on the source mysql instance, resume the program, step and observe the normalize logic from [3] on.

zhjwpku:
@leonardBang can you please help to review this ? 🙃

wuchong:
Thanks @zhjwpku for opening the discussion. 

First of all, Flink CDC incremental snapshot algorithm refers the **idea** of the DBLog paper.
DBLog logic is not compatible with Flink changelog mechanism, so our algorithm is a variant.

I would like to correct 2 important points in your miss-understanding:

1. all the binlog between lw to hw will be upserted to the snapshot records by primary key, not only the binlog of `lw->sr` or `sr->hw`. 
2. Flink CDC will buffer the snapshot records in memory, and then upsert the binlogs (`lw->hw`) to the buffer by PK, and then output the records in buffer. Flink CDC reads snapshot records in chunk level (default 8K rows), so there is no OOM risk. 

Hope that can solve your concerns.

zhjwpku:
> Thanks @zhjwpku for opening the discussion.
> 
> First of all, Flink CDC incremental snapshot algorithm refers the **idea** of the DBLog paper. DBLog logic is not compatible with Flink changelog mechanism, so our algorithm is a variant.
> 
> I would like to correct 2 important points in your miss-understanding:
> 
> 1. all the binlog between lw to hw will be upserted to the snapshot records by primary key, not only the binlog of `lw->sr` or `sr->hw`.
> 2. Flink CDC will buffer the snapshot records in memory, and then upsert the binlogs (`lw->hw`) to the buffer by PK, and then output the records in buffer. Flink CDC reads snapshot records in chunk level (default 8K rows), so there is no OOM risk.
> 
> Hope that can solve your concerns.

Thanks for your explaination, I will try to understand the flink changelog mechanism ;)

wuchong (commented):
Thanks @zhjwpku ,  I think the root cause of the problem is that we didn't follow a true upsert way (`Map.put` & `Map.remove`) to merge snapshot records and binlog records. Currently, we add records into a List which can't provide unique semantic on primary key, and this is very error-prone.

Could you help to refactor the code of `upsertBinlog` to the following logic?
0. add `lowWatermarkEvent` into `normalizedBinlogRecords`.
1. iterate all the records in `binlogRecords`. 
2. for DELETE records, we get the key and `Map.remove` it from `snapshotRecords`. 
3. for all other kind records, we get the key and `Map.put` it into `snapshotRecords` (we still need to decorate the UPDATE event into READ event). 
4. add `snapshotRecords.values()` into `normalizedBinlogRecords`
5. add `highWatermarkEvent` into `normalizedBinlogRecords`.
6. return `normalizedBinlogRecords`

This is just like how the current upsert sink works, and avoid frequently `Map.containsKey`, `Map.remove` invokes.

wuchong:
@zhjwpku I added a test to reproduce this problem to make sure the bug is fixed and won't break in the future. Could you pick the commit (https://github.com/wuchong/flink-cdc-connectors/tree/pull/592) to your branch?

cc @leonardBang , could you help to review it as well?

wuchong (approved):
LGTM

PR #594, opened 2021-11-10
[tests] Add end-to-end tests for cdc connector uber jars

wuchong:
We introduced flink-sql-connectors-xxx for all mysql-cdc, postgres-cdc and mongodb-cdc. This module is used to shade kafka and build an uber jar which is mainly used for SQL clusters. However, we don't have any tests for the uber jars, e.g. does the shading work? do we exclude any needed classes? This can be covered by e2e tests.


wuchong:
This pull request introduced a new module `flink-cdc-e2e-tests` for end-to-end testing purpose. It startup a flink cluster using docker containers, a mysql docker container, and submit SQL job using the compiled mysql-cdc uber jars. This could verify either the uber jar contains all the needed dependencies or class conflicts. 

Currently, this pull request only add e2e tests for mysql-cdc. I will add tests for other cdc connectors later.

wuchong:
The e2e can verify missing class problem of uber jar. See the artifact logs in the build: 

https://artprodsea1.artifacts.visualstudio.com/Ad47aad79-177f-4b1c-917d-79c11fccad9f/dac773cf-e9ac-4cdf-834a-b5afa51a0732/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3ZlcnZlcmljYS9wcm9qZWN0SWQvZGFjNzczY2YtZTlhYy00Y2RmLTgzNGEtYjVhZmE1MWEwNzMyL2J1aWxkSWQvNjkyL2FydGlmYWN0TmFtZS9sb2dzLWNpX2J1aWxkLXRlc3RfY2lfYnVpbGRfZTJlLTE2MzY1NDMxOTk1/content?format=zip

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/utils/JdbcProxy.java
```diff
@@ -0,0 +1,92 @@
...
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/** Proxy to communicate with database using JDBC protocal. */
```
leonardBang, line 33: typo: `protocol`

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/utils/JdbcProxy.java
```diff
@@ -0,0 +1,92 @@
...
+                checkResult(expectedResult, table, fields);
+                result = true;
+                break;
+            } catch (AssertionError | SQLException throwable) {
+                Thread.sleep(1000L);
+            }
+        }
+        if (!result) {
+            checkResult(expectedResult, table, fields);
+        }
```
leonardBang, line 90: A check with timeout exception is better than a final check

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/utils/ParameterProperty.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Retrieves the value of this property.
+     *
+     * @return Optional containing the value of this property
+     */
+    public Optional<V> get() {
+        final String value = System.getProperty(propertyName);
+        return value == null ? Optional.empty() : Optional.of(converter.apply(value));
+    }
```
leonardBang, line 47: looks like we never used these methods, could we delete them?

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/utils/SQLJobSubmission.java
```diff
@@ -0,0 +1,76 @@
...
+            for (Path jarFile : jarFiles) {
+                addJar(jarFile);
+            }
+            return this;
+        }
+
+        public SQLJobSubmissionBuilder addJars(List<Path> jarFiles) {
+            jarFiles.forEach(this::addJar);
+            return this;
+        }
```
leonardBang, line 70: we can safely delete this method, it's duplicate with above?

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/utils/TestUtils.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Copy all the files and sub-directories under source directory to destination directory
+     * recursively.
+     *
+     * @param source directory or file path to copy from.
+     * @param destination directory or file path to copy to.
+     * @return Path of the destination directory.
+     * @throws IOException if any IO error happen.
+     */
+    public static Path copyDirectory(final Path source, final Path destination) throws IOException {
```
leonardBang, line 100: diito

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/MySqlE2eITCase.java
```diff
@@ -0,0 +1,141 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+
+/** End-to-end tests for mysql-cdc connector uber jar. */
+public class MySqlE2eITCase extends FlinkContainerTestEnvironment {
```
leonardBang, line 38: I think end to end test is different with integration test, how about rename it to `MySqlEndToEndTest` ?
and we use `tests` package prefix in Flink, here we use `e2e`, do we need to keep the same style?

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/MySqlE2eITCase.java
```diff
@@ -0,0 +1,141 @@
...
+                        " primary key (`id`) not enforced",
+                        ") WITH (",
+                        " 'connector' = 'mysql-cdc',",
+                        " 'hostname' = '" + INTER_CONTAINER_MYSQL_ALIAS + "',",
+                        " 'port' = '3306',",
+                        " 'username' = '" + MYSQL_TEST_USER + "',",
+                        " 'password' = '" + MYSQL_TEST_PASSWORD + "',",
+                        " 'database-name' = '" + mysqlInventoryDatabase.getDatabaseName() + "',",
+                        " 'table-name' = 'products_source',",
+                        " 'server-id' = '5800-5900',",
```
leonardBang, line 66: the `server-id` range is to large here

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/OracleE2eITCase.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/** End-to-end tests for oracle-cdc connector uber jar. */
+public class OracleE2eITCase extends FlinkContainerTestEnvironment {
+
+    private static final Logger LOG = LoggerFactory.getLogger(OracleE2eITCase.class);
+    private static final String ORACLE_TEST_USER = "dbzuser";
+    private static final String ORACLE_TEST_PASSWORD = "dbz";
+    protected static final String ORACLE_DRIVER_CLASS = "oracle.jdbc.driver.OracleDriver";
```
leonardBang, line 46: `private` is enough

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/PostgresE2eITCase.java
```diff
@@ -0,0 +1,215 @@
...
+                        " 'connector' = 'postgres-cdc',",
+                        " 'hostname' = '" + INTER_CONTAINER_PG_ALIAS + "',",
+                        " 'port' = '" + POSTGRESQL_PORT + "',",
+                        " 'username' = '" + PG_TEST_USER + "',",
+                        " 'password' = '" + PG_TEST_PASSWORD + "',",
+                        " 'database-name' = '" + POSTGRES.getDatabaseName() + "',",
+                        " 'schema-name' = 'inventory',",
+                        " 'table-name' = 'products',",
+                        // dropping the slot allows WAL segments to be discarded by the database
+                        " 'debezium.slot.drop_on_stop' = 'true'",
```
leonardBang, line 103: we can give a certain `slot.name`

leonardBang (commented):
Thanks @wuchong for the great work, the PR looks good totally, I only left some minor comments

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/MySqlE2eITCase.java
wuchong: There is only unit-test and integration-test phase in `maven-surefire-plugin`, no end-to-end notion in it. We matches `*Test.java` suffix as unit test, and `*ITCase.java` as integration-test. So we have to name it ITCase.

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/MySqlE2eITCase.java
wuchong: Is there any problem to use a large range?

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/PostgresE2eITCase.java
wuchong: There is a default `slot.name=flink`.

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/utils/ParameterProperty.java
wuchong: OK. I just copied the files from Flink.

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/utils/JdbcProxy.java
wuchong: I don't think so. A check here can print the mimatched results which is much more meaningful than timeout exception.

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/MySqlE2eITCase.java
leonardBang: I mean this large range looks like we need 100 server ids in e2e test, but actually 4 server ids is enough

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/utils/JdbcProxy.java
leonardBang: I agree with you, I noticed the timeout check logic may problem, if the first attempt we didn't pull enough data(i.e the upstream program is still writing), the check will fail due to we call `checkResult` at first time in and the timeout is useless?

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/MySqlE2eITCase.java
leonardBang: It makes sense to me, how about the package prefix? Do you think we need to keep align with Flink project

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/MySqlE2eITCase.java
wuchong: Yes. But it should still works even we set a large range, right? I think many users will set a large range, because they may change parallelism frequently.

flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/e2e/MySqlE2eITCase.java
wuchong: OK. I will use `tests`  package

wuchong:
Thanks @leonardBang , I have addressed the comments.

leonardBang (approved):
Thanks @wuchong for the update, LGTM

PR #615, opened 2021-11-17
[connector-debezium] Allow user to pass other configs for JsonConvert…

paul8263:
…er in JsonDebeziumDeserializationSchema

This change allows user to add other config for JsonConverter in JsonDebeziumDeserializationSchema.

Background:
By default JsonConverter in org.apache.kafka.connect.json will parse the value typed BigDecimal to Base64 value, as JsonConverterConfig will set 'decimal.format' to 'DecimalFormat.BASE64.name()' as the default value. For example a record with BigDecimal fields will be converter to: {"NAME":"Kate","NUMBER":"FA=="}, which does not meet the requirements. However, currently we cannot control this behavior because there are no configs we can pass to JsonDebeziumDeserializationSchema constructor. This PR is to solve this.


leonardBang:
Thanks @paul8263 for the contribution, sorry for the late reply, could you add a test and documentation for the change ?

paul8263:
Hi @leonardBang ，
I added a unit test but got an error: https://dev.azure.com/ververica/flink-cdc-connectors/_build/results?buildId=1050&view=logs&j=ccedd2bf-e82a-54b8-b75f-7971f1fbbb9f&t=385827ba-1388-5c8f-6875-4ed5108d031f&l=1964
It's about Postgres.
By the way where should I put the documentation?

leonardBang:
Thanks @paul8263 for the contribution, I found the test is blocking thus I help improve the PR a little, please take a more look

paul8263:
Hi @leonardBang ,
Thank your very much. I forget to change the extracted method and the entries in customConverterConfigs were wrongly duplicated.

leonardBang (approved):

PR #638, opened 2021-11-23
[mysql] change MySqlSource constructor's accessibility from package-private to public

zhjwpku:
By applying this change, the user can construct a
MySqlSourceConfigFactory first, and use the constructor
to construct the MySqlSource, which can reduce the work
to maintain some other MySqlSourceConfig code.

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/MySqlSource.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/MySqlSource.java
@@ -75,7 +75,7 @@ import static com.ververica.cdc.connectors.mysql.debezium.DebeziumUtils.openJdbc
  *
  * <pre>{@code
  * MySqlSource
- *     .<RowData>builder()
+ *     .<String>builder()
  *     .hostname("localhost")
  *     .port(3306)
  *     .databaseList("mydb")
@@ -110,7 +110,7 @@ public class MySqlSource<T>
         return new MySqlSourceBuilder<>();
     }
 
-    MySqlSource(
+    public MySqlSource(
             MySqlSourceConfigFactory configFactory,
             DebeziumDeserializationSchema<T> deserializationSchema) {
         this.configFactory = configFactory;


wuchong:
@zhjwpku Thanks for the pull request. However, could you explain a bit why you need the MySqlSource constructor and MySqlSourceConfigFactoy? Why the `MySqlSourceBuilder` can't satisfy your need? In theory, they exposes the same configs.

zhjwpku:
> 

Hi Jark,

What we need is `MySqlSourceConfig`, since we what to use the `TableDiscoveryUtils#discoverCapturedTableSchemas` flink-connector-mysql-cdc provided. We could use the MySqlSourceConfig constructor, but that's more error prone, so we think `MySqlSourceConfigFactory#createConfig` might be a better option.

If we construct the MySqlSourceConfigFactory first, we can use it to create a `MySqlSourceConfig` instance same as used in the `MysqlSource`. In this way, we just need to maintain the database properties in one place.

BR,
Zhao

wuchong:
@zhjwpku , `MySqlSourceConfigFactory` and `MySqlSourceConfigFactory#createConfig` are public. So you are free to use them to create the `MySqlSourceConfig` instance. But why do you need to update `MysqlSource` constructor?

zhjwpku:
@wuchong I know `MySqlSourceConfigFactory` and `MySqlSourceConfigFactory#createConfig` are public, but If `MySqlSource` constructor are package-private, I have to main a `MySqlSourceConfigFactory` and use `MySqlSourceBuilder` to main the db information twice. like:

```
MySqlSource
     .<String>builder()
     .hostname("localhost")
     .port(3306)
     .databaseList("mydb")
     .tableList("mydb.users")
     .username(username)
     .password(password)
     .serverId(5400)
     .deserializer(new JsonDebeziumDeserializationSchema())
     .build();

MySqlSourceConfigFactory factory = new MySqlSourceConfigFactory()
        .hostname("localhost")
        .port(3306)
        .databaseList("mydb")
        .tableList("mydb.users")
        .username(username)
        .password(password);
```

But if the `MySqlSource` constructor is public, I can do this:

```
MySqlSourceConfigFactory factory = new MySqlSourceConfigFactory()
        .hostname("localhost")
        .port(3306)
        .databaseList("mydb")
        .tableList("mydb.users")
        .username(username)
        .password(password);

MySqlSource<String> source = new MySqlSource<>(configFactory, new JsonDebeziumDeserializationSchema());
```

DB infomations just need to type once, which is less error prone.

I am not sure public the `MySqlSource` constructor is the reasonable way to do this, but it's the easiest way. Add a `getMySqlSourceConfig` method to `MySqlSource` is another solution, but I think that might violate the `MySqlSource`'s encapsulation.

wuchong:
Thanks @zhjwpku , but I think maybe it would be better to provide `MySqlSource#getConfigFactory()` method. Because expose the constructor of `MySqlSource` will violate with the builder pattern which we actually don't want to expose the constructor (the constructor may adding more parameters in the future).

zhjwpku:
> Thanks @zhjwpku , but I think maybe it would be better to provide `MySqlSource#getConfigFactory()` method. Because expose the constructor of `MySqlSource` will violate with the builder pattern which we actually don't want to expose the constructor (the constructor may adding more parameters in the future).

ok, I will file another PR then ;)

PR #673, opened 2021-12-02
[build] Remove override definition of maven-surefire-plugin in connectors pom

wuchong:
This enables code coverage works using [JaCoCo](https://www.eclemma.org/jacoco/trunk/doc/maven.html) for the connector modules.


leonardBang (approved):
Thanks @wuchong , LGTM

PR #706, opened 2021-12-09
[oracle] let oracle-cdc connect to PDB CDB old non-CDB successfully

ZhifengSong:
[oracle] The general **dbname** should be SERVICE_NAME but not INSTANCE_NAME in the _lsnrctl status_ output, especially connect to PDB in CDB  it must be SERVICE_NAME now, of course we should support both of the two features in the future.


ZhifengSong:
And with this change can resolve the ERROR: **oracle.net.ns.NetException: Invalid number format for port number** when oracle listener does not support SID like PDB

ruanhang1993:
cc @leonardBang this PR is in version 2.2.0, but not merge.

leonardBang:
Supported in 2.3, see https://github.com/ververica/flink-cdc-connectors/issues/1244

PR #707, opened 2021-12-09
WIP [mysql] add MySqlSourceEnumeratorMetrics

zhjwpku:
It will be great if we expose MySqlSnapshotSplitAssigner's status
to metrics, people can see the snapshot progress on flink's UI.

Unfortunately, `enumContext.metricGroup())` returns null, make
it impossible to register the metrics.

I checked the flink source code a little bit, seems
`SourceCoordinatorContext.metricGroup()` returns null directly.

I open this PR and make it WIP, trying to understand FLIP-27
more deeply, I hope I can get it solved soon.

BTW, if any one think this idea is undoable, I'd be appreciate
if you can tell me the reason.

Signed-off-by: 元组 <zhaojunwang.zjw@alibaba-inc.com>

PR #720, opened 2021-12-14
[mysql] expose Enumerator metrics via Source Event

zhjwpku:
I opened a PR #707 last week, which hope to expose some
Enumerator metrics directly to `SplitEnumeratorContext.metricGroup()`,
but unfortunately, it return null for the `MetricGroup`. After
some investigation of FLIP-27, it says `The SourceCoordinators
will run in the JobMaster, but not as part of the ExecutionGraph`.

So there is noway to expose the Enumerator metrics ;(

In this PR, I use the RPC Gateway, a.k.a SourceEvent to transfer
the Enumerator metrics to each SourceReader.

I add 5 metrics:

- alreadyProcessedTableCount  // Already processed table count for this flink job.
- remainingTableCount	// Remaining table count for this flink job to handle.
- finishedSplitCount	// Processed splits count for this flink job.
- assignedSplitCount	// Assigned but not finished splits count for this flink job.
- remainingSplitCount	// Remaining split count for the currently processing table.

For now I just use the `FinishedSnapshotSplitsAckEvent` to
carry the above metrics, we may alternatively send a
request event to query the metrics in `MySqlSourceReader#start`
and `MySqlSourceReader#onSplitFinished`, that might need
some future discussion.

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlBinlogSplitAssigner.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlBinlogSplitAssigner.java
@@ -114,6 +114,31 @@ public class MySqlBinlogSplitAssigner implements MySqlSplitAssigner {
     @Override
     public void close() {}
 
+    @Override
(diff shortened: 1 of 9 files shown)


zhjwpku:
this PR is moved to #853

PR #752, opened 2021-12-24
[hotfix][mysql] Fix message timestamp of snapshot record to 0L

leonardBang:


leonardBang:
The CI passed, merging..

PR #756, opened 2021-12-28
support flink1.14

wangxiaojing:
support flink1.14


leonardBang:
Thanks @wangxiaojing for the contribution, but the end to end test failed, we need to fix it.

flink-cdc-e2e-tests/pom.xml
```diff
@@ -30,7 +30,7 @@ under the License.
     <packaging>jar</packaging>
 
     <properties>
-        <flink-1.13>1.13.3</flink-1.13>
+        <flink-1.14>1.14.2</flink-1.14>
```
leonardBang, line 33: @wangxiaojing Thanks for the contribution, the user need the CDC Connector can run on both Flink 1.13 cluster and Flink 1.14 Cluster, thus I think what we should need to do is make the the CDC Connector can run on  Flink 1.14 Cluster, so we do not need to change the version, but only need to the incompatibility API that CDC Connector used, we can add e2e test with Flink 1.14 cluster to check this.

leonardBang (changes requested):

sap1ens:
Hey @wangxiaojing, do you need any help with this?

leonardBang:
Looks like @wangxiaojing doesn't have time to update this PR，@sap1ens Would you like to take over this PR and continue contributing base on @wangxiaojing existed work?

leonardBang:
> 2\. Add e2e tests to verify that they pass with both 1.14 and 1.13

@kjatti You're right, would you like to contribute for this ticket?

leonardBang:
I help update this PR as there're too many users are waiting this feature.

leonardBang (approved):
Thanks all for the contribution, LGTM.

leonardBang:
The CI passed, merging...

PR #810, opened 2022-01-18
[mysql] rename dispatchHighWatermark to dispatchBinlogEndWatermark

zhjwpku:
If no need to read binlog, a BINLOG_END event is appended to the
event queue, `dispatchHighWatermark` is kind of confusing, so
change it to `dispatchBinlogEndWatermark`.


leonardBang (approved):
LGTM, I suggest to use `dispatchBinlogEndEvent`

PR #777, opened 2022-01-06
[mysql] add new tables to existing cdc job

zhjwpku:
This PR try to achieve the feature mentioned in [0]

1. MySqlSourceEnumerator
  1.1 check whether there is newly added table in `MySqlSnapshotSplitAssigner#open`
    1.1.1 if assigner not finished, add the tables to remainingTables
          and continue the normal processure
    1.1.2 if assigner has finished, a.k.a the job has enter the incremental phase
      1.1.2.1 set the assigner to suspend status
      1.1.2.2 send `SuspendBinlogSplitReaderRequestEvent` to `SourceReader`
2. MySqlSourceReader
    2.1 on receiving the `SuspendBinlogSplitReaderRequestEvent`
      2.1.2 set a flag to tell MySqlSplitReader to suspend
      2.1.3 `MySqlSplitReader` stop currentReader in `fetch()` and update
            binlog offset in `MySqlRecordEmitter#emitRecord`
      2.1.4 response a `SuspendBinlogSplitReaderResponseEvent` event to
            `MySqlSplitEnumerator` in `MySqlSourceReader#snapshotState`
3. MySqlSourceEnumerator
   3.1 on receiving the `SuspendBinlogSplitReaderResponseEvent`
     3.1.1 save the received offset, and start again the snapshot phase
     3.1.2 when snapshot phase finished, use the saved offset as minBinlogOffset
           in `MySqlHybridSplitAssigner#createBinlogSplit`

[0] https://github.com/ververica/flink-cdc-connectors/issues/513

Signed-off-by: 元组 <zhaojunwang.zjw@alibaba-inc.com>

PR #846, opened 2022-02-10
[mysql] Add new tables to an existing job

leonardBang:
base on https://github.com/ververica/flink-cdc-connectors/pull/777


leonardBang:
Let's track this in #777

Issue #641, opened 2021-11-23
[mysql] Support heartbeat event for table that update infrequently


leonardBang:
If the table updates infrequently, the binlog file may have been cleaned in its last committed binlog position

camelus0211:
I am very interested in this, please assign it to me.

camelus0211:
At present, i investigated two options: one add the heartbeat table of the subscription table, and send data to the heartbeat table regularly to achieve the logic of gtid heartbeat; the other is to monitor the binlog full table events in real time at the bottom layer, and update the latest gtid as heartbeat logic.
@leonardBang Do you have any other suggestions? thanks.

leonardBang:
@camelus0211  Thanks for your investigation very much,  the later one is more graceful, I think we can use the heartbeat event from underlying debeizum reader. Could you check `EventDispatcher#dispatchHeartbeatEvent()` in `MySqlBinlogSplitReadTask` and `MySqlStreamingChangeEventSource`?

camelus0211:
thanks, i will check it.

camelus0211:
I post the design document here, [API design](https://docs.google.com/document/d/1XVS6opo8adTqJK3KiJibKXjloKB_hW22PBNzvHRa0GU/edit#heading=h.259eqklad6hn), feel free to comment. 
@leonardBang thanks.

leonardBang:
> I post the design document here, [API design](https://docs.google.com/document/d/1XVS6opo8adTqJK3KiJibKXjloKB_hW22PBNzvHRa0GU/edit#heading=h.259eqklad6hn), feel free to comment. @leonardBang thanks.

The design makes sense to me @camelus0211 , please open the PR and we can discuss under the PR later.

PR #852, opened 2022-02-12
[mysql] add heartbeat support

camelus0211:
If the table updates infrequently, the binlog file or GTID set may have been cleaned in its last committed binlog position. The CDC job restart fails in this case. So the heartbeat event will update binlog position. #641

### DIFF
--- a/docs/content/connectors/mysql-cdc.md
+++ b/docs/content/connectors/mysql-cdc.md
@@ -375,6 +375,14 @@ After the server you monitored fails in MySQL cluster, you only need to change t
 
 It's recommended to configure a DNS(Domain Name Service) or VIP(Virtual IP Address) for your MySQL cluster, using the DNS or VIP address for ```mysql-cdc``` connector, the DNS or VIP would automatically route the network request to the active MySQL server. In this way, you don't need to modify the address and restart your pipeline anymore.
 
+#### MySQL Heartbeat Event Support
+
+If the table updates infrequently, the binlog file or GTID set may have been cleaned in its last committed binlog position.
+The CDC job restart fails in this case. So the heartbeat event will update binlog position. It's recommended to configure `debezium.heartbeat.interval.ms`.
+```
+debezium.heartbeat.interval.ms=3000
+```
+
 #### How Incremental Snapshot Reading works
 
 When the MySQL CDC source is started, it reads snapshot of table parallelly and then reads binlog of table with single parallelism.
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/RecordUtils.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/RecordUtils.java
@@ -62,6 +62,8 @@ public class RecordUtils {
 
     public static final String SCHEMA_CHANGE_EVENT_KEY_NAME =
             "io.debezium.connector.mysql.SchemaChangeKey";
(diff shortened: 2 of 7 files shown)


docs/content/connectors/mysql-cdc.md
```diff
@@ -361,6 +361,14 @@ After the server you monitored fails in MySQL cluster, you only need to change t
 
 It's recommended to configure a DNS(Domain Name Service) or VIP(Virtual IP Address) for your MySQL cluster, using the DNS or VIP address for ```mysql-cdc``` connector, the DNS or VIP would automatically route the network request to the active MySQL server. In this way, you don't need to modify the address and restart your pipeline anymore.
 
+#### MySQL Heartbeat Event Support
```
leonardBang, line 364: Could we introduce an option `heartbeat.interval.ms` with default value `30s` so that  user can configure this easily?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/RecordUtils.java
```diff
@@ -300,6 +302,15 @@ public static boolean isSchemaChangeEvent(SourceRecord sourceRecord) {
         return false;
     }
 
+    public static boolean isHeartbeatRecord(SourceRecord record) {
```
leonardBang, line 305: hint: 
```suggestion
    public static boolean isHeartbeatEvent(SourceRecord record) {
```

leonardBang (commented):
Thanks for the contribution @camelus0211 , I left some comments.
Could you add a test to cover the change? the test code logic could be that we found the offset in binlog split state is forward after we enabled the heartbeat event.

leonardBang:
@camelus0211 You can rebase(`git rebase master`) latest master branch instead of merge latest master when you want to make you dev branch base on latest master, in this way the github PR will only contains your commits. but now it contains 14 commits, I guess you're using `git merge` command.

camelus0211:
> @camelus0211 You can rebase(`git rebase master`) latest master branch instead of merge latest master when you want to make you dev branch base on latest master, in this way the github PR will only contains your commits. but now it contains 14 commits, I guess you're using `git merge` command.
@leonardBang Thanks for your reminder. I've tried to rebase it, but it contains 10 other commits. Is that ok? or pull new request?

camelus0211:
> @camelus0211 You can rebase(`git rebase master`) latest master branch instead of merge latest master when you want to make you dev branch base on latest master, in this way the github PR will only contains your commits. but now it contains 14 commits, I guess you're using `git merge` command.

@leonardBang Thanks for your reminder. I've tried to rebase it, but it contains 10 other commits. Is that ok? or pull new request?

leonardBang:
> > @camelus0211 You can rebase(`git rebase master`) latest master branch instead of merge latest master when you want to make you dev branch base on latest master, in this way the github PR will only contains your commits. but now it contains 14 commits, I guess you're using `git merge` command.
> 
> @leonardBang Thanks for your reminder. I've tried to rebase it, but it contains 10 other commits. Is that ok? or pull new request?

you should use `git rebase master` instead of `git merge master`

leonardBang:
Thanks for the update @camelus0211 , could you add a test to cover the change?

camelus0211:
> Thanks for the update @camelus0211 , could you add a test to cover the change?

@leonardBang ok, I will add it as soon as possible.

leonardBang:
fixed in #641

Issue #868, opened 2022-02-18
[common][bug] The 'ts_ms' field in snapshot SourceRecord returns non zero milliseconds


leonardBang:
The 'ts_ms' field in snapshot SourceRecord should return  0L but actual it may return a positive long value.

PR #878, opened 2022-02-22
[common] DebeziumChangeFetcher's emit delay and fetch delay should be unavailable during the snapshot phase

fsk119:
Solve #868 .


leonardBang (approved):
LGTM

PR #883, opened 2022-02-23
Added option for MysqlSource to dispatch watermark event at the begin…

qidian99:
When we enter binlog phase in a cdc job, the SplitAssigner only assigns a binlog split to one of the SourceReaders, namely the first reader waiting for splits in the queue. Therefore, there are use cases where we want to automatically scale in or out based on whether the sync job has entered binlog phase. A possible scenario would be as follows:

1. The user start a cdc job to sync two tables in the target database with a parallelism of 16
2. All snapshot splits has been emitted and a complete checkpoint has succeeded so that all records are consumed in the sink
3. A savepoint is triggered and the cdc job is restarted with a parallelism of 1.
4. The user wants to add new tables to the cdc job as in [this PR](https://github.com/ververica/flink-cdc-connectors/pull/777). A savepoint is triggered again and the job is restarted with 100 tables added and a parallelism of 128 for parallel reading splits
5. All the new snapshot splits has been emitted and a complete checkpoint has succeeded. 
6. A savepoint is triggered again and the job is restarted with a parallelism of 1 to consume binlog records


yuxiqian:
Hi @qidian99, sorry for the delay of this PR. Could you please rebase it to latest `master` branch since there’s been lots of changes in Flink CDC repo since your original commit? Kindly reminder that anything under `com.ververica.cdc.connectors.mysql` package has been moved to `org.apache.flink.cdc.connectors.mysql`.

Thanks for your contribution!

PR #886, opened 2022-02-23
[hotfix][mysql] Remove mysql-connector-java dependency from flink-sql-connector-mysql-cdc

### LINKED ISSUE #884: Is there a problem with our license selection?
mysql-connector-java, which flink-sql-connector-mysql-cdc relies on, is a GPL license, so should flink-cdc-connectors also be a GPL license?

image

JingsongLi:
This fixes #884
User should add this dependency by themselves.


wuchong:
We can't simply remove the mysql driver dependency, because the e2e test need the driver to pull mysql data.

JingsongLi:
> We can't simply remove the mysql driver dependency, because the e2e test need the driver to pull mysql data.

Thanks for the information, I'll close this PR

Issue #901, opened 2022-03-01
[task] Remove changelog json format  support


leonardBang:
`changelog-json`  is introduced at the time that Flink didn't offer any changelog format support. Currently Flink offers several changelog i.e.[Debezium CDC, MAXWELL CDC, OGG CDC, CANAL CDC](https://nightlies.apache.org/flink/flink-docs-release-1.14/docs/connectors/table/formats/overview/) that can used in both source and sink, we recommend user to use above CDC formats, user can also obtain the old `changelog-json` format under the Flink CDC version 2.1.

PR #912, opened 2022-03-04
[format][changelog-json] Remove the support of changelog-json format

leonardBang:
[task] Remove changelog json format support #901


PatrickRen (commented):
Thanks for the patch @leonardBang ! 

I made a global search with keyword "changelog-json" and found some remnants in `about.md` :

https://github.com/ververica/flink-cdc-connectors/blob/c77275cb683bdb7aa13f4d6b89baecf9e382a305/docs/content/about.md?plain=1#L22

And the link incorrectly points to an empty page in GitHub wiki. I think we can add a footnote here suggesting `changelog-json` is deprecated, and fix the broken link.

leonardBang:
> And the link incorrectly points to an empty page in GitHub wiki. I think we can add a footnote here suggesting `changelog-json` is deprecated, and fix the broken link.

Nice catch, I decide to delete the piece as they're rarely used( if users use it they have found the link error).

PatrickRen (approved):
Thanks for the update @leonardBang ! LGTM.

PR #922, opened 2022-03-10
[docs] The link of download flink is 404 now.

GOODBOY008:
Fix https://github.com/ververica/flink-cdc-connectors/issues/918.


leonardBang (approved):
Thanks @GOODBOY008 for the fix, LGTM

PR #924, opened 2022-03-10
[mysql-cdc] support polardbx ddl by overwrite the mysql antlr g4

Cleverdada:
major updates:
overwrite the mysql antlr g4(createTable), add content which named with "polardbx" prefix .


leonardBang:
Thanks @Cleverdada  for the great work, but the change is to huge to maintain in the future, could we find a simper way to support PolarDBx ?

Cleverdada:
> 

Yes, I think there are two better ways to support:

1. Add a option to generate the ddl in mysql-cdc when "show create table" failed, such as "desc table" or "select * from information_schema" .
2. Create a polardbx-cdc module to support, but I'm not familiar with it yet .

PR #927, opened 2022-03-11
[MySQL] Add heartbeat event handling logic

PatrickRen:
This pull request add heartbeat event handling logic in `MySqlRecordEmitter` in order to keep offsets in split state tracking the latest in binlog. See #641 for details.

### DIFF
--- a/docs/content/connectors/mysql-cdc.md
+++ b/docs/content/connectors/mysql-cdc.md
@@ -385,7 +385,7 @@ It's recommended to configure a DNS(Domain Name Service) or VIP(Virtual IP Addre
 #### MySQL Heartbeat Event Support
 
 If the table updates infrequently, the binlog file or GTID set may have been cleaned in its last committed binlog position.
-The CDC job restart fails in this case. So the heartbeat event will update binlog position. By default heartbeat event is enabled in MySQL CDC source and the interval is set to 30 seconds. You can specify the interval by using table option ```heartbeat.interval```, or set the option to 0 to disable heartbeat events.
+The CDC job may restart fails in this case. So the heartbeat event will help update binlog position. By default heartbeat event is enabled in MySQL CDC source and the interval is set to 30 seconds. You can specify the interval by using table option ```heartbeat.interval```, or set the option to `0s` to disable heartbeat events.
 
 #### How Incremental Snapshot Reading works
 
--- a/docs/content/connectors/mysql-cdc.md
+++ b/docs/content/connectors/mysql-cdc.md
@@ -253,6 +253,13 @@ During a snapshot operation, the connector will query each included table to pro
           <td>The connection pool size.</td>
     </tr>
     <tr>
+          <td>heartbeat.interval</td>
+          <td>optional</td>
+          <td style="word-wrap: break-word;">30s</td>
+          <td>Duration</td>
+          <td>The interval of sending heartbeat event for tracing the latest available binlog offsets.</td>
+    </tr>
+    <tr>
       <td>debezium.*</td>
       <td>optional</td>
       <td style="word-wrap: break-word;">(none)</td>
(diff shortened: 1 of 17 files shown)


docs/content/connectors/mysql-cdc.md
```diff
@@ -375,6 +375,14 @@ After the server you monitored fails in MySQL cluster, you only need to change t
 
 It's recommended to configure a DNS(Domain Name Service) or VIP(Virtual IP Address) for your MySQL cluster, using the DNS or VIP address for ```mysql-cdc``` connector, the DNS or VIP would automatically route the network request to the active MySQL server. In this way, you don't need to modify the address and restart your pipeline anymore.
 
+#### MySQL Heartbeat Event Support
+
+If the table updates infrequently, the binlog file or GTID set may have been cleaned in its last committed binlog position.
+The CDC job restart fails in this case. So the heartbeat event will update binlog position. It's recommended to configure `debezium.heartbeat.interval.ms`.
+```
+debezium.heartbeat.interval.ms=3000
+```
```
leonardBang, line 384: Also give the SQL API usage way?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/reader/MySqlRecordEmitter.java
```diff
@@ -98,6 +99,11 @@ public void emitRecord(SourceRecord element, SourceOutput<T> output, MySqlSplitS
             }
             reportMetrics(element);
             emitElement(element, output);
+        } else if (isHeartbeatEvent(element)) {
+            if (splitState.isBinlogSplitState()) {
+                BinlogOffset position = getBinlogPosition(element);
+                splitState.asBinlogSplitState().setStartingOffset(position);
```
leonardBang, line 105: minor: introduce a method to reduce the duplicated code ?

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/debezium/reader/BinlogSplitReaderTest.java
```diff
@@ -279,56 +287,130 @@ public void testReadBinlogFromLatestOffset() throws Exception {
...
+                        createBinlogReader(sourceConfig), dataType, sourceConfig, expected.length);
         assertEqualsInOrder(Arrays.asList(expected), actual);
     }
 
-    private List<String> readBinlogSplitsFromLatestOffset(
-            DataType dataType, MySqlSourceConfig sourceConfig, int expectedSize) throws Exception {
-        final StatefulTaskContext statefulTaskContext =
-                new StatefulTaskContext(sourceConfig, binaryLogClient, mySqlConnection);
+    @Test
+    public void testHeartbeatEvent() throws Exception {
```
leonardBang, line 296: Please also test SnapshotSplitReader and I think it will fail as `RecordUtils#normalizedSplitRecords` doesn't consider the nee introduced heartbeat event type. A solid way is that we should disable heartbeat event for snapshot split.

leonardBang (commented):
Thanks @PatrickRen for the contribution, I left some comments

docs/content/connectors/mysql-cdc.md
```diff
@@ -378,10 +385,7 @@ It's recommended to configure a DNS(Domain Name Service) or VIP(Virtual IP Addre
 #### MySQL Heartbeat Event Support
 
 If the table updates infrequently, the binlog file or GTID set may have been cleaned in its last committed binlog position.
-The CDC job restart fails in this case. So the heartbeat event will update binlog position. It's recommended to configure `debezium.heartbeat.interval.ms`.
-```
-debezium.heartbeat.interval.ms=3000
-```
+The CDC job restart fails in this case. So the heartbeat event will update binlog position. By default heartbeat event is enabled in MySQL CDC source and the interval is set to 30 seconds. You can specify the interval by using table option ```heartbeat.interval```, or set the option to 0 to disable heartbeat events.
```
leonardBang, line 388:
```suggestion
The CDC job may restart fails in this case. So the heartbeat event will help update binlog position. By default heartbeat event is enabled in MySQL CDC source and the interval is set to 30 seconds. You can specify the interval by using table option ```heartbeat.interval```, or set the option to `0s` to disable heartbeat events.
```

leonardBang (approved):
LGTM

PR #928, opened 2022-03-11
[mysql] optimize scan newly added tables feature

zhjwpku:
In current implemtation, `MySqlSourceEnumerator` send suspend event to
`MySqlSourceReader` in `start` and `syncWithReader` periodly, there is
a large chance the event sent from `start` be lost if the
`MySqlSourceReader` has not been registered.

In this PR, instead of waiting for the suspend event, `MySqlSourceReader`
suspend the binglog split **eagerly** by checking the `sourceConfig` if
there is newly added table, this simplified the `MySqlSourceReader` logic,
the `MySqlSplitReader` need not to be told to stop.

We use the flink stop savepoint to take care of the binlogSplit state.


zhjwpku:
This patch also fixed the bug mentioned in #960

zhjwpku:
We keep the notification mechanism instead of letting the SourceReader detect whether it should be suspended or not, because in the following Flink release, there will be features that support add table on the fly instead of relying restart from savepoint.

Peng-89:
When restoring from a checkpoint, both newly added tables and existing tables will be initialized?

PR #952, opened 2022-03-16
test ob ci on mac vm

whhe:


whhe:
Give up after many failures.

PR #960, opened 2022-03-18
[mysql] add test cases to trigger newly added table feature bug

zhjwpku:


zhjwpku:
This patch will trigger a bug of the scan newly added table feature, when `MySqlSourceReader` restore from state, it does not contain tableSchemas of the newly added table, debezium will throw an Exception, `address_beijing` is the newly added table.

zhjwpku:
@leonardBang this is the bug I mentioned yestoday, pls take a look.

zhjwpku:
fixed by #965

PR #961, opened 2022-03-18
[docs][site] Introduce the project website

tsreaper:
Hi dear Flink CDC community! In order to celebrate the release of version 2.2 I'd like to introduce a project website for CDC. See https://tsreaper.github.io/flink-cdc-connectors/ for preview.

The squirrel image in the hero banner is from https://pixabay.com/zh/photos/squirrel-fur-cute-forest-mammal-3815235/, which is under the [Pixabay License](https://pixabay.com/zh/service/license/). Seems that we can use it in the project website.

Also note that the download and the documentation links are hard-coded. This should not be a problem unless we change the project name and even if we change that, the github button API will also break so we have to change the site anyway.


leonardBang:
Thanks @tsreaper for the nice work, please check the rat output file there're some files' license is not approved.

leonardBang (approved):
Thanks @tsreaper  for the awesome contribution to let use have a so cool site. 
Give my big +1.

PR #972, opened 2022-03-21
[WIP][mysql] Cleanup the finished splits information from enumerators

### LINKED ISSUE #865: [mysql] Too many splitInfo cause OOM of jobmanager
**Describe the bug**

in the issue https://issues.apache.org/jira/browse/FLINK-24855， have some problems when we sync big tables or full database 

**Environment :**
 - Flink version :  1.13.3
 - Flink CDC version:  2.1.1
 - Database and version: mysql-5.7

**To Reproduce**
1. Thes test data :  The data scale more than billion ，and produce more than one hundred thousand enumerator state
2. The test code :  The multiple tables sync by stream api 
3. The error :
Source Coordinator Thread already exists. There should never be more than one thread driving the actions of a Source Coordinator.

**Additional Description**

minchowang:
Fixes #865

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
@@ -24,7 +24,6 @@ import com.ververica.cdc.connectors.mysql.source.config.MySqlSourceConfig;
 import com.ververica.cdc.connectors.mysql.source.offset.BinlogOffset;
 import com.ververica.cdc.connectors.mysql.source.split.FinishedSnapshotSplitInfo;
 import com.ververica.cdc.connectors.mysql.source.split.MySqlBinlogSplit;
-import com.ververica.cdc.connectors.mysql.source.split.MySqlSnapshotSplit;
 import com.ververica.cdc.connectors.mysql.source.split.MySqlSplit;
 import io.debezium.relational.TableId;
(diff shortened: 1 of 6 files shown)


minchowang:
We need to clenaup `MySqlSnapshotSplitAssigner`.`assignedSplits`、 `MySqlSnapshotSplitAssigner`.`splitFinishedOffsets`  at the right time.

minchowang:
@zhjwpku @leonardBang Big-olds, What do you think about this patch?

minchowang:
I think we have two the way to improvement it. 
At the mutipl tables sync case：
1. when whole phase change to binlog starting,  we can cleanup finished snapshotSplits. (the way this patch is implemention, during snapshot still using a lot of memory of jobmanager, however this implementation just need a little to modify the code)
2. when all splits of single table have been read completed, we just retain a split for this table (the split is min binlog start position)， this way can during snapshot  forever using steady state value memory of jobmanager.

zhjwpku:
> 

For case 2, do you mean that  we generate a split that only contains the lowest binlog starting point?
I'm not sure this works because for each chunk of this table, it's high watermark should be retained.
Maybe we can use a list(or a more efficient data structure) to maintain all the high watermakrs in a
single split?

minchowang:
> > 
> 
> For case 2, do you mean that we generate a split that only contains the lowest binlog starting point? I'm not sure this works because for each chunk of this table, it's high watermark should be retained. Maybe we can use a list(or a more efficient data structure) to maintain all the high watermakrs in a single split?

i updated case 2, please look again

ruanhang1993:
#865 has been completed. Close this PR.

PR #973, opened 2022-03-22
[mysql] accelerate suspend speed

zhjwpku:
`suspendBinlogReaderIfNeed` broadcast `SuspendBinlogReaderEvent`
when splitAssigner is in suspend status, but there are chances
that this happens before `SourceReader` registered to
`SourceEnumerator`, in that case, we will need to wait
`CHECK_EVENT_INTERVAL` a.k.a 30s to resend the event message,
which can be annoying.

In this patch, simply send `SuspendBinlogReaderEvent` message
when `SourceReader` registers if the assigner's status is
suspended.


leonardBang (approved):
LGTM

PR #991, opened 2022-03-23
[build] Decrease the azure job timeout to reasonable 40 minutes

leonardBang:
Current azure job timeout is 240 mins which is too long which lead to some bad case hang and waste the machine resource. As the max cost time of job is about  22 minutes now, I suggest reducing to 40 minutes. 

Normal case : https://dev.azure.com/ververica/flink-cdc-connectors/_build/results?buildId=1342&view=results

Bad case: https://dev.azure.com/ververica/flink-cdc-connectors/_build/results?buildId=1343&view=results


wuchong (approved):
LGTM.

PR #992, opened 2022-03-23
[common] Bump debezium version to 1.6.4.Final

leonardBang:


JesseAtSZ:
In the 1.5.4.Final version of Debezium Engine currently used, there will be an error in SourceRecord when the provide.transaction.metadata is set. 

https://issues.redhat.com/browse/DBZ-4909

https://github.com/ververica/flink-cdc-connectors/issues/1002

It has been verified that using 1.6.4 version can solve this problem

leonardBang:
The CI passed, the failed tests is related to https://issues.redhat.com/browse/DBZ-4997, I've ignore it temporarily.

PR #1005, opened 2022-03-25
Remove redundant environment variable in tutorial

xuebingli:
In `mysql-postgres-tutorial.md`, environment variable `POSTGRES_PASSWORD` is defined twice with different values. This is not only redundant, but it could also lead to errors if/when the wrong value is used.

This PR removes the wrong value and only keeps the right one.


leonardBang (approved):
Thanks @xuebingli for the contribution, LGTM

Issue #963, opened 2022-03-18
We should make some improvements to the Flink CDC document


rookiegao:
These areas that need to be improved often confuse beginners.

-  The Flink CDC datastream API does not need to set the tablelist parameter when synchronizing the whole database.
-  Flink Oracle CDC wiki Q4: Oracle Sid and schema name should be distinguished by punctuation to avoid ambiguity.
-  After 2.1.0, the Flink MySQL CDC includes the MySQL driver package by default. We don't need to provide it ourselves.
-  The Flink CDC connector is a source connector rather than a sink connector. The project is only responsible for reading data from the source and does not support the sink function.
-  Specific instructions should be added to the Oracle CDC SQL table creation example on the official website.

For example, on the official website:

CREATE TABLE products (
     ID INT NOT NULL,
     NAME STRING,
     DESCRIPTION STRING,
     WEIGHT DECIMAL(10, 3),
     PRIMARY KEY(id) NOT ENFORCED
     ) WITH (
     'connector' = 'oracle-cdc',
     'hostname' = 'localhost',
     'port' = '1521',
     'username' = 'flinkuser',
     'password' = 'flinkpw',
     'database-name' = 'XE',
     'schema-name' = 'inventory', 
     'table-name' = 'products');

New description

CREATE TABLE products (
     ID INT NOT NULL,
     NAME STRING,
     DESCRIPTION STRING,
     WEIGHT DECIMAL(10, 3),
     PRIMARY KEY(id) NOT ENFORCED
     ) WITH (
     'connector' = 'oracle-cdc',
     'hostname' = 'localhost',
     'port' = '1521',
     'username' = 'flinkuser',
     'password' = 'flinkpw',
     'database-name' = 'XE',  -- XE is oracle SID
     'schema-name' = 'inventory', -- inventory is oracle user
     'table-name' = 'products');
     
- Other areas that need improvement are welcome to discuss.

leonardBang:
Thanks for opening this issue, would you like to open a PR to improve the document ? @rookiegao

rookiegao:
@leonardBang Yes, I'm willing to improve it. Do you have any other good suggestions？

PR #1014, opened 2022-03-28
[docs] CDC document update #963

rookiegao:
Documentation will be more convenient for beginners.


rookiegao:
I noticed that the snapshot version has been explained in the quick start doc. Do I need to keep this note?

PR #1087, opened 2022-04-19
[hotfix][mysql] Exception capture is added to the ExecutorService

### LINKED ISSUE #1081: 监视千万级别的MYSQL表，全量阶段没有反应。
**Describe the bug(Please use English)**
Use the Initia mode to monitor millions of tables, do not read data, no response;

**Environment :**
 - Flink version :  1.14.3
 - Flink CDC version: 2.2.0
 - Database and version:  MYSQL 8.0.15

**To Reproduce**
1. Thes test data :
2. The test code :
3. The error :

**Additional Description**
同步ersdata.book_datatype_rel这个1000多万的表，启动后没有任何反应；其他的数据量没有呢么大的表非常正常；

image

rookiegao:
fix #1081

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
@@ -200,17 +200,21 @@ public class MySqlSnapshotSplitAssigner implements MySqlSplitAssigner {
 
             executor.submit(
                     () -> {
-                        Iterator<TableId> iterator = remainingTables.iterator();
-                        while (iterator.hasNext()) {
-                            TableId nextTable = iterator.next();
-                            // split the given table into chunks (snapshot splits)
-                            Collection<MySqlSnapshotSplit> splits =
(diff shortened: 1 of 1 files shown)


rookiegao:
cc @leonardBang please review it

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
```diff
@@ -198,21 +199,28 @@ private void startAsynchronouslySplit() {
...
+                                    synchronized (lock) {
+                                        remainingSplits.addAll(splits);
+                                        remainingTables.remove(nextTable);
+                                        lock.notify();
+                                    }
+                                }
+                            });
+
+            try {
+                future.get();
```
PatrickRen, line 220: The chunk splitting logic is designed to run asynchronously as an optimization. Waiting for the future to finish here will break the async mechanism.

PatrickRen (changes requested):
Thanks for the patch @rookiegao but I'm afraid the implementation will break the async design here. 

What about holding the exception thrown inside the executor in a volatile variable in `SplitAssigner` and throw out in `getNext()`?

PR #1089, opened 2022-04-20
[docs][mongodb] Update document about MongoDB database and collection regex filter for master.

Jiabao-Sun:
Update document about MongoDB database and collection regex filter.


leonardBang (approved):
LGTM +1

PR #1098, opened 2022-04-20
[mysql] Throw RuntimeException timely in snapshot scan phase

leonardBang:
Throw RuntimeException timely in snapshot scan phase


lsyldliu (approved):
Thanks for your contribution @leonardBang, LGTM

PR #1099, opened 2022-04-20
[mysql] Throw RuntimeException timely in snapshot scan phase

leonardBang:


leonardBang:
The failed test is not related to current PR, merging...

PR #1103, opened 2022-04-21
[hotfix][tidb] fix RowKind.INSERT judgement condition in DeserializatonSchema

rinkako:
RowKind.INSERT is decided via `if (row.getOldValue() == null)` in `RowDataTiKVChangeEventDeserializationSchema:84`. 

But when performing INSERT statements in TiDB, it produced a Row record with NOT NULL return value from  `row.getOldValue()` but `Internal.EMPTY_BYTE_ARRAY`(a String[0] array) return value from `row.getOldValue().toByteArray()`, which will produce wrong RowKind (expect `INSERT` but actual `UPDATE_AFTER`) of the emitted RowData.

Hence it should add a length judgement condition for INSERT.

cc @leonardBang


rinkako:
sorry, I made a wrong PR from master branch, I will re-open one

PR #1107, opened 2022-04-22
[mysql] Fix dead loop when splitting chunks.

GOODBOY008:
Link issue https://github.com/ververica/flink-cdc-connectors/issues/1031.

### DIFF
--- a/flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/utils/ObjectUtils.java
+++ b/flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/utils/ObjectUtils.java
@@ -26,13 +26,13 @@ public class ObjectUtils {
 
     /**
      * Returns a number {@code Object} whose value is {@code (number + augend)}, Note: This method
-     * does not consider number overflow because we don't want to change the object type.
+     * will throw {@link ArithmeticException} if number overflows.
      */
-    public static Object plus(Object number, int augend) {
+    public static Object plus(Object number, int augend) throws ArithmeticException {
         if (number instanceof Integer) {
-            return (int) number + augend;
+            return Math.addExact((Integer) number, augend);
         } else if (number instanceof Long) {
-            return (long) number + augend;
+            return Math.addExact((Long) number, augend);
         } else if (number instanceof BigInteger) {
             return ((BigInteger) number).add(BigInteger.valueOf(augend));
         } else if (number instanceof BigDecimal) {
--- a/flink-cdc-base/src/test/java/com/ververica/cdc/connectors/base/experimental/MySqlChunkSplitter.java
+++ b/flink-cdc-base/src/test/java/com/ververica/cdc/connectors/base/experimental/MySqlChunkSplitter.java
@@ -220,7 +220,12 @@ public class MySqlChunkSplitter implements JdbcSourceChunkSplitter {
         while (ObjectUtils.compare(chunkEnd, max) <= 0) {
             splits.add(ChunkRange.of(chunkStart, chunkEnd));
             chunkStart = chunkEnd;
-            chunkEnd = ObjectUtils.plus(chunkEnd, chunkSize);
+            try {
+                chunkEnd = ObjectUtils.plus(chunkEnd, chunkSize);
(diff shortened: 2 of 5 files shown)


flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/utils/ObjectUtils.java
```diff
@@ -26,13 +26,21 @@ public class ObjectUtils {
...
     /**
      * Returns a number {@code Object} whose value is {@code (number + augend)}, Note: This method
-     * does not consider number overflow because we don't want to change the object type.
+     * will return max value of number type,if number overflows.
      */
     public static Object plus(Object number, int augend) {
         if (number instanceof Integer) {
-            return (int) number + augend;
+            try {
+                return Math.addExact((Integer) number, augend);
```
PatrickRen, line 34: What about throwing the `ArithmeticException` directly if the result overflows, and catch the exception in `ChunkSplitter`? Returning Integer.MAX or Long.MAX on overflow is a bit confusing and ambiguous.

PatrickRen (commented):
Thanks for the patch @GOODBOY008 ! I left a comment

flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/utils/ObjectUtils.java
GOODBOY008: Thks review,I will fix this.

flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/utils/ObjectUtils.java
```diff
@@ -26,13 +26,13 @@ public class ObjectUtils {
 
     /**
      * Returns a number {@code Object} whose value is {@code (number + augend)}, Note: This method
-     * does not consider number overflow because we don't want to change the object type.
+     * will throw {@link ArithmeticException} if number overflows.
      */
     public static Object plus(Object number, int augend) {
```
PatrickRen, line 31: What about adding `ArithmeticException` to the method signature directly? Like:
```java
public static Object plus(Object number, int augend) throws ArithmeticException
```

flink-cdc-base/src/test/java/com/ververica/cdc/connectors/base/experimental/MySqlChunkSplitter.java
```diff
@@ -220,7 +220,17 @@ private List<ChunkRange> splitEvenlySizedChunks(
...
+                chunkEnd = ObjectUtils.plus(chunkEnd, chunkSize);
+            } catch (ArithmeticException e) {
+                // Set chunkEnd to the max value of number type to avoid dead loop if number
+                // overflows.
+                chunkEnd = chunkEnd instanceof Integer ? Integer.MAX_VALUE : Long.MAX_VALUE;
+            }
+            // If chunkStart equals chunkEnd,should stop splitting chunk.
+            if (ObjectUtils.compare(chunkStart, chunkEnd) == 0) {
+                break;
+            }
```
PatrickRen, line 233: Could we jump out of the loop here directly in the catch clause? Then all records greater than `chunkStart` (`chunkStart` to `MAX` in overflow case) will be treated as the ending split.

PatrickRen (commented):
Thanks @GOODBOY008 for the update. I left some comments.

flink-cdc-base/src/test/java/com/ververica/cdc/connectors/base/experimental/MySqlChunkSplitter.java
GOODBOY008: @PatrickRen I have fixed.Thks for your tips.

ruanhang1993 (commented):
Thanks for your work. I find something to fix.

GOODBOY008:
@ruanhang1993 Thank you very much for your suggestions.I had add test and make method to public.

ruanhang1993 (approved):
LGTM. It is better to add another test which does not overflow.

PR #1109, opened 2022-04-22
[mysql] Catch and hold exception in chunk splitter thread and rethrow in getNext()

PatrickRen:
This pull request fixes #1081 that an exception should be thrown in coordinator thread if the consuming table doesn't have primary key.

PR #1110, opened 2022-04-22
[BP-2.2][MySQL] Catch and hold exception in chunk splitter thread and rethrow in getNext()

PatrickRen:
Unchanged back port of #1109 on release-2.2


leonardBang (approved):
+1

PR #761, opened 2021-12-30
[mysql] update check gtid set

camelus0211:
Flink job could not restart from checkpoint when some binlog files in the new master were cleared during the master/slave switchover. We add the gtid logic, it works. #743

PR #1111, opened 2022-04-22
[mysql] Use the gtid set firstly when compare the binlog offset (#761)

PatrickRen:
This is an unchanged back port of #761 on release-2.2


leonardBang (approved):
+1

PR #1116, opened 2022-04-25
[mysql] Add heartbeatInterval field to equals() and hashCode()

PatrickRen:
This pull request adds a missing field `heartbeatInterval` into MySqlTableSource.equals() and hashCode().

PR #1118, opened 2022-04-25
[mysql] Add heartbeatInterval field to equals() and hashCode()

PatrickRen:
Unchanged back port of #1116 on release-2.2


leonardBang (approved):
+1

PR #1161, opened 2022-05-09
[PostgreSQL] Add DeserializationConverterFactory for PostGIS schemas

Elephantusparvus:
This adds a simple DeserializationConverterFactory for PostgisGeometry objects, like it is done in with the com.ververica.cdc.connectors.mysql.table.MySqlDeserializationConverterFactory.
But in the case of PostGIS the data is encoded in ewkb, which should already include the srid (see here https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java and https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgisGeometry.java), thus converting it to GeoJSON would be an 
In the string case this converter returns an hex encoded string of the ewkb, which could be directly transformed to a corresponding geometry type column of apache sedona (https://sedona.apache.org/tutorial/flink/sql/) with the 'ST_GeomFromWKB' function. In the byte case the raw ewkb byte array is directly returned. This array again can easily be converted into the corresponding geometry object e.g. by using https://locationtech.github.io/jts/javadoc/org/locationtech/jts/io/WKBReader.html#read-byte:A-.

### DIFF
--- a/flink-connector-postgres-cdc/src/main/java/com/ververica/cdc/connectors/postgres/table/PostgreSQLTableSource.java
+++ b/flink-connector-postgres-cdc/src/main/java/com/ververica/cdc/connectors/postgres/table/PostgreSQLTableSource.java
@@ -124,6 +124,8 @@ public class PostgreSQLTableSource implements ScanTableSource, SupportsReadingMe
                         .setMetadataConverters(metadataConverters)
                         .setResultTypeInfo(typeInfo)
                         .setValueValidator(new PostgresValueValidator(schemaName, tableName))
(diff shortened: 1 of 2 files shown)


leonardBang (commented):
@Elephantusparvus Thanks for the contribution, the changes looks good to me overall, Could you add a test for the change?

yuxiqian:
Hi @Elephantusparvus, thanks for your contribution! Seems this issue has been addressed by #1554, would you mind if I close this PR?

PR #1188, opened 2022-05-17
[mysql][polardb-x] Support all charsets  for MySQL CDC Connector

ruanhang1993:
This PR fixed the encoding bug when a mysql table uses a different charset encoding in the column level.

Changes:
- Debezium `MySqlValueConverters#charsetFor(Column)` is reused to get the column charset.
- For String type in mysql, first encoding with the jdbc connection charset(UTF_8). Then use the column charset.

### DIFF
--- a/flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/polardbx/PolardbxSourceITCase.java
+++ b/flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/polardbx/PolardbxSourceITCase.java
@@ -77,7 +77,7 @@ public class PolardbxSourceITCase extends AbstractTestBase {
     private static final String USER_NAME = "polardbx_root";
     private static final String PASSWORD = "123456";
     private static final String DATABASE = "polardbx_ddl_test";
-    private static final String IMAGE_VERSION = "2.0.1";
+    private static final String IMAGE_VERSION = "2.1.0";
     private static final DockerImageName POLARDBX_IMAGE =
             DockerImageName.parse("polardbx/polardb-x:" + IMAGE_VERSION);
 


ruanhang1993:
@fsk119 Please take a look. Thanks.
BTW, I only add four charset tests in this PR. Do we need to add all of the charsets in the mysql ?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
```diff
@@ -313,7 +316,14 @@ else if (actualColumn.jdbcType() == Types.TINYINT
         else if ("CHAR".equals(actualColumn.typeName())
                 || "VARCHAR".equals(actualColumn.typeName())
                 || "TEXT".equals(actualColumn.typeName())) {
-            return rs.getBytes(fieldNo);
+            Charset columnCharset = DebeziumUtils.charsetFor(actualColumn);
+            final byte[] stringColumnBytes = rs.getBytes(fieldNo);
+            if (stringColumnBytes != null && columnCharset != null) {
+                // UTF_8 for the jdbc connection charset encoding
+                return new String(stringColumnBytes, StandardCharsets.UTF_8)
```
fsk119, line 323: I find currently users can set the jdbc parameters in the with option `jdbc.properties.*`[1]. Therefore, the charset encoding may not  UTF_8 here.

I just wonder whether we should use the default encoding format here.  The parameter `characterSetResults` means return the result with the default value if no set. Could you check what's the jdbc url in the debezium?

[1] https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mysql-cdc.html#connector-options
[2] https://docs.oracle.com/cd/E17952_01/connector-j-8.0-en/connector-j-connp-props-session.html#cj-conn-prop_characterSetResults

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
```diff
@@ -1423,6 +1426,191 @@ public void testColumnOptionalWithDefaultValue() throws Exception {
         result.getJobClient().get().cancel().get();
     }
 
+    @Test
+    public void testUcs2Charset() throws Exception {
```
fsk119, line 1430: I think we can move the charset tests in a single file. I think it would be much clean.

fsk119 (commented):
Thanks for your great work. I leave some comments.

ruanhang1993:
I find that the charset properties are not using in the jdbc connection rightly.

This issue should wait https://github.com/ververica/flink-cdc-connectors/issues/674 to be fixed. After setting the properties rightly, we could pass the new tests that use different charset.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
ruanhang1993: in the debezium class MysqlConnection, the jdbc url pattern is : (characterEncoding=UTF-8&characterSetResults=UTF-8)
```
protected static final String URL_PATTERN = "jdbc:mysql://${hostname}:${port}/?useInformationSchema=true&nullCatalogMeansCurrent=false&useSSL=${useSSL}&useUnicode=true&characterEncoding=UTF-8&characterSetResults=UTF-8&zeroDateTimeBehavior=CONVERT_TO_NULL&connectTimeout=${connectTimeout}";
```

So I think it is ok to use the default UTF-8.

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
ruanhang1993: OK, I will fix it.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
```diff
@@ -313,7 +316,14 @@ else if (actualColumn.jdbcType() == Types.TINYINT
         else if ("CHAR".equals(actualColumn.typeName())
                 || "VARCHAR".equals(actualColumn.typeName())
                 || "TEXT".equals(actualColumn.typeName())) {
-            return rs.getBytes(fieldNo);
+            Charset columnCharset = DebeziumUtils.charsetFor(actualColumn);
+            final byte[] stringColumnBytes = rs.getBytes(fieldNo);
```
fsk119, line 320: After investigate, I find `return rs.getString()` is enough.

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MysqlConnectorCharsetITCase.java
```diff
@@ -0,0 +1,213 @@
...
+            }
+        };
+    }
+
+    @Before
+    public void before() {
+        TestValuesTableFactory.clearAllData();
+        env.setParallelism(DEFAULT_PARALLELISM);
+        env.enableCheckpointing(200);
+        charsetTestDatabase.createAndInitialize();
```
fsk119, line 148: we don't need initialize the database before every test. Move to 

```
@BeforeClass
public static void beforeClass() {
        charsetTestDatabase.createAndInitialize();
}
```

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MysqlConnectorCharsetITCase.java
```diff
@@ -0,0 +1,213 @@
...
+                        4);
+        tEnv.executeSql(sourceDDL);
+        // async submit job
+        TableResult result =
+                tEnv.executeSql(String.format("SELECT table_id,table_name FROM %s", testName));
+
+        CloseableIterator<Row> iterator = result.collect();
+        waitForSnapshotStarted(iterator);
+        assertEqualsInAnyOrder(Arrays.asList(expected), fetchRows(iterator, expected.length));
+        result.getJobClient().get().cancel().get();
```
fsk119, line 189: getJobClient().ifPresent(client -> client.cancel())

fsk119 (commented):
Thanks for update. I left some comments.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
ruanhang1993: It seems that the ResultSetImpl.getString will convert the result with the right configuration. But I read this comment:
```
// DBZ-2673
        // It is necessary to check the type names as types like ENUM and SET are
        // also reported as JDBC type char
        else if ("CHAR".equals(actualColumn.typeName())
                || "VARCHAR".equals(actualColumn.typeName())
                || "TEXT".equals(actualColumn.typeName())) {
            return rs.getString(fieldNo);
        }
```
Will some other bugs occur by this change for the types like ENUM and SET ?

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MysqlConnectorCharsetITCase.java
ruanhang1993: fixed

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MysqlConnectorCharsetITCase.java
ruanhang1993: fixed

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
fsk119: DBZ-2673[1] is not about the SET or ENUM type and it is used to fix the charset problems. DBZ-2673 wants to get the raw bytes for the column with latin1 charset. It made changes:

1. override the jdbc url to set characterSetResults null(utf-8 by default)
2. get the rawbytes using rs.getBytes.

However, if the characterSetResults is utf-8, the rs.getBytes is equal to the rs.getString().getBytes("UTF-8"), which causes the problem here.

After read the fix in the DBZ-3700, I think it's on the right way to solve all the problem here[2,3]. And add more tests to verify.

[1] https://issues.redhat.com/browse/DBZ-2673
[2] https://issues.redhat.com/browse/DBZ-3700
[3] https://github.com/debezium/debezium/pull/3128/commits/6ee9ffca073be66ecce31712789672ea6df0f5c4

fsk119 (approved):
LGTM

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
ruanhang1993: I refactor the code with the DBZ-3700.
This api can not be found : connectorConfig.customConverterRegistry().isEmpty(). Do I need to revert the change in commit 7f79cb9 ?

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MysqlConnectorCharsetITCase.java
```diff
@@ -0,0 +1,218 @@
...
+                        true,
+                        getServerId(),
+                        4);
+        tEnv.executeSql(sourceDDL);
+        // async submit job
+        TableResult result =
+                tEnv.executeSql(String.format("SELECT table_id,table_name FROM %s", testName));
+
+        CloseableIterator<Row> iterator = result.collect();
+        waitForSnapshotStarted(iterator);
```
leonardBang, line 192: Make some DDL change to cover binlog reading? we can simply let `id = id + 1`

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MysqlConnectorCharsetITCase.java
```diff
@@ -0,0 +1,218 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Random;
+
+/** Integration tests for MySQL Table source. */
```
leonardBang, line 43: Give a meaningful note?

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MysqlConnectorCharsetITCase.java
```diff
@@ -0,0 +1,218 @@
...
+@RunWith(Parameterized.class)
+public class MysqlConnectorCharsetITCase extends MySqlSourceTestBase {
```
leonardBang, line 45: Could you also add a charset integration test for PolarDB-x which should be covered too?

leonardBang (commented):
Thanks @ruanhang1993 for the contribution, I left some comments

qidian99:
@ruanhang1993 Hi Ruan, I pm you some question regarding the details of this PR. I'll copy my question below -- could you also check your email? 

IMHO, there are two charset transformations happening if the charset used in JDBC connection is different from that of table columns. For instance if the column uses latin1 and jdbc uses utf8, when we read the records the following transformation will occur:
latin1 -> utf8 -> bytes

So if we directly call getBytes and let the use convert it to latin1, the first utf8 transform is missing. Therefore we should let jdbc handle charset conversion for us by calling getObject and underlyingly these transformations will happen:
bytes -> utf8 -> latin1

Please correct me if there's any misunderstanding.

ruanhang1993:
> @ruanhang1993 Hi Ruan, I pm you some question regarding the details of this PR. I'll copy my question below -- could you also check your email?
> 
> IMHO, there are two charset transformations happening if the charset used in JDBC connection is different from that of table columns. For instance if the column uses latin1 and jdbc uses utf8, when we read the records the following transformations will occur: latin1 -> utf8 -> bytes
> 
> So if we directly call getBytes and let the use convert it to latin1, the first utf8 transformation is missing. Therefore we should let jdbc handle charset conversion for us by calling getObject and underlyingly these transformations will happen: bytes -> utf8 -> latin1
> 
> Please correct me if there's any misunderstanding.

I think there is something you misunderstand.

`getObject` only helps to complete the job that converts the value by a right connection charset(default utf-8) which is set by `characterSetResults`.The column charset `latin1` only means the charset in the table column. The returned result is depended by the `characterSetResults`.

This bug is that if we return a byte[] object which is UTF-8 encoding, the debezium will convert this value to String by a `latin1` encoding. So we could return a String and the debezium will not do these processing.`getObject` will invoke `getString` for these char types finally. So we do not need the `if` statement by using it.

leonardBang:
@ruanhang1993 I open a new PR #1468 to fix the issue base on your work, please do not use you master branch as a develop branch which others can not append commits.

PR #1192, opened 2022-05-18
[mysql] Release the debezium reader thread resources after reading finished

ruanhang1993:
This PR make the reader return the thread resources after reading finished in the mysql cdc connector.


ruanhang1993:
Duplicate to #1358

PR #1193, opened 2022-05-18
[mysql] Generates multiple chunks when approximate row count is bigger than chunk size

ruanhang1993:
This PR fix the bug #1349 in splitting the chunks. The splitter should use dynamic chunk size instead of chunk size.


leonardBang (approved):
Thanks @ruanhang1993 for the work, LGTM

PR #996, opened 2022-03-24
[mysql][hotfix] serialize isSuspended flag for MySqlBinlogSplit

zhjwpku:
Signed-off-by: 元组 <zhaojunwang.zjw@alibaba-inc.com>

PR #1211, opened 2022-05-22
[mysql] Fix the binlogSplit state incompatibility when upgrading from 2.2.0

lzshlzsh:
EOFException will happen when upgrading from mysql-cdc 2.2.0 and restoring from a savepoint. The exception stack is ac follows:
`java.io.EOFException
	at org.apache.flink.core.memory.DataInputDeserializer.readBoolean(DataInputDeserializer.java:125)
	at com.ververica.cdc.connectors.mysql.source.split.MySqlSplitSerializer.deserializeSplit(MySqlSplitSerializer.java:165)
	at com.ververica.cdc.connectors.mysql.source.split.MySqlSplitSerializer.deserialize(MySqlSplitSerializer.java:124)` 

The reason is that isSuspended flag is serialized for MySqlBinlogSplit  in #996 which results in state incompatibility and we should increase serializer version normally, but we did not.

We can fix the problem simply by checking whether there is data left in buffer as  isSuspended flag is the last field.

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/split/MySqlSplitSerializer.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/split/MySqlSplitSerializer.java
@@ -161,7 +161,7 @@ public final class MySqlSplitSerializer implements SimpleVersionedSerializer<MyS
             boolean isSuspended = false;
             if (version >= 3) {
                 totalFinishedSplitSize = in.readInt();
-                if (version > 3) {
+                if (version > 3 && in.available() > 0) {
                     isSuspended = in.readBoolean();
                 }
             }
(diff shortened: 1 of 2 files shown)


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/split/MySqlSplitSerializer.java
```diff
@@ -161,7 +161,7 @@ public MySqlSplit deserializeSplit(int version, byte[] serialized) throws IOExce
             boolean isSuspended = false;
             if (version >= 3) {
                 totalFinishedSplitSize = in.readInt();
-                if (version > 3) {
+                if (version > 3 && in.available() > 0) {
```
ruanhang1993, line 164: I think this code should be like :
```
if (version > 3) {
    // Descirbe the bug in PR#996.
    if (in.available() <= 0 && version != 4) {
        throw new IllegalStateException("Expected more info for the split.");
    }
    isSuspended = in.readBoolean();
}
```
IMO, change like this will not hide bugs in other future versions. But the code I provide don't look pretty. ;)

ruanhang1993 (approved):
Thanks for your work, @lzshlzsh . LGTM.

yuxiqian:
Hi @lzshlzsh, sorry for the delay of this PR. Could you please rebase it to latest `master` branch since there’s been lots of changes in Flink CDC repo since your original commit? Kindly reminder that `com.ververica.cdc.connectors.mysql` package has been moved to `org.apache.flink.cdc.connectors.mysql`.

PR #1257, opened 2022-06-09
[update] update README.md for quick start

hunter-cloud09:
When I am using flink1.14+flink cdc 2.2.1, because the flink version depends on the source code is 1.13.5, it will prompt that the dependency org.apache.flink:flink-shaded-guava:18.0-13.0 is missing, if you refer to flink-sql-connector-mysql-cdc, these problems will not appear


ruanhang1993:
cc @leonardBang

rookiegao:
cc @leonardBang

PR #1284, opened 2022-06-16
[common] Bump Flink to 1.14.4

snuyanzin:
The PR bumps Flink version to 1.14.4
Flink's guava is bumping from 18 to 30
`blink` name is removed from dependencies

### DIFF
--- a/flink-cdc-base/pom.xml
+++ b/flink-cdc-base/pom.xml
@@ -49,14 +49,14 @@ under the License.
 
         <dependency>
             <groupId>org.apache.flink</groupId>
-            <artifactId>flink-table-planner-blink_${scala.binary.version}</artifactId>
+            <artifactId>flink-table-planner_${scala.binary.version}</artifactId>
             <version>${flink.version}</version>
             <scope>test</scope>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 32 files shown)


leonardBang (commented):
Thanks @snuyanzin for the contribution, the PR is really good, and I only left one minor comment.  Could you also update the note in all sql connector pom files[1] ?
 
[1]https://github.com/snuyanzin/flink-cdc-connectors/blob/flink144/flink-sql-connector-postgres-cdc/pom.xml#L67

snuyanzin:
thanks for pointing this
I've updated the notes.

Hmm ... after updating only the notes the build starts failing.... @leonardBang could you please tell if there is a way to restart ci without making empty commits?

leonardBang:
> thanks for pointing this I've updated the notes.
> 
> Hmm ... after updating only the notes the build starts failing.... @leonardBang could you please tell if there is a way to restart ci without making empty commits?

I did not set robot to retrigger the CI, I've rerun the failed tests, let me try to add a robot which supports re-trigger the CI

leonardBang (approved):
Thank @snuyanzin for the great work, LGTM

PR #1306, opened 2022-06-22
[oceanbase] remove mysql connector dependency and add fall back for driver name

whhe:
Update about MySQL driver.
- Remove the `mysql-connector-java` dependency from oceanbase cdc connector, this makes users be able to choose the driver according to their requirements in the future. 
- Add a fall back logic for driver class selection, and a startup option like `jdbc.driverClassName` should be added after other drivers are fully tested.

### DIFF
--- a/docs/content/connectors/oceanbase-cdc.md
+++ b/docs/content/connectors/oceanbase-cdc.md
@@ -7,6 +7,10 @@ Dependencies
 
 In order to setup the OceanBase CDC connector, the following table provides dependency information for both projects using a build automation tool (such as Maven or SBT) and SQL Client with SQL JAR bundles.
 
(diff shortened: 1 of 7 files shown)


leonardBang:
@whhe Thanks for opening this PR, I think keep a proper driver dependency is important especially for SQL API users. Do we have any reason that we have to remove the driver？ DataStream API user can also exclude the default driver and then use their customer driver.

whhe:
> @whhe Thanks for opening this PR, I think keep a proper driver dependency is important especially for SQL API users. Do we have any reason that we have to remove the driver？ DataStream API user can also exclude the default driver and then use their customer driver.

The reason of this change is that the future versions of OceanBase may require a different version of driver, and the driver may not be compatible with current database servers, so I think it's better to configure the driver by user themselves. Since this is a change in usage, it should be done as soon as possible, otherwise it may affect users who have already applied it in production at the future time.

I'm not sure how other connectors solve this problem, please tell me if you have any suggestion on this.

Issue #1345, opened 2022-07-05
[test] OracleE2eITCase.testOracleCDC failed in CI


leonardBang:
```
Jul 01 14:20:54 14:20:54.273 [ERROR] Failures: 
Jul 01 14:20:54 14:20:54.274 [ERROR]   OracleE2eITCase.testOracleCDC:170 expected:<[101,scooter,Small 2-wheel scooter,3.14, 102,car battery,12V car battery,8.1, 103,12-pack drill bits,12-pack of drill bits with sizes ranging from #40 to #3,0.8, 104,hammer,12oz carpenters hammer,0.75, 105,hammer,14oz carpenters hammer,0.875, 106,hammer,18oz carpenter hammer,1.0, 107,rocks,box of assorted rocks,5.1, 108,jacket,water resistent black wind breaker,0.1, 109,spare tire,24 inch spare tire,22.2, 111,jacket,new water resistent white wind breaker,0.5]> but was:<[101,scooter,Small 2-wheel scooter,3.14, 102,car battery,12V car battery,8.1, 103,12-pack drill bits,12-pack of drill bits with sizes ranging from #40 to #3,0.8, 104,hammer,12oz carpenters hammer,0.75, 105,hammer,14oz carpenters hammer,0.875, 106,hammer,16oz carpenters hammer,1.0, 107,rocks,box of assorted rocks,5.1, 108,jacket,water resistent black wind breaker,0.1, 109,spare tire,24 inch spare tire,22.2, 111,jacket,new water resistent white wind breaker,0.5]>
Jul 01 14:20:54 14:20:54.274 [INFO] 
Jul 01 14:20:54 14:20:54.276 [ERROR] Tests run: 12, Failures: 1, Errors: 0, Skipped: 0
```

https://dev.azure.com/ververica/flink-cdc-connectors/_build/results?buildId=1808&view=logs&j=ccedd2bf-e82a-54b8-b75f-7971f1fbbb9f&t=385827ba-1388-5c8f-6875-4ed5108d031f

PR #1348, opened 2022-07-06
[test][oracle] Fix unstable tests OracleE2eITCase due to reuse container

leonardBang:
try to fix unstable tests #1345


fsk119 (approved):
Thanks for your work. LGTM

Issue #1347, opened 2022-07-06
Flink


xuexiaowengh:
**Describe the bug(Please use English)**

**Environment :**
 - Flink version :  
 - Flink CDC version: 
 - Database and version: 

**To Reproduce**
1. Thes test data :
2. The test code :
3. The error :

**Additional Description**

xuexiaowengh:
I tried to fix the problem by:
1. Create a function to close all the connections of pools
![image](https://user-images.githubusercontent.com/64528302/177527430-91ca68fc-7be3-484e-ac33-8e24fee52d84.png)

2.call the close() when the MySqlSourceEnumerator is closing
![image](https://user-images.githubusercontent.com/64528302/177527844-7501acc0-5d3f-45e8-8077-96ebc1f608b0.png)

My test results show that this method fixes the problem.

PatrickRen:
Closing this issue because it was created before version 2.3.0 (2022-11-10). Please try the latest version of Flink CDC to see if the issue has been resolved. If the issue is still valid, kindly report it on [Apache Jira](https://issues.apache.org/jira) under project `Flink` with component tag `Flink CDC`. Thank you!

PR #1354, opened 2022-07-06
[mysql]Fix jdbc connections not be closed when job was cancelled (#1347)

xuexiaowengh:
[mysql]Fix jdbc connections not be closed when job was cancelled (#1347)

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/connection/JdbcConnectionPools.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/connection/JdbcConnectionPools.java
@@ -53,4 +53,12 @@ public class JdbcConnectionPools implements ConnectionPools {
             return pools.get(poolId);
         }
     }
+
+    public void close() {
+        for (HikariDataSource hikariDataSource : pools.values()) {
+            if (hikariDataSource != null && !hikariDataSource.isClosed()) {
+                hikariDataSource.close();
+            }
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


Matrix42:
@leonardBang Hi leonard I don't konw why mysql cdc connector use a connection pool, other connector's connection are created through the DriverManger, can you tell me why?

Matrix42:
In this issue  we can't close connection pool directly, the `JdbcConnectionPools` is a constant, it's used by all jobs in current jvm, I think we can put the `pools` to a `ThreadLocal` we can close the connection pool used by current Thread. And the connection poll not only used by `MySqlSourceEnumerator`,other place like `JdbcRowDataInputFormat`,  `JdbcOutputFormat`, `MySqlValidator` use it as well.

xuexiaowengh:
Thank you for your reply, I understand, will someone solve this, or should I try to solve it and resubmit this pr? @Matrix42

Matrix42:
I'am not the Maintainer of this repository , I think you can ask  the opinion of leonardBang @xuexiaowengh

xuexiaowengh:
Hello,Matrix42,I just tested it and found that this statement may not be true: 
“the JdbcConnectionPools is a constant, it's used by all jobs in current jvm”
When I started two FlinkCDC jobs to read the same table, and then cancelled the two jobs in turn, I found that pools were not the same:
image
image
image

so I think pools are not shared by all jobs in the JVM, or do you have another way to test the problem? @Matrix42

Matrix42:
Hi xuexiaowengh, the `JdbcConnectionPools` is single instance and thread safe, if in the same jvm it'll only have one instance, so if your job run with same taskmanger it must use the same `JdbcConnectionPools` .
I'v tested it using a simple way, tow thread in the same jvm get the `JdbcConnectionPools` ,i got the same hash code.
> Two slots in a jvm just like two thread in a jvm
image

xuexiaowengh:
Thx, Matrix42, I can see why there are two different pools in my test result because I used datastream API and included flink-connector-mysql-cdc when packing job jar. This results in a separate JdbcConnectionPools for each job; When I put flink-connector-mysql-cdc into FLINK_HOME/lib/ to test again, I can see that the two job closures are the same pools. Thank you for your previous reply. @Matrix42

leonardBang:
@xuexiaowengh I close the PR after read your discussion.

844028312:
> 谢谢，Matrix42，我明明为什么我的测试结果中有两个不同的池，因为我在打包作业jar时使用了数据流API并包含了flink-connector-mysql-cdc 。这导致每个作业都有一个单独的JdbcConnectionPools；当我将flink-connector-mysql-cdc放入FLINK_HOME/lib/ 再次测试时，可以看到这两个工作屏蔽包是同一个池。谢谢您之前的回复。@Matrix42

你好，我试了把flink-connector-mysql-cdc放入FLINK_HOME/lib/ ，但是JdbcConnectionPools不同的作业还不不共享，是需要注意什么吗

Issue #317, opened 2021-08-13
Flink-CDC  在同步字段为timestamp类型的数据时，初始化数据和增量数据时区不一致


ghost:
在使用flinkcdc stream处理时，字段类型为timestamp的字段，初始化的数据和增量binlog的数据里的timestamp类型数据时区不一致。我是临时通过isSnapshot这个字段处理了一下，希望可以修复一下这个bug。flink版本1.2.0，flinkcdc1.4.0

leonardBang:
@xiaoshijiang523  Could you post a simple demo to reproduce the issue? thanks

wuqingquan-github:
env.addSource(getCDC(DATABASE, TABLE)).print();

mysql中数据：start_time和end_time都是datetime类型
{"id_card":"362324","end_time":2021-08-14 08:55:11,"start_time":2021-08-14 08:44:08,"id":72}

不管是否添加配置：mysql.serverTimezone=Asia/Shanghai，结果都一样
"data":{"id_card":"362324","end_time":1628931311000,"start_time":1628930648000,"id":72}
打印的结果转换为LocalDateTime分别是2021-08-14 16:44:08和2021-08-14 16:55:11

 public static DebeziumSourceFunction<JSONObject> getCDC(String databaseList, String tableList) {
        return MySQLSource.<JSONObject>builder()
                .hostname(HOSTNAME)
                .port(3306)
                .username(USERNAME)
                .password(PASSWORD)
                .databaseList(databaseList)
                .tableList(tableList)
                .deserializer(new MyDeserializer())
                .build();
}

wuqingquan-github:
env.addSource(getCDC(DATABASE, TABLE)).print();

mysql中数据：start_time和end_time都是datetime类型
{"id_card":"362324","end_time":2021-08-14 08:55:11,"start_time":2021-08-14 08:44:08,"id":72}

不管是否添加配置：SERVER_TIME_ZONE=Asia/Shanghai，结果都一样
"data":{"id_card":"362324","end_time":1628931311000,"start_time":1628930648000,"id":72}
打印的结果转换为LocalDateTime分别是2021-08-14 16:44:08和2021-08-14 16:55:11

    public static DebeziumSourceFunction<JSONObject> getCDC(String databaseList, String tableList) {
        return MySQLSource.<JSONObject>builder()
                .hostname(HOSTNAME)
                .port(3306)
                .username(USERNAME)
                .password(PASSWORD)
[...]

PR #1366, opened 2022-07-12
[Debezium] Fix the server timeZone bug of debeziumDeserialization when using dataSream API（#317）

youyangkou:
Fix the server timeZone bug of debeziumDeserialization when using dataSream API（#317）


youyangkou:
@leonardBang please help to review this pr. The test failure seems not related to my modification.

flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/JsonDebeziumDeserializationSchema.java
```diff
@@ -46,35 +47,53 @@ public class JsonDebeziumDeserializationSchema implements DebeziumDeserializatio
...
+     * The custom configurations for {@link JsonConverter}.
+     */
     private Map<String, Object> customConverterConfigs;
 
     public JsonDebeziumDeserializationSchema() {
         this(false);
     }
 
     public JsonDebeziumDeserializationSchema(Boolean includeSchema) {
+        this(includeSchema, 8);
```
leonardBang, line 62: Thanks @youyangkou for the contribution, but the PR looks like work in a wrong way. 
Firstly, the  timeZoneOffset `8` works is  that you happen to be in GMT+8 timezone, for other timezone it doesn't work.
Secondly, Not all source need consider the timezone issue, for example, pg doesn't exist the timezone issue, and even in MySQL, only the record from changelog need consider the timezone offset.

yuxiqian:
Hi @XXXX, thanks for your contribution! Seems this issue has been fixed in #1407, so I'm going to close this PR. Feel free to reopen it if you have other concerns.

PR #1369, opened 2022-07-13
[oceanbase] fix regex for change events

whhe:
`fnmatch` is used to match pattern of change events, while `regex` is used for snapshot reading. 

In this pr, I make the log client to fetch all the data at first and do filter with regex in the process part.


whhe:
Please help review this pr when you are available. @leonardBang @GOODBOY008 @Jiabao-Sun 

The ci failure is not related to my modification.

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/OceanBaseSource.java
```diff
@@ -219,7 +217,7 @@ public SourceFunction<T> build() {
             }
             obReaderConfig.setUsername(username);
             obReaderConfig.setPassword(password);
-            obReaderConfig.setTableWhiteList(tableWhiteList);
+            obReaderConfig.setTableWhiteList(String.format("%s.*.*", tenantName));
```
Jiabao-Sun, line 220: Can we specify an explicit database or table name, when the regex matches only one database or table?

Jiabao-Sun (commented):
Thanks @whhe for the excellent work to push this feature forward.

From my point of view, I think we can do some optimizations mentioned in the comments, especially in the scenario of a single database or a single table. 

In addition, using not fully qualified table regex does not seem to be able to partially filter the tables with same name but in different databases. Shall we use the fully qualified `tableList` like MySQL CDC connector?

CC @leonardBang

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/OceanBaseSource.java
whhe: Change events reading is started before snapshot reading to ensure no data loss for now, so it's hard do what you say.

whhe:
> Thanks @whhe for the excellent work to push this feature forward.
> 
> 
> 
> From my point of view, I think we can do some optimizations mentioned in the comments, especially in the scenario of a single database or a single table. 
> 
> 
> 
> In addition, using not fully qualified table regex does not seem to be able to partially filter the tables with same name but in different databases. Shall we use the fully qualified `tableList` like MySQL CDC connector?
> 
> 
> 
> CC @leonardBang 

Thanks for your advice. 

I used regex here at the beginning and so I keep it in this pr. And it make sense to use fully qualified table list, the monitor list can be more accurate in this way. I would change it asap.

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/OceanBaseSource.java
whhe: I reconsidered your suggestion, we can get the table list at the begining and then start to read change events. I will try to support this.

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/OceanBaseSource.java
```diff
@@ -54,6 +54,7 @@ public static <T> Builder<T> builder() {
         private String tenantName;
         private String databaseName;
         private String tableName;
+        private String tableList;
```
Jiabao-Sun, line 57: Hi @whhe, shall we use `databaseList` and `tableList` to keep the same form and the tableName field seems no longer needed.

https://github.com/ververica/flink-cdc-connectors/blob/02eca305d2682fcc0bfd626adb6db1d8b3213a9d/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java#L55-L56

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/table/OceanBaseTableSourceFactory.java
```diff
@@ -64,13 +64,22 @@ public class OceanBaseTableSourceFactory implements DynamicTableSourceFactory {
...
-                    .withDescription("Table name of OceanBase to monitor.");
+                    .withDescription(
+                            "Table name of OceanBase to monitor, should be regular expression. Only can be used with 'initial' mode.");
+
+    public static final ConfigOption<String> TABLE_LIST =
+            ConfigOptions.key("table-list")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "List of full names of tables, separated by commas, e.g. \"db1.table1, db2.table2\".");
```
Jiabao-Sun, line 82: For some compatibility reasons, I think we can just keep `TABLE_NAME` config option to accept "db1.table1, db2.table2".

Jiabao-Sun (commented):
Thanks @whhe. I left a few comments.

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/OceanBaseSource.java
whhe: I noticed that MySqlSource is using `databaseList` and `tableList`, but I'm still not clear about why do we need a `databaseList` while `tableList` already contains the db name in every item. Is there something I missed?

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/table/OceanBaseTableSourceFactory.java
whhe: `database-name` and `table-name` are used as regex, and `table-list` is used as exact value, and they can also be used together. That's my original plan.

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/OceanBaseSource.java
Jiabao-Sun: The `databaseList` and `tableList` are carried from debezium then other CDC connectors also keep the same form.

| Property | Description  |
|:-:|:-:|
|database.include.list | An optional, comma-separated list of regular expressions that match the names of the databases for which to capture changes. The connector does not capture changes in any database whose name is not in database.include.list. By default, the connector captures changes in all databases. Do not also set the database.exclude.list connector confiuration property. |
| table.include.list | An optional, comma-separated list of regular expressions that match fully-qualified table identifiers of tables whose changes you want to capture. The connector does not capture changes in any table not included in table.include.list. Each identifier is of the form databaseName.tableName. By default, the connector captures changes in every non-system table in each database whose changes are being captured. Do not also specify the table.exclude.list connector configuration property. |

Please refer to https://debezium.io/documentation/reference/1.6/connectors/mysql.htm.

Jiabao-Sun: The database(s) and table(s) regex feature is mostly used in the stream api. 
The `databaseList` or `tableList` can be used independently as described above.

whhe: So `databaseList` is used to sync the whole database in the list? If it's ture, then `databaseList` might be similar with `table-name` regex in usage. If we keep the options like `database-name` and `table-name`, `databaseList` would be unnecessary.

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/table/OceanBaseTableSourceFactory.java
whhe: Keep `database-name` and `table-name` in sql option will also make the configuration the same with stream api, which may be better than using different config in different mode.

Jiabao-Sun: > Keep `database-name` and `table-name` in sql option will also make the configuration the same with stream api, which may be better than using different config in different mode.

Yes, I agree with that. 
However, for users, there will be some differences using different connectors, which may increase the learning cost.

whhe: > > Keep `database-name` and `table-name` in sql option will also make the configuration the same with stream api, which may be better than using different config in different mode.
> 
> Yes, I agree with that. However, for users, there will be some differences using different connectors, which may increase the learning cost.

I understand your concern, I will add instruction in document to help users figure out how to set monitor list.

whhe:
@Jiabao-Sun Docs updated, please take a look when you are available.

Jiabao-Sun:
Thanks @whhe for the update.
Hi @leonardBang, do you have some suggestions?

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/table/OceanBaseTableSourceFactory.java
```diff
@@ -163,6 +172,8 @@ public DynamicTableSource createDynamicTableSource(Context context) {
         String tenantName = config.get(TENANT_NAME);
         String databaseName = config.get(DATABASE_NAME);
         String tableName = config.get(TABLE_NAME);
+        String tableList = config.get(TABLE_LIST);
```
leonardBang, line 175: we can add validation for this three options like  (1) `tableName ` is required when `databaseName` is configured, (2) `tableList` is required when `databaseName` and `tableName` are not configured
https://github.com/ververica/flink-cdc-connectors/blob/7de1587ca83fa98a0ab6cbc54b13c80f620e7f51/flink-connector-oracle-cdc/src/main/java/com/ververica/cdc/connectors/oracle/OracleSource.java#L190

docs/content/connectors/oceanbase-cdc.md
```diff
@@ -94,6 +94,12 @@ Flink SQL> SELECT * FROM orders;
 Connector Options
 ----------------
 
+The OceanBase CDC Connector contains some options for both sql and stream api as the following sheet. 
+
+*Note*: The connector supports two ways to specify the table list to listen to, and will get the union of the results when both way are used at the same time.
+1. Use `database-name` and `table-name` to match database and table names in regex. Can only be used in `initial` startup mode.
```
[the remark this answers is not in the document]
whhe: Because the 'liboblog' which used to fetch change events log from OceanBase can only support fnmatch for now. If we want to use regex for change events, we must fetch all the log data firstly and do filter in the connector.

flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/table/OceanBaseTableSourceFactory.java
whhe: Done.

docs/content/connectors/oceanbase-cdc.md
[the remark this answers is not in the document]
leonardBang: Got it， Could we also add the background to the note?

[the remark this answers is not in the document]
whhe: OK.

whhe:
@leonardBang Comments addressed. PTAL.

leonardBang (approved):
Thanks @whhe for the update, LGTM. Btw, the logic in `OceanBaseRichSourceFunction` is too long to read, we can split it in the future.

PR #1392, opened 2022-07-22
[mysql] Optimize pure binlog phase check logic to improve performance

lzshlzsh:
**Pure binlog phase performace opt**
Our online business found that the performance of the binlog phase of Flink jobs could not meet the demand. Through performance analysis, we found that the performance bottleneck in the phase was the comparison of binlog offset. For each binlog data, mysql-cdc needs to judge whether the offset of the binlog data is after the end of the full snapshot phase (max split high watermark). If so, it is in pure binlog phase, and can be directly output to the downstream. The comparison between binlog data offset and max split high watermark consumes CPU very much and has become a performance bottleneck (see the following figure).
Further analysis of the internal logic of mysql-cdc shows that for each mysql-cdc table, the state of incremental synchronization will remain unchanged after entering the pure binlog phase. Therefore, it is sufficient to keep a flag for each table to judge whether the table has entered pure binlog phase, so as to avoid binlog offset comparison of each data in the pure binlog phase, Improve the performance of incremental data synchronization.
image

**Performance improvement**
In the actual online scenario test (more than 180 tables are synchronized with many table fields in a flink job), the performance is improved by 3 times (from 5k/s to 2w/s), which meets the real-time synchronization needs of our business.

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
@@ -43,9 +43,11 @@ import javax.annotation.Nullable;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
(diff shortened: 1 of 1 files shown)


lzshlzsh:
@leonardBang Would you have a look at if there is any problem with this optimization?

leonardBang (approved):
Thanks @lzshlzsh for the great work, I like your detail analysis, the optimization makes sense to me, LGTM

PR #1407, opened 2022-07-26
[mysql] Use local timezone as the default server timezone

kylemeow:
Due to the issue mentioned in https://github.com/ververica/flink-cdc-connectors/issues/1406, Flink CDC Connectors might treat Timestamp fields with different behaviors during snapshot and streaming phases.

Here I propose changing the default value of `server-time-zone` to local timezone so that most users would not need to manually set timezone to get the correct results.


tisonkun:
Spotless failed. It seems you should run `mvn spotless:apply` and push the diff :)

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/table/MySqlTableSourceFactory.java
```diff
@@ -305,4 +308,20 @@ && doubleCompare(distributionFactorLower, 1.0d) <= 0,
                         1.0d,
                         distributionFactorLower));
     }
+
+    /** Replaces the default timezone placeholder with local timezone, if applicable. */
+    private static ZoneId getServerTimeZone(ReadableConfig config) {
+        String timeZone = config.get(SERVER_TIME_ZONE);
```
syhily, line 314: `config.contains(SERVER_TIME_ZONE)` could be better?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceOptions.java
```diff
@@ -68,8 +68,10 @@ public class MySqlSourceOptions {
     public static final ConfigOption<String> SERVER_TIME_ZONE =
             ConfigOptions.key("server-time-zone")
                     .stringType()
-                    .defaultValue("UTC")
-                    .withDescription("The session time zone in database server.");
+                    .defaultValue("default")
```
syhily, line 71: Use `noDefaultValue()` instead of `.defaultValue("default")`?

kylemeow:
OracleE2eITCase.testOracleCDC failed, but it seems unrelated to the changes made.

leonardBang:
> OracleE2eITCase.testOracleCDC failed, but it seems unrelated to the changes made.

@kylemeow Thanks for the contribution, I've fixed the test, could you rebase master and try again?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceOptions.java
```diff
@@ -68,8 +68,10 @@ public class MySqlSourceOptions {
     public static final ConfigOption<String> SERVER_TIME_ZONE =
             ConfigOptions.key("server-time-zone")
                     .stringType()
-                    .defaultValue("UTC")
-                    .withDescription("The session time zone in database server.");
+                    .noDefaultValue()
+                    .withDescription(
+                            "The session time zone in database server. If not set, then "
```
leonardBang, line 73: I've considered this pain spot too, about the inferred default timezone, how about use server's timezone as default value firstly and then use the current system timezone. Of course that user's configuration own the most high level priority.

leonardBang: we can use `SELECT @@GLOBAL.time_zone;` statement to query the server timezone
https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html

leonardBang (commented):
Thanks @kylemeow for the contribution, I left some comments

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceOptions.java
kylemeow: Hi @leonardBang , thanks for the insightful comment, and I am looking into this issue but have some doubts about this solution.

One thing is that, as the `time_zone` returned by MySQL could be an offset like '+8:00', there are lots of different timezone names matching this offset like 'Asia/Chongqing', 'Asia/Singapore', 'Australia/Perth', etc.  If we arbitrarily choose a timezone name matching the offset, we may miss some traits like 'Daylight saving time' or other local time adjustments. Also, 'CST' could mean 'China Standard Time' or 'Central Standard Time' in the US, so the time_zone value returned by MySQL might not be enough to accurately determine this value.

Also, in some cases, the client might not be able to directly connect to the MySQL server during Logical Plan generation, thus if we make active probes to the remote server at that time, Flink may get stuck for a long time, which is not friendly to users. Alternatively, if we move the timezone detection logic to the runtime, this is not trivial work anyway.

What do you think about these? Thank you again for the review : )

leonardBang: @kylemeow Thanks for sharing your opinions, Happy to see you've considered the timezone name and DST mechanism.  Firstly, We can use `SELECT TIMEDIFF(NOW(), UTC_TIMESTAMP)` to get the  accurate timezone offset which has been considered the  'CST' that may mean 'China Standard Time' or 'Central Standard issue, so CST is not a problem here. 

About DST issue, as the returned offset may change due to the DST exchange, the client need to know the adjust the change dynamically, but unfortunately we(flink cdc or flink) only load the timezone or time offset once, thus the timestamp conversion would be wrong after the DST coming or leaving. Although MySQL deals temporal function and timestamp wrongly under DST zones until `8.0.26` IIUC [1],  it's still an issue in the long term. In a word, query server timezone might get wrong result if MySQL server's timezone used DST. 

Using system's default server timezone also exists some issue if user's database and flink cluster are deployed in different timezone. 

>MySQL server during Logical Plan generation

We've visited MySQL server  in compile phase(validator), it is not a new introducing issue.
[1] https://bugs.mysql.com/bug.php?id=94257

leonardBang: @kylemeow Just like we have two poisons, one is wrong result by DST, the other is wrong result by  system's default timezone. Let's drink the later one, maybe we can reduce the influence by add validation for the MySQL server  and Flink Cluster are deployed under different timezone.

kylemeow: Thanks @leonardBang for the detailed explanation. As for the 'Logical Plan generation' part, sorry for the ambiguity, I originally meant that when `MySqlTableSourceFactory` generates timezone settings and Debezium properties during Flink logical plan generation process, the client-side JVM might not be able to directly connect to the MySQL server to execute SQL queries.

But anyway, users must guarantee that MySQL server is accessible at runtime, thus we could do the timezone verification at runtime, e.g. in `com.ververica.cdc.debezium.DebeziumSourceFunction#open`, and when we find there is a difference between the inferred timezone and actual MySQL timezone, we could print a warning message or throw an exception.

leonardBang: Using system's timezone as default timezone and validating it during runtime makes sense to me, An exception is better as it will lead to the data correctness.

kylemeow: Added runtime timezone check as recommended : )

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/MySqlTestUtils.java
```diff
@@ -59,6 +59,7 @@ public static MySqlSource.Builder<SourceRecord> basicSourceBuilder(
                 .username(database.getUsername())
                 .password(database.getPassword())
                 .deserializer(new ForwardDeserializeSchema())
+                .serverTimeZone("UTC")
```
leonardBang, line 62: minor: only `testValidateTimezone` need set the parameter, but we set it for all legacy tests.

leonardBang, line 62: minor: only `testValidateTimezone` need this parameter, but we set it as a fixed value for all legacy tests.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/MySqlValidator.java
```diff
@@ -157,4 +161,36 @@ private void checkBinlogRowImage(JdbcConnection connection) throws SQLException
...
+        if (timeZoneProperty == null) {
+            LOG.warn(
+                    "{} is not set, which might cause data inconsistencies for time-related fields.",
+                    SERVER_TIME_ZONE.key());
+            return;
+        }
+
+        ZoneId timeZone = ZoneId.of(timeZoneProperty);
+        int timeZoneOffsetInSeconds =
+                timeZone.getRules().getOffset(LocalDateTime.now()).getTotalSeconds();
```
leonardBang, line 177: I have some concerns about the validation:
(1) The timezone may have multiple validate offsets.
(2)`timeDiffInSeconds != timeZoneOffsetInSeconds` it too strictly because the returned value of `TIME_TO_SEC(TIMEDIFF(NOW(), UTC_TIMESTAMP))` may be different seconds in some DBMS.

leonardBang (commented):
Thanks @kylemeow for the contribution, I left a few comments

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/MySqlTestUtils.java
kylemeow: Hi @leonardBang , the reason for setting `serverTimezone` to UTC is that the containerized MySQL for testing runs in UTC timezone by default, however, the test cases might be run on computers at different timezones, which could eventually cause test failures, and this is actually desired as we need to throw an exception when the local timezone is different from the server and no serverTimeZone is set by the user.

One possible solution is to dynamically append the local `default-time-zone`  settings to `test/resources/docker/server/my.cnf` and `docker/server-gtids/my.cnf`, so that the containerized MySQL runs in the local timezone and the test cases pass.

Another solution is to manually set the `serverTimezone` to UTC, so that the source has the same timezone as the server. As we have already tested timezone auto-detection cases in 'MySqlTableSourceFactoryTest.java', personally I think this approach is simpler.

Just my two cents, and looking for your reply :)

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/MySqlValidator.java
kylemeow: Thanks for the insights.

Yes, for question 1), take Los Angeles for example, when Daylight Saving Time begins, the actual timezone shifts from 'Pacific Standard Time (UTC-8)' to 'Pacific Daylight Time (UTC-7)', and some DBAs choose to change timezone accordingly, others might choose to stay with the standard timezone all year long. Thus we need to allow both cases, and I have loosened the check in the latest commit.

For the question 2), since this MR only focuses on MySQL, I have tested on 5.6, 5.7, 8.0, it seems that the returned value of `TIME_TO_SEC(TIMEDIFF(NOW(), UTC_TIMESTAMP))` behaves as expected so far. I wonder if there are other cases that I haven't found out?

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/MySqlTestUtils.java
leonardBang: I mean MySqlTestUtils#basicSourceBuilder method should has a `serverTimezone` parameter which is more flexible

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/MySqlValidator.java
leonardBang: For the question 2  I mean other MySQL vendors like PolarDB-x/TDSQL and so on. But I think we can focus on MySQL currently and adjust our strategy in the fututre if other vendors can not pass the check.

kylemeow: Hi @leonardBang , you are right that the newly added PolarDB-X test case throw exceptions like '14c6666538c00000][172.17.0.2:8527][polardbx]ERR-CODE: [PXC-4518][ERR_VALIDATE] : Column 'UTC_TIMESTAMP' not found in any table' when executing TIMEDIFF queries, because it mistakenly treats `UTC_TIMESATMP` as a column name.

After changing `UTC_TIMESTAMP` to `UTC_TIMESTAMP()`, this problem is settled. Anyway, TDSQL-C behaves the same as MySQL and treats `UTC_TIMESTAMP` the same as `UTC_TIMESTAMP()` call.

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/MySqlTestUtils.java
kylemeow: Hi @leonardBang , I have added a `serverTimezone` parameter in this method as recommended. However, AFAIK when running test suites in non-UTC timezones, most of the existing MySQL test cases need to be explicitly set to UTC timezone in accordance with the MySQL server timezone settings, so the actual argument for this method should still be 'UTC' : )

leonardBang (approved):
Thanks @kylemeow for the nice work, so happy to discuss with you, LGTM.

Issue #1410, opened 2022-07-26
[TiDB] TiKVRichParallelSourceFunction NPE if table not found


link3280:
**Describe the bug(Please use English)**

**Environment :**
 - Flink version :  1.14.0
 - Flink CDC version: 2.2.1
 - Database and version: TiDB 5.1.4

**To Reproduce**
1. The test data : read a non-exist table via TiDB CDC connector
2. The test code :
3. The error : 
```
java.lang.NullPointerException
	at com.ververica.cdc.connectors.tidb.TiKVRichParallelSourceFunction.open(TiKVRichParallelSourceFunction.java:104)
	at org.apache.flink.api.common.functions.util.FunctionUtils.openFunction(FunctionUtils.java:34)
	at org.apache.flink.streaming.api.operators.AbstractUdfStreamOperator.open(AbstractUdfStreamOperator.java:100)
	at org.apache.flink.streaming.runtime.tasks.RegularOperatorChain.initializeStateAndOpenOperators(RegularOperatorChain.java:110)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restoreGates(StreamTask.java:711)
	at org.apache.flink.streaming.runtime.tasks.StreamTaskActionExecutor$SynchronizedStreamTaskActionExecutor.call(StreamTaskActionExecutor.java:100)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restoreInternal(StreamTask.java:687)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restore(StreamTask.java:654)
	at org.apache.flink.runtime.taskmanager.Task.runWithSystemExitMonitoring(Task.java:958)
	at org.apache.flink.runtime.taskmanager.Task.restoreAndInvoke(Task.java:927)
	at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:766)
	at org.apache.flink.runtime.taskmanager.Task.run(Task.java:575)
	at java.lang.Thread.run(Thread.java:748)
```

**Additional Description**

PR #1411, opened 2022-07-26
[tidb] Fix NPE when the specified table not found (#1410)

link3280:
As indicated by the title.


link3280:
cc @leonardBang @GOODBOY008. Should be a quick one, thanks!

leonardBang (approved):
Thanks @link3280 for the fix, LGTM

PR #1431, opened 2022-07-29
[metric] fix flink cdc metric conflict with flink built in metric

zhmin:
CurrentEmitEventTimeLag and currentFetchEventTimeLag metrics are both registered in flink cdc and flink framework.Flink MetricRegistryImpl is responsible for metric registering, when the metric with same name is registered twice, only the first will be successfully. 

Flink cdc metric register after flink framework, so  flink cdc metric will be ignored. The following log proves it: 
2022-07-29 11:41:58.031 WARN org.apache.flink.metrics.MetricGroup     [] - Name collision: Group already contains a Metric with the name 'currentEmitEventTimeLag'.Metric will not be reported.

To deal with the conflict, i suggest  add cdc prefix to metric name.

### DIFF
--- a/flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/DebeziumSourceFunction.java
+++ b/flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/DebeziumSourceFunction.java
@@ -433,7 +433,7 @@ public class DebeziumSourceFunction<T> extends RichSourceFunction<T>
                 getRuntimeContext().getClass().getMethod("getMetricGroup");
         getMetricGroupMethod.setAccessible(true);
         final MetricGroup metricGroup =
-                (MetricGroup) getMetricGroupMethod.invoke(getRuntimeContext());
+                ((MetricGroup) getMetricGroupMethod.invoke(getRuntimeContext())).addGroup("cdc");
 
         metricGroup.gauge(
                 "currentFetchEventTimeLag",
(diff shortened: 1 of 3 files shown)


ruanhang1993:
Thanks for your work , @zhmin .
I run the test `MySqlConnectorITCase#testConsumingAllEvents` in debug mode. It seems that both `sourceIdleTime` and `currentEmitEventTimeLag` will be ignored, which are registered in the parent metric group. But the `currentFetchEventTimeLag ` metric is not ignored when I run the test. The code in Flink lies in the `InternalSourceReaderMetricGroup.java`.
截屏2022-07-29 下午5 50 17

I don't think renaming is the best way to fix this. Maybe we need a new own metric group for our metrics. Or we should use the metrics in the parent metric group. 
@leonardBang What do you think ?

zhmin:
yes, you are right! Flink cdc metric is registerde in SourceOperator.open method, while flink metric is registerde in SourceOperator.setup method.
If we don't rename metric, creating sub metric group for cdc metric, is very good idea.
The code we modifide is less,  just call addGroup method to create new metric group.
```java 
MetricGroup taskManagerMetricGroup= (MetricGroup) metricGroupMethod.invoke(readerContext);
final MetricGroup metricGroup = taskManagerMetricGroup.addGroup("cdc")
```

lvyanquan:
Meet the same situation, Is this PR still in progress? willing to complete the subsequent development work If need.

zhmin:
@ruanhang1993 @lvyanquan can anyone help with code review？

lvyanquan:
It works for me.

yuxiqian:
Hi @zhmin, thanks for your contribution! Seems this issue has been fixed in #2415, so I'm going to close this PR. Feel free to reopen it if you have other concerns.

PR #1443, opened 2022-08-03
[oceanbase] use utc offset as timezone by default

whhe:
Similar to [MySQL](https://dev.mysql.com/doc/refman/5.7/en/time-zone-support.html#time-zone-variables), OceanBase may not support named timezone like 'UTC' and 'PST'. So it's better to use UTC offset here.

### DIFF
--- a/docs/content/connectors/oceanbase-cdc.md
+++ b/docs/content/connectors/oceanbase-cdc.md
@@ -201,9 +201,9 @@ The OceanBase CDC Connector contains some options for both sql and stream api as
             <tr>
                 <td>server-time-zone</td>
                 <td>optional</td>
-                <td style="word-wrap: break-word;">UTC</td>
+                <td style="word-wrap: break-word;">+00:00</td>
                 <td>String</td>
-                <td>The session time zone in database server, e.g. "Asia/Shanghai". It controls how the TIMESTAMP type in OceanBase converted to STRING in snapshot reading, please make sure to set it same with the timezone of `oblogproxy` deployment. </td>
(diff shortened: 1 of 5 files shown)


docs/content/connectors/oceanbase-cdc.md
```diff
@@ -201,9 +201,9 @@ The OceanBase CDC Connector contains some options for both sql and stream api as
             <tr>
                 <td>server-time-zone</td>
                 <td>optional</td>
-                <td style="word-wrap: break-word;">UTC</td>
+                <td style="word-wrap: break-word;">+00:00</td>
                 <td>String</td>
-                <td>The session time zone in database server, e.g. "Asia/Shanghai". It controls how the TIMESTAMP type in OceanBase converted to STRING in snapshot reading, please make sure to set it same with the timezone of `oblogproxy` deployment. </td>
+                <td>The session timezone which controls how temporal types are converted to STRING in OceanBase. Can be UTC offset in format "±hh:mm", or named time zones if the time zone information tables in the mysql database have been created and populated.</td>
```
leonardBang, line 206: The offset way may lead to incorrect  result under the timezone which uses DayLightSaving, these area will have multiple offsets when coming or leaving  DayLightSaving. Thus maybe named timezone is better.

whhe: > The offset way may lead to incorrect result under the timezone which uses DayLightSaving, these area will have multiple offsets when coming or leaving DayLightSaving. Thus maybe named timezone is better.

Yes, named timezone is better, and this modification still supports it if these time zone information tables like `mysql.time_zone_name` are populated. Here we only changed the default value in the logic, as these `mysql.*` tables are empty by default.

leonardBang (approved):
Thanks @whhe for the contribution, LGTM

PR #1446, opened 2022-08-04
[hotfix][test] fix the error in MySqlSourceITCase when adding the savepoint path

ruanhang1993:
`StreamExecutionEnvironment#getConfiguration` now returns an unmodifiable configuration, which will cause test failing.


leonardBang (approved):

PR #1451, opened 2022-08-05
[hotfix][test] Fix testMetadataColumns error in xxxConnectorITCase introduced by FLINK-20370, which will send -U records when primary keys of source and sink is different

ruanhang1993:
Fix testMetadataColumns error in xxxConnectorITCase introduced by FLINK-20370, which will send -U records when primary keys of source and sink is different


leonardBang (approved):

PR #1452, opened 2022-08-07
[shell] fix shell script 

legendtkl:
Minor change: the '#!/bin/bash' line in shell script must be set at first line

### DIFF
--- a/flink-connector-oracle-cdc/src/test/resources/docker/assets/activate-archivelog.sh
+++ b/flink-connector-oracle-cdc/src/test/resources/docker/assets/activate-archivelog.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 #############################################################
 #  Copyright 2022 Ververica Inc.
 #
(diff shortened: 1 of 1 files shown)


flink-connector-oracle-cdc/src/test/resources/docker/assets/activate-archivelog.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
leonardBang, line 1: Thanks @legendtkl for the contribution, but I checked the scripts works well even the `#!/bin/bash` is not placed on the first line. Could you explain more about the limitation?

legendtkl: ref this: https://stackoverflow.com/questions/8967902/why-do-you-need-to-put-bin-bash-at-the-beginning-of-a-script-file

leonardBang: Thanks @legendtkl for the clarification

leonardBang (approved):
LGTM

PR #1461, opened 2022-08-09
[mongodb] Expose updateLookup option for stream source.

Jiabao-Sun:
When sinking to a store that supports partial updates, such as Elasticsearch, we may not need to get the full document of the changed stream record.
Therefore, the configuration of updateLookup is exposed for the stream api to reduce the overhead brought by update lookup.


Jiabao-Sun:
Hi @GOODBOY008, please help review it when you have time.

GOODBOY008 (approved):
@Jiabao-Sun Thanks for your contribution,LGTM

PR #1157, opened 2022-05-07
[mysql] startup mode support of  timestamp and earliest

hiliuxg:
### feature 
1. startup mode support of  timestamp and earliest
2. support of gtid mode

### demo in java
```java
MySqlSource<String> mySqlSource =
                MySqlSource.<String>builder()
                        .hostname("172.23.218.236")
                        .port(3308)
                        .databaseList("db_test") // set captured database
                        .tableList("db_test.t_user") // set captured table
                        .username("test")
                        .password("123")
                        .serverId("5401-5404")
                        .startupOptions(StartupOptions.earliest())
                        .deserializer(new JsonDebeziumDeserializationSchema())
                        .includeSchemaChanges(true) // output the schema changes as well
                        .build();

StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
env.enableCheckpointing(10000);

env.fromSource(mySqlSource, WatermarkStrategy.noWatermarks(), "mysqlBinlogSource") .print() .setParallelism(1);

env.execute("Print MySQL Snapshot + Binlog");
```

### demo in sql
```sql
CREATE TABLE t_user (
    id int,
    name varchar,
    age int,
    create_time timestamp,
    PRIMARY KEY(id) NOT ENFORCED
) WITH (
    'connector' = 'mysql-cdc',
    'scan.startup.mode' = 'timestamp',
    'scan.startup.timestamp-millis' = '1652068405000',
    'hostname' = '172.23.211.64',
    'port' = '3307',
    'username' = 'test',
    'password' = '123',
    'database-name' = 'db_test',
    'table-name' = '.*'
)
```

PR #1465, opened 2022-08-10
[WIP] [mysql] Support starting from specific timestamp in the new MySqlSource

qidian99:


ruanhang1993:
Duplicated to #1157.
Besides that, this code change filters the records by the timestamp instead of reading from the timestamp.

ruanhang1993:
cc @leonardBang

PR #1493, opened 2022-08-19
[test] Remove usage of deprecated EnvironmentSettings.Builder#useBlinkPlanner

snuyanzin:
`org.apache.flink.table.api.EnvironmentSettings.Builder#useBlinkPlanner` became deprecated in 1.14.x when there is only one planner left.

In 1.15.x it is removed => makes sense remove here to simplify future updates


leonardBang (approved):
Thanks @snuyanzin for the contribution, LGTM. 
Btw, blink planner is officially as the default planner since Flink 1.13 by [FLINK-21709](https://issues.apache.org/jira/browse/FLINK-21709)

PR #1499, opened 2022-08-23
[cdc-base][mysql] Fix split key comparison when may contain different types

### LINKED ISSUE #1496: The split key comparison is problematic if the key may contains different types
This is related to #1414

leonardBang:
The PR aims to fix #1496


ruanhang1993 (approved):
LGTM.

Issue #1363, opened 2022-07-12
Support for Flink 1.15


tigrulya-exe:
Hello! Are there any plans in the near future to support version 1.15 of Flink?

leonardBang:
Yes, we will support 1.15 @tigrulya-exe, let me think how to support multiple versions(1.13&1.14&1.15)

tigrulya-exe:
Hello, @leonardBang! Thanks for reply, but is it even possible to support all of this Flink versions in one CDC-connectors version without backporting Flink 1.15 features to older versions of Flink? At first, I thought about using different maven profiles in CDC for different Flink versions, but, as far as I understand, Flink 1.15 also brings some code, that breaks backward compatibility (e.g. removing old planner, LogicalTypeChecks.hasFamily removal, etc.). So, maybe it will be more simpler to introduce separate version for 1.15+?

tigrulya-exe:
Hello, @leonardBang! I've added PR with Maven profiles approach - #1504. Could you take a look, please?

leonardBang:
Thanks @tigrulya-exe for opening this PR， @PatrickRen Could you help review this PR？

PR #1504, opened 2022-08-24
[common] Add Flink 1.15 support. 

tigrulya-exe:
Bumped Flink version to 1.15.1. 
Added Maven profile for Flink 1.14 compatibility.
Relates to #1363

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/table/MySqlDeserializationConverterFactory.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/table/MySqlDeserializationConverterFactory.java
@@ -21,7 +21,6 @@ import org.apache.flink.table.data.StringData;
 import org.apache.flink.table.types.logical.ArrayType;
 import org.apache.flink.table.types.logical.LogicalType;
 import org.apache.flink.table.types.logical.LogicalTypeFamily;
-import org.apache.flink.table.types.logical.utils.LogicalTypeChecks;
 
 import com.esri.core.geometry.ogc.OGCGeometry;
 import com.fasterxml.jackson.databind.JsonNode;
@@ -115,8 +114,7 @@ public class MySqlDeserializationConverterFactory {
 
     private static Optional<DeserializationRuntimeConverter> createArrayConverter(
             ArrayType arrayType) {
-        if (LogicalTypeChecks.hasFamily(
-                arrayType.getElementType(), LogicalTypeFamily.CHARACTER_STRING)) {
+        if (hasFamily(arrayType.getElementType(), LogicalTypeFamily.CHARACTER_STRING)) {
             // only map MySQL SET type to Flink ARRAY<STRING> type
             return Optional.of(
                     new DeserializationRuntimeConverter() {
@@ -147,5 +145,9 @@ public class MySqlDeserializationConverterFactory {
             // otherwise, fallback to default converter
             return Optional.empty();
         }
+    }
+
+    private static boolean hasFamily(LogicalType logicalType, LogicalTypeFamily family) {
+        return logicalType.getTypeRoot().getFamilies().contains(family);
     }
 }
--- a/flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
+++ b/flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
@@ -57,8 +57,8 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static com.ververica.cdc.connectors.mysql.LegacyMySqlSourceTest.currentMySqlLatestOffset;
+import static com.ververica.cdc.connectors.mysql.MySqlTestUtils.waitForJobStatus;
 import static org.apache.flink.api.common.JobStatus.RUNNING;
-import static org.apache.flink.runtime.testutils.CommonTestUtils.waitForJobStatus;
 import static org.junit.Assert.assertEquals;
 
 /** Integration tests for MySQL Table source. */
--- a/flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
+++ b/flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
@@ -70,8 +70,7 @@ public class MySqlConnectorITCase extends MySqlSourceTestBase {
     private static final String TEST_USER = "mysqluser";
     private static final String TEST_PASSWORD = "mysqlpw";
 
-    private static final MySqlContainer MYSQL8_CONTAINER =
-            (MySqlContainer) createMySqlContainer(MySqlVersion.V8_0).withExposedPorts(3307);
+    private static final MySqlContainer MYSQL8_CONTAINER = createMySqlContainer(MySqlVersion.V8_0);
 
     private final UniqueDatabase inventoryDatabase =
             new UniqueDatabase(MYSQL_CONTAINER, "inventory", TEST_USER, TEST_PASSWORD);
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,6 @@ under the License.
 
     <properties>
         <flink.version>1.15.2</flink.version>
-        <flink.submodule.postfix></flink.submodule.postfix>
         <debezium.version>1.6.4.Final</debezium.version>
         <tikv.version>3.2.0</tikv.version>
         <geometry.version>2.2.0</geometry.version>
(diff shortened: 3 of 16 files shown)


tigrulya-exe:
I've tried to reproduce oceanbase test failure locally and found out this issue: https://github.com/testcontainers/testcontainers-java/issues/5151. In `OceanBaseTestBase` we create two containers with host network mode and expose ports, like in issue above and Flink 1.15.1 uses version `1.6.2` of TestContainers. I'll try to fix it

tigrulya-exe:
The cause of failing my-sql tests can be https://issues.apache.org/jira/browse/FLINK-28861. I've debugged `MySqlSourceITCase` with Flink 1.15.1 and can acknowledge, that OperatorIDs of same operators were different before and after the savepoint.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/table/MySqlDeserializationConverterFactory.java
```diff
@@ -115,8 +114,7 @@ public Object convert(Object dbzObj, Schema schema) throws Exception {
 
     private static Optional<DeserializationRuntimeConverter> createArrayConverter(
             ArrayType arrayType) {
-        if (LogicalTypeChecks.hasFamily(
-                arrayType.getElementType(), LogicalTypeFamily.CHARACTER_STRING)) {
+        if (hasFamily(arrayType.getElementType(), LogicalTypeFamily.CHARACTER_STRING)) {
```
ruanhang1993, line 117: arrayType.getElementType().is(LogicalTypeFamily.CHARACTER_STRING)

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/table/MySqlDeserializationConverterFactory.java
```diff
@@ -148,4 +146,8 @@ public Object convert(Object dbzObj, Schema schema) throws Exception {
             return Optional.empty();
         }
     }
+
+    private static boolean hasFamily(LogicalType logicalType, LogicalTypeFamily family) {
```
ruanhang1993, line 150: This is replaced by `arrayType.getElementType().is(LogicalTypeFamily.CHARACTER_STRING)`.

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
```diff
@@ -65,8 +65,7 @@ public class MySqlConnectorITCase extends MySqlSourceTestBase {
     private static final String TEST_USER = "mysqluser";
     private static final String TEST_PASSWORD = "mysqlpw";
 
-    private static final MySqlContainer MYSQL8_CONTAINER =
-            (MySqlContainer) createMySqlContainer(MySqlVersion.V8_0).withExposedPorts(3307);
+    private static final MySqlContainer MYSQL8_CONTAINER = createMySqlContainer(MySqlVersion.V8_0);
```
ruanhang1993, line 68: Why need change this line?

pom.xml
```diff
@@ -467,6 +474,14 @@ under the License.
     </build>
 
     <profiles>
+        <profile>
+            <id>flink-1.14</id>
```
ruanhang1993, line 478: I think a cdc version only need support one flink version.

ruanhang1993 (commented):
Thanks for your contribution. I left some comments.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/table/MySqlDeserializationConverterFactory.java
tigrulya-exe: It was introduced in Flink 1.15 (https://issues.apache.org/jira/browse/FLINK-24599), so it won't compile with enabled Flink-1.14 profile.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/table/MySqlDeserializationConverterFactory.java
tigrulya-exe: Same as above

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/table/MySqlConnectorITCase.java
tigrulya-exe: I changed this line when I fixed a falling test, where this container was used. This port exposure does not make sense, but only confuses the reader of the test.

ruanhang1993: The default port 3306 has been used in the `MySqlSourceTestBase` by the `MYSQL_CONTAINER` whose mysql version is 5.7. So we need to keep this setting.

tigrulya-exe: As I said it confuses the reader of the test. In tests with `MYSQL8_CONTAINER`, when we get database port in table options via `MySqlContainer.getDatabasePort()` we get mapped port for a given port exposed by the container (it's always `MYSQL_PORT`(3306), even before my changes). So, there would be no port collision with `MYSQL_CONTAINER`, because they both use 3306 **inside** of the container, but mapped **host** ports are different.

tigrulya-exe:
Hi, @ruanhang1993 ! Thanks for comments! I've fixed all issues, can you recheck PR, please?

ruanhang1993 (approved):
LGTM, thanks for your contribution.

pom.xml
```diff
@@ -467,6 +474,14 @@ under the License.
...
+            <properties>
+                <flink.version>1.14.4</flink.version>
```
snuyanzin, line 480: Is it required to still have 1.14.x support?
Why just not to update to 1.15.x?
Or otherwise somehow need to guarantee that future changes will not add some 1.15.x specific api which could break 1.14.x usage... It seems currently ci does not check this profile... or did I miss something?

ruanhang1993:
@tigrulya-exe I think we should add the test environment in cdc e2e tests to verify the flink version 1.15.

pom.xml
```diff
@@ -71,7 +71,8 @@ under the License.
     </distributionManagement>
 
     <properties>
-        <flink.version>1.14.4</flink.version>
+        <flink.version>1.15.2</flink.version>
+        <flink.submodule.postfix></flink.submodule.postfix>
```
leonardBang, line 75: The property is empty?

pom.xml
leonardBang, line 478: Why we need this  @ruanhang1993 , Could we build a connector base on Flink 1.15 and then run it on Flink 1.14 cluster？ I think this is what we want.

pom.xml
ruanhang1993: Yes, the flink 1.15.2 need not scala suffix. I will try to remove this config and 1.14 profile.

pom.xml
ruanhang1993: I will try to remove 1.14 profile.

pom.xml
ruanhang1993: Thanks for your option, @snuyanzin. I have removed this profile to have a try.

leonardBang (approved):
Thanks @tigrulya-exe and @ruanhang1993 for the contribution, LGTM

CaoYunzhou:
when support cdc to elasticsearch 8.x for flink cdc ?

leonardBang:
> when support cdc to elasticsearch 8.x for flink cdc ?

You can open a issue firstly @CaoYunzhou

tigrulya-exe:
Hello, @ruanhang1993 ! Sorry for late response, but why did we remove maven module for flink 1.14 version? As far as I remember you said in #1363, that we need to support multiple Flink versions - 1.13, 1.14 and 1.15. For this purpose I've added module, because we can't just use new version of cdc with older Flink distributions due to incompatible changes in 1.15.

ruanhang1993:
Hi, @tigrulya-exe .  Sorry for the late response.
Actually we do not need to use the Flink 1.14 in the package phase in order to support the Flink version 1.14.
What we should do is to make sure that cdc connectors compiled by Flink 1.15 could run right in the Flink 1.14/1.13 environment. So the changes you provided in the [MySqlDeserializationConverterFactory.java](https://github.com/ververica/flink-cdc-connectors/pull/1504/files#diff-0b42bf52d2c60b5804d9b1359cba1abff8b00600a492a3f145715c32d05edba8) is needed. The API is not contained in Flink 1.15. We provide a method that can be used in Flink 1.13, 1.14 and 1.15, which provides the ability to support those Flink versions.

PR #1449, opened 2022-08-05
 dont need captureNewlyAddedTables when isRemainingTablesCheckpointed is false and assign isnt finished on MySqlSnapshotSplitAssigner

lidoudou1993:
dont need captureNewlyAddedTables when isRemainingTablesCheckpointed is false and assign isnt finished on MySqlSnapshotSplitAssigner

PR #1507, opened 2022-08-24
[mysql] Enumerate captured tables lazily

leonardBang:
This PR aims to delay the captured tables enumeration from startup to runtime


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
```diff
@@ -144,9 +144,30 @@ private MySqlSnapshotSplitAssigner(
     @Override
     public void open() {
         chunkSplitter = createChunkSplitter(sourceConfig, isTableIdCaseSensitive);
+        discoveryCaptureTables();
```
ruanhang1993, line 147: The changes in this PR #1449 missed.

leonardBang: After rethink I found the optimization only happen when upgrade from 2.0 to 2.3, the benefit is too small and it also let the code too difficult to understand. Thus I decide to ignore it.

PR #1508, opened 2022-08-25
fix startupOptions mode bug 

TeriMoni:
the class MySqlSourceConfigFactory is not support the new startupOptions


ruanhang1993:
Other startup modes are note supported now.

leonardBang:
@TeriMoni Thanks for your contribution. But we need to improve the PR  to really support these options, currently the underlying code does not support these options, thus your change is not enough.

ruanhang1993:
These start up modes have been added in #1661. Close this PR.

PR #1513, opened 2022-08-28
There is a typo in the comments in MysqlSnapshotSplit.

empcl:
should be "The high watermark is not null when the split read finished.", but not 'bull'.


leonardBang (approved):
LGTM

PR #1521, opened 2022-08-30
[mysql][support-multi-table] support set separator to distinguish different tables

LXin96:
1、Currently, doesn't support separator to set multi-table-name. Howere, when the tablename is different and can't use a regex to describle, so i think maybe we can support a separator. eg: table-name=t[1-4],xx1,ab2
2、We don't care about regex and normal tablename cause a confict because tableFilter can  help to deal with repeat table
according to tableId.


LXin96:
@leonardBang could you please review this pr when you have a time?

LXin96:
regex can use "|" to union different database name and table name

leonardBang:
@LXin96  Could you open a PR to add some regex patterns for multiple tables to the document?

PR #1523, opened 2022-08-30
[mysql] Reduce the memory usage of JM by sharing table schemas between splits

### LINKED ISSUE #865: [mysql] Too many splitInfo cause OOM of jobmanager

leonardBang:
The optimization to reduce JM memory usage, this PR aims to resolve #865

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializer.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializer.java
@@ -266,7 +266,7 @@ public class PendingSplitsStateSerializer implements SimpleVersionedSerializer<P
                             assignedSchemaLessSnapshotSplits.put(
                                     entry.getKey(), entry.getValue().toSchemaLessSnapshotSplit());
                         });
-        if (version == 4) {
+        if (version >= 4) {
             tableSchemas.putAll(readTableSchemas(splitVersion, in));
         }
         return new SnapshotPendingSplitsState(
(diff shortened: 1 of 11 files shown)


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializer.java
```diff
@@ -205,29 +230,50 @@ private HybridPendingSplitsState deserializeLegacyHybridPendingSplitsState(
...
+                });
+        assignedSnapshotSplits
+                .entrySet()
+                .forEach(
+                        entry -> {
+                            tableSchemas.putAll(entry.getValue().getTableSchemas());
+                            assignedSchemaLessSnapshotSplits.put(
+                                    entry.getKey(), entry.getValue().toSchemaLessSnapshotSplit());
+                        });
+        if (version == 4) {
```
ruanhang1993, line 269: The condition `version >= 4` may be better?

PatrickRen (approved):
Thanks for the patch @leonardBang! LGTM. Just a minor comment on the serializer that maybe we could add a backward compatibility test for validating new serializers could deserialize state with earlier version.

PR #1356, opened 2022-07-07
[oceanbase] Add new deserialization schema with runtime converter

whhe:
The previous logic is to do the first conversion with OceanBaseJdbcConverter helper methods, and then use runtime converters in JdbcValueConverters to do the second conversion. Now we need only do conversion one time with the new deserialization schema.

close #980 close #1289

PR #1524, opened 2022-08-30
[oceanbase] fix npe in getTableSchema from UPDATE logmessage

whhe:
Half of the elements in `columnNames` will be null if we try to generate table schema from a `UPDATE` type log message, so we should skip these null values when create column instances.


ruanhang1993:
@whhe Are there some tests that already produce this error?

whhe:
> @whhe Are there some tests that already produce this error?

There is a user in the DingTalk group of Flink CDC who meets this issue, it's not covered by the integration tests for now.

whhe:
BTW, this fix will not be required after #1356 is merged.

PR #1548, opened 2022-09-09
[oracle]oracle cdc supports synchronizing oracle slave nodes

waywtdcc:
Need to assign before GRANT CREATE TABLE TO flinkuser2;
   GRANT ALTER ANY TABLE TO flinkuser2;
   GRANT CREATE SEQUENCE TO flinkuser2; three permissions. But doing so has flaws
1. It cannot be realized in the Oracle slave library cdc synchronization, because the slave library cannot write data. Therefore, the cdc data of the oracle slave library cannot be synchronized
2. There is no need to create and write the LOG_MINING_FLUSH table when using flink cdc, and there is also sdn related information in the checkpoint.

I have removed the creation and use of the LOG_MINING_FLUSH table here, because I think that using flink cdc does not need to record sdn in the LOG_MINING_FLUSH table

### DIFF
--- a/docs/content/connectors/index.md
+++ b/docs/content/connectors/index.md
@@ -5,6 +5,7 @@
 
 mongodb-cdc
 mysql-cdc
+mysql-cdc(ZH)
 oceanbase-cdc
 oracle-cdc
 postgres-cdc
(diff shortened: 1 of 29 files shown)


leonardBang:
Thanks @waywtdcc for opening this PR, but could you open an issue to describe the feature before open a PR？

waywtdcc:
[https://github.com/ververica/flink-cdc-connectors/issues/1574](url)

zhbdesign:
2.3支持Oracle 从库同步了吗

PR #1554, opened 2022-09-14
[pg-connector] add geometry type support.

GOODBOY008:
[[add geometry type support.](https://github.com/ververica/flink-cdc-connectors/issues/1539)](https://github.com/ververica/flink-cdc-connectors/issues/1539)


Jiabao-Sun (approved):
Thanks @GOODBOY008, LGTM.

Issue #1558, opened 2022-09-16
e2e-tests optimize for arm


deemogsw:
I want to debug the code with module flink-cdc-e2e-tests in IDE.
While,although I have run the docker desktop in my mac.An exception still been throwed like this
<img width="1381" alt="image" src="https://user-images.githubusercontent.com/48081481/190575322-457a51ba-341b-4a3c-b676-804560786c48.png">
The current code can't work in Mac with apple silicon chip. 

Besides teste moudle use docker container to run JM & TM,testContainer would assign  a random port for the exposed Port.So when I run the test case,I must find out which port is currently bound to 8081.so that I can visit flink web.
This is not friendly for those people who has less idea with docker.

deemogsw:
@leonardBang May you have a look.

PatrickRen:
Closing this issue because it was created before version 2.3.0 (2022-11-10). Please try the latest version of Flink CDC to see if the issue has been resolved. If the issue is still valid, kindly report it on [Apache Jira](https://issues.apache.org/jira) under project `Flink` with component tag `Flink CDC`. Thank you!

PR #1560, opened 2022-09-16
[e2e] optimization for arm64

deemogsw:
linked #1558
1 make sure mac with apple silicon chip can run the test case.
2 define the expose port 8081:8081 in JobManager's container.So when container is running, Flink web is avaliable in http://localhost:8081


yuxiqian:
Hi @deemogsw, thanks for your great contribution!

Currently MySQL has upgraded docker image tag to `mysql:8.0`, which should run on Apple Silicon chips natively.

As for static port mappings, my concern is 8081 port isn't always available, and there might be some problem if several e2e test cases run simultaneously. Currently TestContainers always maps them to a random free port without explicit `withPortBindings` command, and users could use `docker port <container_id>` or `Container.getMappedPort()` to access the mapped port info.

I'm going to close this PR since these problems have been addressed in latest `master` branch. Feel free to reopen it if you have other concerns.

Issue #1582, opened 2022-09-26
The fetchSize parameter is not used


fuyun2024:
**Describe the bug(Please use English)**
When I set fetchSize=10, use MySqlConnectorConfig The value obtained by the getQueryFetchSize() method is 0

**Environment :**
 - Flink version :  Flink 1.14
 - Flink CDC version:  2.3-SNAPSHOT
 - Database and version: 

**To Reproduce**
1. The test data :
2. The test code :
3. The error :

**Additional Description**

PR #1583, opened 2022-09-26
[mysql] Fix the fetchSize parameter is not used

fuyun2024:
issues : #1582

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
@@ -320,9 +320,11 @@ public class MySqlSourceConfigFactory implements Serializable {
 
         // use cursor-based streaming to retrieve a set number of rows
         // https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-implementation-notes.html
+        // https://github.com/debezium/debezium/pull/2245
         props.setProperty(SNAPSHOT_FETCH_SIZE.name(), String.valueOf(fetchSize));
         if (fetchSize > 0) {
             jdbcProperties.setProperty("useCursorFetch", "true");
+            props.setProperty("database.useCursorFetch", String.valueOf(true));
         }
 
         return new MySqlSourceConfig(
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
@@ -39,7 +39,7 @@ import static com.ververica.cdc.connectors.mysql.source.config.MySqlSourceOption
 import static com.ververica.cdc.connectors.mysql.source.config.MySqlSourceOptions.HEARTBEAT_INTERVAL;
 import static com.ververica.cdc.connectors.mysql.source.config.MySqlSourceOptions.SCAN_INCREMENTAL_SNAPSHOT_CHUNK_SIZE;
 import static com.ververica.cdc.connectors.mysql.source.config.MySqlSourceOptions.SCAN_SNAPSHOT_FETCH_SIZE;
-import static io.debezium.config.CommonConnectorConfig.QUERY_FETCH_SIZE;
+import static io.debezium.config.CommonConnectorConfig.SNAPSHOT_FETCH_SIZE;
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
 /** A factory to construct {@link MySqlSourceConfig}. */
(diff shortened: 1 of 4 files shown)


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
```diff
@@ -270,7 +270,7 @@ public MySqlSourceConfig createConfig(int subtaskId) {
         props.setProperty("database.user", checkNotNull(username));
         props.setProperty("database.password", checkNotNull(password));
         props.setProperty("database.port", String.valueOf(port));
-        props.setProperty("database.fetchSize", String.valueOf(fetchSize));
```
ruanhang1993, line 273:
```java
props.setProperty(QUERY_FETCH_SIZE.name(), String.valueOf(fetchSize));
if (fetchSize > 0) {
    // https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-implementation-notes.html
    jdbcProperties.setProperty("useCursorFetch", "true");
}
```

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
```diff
@@ -317,6 +317,8 @@ public MySqlSourceConfig createConfig(int subtaskId) {
         if (jdbcProperties == null) {
             jdbcProperties = new Properties();
         }
+        // https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-implementation-notes.html
+        jdbcProperties.setProperty("useCursorFetch", "true");
```
ruanhang1993, line 321: Move this line near its relative code.

ruanhang1993 (commented):
Thanks for fixing this bug, @fuyun2024

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
fuyun2024: Hi @ruanhang1993  Thank you for your advice, but 'jdbcProperties' is initialized here. Do I need to move it here?

ruanhang1993: I think we could initialize the `jdbcProperties` earlier.

fuyun2024: hi @ruanhang1993 , I think the important code is at the top, but it is not the most important, so I put it at the end

GOODBOY008:
@fuyun2024  `MySqlTimezoneITCase` hangs in test.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
```diff
@@ -318,6 +318,13 @@ public MySqlSourceConfig createConfig(int subtaskId) {
             jdbcProperties = new Properties();
         }
 
+        // use cursor-based streaming to retrieve a set number of rows
+        // https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-implementation-notes.html
+        if (fetchSize > 0) {
+            jdbcProperties.setProperty("useCursorFetch", "true");
+            props.setProperty(QUERY_FETCH_SIZE.name(), String.valueOf(fetchSize));
```
GOODBOY008, line 325: The `QUERY_FETCH_SIZE` property is useless for mysql.

GOODBOY008 (commented):
@fuyun2024  It‘s not easy to support `useCursorFetch`.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
fuyun2024: Hi @GOODBOY008  I use the mysql 8.0.21 driver and it works

ruanhang1993:
@fuyun2024 Maybe we need to add some tests to verify the bug is fixed.

ruanhang1993:
If we open the `useCursorFetch`, the tests in the  `MySqlTimezoneITCase` will fail.
It seems that the date type field can not be parsed right when `useCursorFetch` is true.
截屏2022-10-31 下午7 28 17

ruanhang1993:
> If we open the `useCursorFetch`, the tests in the `MySqlTimezoneITCase` will fail. It seems that the date type field can not be parsed right when `useCursorFetch` is true. 截屏2022-10-31 下午7 28 17

This problem has been fixed in DBZ-3238(https://github.com/debezium/debezium/pull/2245).
So I have applied these changes to mysql cdc connector.

fuyun2024:
HI @ruanhang1993 , Thank you for your time and suggestions,  @GOODBOY008  has submitted a new PR  #1629 , and I will close this submission later.

PR #1505, opened 2022-08-24
[mysql] Let the records of snapshot split don't cross checkpoints

leonardBang:
The output records of a snapshot split should not cross checkpoints, this may lead to data duplication when failover happen, because partial snapshot records of one split may sent in last checkpoint, but this snapshot split may be re-read all snapshot records after failover, this behavior bring potential data duplication issue.

PR #1622, opened 2022-10-18
[cdc-base] Let all records of one snapshot split don't cross checkpoints

ruanhang1993:
Apply changes in #1505 to cdc-base.


leonardBang (approved):
Thanks @ruanhang1993 for the pick, LGTM

Issue #1375, opened 2022-07-14
about the last one MySqlSnapshotSplit


jiangjin-f:
**Describe the bug(Please use English)**

**Environment :**
 - Flink version :  1.13.5
 - Flink CDC version: 2.2.1
 - Database and version: mysql 

**To Reproduce**
1. Thes test data :
2. The test code :
3. The error :

**Additional Description**

When the last split is read, if the full amount of data is large and a lot of data is written per second, the last split directly uses the primary key > =?, The data volume is too large and the task fails

![Uploading 719fad9b-bcbd-465e-acc8-4d6e35e6a822.png…]()
![企业微信截图_c0903bba-5a03-454e-bd2e-5cb100ff2ed8](https://user-images.githubusercontent.com/46118864/178945021-4c62d210-1fb5-45c9-8489-e30de9f1b0f3.png)

JustinLeesin:
I also encountered such problem, the task manager OOM, looking for a solution

MeepoD:
I also encountered the same problem，When table is very large, it's oom

PatrickRen:
Closing this issue because it was created before version 2.3.0 (2022-11-10). Please try the latest version of Flink CDC to see if the issue has been resolved. If the issue is still valid, kindly report it on [Apache Jira](https://issues.apache.org/jira) under project `Flink` with component tag `Flink CDC`. Thank you!

PR #1625, opened 2022-10-18
support table splitting process control when read table snapshot

JustinLeesin:
This resolves a problem encountered in our production environment when there are too many tables to be read on the snapshot phase and each takes quite a lot of time ,  as a result, the process of data reading falls behind the process of chunk splitting too much. 

To be more specific, say there are 100 tables to be read, the current process is:
1、 split all the tables into chunks one by one, which may last several hours, as it's a lightweight job.
2、read the chunks from procedure 1, which may last several days, as it's a heavyweight job.

When start reading the last split of the last table,  executing 'select from table where primary key > *** ',  too much data has accumulated for this chunk, because a long time has passed since it has been split, then the job will fail .

There are someone else who also encountered this problem (url) #1375 

As a  solution,  we can control the speed of chunk splitting thread, to prevent it from going too much ahead. when it reaches the threshold value, it will wait for the data reading thread.

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/MySqlSourceBuilder.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/MySqlSourceBuilder.java
@@ -170,6 +170,17 @@ public class MySqlSourceBuilder<T> {
         return this;
     }
 
+    public MySqlSourceBuilder<T> splitTableProcessControlEnabled(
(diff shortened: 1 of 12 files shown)


yuxiqian:
Hi @JustinLeesin , sorry for the delay of this PR. Could you please rebase it to latest `master` branch since there’s been lots of changes in Flink CDC repo since your original commit? Kindly reminder that `com.ververica.cdc.connectors.mysql` package has been moved to `org.apache.flink.cdc.connectors.mysql`.

cc @leonardBang

PR #1661, opened 2022-10-26
[mysql] Support starting reading binlog from earliest offset / specific offset / timestamp

PatrickRen:
This PR introduces a new feature #254 for the new MySQL CDC source, that users could specify starting point of binlog reading, including earliest offset, specific offset and timestamp.

PR #1724, opened 2022-11-09
[mysql] Use event filter to drop events earlier than the specified timestamp in timestamp startup mode

PatrickRen:
This pull request reworks the logic of timestamp startup mode, which uses a event filter in `MySqlBinlogSplitReadTask` to drop events earlier than the specified startup timestamp. 

Compared to the previous implementation #1661 in which the filter was applied on `ChangeEventCreator`, the filter is applied right after the binlog event is received, so that the diff in schema will not be complained by Debezium reader under the case that schema change happens before the specified timestamp. 

This PR also improves the error message under the case that starting from earliest offset but schema change happened before.


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
```diff
@@ -299,6 +304,21 @@ private void configureFilter() {
...
     }
 
+    private Predicate<Event> createEventFilter(BinlogOffset startingOffset) {
+        // If the startup mode is set as TIMESTAMP, we need to apply a filter on event to drop
+        // events earlier than the specified timestamp.
+        if (BinlogOffsetKind.TIMESTAMP.equals(startingOffset.getOffsetKind())) {
+            long startTimestampSec = startingOffset.getTimestampSec();
+            // Notes:
+            // 1. Heartbeat event doesn't contain timestamp, so we just keep it
+            // 2. Timestamp of event is in millisecond
```
leonardBang, line 314:
```suggestion
            // 2. Timestamp of event is in epoch millisecond
```

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/task/context/MySqlErrorHandler.java
```diff
@@ -53,20 +63,48 @@ protected boolean isRetriable(Throwable throwable) {
...
+    }
+
+    private boolean isSchemaOutOfSyncWithEarliestStartupMode(Throwable t) {
+        Throwable rootCause = ExceptionUtils.getRootCause(t);
+        return rootCause instanceof ConnectException
+                && rootCause
+                        .getMessage()
+                        .endsWith(
+                                "internal schema representation is probably out of sync with real database schema")
+                && sourceConfig.getStartupOptions().startupMode == StartupMode.EARLIEST_OFFSET;
```
leonardBang, line 108: The `specific` offset and `timestamp` modes need to be considered too.

leonardBang (commented):
Thanks @PatrickRen for the contribution, I left some comments

flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/debezium/reader/BinlogSplitReaderTest.java
```diff
@@ -363,6 +366,39 @@ public void testReadBinlogFromEarliestOffset() throws Exception {
...
+        MySqlBinlogSplit split = createBinlogSplit(sourceConfig);
+        BinlogSplitReader reader = createBinlogReader(sourceConfig);
+        reader.submitSplit(split);
+
+        // An exception is expected here because the table schema is changed, which is not allowed
+        // under
+        // earliest startup mode.
+        Throwable throwable =
+                assertThrows(Throwable.class, () -> readBinlogSplits(dataType, reader, 1));
+        assertTrue(
```
leonardBang, line 397: Could you also check the exception message?

leonardBang (approved):
Thanks @PatrickRen for the update, LGTM

leonardBang (approved):
Thanks @PatrickRen for the update, LGTM

PR #1739, opened 2022-11-15
about MYSQL GMT+8 BUG

kahineasin:
about the mysql(GMT+8) Datetime column has 8 more hours bug.

maybe can use this update to fix it. 
(1.use userTimezone when read from database; 2.convert Timestamp to "timestamp long vlaue" but not "LocalDateTime")

I have test on my pc, it fix this bug all right.

### DIFF
--- a/docs/content/connectors/index.md
+++ b/docs/content/connectors/index.md
@@ -7,6 +7,7 @@ mongodb-cdc
 mysql-cdc
 mysql-cdc(ZH)
 oceanbase-cdc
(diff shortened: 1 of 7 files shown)


leonardBang:
Thanks @kahineasin for the contribution, But we have fixed this issue in 2.3.0 version, you can check out it. 
You could know more details refer this PT https://github.com/ververica/flink-cdc-connectors/pull/1407

Issue #1740, opened 2022-11-15
The parameter types described in the MySQL CDC document are inconsistent with those defined in the code


GeiChan:
The parameter types described in the MySQL CDC document are inconsistent with those defined in the code

![mysql-cdc-zh](https://user-images.githubusercontent.com/42415224/201820716-9ea27660-9795-4fb9-a83a-67bde9582024.png)
![mysql-cdc](https://user-images.githubusercontent.com/42415224/201820542-80c48210-af90-481e-b895-a72ac2819c02.png)
![mysql-cdc-code](https://user-images.githubusercontent.com/42415224/201820640-bb205236-a625-4608-8a02-49345a57d82f.png)

PR #1741, opened 2022-11-15
fix the parameter types described in the MySQL CDC document are incon…

GeiChan:
fix the parameter types described in the MySQL CDC document are inconsistent with those defined in the code #1740


rookiegao:
Thanks @GeiChan the contribution

image

rookiegao:
cc @GOODBOY008 PTAL

leonardBang (approved):
Thanks @GeiChan and @rookiegao for the fix and review, LGTM

rookiegao:
@GeiChan Can you fix this problem for matser branches as well？

Issue #1770, opened 2022-11-23
[MySQL-CDC] serverId parameter out of Integer range


legendtkl:
**Describe the bug(Please use English)**
Now the 'server_id' is store as int, but the value of parameter 'server_id' would be larger than INT_MAX, this would cause IntegerParseException

**Environment :**
 - Flink version :  1.16
 - Flink CDC version: 2.3.0
 - Database and version: mysql 5.7

**To Reproduce**
1. The test data :
MySQL instance in Aliyun Cloud, with a large server_id (out of INT range)
```bash
mysql> show variables like '%server_id%';
+----------------+------------+
| Variable_name  | Value      |
+----------------+------------+
| server_id      | 2437466879 |
| server_id_bits | 32         |
+----------------+------------+

```

2. The test code :
```java
public class MySqlSourceExample {
  public static void main(String[] args) throws Exception {
    MySqlSource<String> mySqlSource = MySqlSource.<String>builder()
      .hostname("host")
      .port(3306)
      .databaseList("dbs") // set captured database
      .tableList("tables") // set captured table
      .username("user")
      .password("password")
      .serverId("1-2437466879")
      .deserializer(new JsonDebeziumDeserializationSchema()) // converts SourceRecord to JSON String
      .build();

    StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();

    env
      .fromSource(mySqlSource, WatermarkStrategy.noWatermarks(), "MySQL Source")
      // set 4 parallel source tasks
      .setParallelism(4)
      .print().setParallelism(1); // use parallelism 1 for sink to keep message ordering

     env.execute("Print MySQL Snapshot + Binlog");
  }
}
```

3. The error :
```bash
Exception in thread "main" java.lang.IllegalStateException: The server id 2437466879 is not a valid numeric.
	at com.ververica.cdc.connectors.mysql.source.config.ServerIdRange.parseServerId(ServerIdRange.java:108)
	at com.ververica.cdc.connectors.mysql.source.config.ServerIdRange.from(ServerIdRange.java:96)
	at com.ververica.cdc.connectors.mysql.source.config.MySqlSourceConfigFactory.serverId(MySqlSourceConfigFactory.java:130)
	at com.ververica.cdc.connectors.mysql.source.MySqlSourceBuilder.serverId(MySqlSourceBuilder.java:108)
[...]

PR #1771, opened 2022-11-23
[#1770] fix NumberFormatException when mysql server_id is larger than INT_MAX

legendtkl:
this close issue https://github.com/ververica/flink-cdc-connectors/issues/1770

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
@@ -288,7 +288,7 @@ public class MySqlSourceConfigFactory implements Serializable {
         props.put("bigint.unsigned.handling.mode", "precise");
 
         if (serverIdRange != null) {
-            int serverId = serverIdRange.getServerId(subtaskId);
+            long serverId = serverIdRange.getServerId(subtaskId);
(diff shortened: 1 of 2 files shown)


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
```diff
@@ -288,7 +288,7 @@ public MySqlSourceConfig createConfig(int subtaskId) {
         props.put("bigint.unsigned.handling.mode", "precise");
 
         if (serverIdRange != null) {
-            int serverId = serverIdRange.getServerId(subtaskId);
+            long serverId = serverIdRange.getServerId(subtaskId);
```
ruanhang1993, line 291: Server id has a fixed range in mysql (https://dev.mysql.com/doc/refman/5.7/en/replication-options.html).
I think we should check the config option of the server id range at the method `ServerIdRange#from`.

ruanhang1993 (commented):
Thanks for contributing. I have left some comments.
I think we should modify the mysql cdc document about the available server id range.

yuxiqian:
Hi @legendtkl, sorry for the delay of this PR. Could you please rebase it to latest `master` branch since there’s been lots of changes in Flink CDC repo since your original commit? Kindly reminder that `com.ververica.cdc.connectors.mysql` package has been moved to `org.apache.flink.cdc.connectors.mysql`.

Thanks for your contribution!

cc @ruanhang1993

PR #1773, opened 2022-11-23
Upgrade postgresql jdbc from 42.2.26 to 42.2.27 to avoid CVE-2022-41946

snuyanzin:
There is a CVE-2022-41946 fixed in latest fixes for each branch
The PR updates pgjdbc version to avoid this CVE
Also updates pgjdbc version in README table


leonardBang (approved):
+1

PR #1808, opened 2022-12-07
[Flink][Improvement]Throw an exception when the table has no primary key set instead of catching it

JasonLeeCoding:
Throw an exception when the table has no primary key set instead of catching it,In this way, the flink job will failover, and users can find the problem faster.

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
@@ -275,9 +275,9 @@ public class MySqlSnapshotSplitAssigner implements MySqlSplitAssigner {
 
     @Override
     public Optional<MySqlSplit> getNext() {
-        checkSplitterErrors();
         waitTableDiscoveryReady();
         synchronized (lock) {
+            checkSplitterErrors();
             if (!remainingSplits.isEmpty()) {
                 // return remaining splits firstly
                 Iterator<MySqlSchemalessSnapshotSplit> iterator = remainingSplits.iterator();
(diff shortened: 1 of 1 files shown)


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
```diff
@@ -275,9 +275,9 @@ private void splitTable(TableId nextTable) {
 
     @Override
     public Optional<MySqlSplit> getNext() {
-        checkSplitterErrors();
         waitTableDiscoveryReady();
         synchronized (lock) {
+            checkSplitterErrors();
```
ruanhang1993, line 280: We should also make the assignment logic of the `uncaughtSplitterException` guarded by the `notificationLock` in the method `splitChunksForRemainingTables`.

ruanhang1993 (commented):
@JasonLeeCoding Thanks for contributing. I left some comments.
Would you mind change the commit message to '[mysql] Fix race condition between split and coordinator thread that async exception might not be caught'?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
JasonLeeCoding: hi @ruanhang1993 I've already modified it,Please help to review it when you have time,thanks.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
ruanhang1993, line 1: Thanks for your quick reply, @JasonLeeCoding .
Actually we do not need to put all parts of `splitChunksForRemainingTables` under the lock.
We could modify like this.
```java
private void splitChunksForRemainingTables() {
    try {
        ......
    } catch (Throwable e) {
        // Release the potential waiting getNext() call
        synchronized (lock) {
            if (uncaughtSplitterException == null) {
                uncaughtSplitterException = e;
            } else {
                uncaughtSplitterException.addSuppressed(e);
            }
            lock.notify();
        }
    }
}
```

ruanhang1993 (approved):
LGTM. Thanks~

PR #1821, opened 2022-12-11
[common] Improve implementation of offset backing store

wuzhenhua01:
Simplified the offset backing store


yuxiqian:
Thanks for your contribution! Before this PR could be merged, could you please rebase it with latest `master` branch?

PR #1830, opened 2022-12-14
[meta] Use issue form templates

tisonkun:
See https://github.com/tisonkun/flink-cdc-connectors/issues/new/choose as a result preview.

Now we require version info and minor reproduce steps for bug reports. Of course, the reporter can fake the input, but it's a stronger hint.

image
image
image


tisonkun:
cc @wuchong @leonardBang any thoughts?

leonardBang (approved):
Thanks @tisonkun for the nice templates, sorry for the late reply due to my sick and vocation, LGTM

PR #1836, opened 2022-12-21
[Fix a Bug] add supported for Oracle CDB mode at OracleIncrementalSource

luckes-yang:
it could not read the PDB tables when using OracleIncrementalSource at Oracle CDB mode,Therefore, check whether the pdb is empty when creating a connection. If it is not empty, switch to the pdb session


yuxiqian:
Thanks for your contribution! Before this PR could be merged, could you please rebase it with latest `master` branch?

PR #1852, opened 2023-01-03
[mysql] a new way to find binlog/positon in timestamp startupmode

lidoudou1993:
reverse traverse all binlog file in database, find binlog position by compare binlog event timestamp and specified timestamp, then start read binlog from this binlog/position


z-haohao:
Whether it is possible to use the dichotomy search method to find the timestamp, because it is the front traversal search and the back traversal search, will take a long time.

yuxiqian:
Closing this PR since it has been implemented by #2924.

PR #1911, opened 2023-02-13
Update build real time data lake tutorial

911432:
I deleted Delete 'catalog-type'='hadoop'.
Because it's optional.
https://iceberg.apache.org/docs/latest/flink/#catalog-configuration

And it has been changed so that it can be downloaded as the latest version of Docker Image. https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/resource-providers/standalone/docker/#flink-docker-images


911432:
I will fix about required dependencies.

PR #1917, opened 2023-02-14
[mysql-cdc] Fix newly-added-table failed if job failover before a success checkpoint

lzshlzsh:
When the newly added table finishes (Assigner status is NEWLY_ADDED_ASSIGNING_FINISHED) and then job failsover before a success checkpoint, the snapshotSplits of newly added table will be added back, and are assigned to taskmanager later. However, taskmanager will first start MySqlBinlogSplit read task after failover, the the later assigned snapshotSplits will never be processed if the job parallelism is 1.  Also the binlog of newly added table will never be processed, as shown in the log below。

```
2023-01-31 23:40:45.283 [blc-9.134.34.15:3306] MySqlStreamingChangeEventSource.java:1222 ERROR io.debezium.connector.mysql.MySqlStreamingChangeEventSource - Error during binlog processing. Last offset stored = null, binlog reader near position = mysql-bin.000101/143162238
2023-01-31 23:40:45.283 [blc-9.134.34.15:3306] MySqlErrorHandler.java:67 WARN  com.ververica.cdc.connectors.mysql.debezium.task.context.MySqlErrorHandler - Schema for table cdc_basic_source.10_random_source_4 is null
2023-01-31 23:40:45.283 [blc-9.134.34.15:3306] MySqlStreamingChangeEventSource.java:429 INFO  io.debezium.connector.mysql.MySqlStreamingChangeEventSource - Error processing binlog event, and propagating to Kafka Connect so it stops this connector. Future binlog events read before connector is shutdown will be ignored.
```

So we should suspend binlogReader and read the snapshotSplits first.

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/enumerator/MySqlSourceEnumerator.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/enumerator/MySqlSourceEnumerator.java
@@ -106,7 +106,7 @@ public class MySqlSourceEnumerator implements SplitEnumerator<MySqlSplit, Pendin
 
     @Override
     public void addSplitsBack(List<MySqlSplit> splits, int subtaskId) {
-        LOG.debug("MySQL Source Enumerator adds splits back: {}", splits);
+        LOG.debug("The enumerator adds splits back: {}", splits);
         splitAssigner.addSplits(splits);
     }
 
@@ -123,7 +123,8 @@ public class MySqlSourceEnumerator implements SplitEnumerator<MySqlSplit, Pendin
     public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
         if (sourceEvent instanceof FinishedSnapshotSplitsReportEvent) {
             LOG.info(
-                    "The enumerator receives finished split offsets {} from subtask {}.",
+                    "The enumerator under {} receives finished split offsets {} from subtask {}.",
+                    splitAssigner.getAssignerStatus(),
                     sourceEvent,
                     subtaskId);
             FinishedSnapshotSplitsReportEvent reportEvent =
@@ -192,7 +193,7 @@ public class MySqlSourceEnumerator implements SplitEnumerator<MySqlSplit, Pendin
                 final MySqlSplit mySqlSplit = split.get();
                 context.assignSplit(mySqlSplit, nextAwaiting);
                 awaitingReader.remove();
-                LOG.info("Assign split {} to subtask {}", mySqlSplit, nextAwaiting);
+                LOG.info("The enumerator assigns split {} to subtask {}", mySqlSplit, nextAwaiting);
             } else {
                 // there is no available splits by now, skip assigning
                 requestBinlogSplitUpdateIfNeed();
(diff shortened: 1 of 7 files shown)


lzshlzsh:
@leonardBang @minchowang @kylemeow Would you help to review the PR?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/enumerator/MySqlSourceEnumerator.java
```diff
@@ -121,6 +122,17 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
     public void addSplitsBack(List<MySqlSplit> splits, int subtaskId) {
         LOG.debug("MySQL Source Enumerator adds splits back: {}", splits);
         splitAssigner.addSplits(splits);
+
+        if (isNewlyAddedAssigningFinished(splitAssigner.getAssignerStatus())) {
```
leonardBang, line 126: Thanks @lzshlzsh for the contribution, but I recently improve the newly added table process. Could you help check is the issue still exits in current codebase?

lzshlzsh: I see #1927，it is a great job. I will take a closer look at it and verify it.  Thanks !

lzshlzsh: @leonardBang After carefully reviewing and testing the non blocking newly add table process #1927, I found a problem, perhaps a design trade-off. Could you explain it in detail？
The in turn execution of BinlogSplitReader and SnapshotSplitReader has achieved the goal of non blocking, but the data read by BinlogSplitReader has not been modified to filter out the data of the newly added table. This may not be a problem in the ETL scenario, but it can lead to incorrect results in scenarios such as group by.

Test case：
1、do savepoint
2、Insert (2001, 17, da) into the newly added table, table schema is:
```
CREATE TABLE `mysql_cdc`
(
    `f_sequence`   INT,
    `f_random`     INT,
    `f_random_str` VARCHAR,
    PRIMARY KEY (`f_sequence`) NOT ENFORCED
) WITH (
  'connector' = 'mysql-cdc',
  ...
);
```
3、restore from savepoint，output duplicated data of key 2001. One emitted by BinlogSplitReader, and the other by SnapshotSplitReader.
```
 +I[2001, 17, da]
 +I[2001, 17, da]
```

leonardBang:
@lzshlzsh Thanks for you reporting this issue and opening a PR ！ I'll take a look this PR, could you rebase the latest master?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/enumerator/MySqlSourceEnumerator.java
leonardBang: I didn't reproduce this case @lzshlzsh , and the case should be covered by NewlyAddedTableITCase#testNewlyAddedTableForExistsPipelineSingleParallelismWithAheadBinlog()

leonardBang (approved):

leonardBang:
All CI passed, merging...

Issue #1886, opened 2023-01-30
Oracle CDC 2.3 can only read one record during snapshot reading


e-mhui:
**Describe the bug(Please use English)**

Oracle CDC 2.3 can only read one record during snapshot reading.

**Environment :**
 - Flink version :  1.13.5
 - Flink CDC version: 2.3
 - Database and version: oracle 11g

**To Reproduce**
1. The test data : 

<img width="254" alt="image" src="https://user-images.githubusercontent.com/111486498/215480063-6a6ac56f-93de-49e6-85bf-3cd541323a8d.png">

3. The test code :

```java

       StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
        env.setParallelism(1);
        env.disableOperatorChaining();

        StreamTableEnvironment tableEnv = StreamTableEnvironment.create(env);

        tableEnv.executeSql(
                    "CREATE TABLE `table_1`(\n"
                            + "    `ID` DECIMAL(10, 2),\n"
                            + "    `NAME` STRING,\n"
                            + "    `PROCESS_TIME` AS PROCTIME())\n"
                            + "    WITH (\n"
                            + "    'connector' = 'oracle-cdc',\n"
                            + "    'hostname' = 'localhost',\n"
                            + "    'username' = 'flinkuser',\n"
                            + "    'password' = 'flinkpw',\n"
                            + "    'database-name' = 'XE',\n"
                            + "    'schema-name' = 'FLINKUSER',\n"
                            + "    'table-name' = 'TB2',\n"
                            + "    'debezium.database.tablename.case.insensitive' = 'false',\n"
                            + "    'debezium.log.mining.strategy' = 'redo_log_catalog',\n"
                            + "    'debezium.log.mining.continuous.mine' = 'false',\n"
                            + "    'scan.incremental.snapshot.enabled' = 'true',\n" // default true
                            + "    'port' = '1521',\n"
                            + "    'scan.startup.mode' = 'initial'\n"
                            + ")"
                );

        TableResult tableResult = tableEnv.executeSql("select * from table_1");
        tableResult.print();
        try {
            env.execute();
            Thread.sleep(1000 * 1000);
        } catch (Exception e) {
            e.printStackTrace();
        }

```

5. The error :

**Additional Description**

1. When I close incremental snapshot, set `scan.incremental.snapshot.enabled=false`, it'll read all records.

[...]

PR #1918, opened 2023-02-14
[oracle-cdc] Fix read one record in the incremental snapshot

e-mhui:
Fix bug: #1886 

In the snapshot, `record.key()` is null, so it will overwrite the previous record. So we can to add record without primary key into list.

```java
  if (!reachChangeLogStart) {
      if (record.key() == null) {
          snapshotRecordsWithoutKey.add(record);
      } else {
          outputBuffer.put((Struct) record.key(), record);
      }
  }
```


EMsnap:
The primary key is vital for snapshot, so this problem cannot happen

ruanhang1993:
Maybe this is a new feature that make oracle cdc support the tables without primary keys.

yuxiqian:
Thanks @e-mhui for the great work! Before this PR could be merged, could you please rebase it with latest `master` branch?

cc @ruanhang1993

Issue #1934, opened 2023-02-23
[Bug][mongodb] Attempt to switch database target during SASL authentication error when using a non admin auth source and duplicated usernames.


Jiabao-Sun:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

Flink 1.13.6

### Flink CDC version

2.3.0

### Database and its version

MongoDB 4.4.1

### Minimal reproduce step

1. Have multiple similar accounts (with same credentials), one in "admin" and the other in another database.
2. Intend to log in with the other account not in the "admin" database.
3. Have SASL enabled.

### What did you expect to see?

Correctly connected mongodb.

### What did you see instead?

Command failed with error 17 (ProtocolError): 'Attempt to switch database target during SASL authentication.

### Anything else?

https://jira.mongodb.org/browse/JAVA-4290

> Speculative authentication always uses "admin" as the source.
>
> In certain circumstances this can lead to a Attempt to switch database target during SASL authentication error when using > a non admin auth source and duplicated usernames.

We may upgrade the mongo driver version greater or equals than 4.3.2 to solve this problem.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #1935, opened 2023-02-23
[mongodb][hotfix] Fix authentication error when using a non admin auth source and duplicated usernames

Jiabao-Sun:
Fix authentication error when using a non admin auth source and duplicated usernames.
#1934


GOODBOY008 (approved):
LGTM

PR #2007, opened 2023-03-15
[oracle-cdc]Optimized slow oracle real-time synchronization

weAreFriendYo:
#https://github.com/ververica/flink-cdc-connectors/issues/1999
To solve the problem of slow real-time oracle synchronization, add a table name as a query condition, improving query efficiency and reducing memory usage


weAreFriendYo:
没改全

e-mhui:
提高了查询元数据的速度，能够提高同步速度吗？为什么呀

PR #2037, opened 2023-03-23
fix oracle test bug.

empcl:
fix oracle test bug.


ruanhang1993:
This test has been fixed.

Issue #2034, opened 2023-03-22
[Bug] TiDB Flink CDC not support exactly once


ArainChen:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.13.6

### Flink CDC version

I had try three version:
2.3.0
2.4-SNAPSHOT
2.2.1

### Database and its version

I had try tow version:
tidb v5.4.3
tidb v6.1.3

### Minimal reproduce step

(1)Start a Flink CDC program locally, read a table in tidb, and enable checkpoints in the Flink program to remain in a directory (I tested the Flink program locally, so the checkpoints save location uses the local directory).
(2)Stop this Flink CDC program to simulate an exception scenario, and then insert, delete, and update this table in TiDB.
(3)Restoring the Flink cdc program from checkpoints will trace the operation of the Tidb table after the program experienced an exception. Insertion and update operations on the Tidb table can be restored to the Flink program for processing, but the deletion operation on the Tidb table has been lost.

### What did you expect to see?

Restoring the Flink cdc program from checkpoints will trace the operations of the Tidb table after the program experienced an exception. Insertion, update, and deletion operations on the Tidb table can be restored to the Flink program for processing, which means that the deletion operation will not be lost.

### What did you see instead?

Restoring the Flink cdc program from checkpoints will trace the operation of the Tidb table after the program experienced an exception. Insertion and update operations on the Tidb table can be restored to the Flink program for processing, but the deletion operation on the Tidb table has been lost.

### Anything else?

Does Flink cdc tidb really implement end-to-end exactly once?

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

CtrlSorcerer:
Have you tested it? If you initially set 'scan. startup. mode' to 'latest offset', stop the task, generate a savepoint, and use this savepoint to start, the previous historical data will still be synchronized.
In my understanding, he should not start synchronizing from the latest time, why would he still synchronize historical data.

ArainChen:
> Have you tested it? If you initially set 'scan. startup. mode' to 'latest offset', stop the task, generate a savepoint, and use this savepoint to start, the previous historical data will still be synchronized. In my understanding, he should not start synchronizing from the latest time, why would he still synchronize historical data.

[...]

PR #2038, opened 2023-03-23
[tidb]fix the tidb checkpoints to exactly once.

ArainChen:
#2034 I fix that bug.


524687460:
I used the same method, but there is another problem: when starting from checkpoint, I get an exception occasionally
捕获

yuxiqian:
Thanks @ArainChen for the great work! Before this PR could be merged, could you please rebase it with latest `master` branch?

PR #2044, opened 2023-03-26
feat comment

tyat:
if the difference between the current ID and the minimum ID is large, but the amount of data in the table is relatively small(for example, 10W), it will trigger the reverse optimization of the Flink cdc


ruanhang1993:
@tyat Thanks for contribution. 
Actually we should add some docs for these features. I have opened a issue #2164 to trace it.
Would you like to help? Let raise a new PR about this.

Best.

PR #2046, opened 2023-03-27
[ISSUE-1827][Mysql] Mysql connector support parallel snapshot when th…

### LINKED ISSUE #1827: Mysql connector support parallel snapshot when there are no primary key for tables
**Is your feature request related to a problem? Please describe.**

Mysql connector now only support parallel snapshot for tables with primary key which means that 
we can only use cdc-version1  to support multiple tables read when there are some tables are without primary key 
Read multiple tables using just one subtask is pretty slow and unbearable for the users

**Describe the solution you'd like**

1、remove the constraint in mysql-cdc 2.0 for primary key
2、for those tables which don't contain primary key, split the whole table as a chunk during split stage
3、since there's no need for the SnapshotSplitReader to upsert binlog with the snapshot data for a whole table
split , just let the highwatermark equals to the lowwatermark to escape the upsert stage.

zstraw:
Resolve #1827

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
@@ -17,6 +17,7 @@
 package com.ververica.cdc.connectors.mysql.debezium.reader;
 
 import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.util.CollectionUtil;
 import org.apache.flink.util.FlinkRuntimeException;
 
 import org.apache.flink.shaded.guava30.com.google.common.util.concurrent.ThreadFactoryBuilder;
@@ -49,8 +50,10 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
@@ -248,7 +251,11 @@ public class SnapshotSplitReader implements DebeziumReader<SourceRecords, MySqlS
             boolean reachBinlogEnd = false;
             SourceRecord lowWatermark = null;
             SourceRecord highWatermark = null;
-            Map<Struct, SourceRecord> snapshotRecords = new LinkedHashMap<>();
+
+            SnapshotRecords snapshotRecords =
+                    containsPrimaryKey()
+                            ? new PrimaryKeySnapshotRecords()
+                            : new NoPrimaryKeySnapshotRecords();
             while (!reachBinlogEnd) {
                 checkReadException();
                 List<DataChangeEvent> batch = queue.poll();
@@ -273,14 +280,7 @@ public class SnapshotSplitReader implements DebeziumReader<SourceRecords, MySqlS
                         break;
                     }
 
-                    if (!reachBinlogStart) {
-                        snapshotRecords.put((Struct) record.key(), record);
-                    } else {
-                        if (isRequiredBinlogRecord(record)) {
-                            // upsert binlog events through the record key
-                            upsertBinlog(snapshotRecords, record);
-                        }
-                    }
+                    snapshotRecords.collect(record, reachBinlogStart);
                 }
             }
             // snapshot split return its data once
@@ -288,7 +288,7 @@ public class SnapshotSplitReader implements DebeziumReader<SourceRecords, MySqlS
 
             final List<SourceRecord> normalizedRecords = new ArrayList<>();
             normalizedRecords.add(lowWatermark);
-            normalizedRecords.addAll(formatMessageTimestamp(snapshotRecords.values()));
+            normalizedRecords.addAll(formatMessageTimestamp(snapshotRecords.getRecords()));
             normalizedRecords.add(highWatermark);
 
             final List<SourceRecords> sourceRecordsSet = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
```diff
@@ -403,4 +403,58 @@ public boolean isRunning() {
             return currentTaskRunning;
         }
     }
+
+    private boolean containsPrimaryKey() {
+        return !CollectionUtil.isNullOrEmpty(currentSnapshotSplit.getSplitKeyType().getFields());
```
ruanhang1993, line 408: It is better not to judge in this way. The chunk key will be set when we have to handle the big table in some cases to provide the at-least-once guarantee.

ruanhang1993: We could init the SnapshotRecordCollector when the `record.key()` returns null.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
```diff
@@ -403,4 +403,58 @@ public boolean isRunning() {
...
+    }
+
+    /** Collect records need to be sent, except low/high watermark. */
+    interface SnapshotRecords {
```
ruanhang1993, line 412: SnapshotRecords -> SnapshotRecordCollector

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlChunkSplitter.java
```diff
@@ -148,7 +148,12 @@ private void analyzeTable(TableId tableId) {
                     ChunkUtils.getChunkKeyColumn(
                             currentSplittingTable, sourceConfig.getChunkKeyColumn());
             splitType = ChunkUtils.getChunkKeyColumnType(splitColumn);
-            minMaxOfSplitColumn = queryMinMax(jdbcConnection, tableId, splitColumn.name());
+            // split column would be null if table has no primary key.
+            if (splitColumn != null) {
+                minMaxOfSplitColumn = queryMinMax(jdbcConnection, tableId, splitColumn.name());
+            } else {
+                minMaxOfSplitColumn = new Object[2];
+            }
```
ruanhang1993, line 156: The chunk splitter should return a total split immediately when the table does not have the primary key. This part should not be changed.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/ChunkUtils.java
```diff
@@ -42,20 +42,20 @@ public static RowType getChunkKeyColumnType(Table table, @Nullable String chunkK
         return getChunkKeyColumnType(getChunkKeyColumn(table, chunkKeyColumn));
     }
 
-    public static RowType getChunkKeyColumnType(Column chunkKeyColumn) {
+    public static RowType getChunkKeyColumnType(@Nullable Column chunkKeyColumn) {
+        if (chunkKeyColumn == null) {
+            return (RowType) ROW().getLogicalType();
+        }
```
ruanhang1993, line 48: This should not be changed as this. I think we could use the first column as the chunk key if the chunk key is not set in the table options.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
```diff
@@ -403,4 +403,58 @@ public boolean isRunning() {
...
+    }
+
+    /** Collect records without primary key. There are no binlog events. */
+    static class NoPrimaryKeySnapshotRecords implements SnapshotRecords {
+        private final List<SourceRecord> snapshotRecords = new LinkedList<>();
+
+        @Override
+        public void collect(SourceRecord record, boolean reachBinlogStart) {
+            if (!reachBinlogStart) {
+                snapshotRecords.add(record);
```
ruanhang1993, line 447: How to handle the delete and update events?

ruanhang1993 (commented):
Thanks for the contribution. There are still some missing places to handle.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
zstraw: > It is better not to judge in this way. The chunk key will be set when we have to handle the big table in some cases to provide the at-least-once guarantee.

Thanks for your review. You mean that the chunk key may be set even though the table has no primary keys?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/ChunkUtils.java
zstraw: You mean using the first column as chunk key to split the table if it has no primary key ?

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
ruanhang1993: Yes. If the table contains too much data, there is a performance problem with this solution.
In this case, we may provide the at-least-once semantics instead of the exactly once semantics.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
zstraw: IMHO, the snapshot stage for the table without primary key only read snapshot data, then binlog data would be read only in the incremental stage. So there are no delete and update events in SnapshotSplitReader if the table has no primary key. Please correct me.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
zstraw: Is it possible that the exactly once semantic is required in some scenarios ? How about to provide an option to make it optional for users who can choose between better performance or exactly once semantic?

ruanhang1993: I think we should control this by the experimental option `scan.incremental.snapshot.chunk.key-column`.
By default, we will return a single split. If users set the `scan.incremental.snapshot.chunk.key-column`, we will split the table into multi splits.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/ChunkUtils.java
ruanhang1993: I replied about this part at the first comment.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
ruanhang1993: It is a good idea to skip the snapshot backfill task when there is a single split. I think you should set this in source but I do not see the relative changes.

ruanhang1993:
Hi, @zstraw ,
I have opened a new PR to implement this based on this PR. Maybe you could help to review it.
Thanks ~

PR #2052, opened 2023-03-28
fix(db2-doc): for db2 , char should be CHARACTER

ZhongLinLeo:
db2 cdc doc modify , char type should be character type . linked below

https://www.ibm.com/docs/en/db2-for-zos/12?topic=columns-string-data-types


GOODBOY008 (approved):
LGTM @ZhongLinLeo Thanks for your contribution.

PR #2063, opened 2023-04-01
[mysql] Fix issue #1944: Initialize complete GTIDs to ensure subsequent recovery from checkpoint

### LINKED ISSUE #1944: [Bug] Flink CDC 2.3.0 set startupOptions = specificOffset set specificOffsetFile and specificOffsetPos then can not start from checkpoint
### Search before asking

### Flink version

1.16.0

### Flink CDC version

2.3.0

### Database and its version

mysql5.7

### Minimal reproduce step

if i restart job from checkpoint can not work

2023-02-21 14:08:29,713 INFO com.ververica.cdc.connectors.mysql.debezium.task.context.StatefulTaskContext [] - MySQL current GTID set 106a4bb6-ec0d-11ec-a2d4-00163e279211:1-204182899,7aec1281-719c-11eb-afcf-00163e06a35c:1-147359662 does contain the GTID set 106a4bb6-ec0d-11ec-a2d4-00163e279211:203495054-204182173 required by the connector.
2023-02-21 14:08:29,801 INFO com.ververica.cdc.connectors.mysql.debezium.task.context.StatefulTaskContext [] - Server has already purged 106a4bb6-ec0d-11ec-a2d4-00163e279211:1-203495053,7aec1281-719c-11eb-afcf-00163e06a35c:1-147359662 GTIDs
2023-02-21 14:08:29,802 WARN com.ververica.cdc.connectors.mysql.debezium.task.context.StatefulTaskContext [] - Some of the GTIDs needed to replicate have been already purged
2023-02-21 14:08:29,803 ERROR org.apache.flink.connector.base.source.reader.fetcher.SplitFetcherManager [] - Received uncaught exception.
java.lang.RuntimeException: SplitFetcher thread 0 received unexpected exception while polling the records
at org.apache.flink.connector.base.source.reader.fetcher.SplitFetcher.runOnce(SplitFetcher.java:150) ~[flink-connector-files-1.16.1.jar:1.16.1]
at org.apache.flink.connector.base.source.reader.fetcher.SplitFetcher.run(SplitFetcher.java:105) [flink-connector-files-1.16.1.jar:1.16.1]
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_362]
at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:1.8.0_362]
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_362]
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_362]
at java.lang.Thread.run(Thread.java:750) [?:1.8.0_362]
Caused by: java.lang.IllegalStateException: The connector is trying to read binlog starting at Struct{version=1.6.4.Final,connector=mysql,name=mysql_binlog_source,ts_ms=1676959709803,db=,server_id=0,file=mysql-bin.005893,pos=15052069,row=0}, but this is no longer available on the server. Reconfigure the connector to use a snapshot when needed.
at com.ververica.cdc.connectors.mysql.debezium.task.context.StatefulTaskContext.loadStartingOffsetState(StatefulTaskContext.java:194) ~[blob_p-384c1d850f997d73acba83fc72dc8cfb8bead162-07d02e9eebcf14d5793c784b1d89c6cc:?]
at com.ververica.cdc.connectors.mysql.debezium.task.context.StatefulTaskContext.configure(StatefulTaskContext.java:117) ~[blob_p-384c1d850f997d73acba83fc72dc8cfb8bead162-07d02e9eebcf14d5793c784b1d89c6cc:?]
at com.ververica.cdc.connectors.mysql.debezium.reader.BinlogSplitReader.submitSplit(BinlogSplitReader.java:103) ~[blob_p-384c1d850f997d73acba83fc72dc8cfb8bead162-07d02e9eebcf14d5793c784b1d89c6cc:?]
at com.ververica.cdc.connectors.mysql.debezium.reader.BinlogSplitReader.submitSplit(BinlogSplitReader.java:71) ~[blob_p-384c1d850f997d73acba83fc72dc8cfb8bead162-07d02e9eebcf14d5793c784b1d89c6cc:?]
at com.ververica.cdc.connectors.mysql.source.reader.MySqlSplitReader.checkSplitOrStartNext(MySqlSplitReader.java:159) ~[blob_p-384c1d850f997d73acba83fc72dc8cfb8bead162-07d02e9eebcf14d5793c784b1d89c6cc:?]
at com.ververica.cdc.connectors.mysql.source.reader.MySqlSplitReader.fetch(MySqlSplitReader.java:71) ~[blob_p-384c1d850f997d73acba83fc72dc8cfb8bead162-07d02e9eebcf14d5793c784b1d89c6cc:?]
at org.apache.flink.connector.base.source.reader.fetcher.FetchTask.run(FetchTask.java:58) ~[flink-connector-files-1.16.1.jar:1.16.1]
at org.apache.flink.connector.base.source.reader.fetcher.SplitFetcher.runOnce(SplitFetcher.java:142) ~[flink-connector-files-1.16.1.jar:1.16.1]
... 6 more

gtids in checkpoint:"file":"mysql-bin.005893","pos":"15052069","kind":"SPECIFIC","gtids":"106a4bb6-ec0d-11ec-a2d4-00163e279211:203495054-204182173"
show master status gtids :106a4bb6-ec0d-11ec-a2d4-00163e279211:1-204479617,
7aec1281-719c-11eb-afcf-00163e06a35c:1-147359662
show binary logs：
企业微信截图_16772288204747

if first set specificOffset then can work from ck

### What did you see instead?

if first set specificOffset then can not work from ck

### Anything else?

_No response_

wallkop:
Resolve #1944

> <b>Problem analysis</b>

Issue #1944 reports a bug where recovery from checkpoint fails when starting from offset-binlogfile&pos. After conducting further tests, I found that this bug can occur in Earliest, Timestamp, BinlogFile&Pos startup modes, as well as when incomplete Gtids are incorrectly set. The root cause of this bug is that when MySQL has GTID enabled and the aforementioned startup modes are used, the complete GTIDs cannot be recorded in the offsetContext. Instead, an empty GtidSet is initialized and subsequently appended by received Gtid Events. This leads to recovery failure when restoring from checkpoint due to incomplete GTIDs being recorded.

_Test premise: MySQL needs to have GTID enabled, and due to settings such as master-slave switching and multiple masters, there are multiple sourceIDs for GTIDs_
|startupMode|How to initialize GTID|Restore from a checkpoint|
|-|-|-|
| initial | Obtain the GTID value through SHOW MASTER STATUS before starting the snapshot | √ |
| latest | Obtain the GTID value through SHOW MASTER STATUS before starting to consume | √ |
| offset-gtid | Use the user-set GTID before starting to consume.<b>Note: The user needs to set the complete GTID, otherwise it will not pass the startup verification</b> | √ |
| offset-binlogfile&pos | Set the consumption anchor of the binlog through binlog_file='xxx', offset=yyy before starting to consume, but the GTID is null after initialization | X |
| earliest | Set the consumption anchor of the binlog through binlog_file='', offset=0 before starting to consume, but the GTID is null after initialization | X |
| timestamp | Set the consumption anchor of the binlog through binlog_file='', offset=0 before starting to consume, but the GTID is null after initialization. After starting the consumption, all binlog events with timestamps less than the specified timestamp will be skipped until an event with a timestamp greater than or equal to timestamp is found, and the GTID will be recorded at that point | X |

> <b>Solution approach</b>

There are two solutions to fix this bug:

* <b>Solution A</b>: When recovering from checkpoint, check whether the startup is based on GTID or binlogFile&Pos settings. If it is based on latest/initial/offset-gitds startup, recover from GTID. If it is based on earliest/timestamp/binlogFile&Pos startup, recover from binlogFile&Pos.

* <b>Solution B</b>: Attempt to initialize the complete GTIDs when starting from earliest/timestamp/binlogFile&Pos, and subsequently recover from GTID in all checkpoints.

Personally, I am more inclined towards Solution B because in complex production environments of MySQL, there are situations such as master-slave architecture, multi-master architecture, failover, etc. Using GTID may avoid more unpredictable problems than using binlogFile&Pos. Therefore, this PR is also implemented based on Solution B.

> <b>Implementation logic</b>

Copied the GtidSet class from com.github.shyiko.mysql.binlog and added the overwriteThenAdd method:
In the CDC scenario, if starting from a certain offset, it can be assumed that the data before that offset will not be consumed again. Therefore, when initializing the GtidSet, if the first GTID event is received, the overwrite method can be used to correct the GtidSet to be complete and continuous. For example, if the initial gtidSet is `A:1-100,B:1-20`, but due to filtering, the first received GTID event is `B:50`, the overwrite method will correct the gtidSet to `A:1-100,B:1-50` to ensure the completeness and continuity of the GtidSet. In the GtidSet, each sourceID has an independent trigger switch to ensure that each SourceID triggers an overwrite when first written, and then uses the add method for updates.

Made some changes in the MySqlStreamingChangeEventSource class and adjusted the logic for initializing the GtidSet:
Starting from earliest/timestamp: The client startup anchor/offsetContext's gtidSet is set to purgedGtids, so CDC will consume from the valid GTID and record the correct and complete GtidSet. For timestamp, because a batch of data (<timestamp) will be skipped after startup, the overwriteThenAppend method will be used to correct it when the first GTID event is received.

Starting from offset-binlogFile&Pos: The client startup anchor is set to binlogFile&Pos. Since it is not cost-effective to obtain accurate corresponding Gtids, the initial value of the gtidSet is set to executedGTIDs (obtained from show master status). At this time, the gtidSet is incorrect, but it will not record the checkpoint. The overwriteThenAppend method will be used to correct it when the first GTID event is received. For example, if the initialized gtidSet is `A:1-100,B:1-100`, and the first received GTID event is `B:50`, the overwrite method will correct it to `A:1-100,B:1-50`.

Starting from offset-gtids: No modifications. As MySQL-CDC will verify the user-set offset-gtids, the task cannot be started if it is not set correctly.

Starting from latest/initial: No modifications. This is because when starting through these two modes, the gtidSet will be initialized through show master status to obtain the complete GTIDs.

> <b>Test cases</b>

Three unit tests have been added for the newly added "overwriteThenAdd" method in the "GtidSet" class to ensure that it works correctly.

Due to the complexity of the modifications made to the "MySqlStreamingChangeEventSource" class, a set of integration tests based on docker-mysql have been written. Eighteen test cases have been written for the initial/latest/earliest/timestamp/offset-gtids/offset-binlogfile&pos startup modes to ensure that the code works correctly.

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/github/shyiko/mysql/binlog/GtidSet.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/github/shyiko/mysql/binlog/GtidSet.java
@@ -1,0 +1,448 @@
+/*
+ * Copyright 2015 Stanley Shyiko
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.shyiko.mysql.binlog;
+
+import java.util.*;
+
+/**
+ * Copied from https://github.com/osheroff/mysql-binlog-connector-java project to fix
+ * https://github.com/ververica/flink-cdc-connectors/issues/1944
+ *
+ * <p>Line 106 ~ 120: A new method called "overwriteThenAdd" has been added to address the incomplete Gtid storage
+ * issue in MySQL-CDC when storing checkpoints. In the MySQL-CDC scenario, consuming from a certain GTID should be
+ * ordered, and GTIDs before this should not be consumed again. Therefore, it is assumed that when recording a
+ * checkpoint, the historical offset should also be recorded as processed.
+ */
+public class GtidSet {
+
+    private final Map<String, UUIDSet> map = new LinkedHashMap<String, UUIDSet>();
+
+    /**
+     * @param gtidSet gtid set comprised of closed intervals (like MySQL's executed_gtid_set).
+     */
+    public GtidSet(String gtidSet) {
+        String[] uuidSets = (gtidSet == null || gtidSet.isEmpty()) ? new String[0] :
+                gtidSet.replace("\n", "").split(",");
+        for (String uuidSet : uuidSets) {
+            int uuidSeparatorIndex = uuidSet.indexOf(":");
+            String sourceId = uuidSet.substring(0, uuidSeparatorIndex);
+            List<Interval> intervals = new ArrayList<Interval>();
+            String[] rawIntervals = uuidSet.substring(uuidSeparatorIndex + 1).split(":");
+            for (String interval : rawIntervals) {
+                String[] is = interval.split("-");
+                long[] split = new long[is.length];
+                for (int i = 0, e = is.length; i < e; i++) {
+                    split[i] = Long.parseLong(is[i]);
+                }
+                if (split.length == 1) {
+                    split = new long[] {split[0], split[0]};
+                }
+                intervals.add(new Interval(split[0], split[1]));
+            }
+            map.put(sourceId, new UUIDSet(sourceId, intervals));
+        }
+    }
+
+    /**
+     * Get an immutable collection of the {@link UUIDSet range of GTIDs for a single server}.
+     * @return the {@link UUIDSet GTID ranges for each server}; never null
+     */
+    public Collection<UUIDSet> getUUIDSets() {
+        return Collections.unmodifiableCollection(map.values());
+    }
+
+    /**
+     * Find the {@link UUIDSet} for the server with the specified UUID.
+     * @param uuid the UUID of the server
+     * @return the {@link UUIDSet} for the identified server, or {@code null} if there are no GTIDs from that server.
+     */
+    public UUIDSet getUUIDSet(String uuid) {
+        return map.get(uuid);
+    }
+
+    /**
+     * Add or replace the UUIDSet
+     * @param uuidSet UUIDSet to be added
+     * @return the old {@link UUIDSet} for the server given in uuidSet param,
+     *         or {@code null} if there are no UUIDSet for the given server.
+     */
+    public UUIDSet putUUIDSet(UUIDSet uuidSet) {
+        return map.put(uuidSet.getUUID(), uuidSet);
+    }
+
+    /**
+     * @param gtid GTID ("source_id:transaction_id")
+     * @return whether or not gtid was added to the set (false if it was already there)
+     */
+    public boolean add(String gtid) {
+        String[] split = gtid.split(":");
+        String sourceId = split[0];
+        long transactionId = Long.parseLong(split[1]);
+        UUIDSet uuidSet = map.get(sourceId);
+        if (uuidSet == null) {
+            map.put(sourceId, uuidSet = new UUIDSet(sourceId, new ArrayList<Interval>()));
+        }
+        return uuidSet.add(transactionId);
+    }
+
+    /**
+     * Add a Gtid string to the current GtidSet. If the corresponding sourceId is being added for the first time,
+     * overwrite it; otherwise, add it.
+     * Overwrite: For example, if the current GtidSet is A:1-100, and a Gtid A:150 is added with overwrite, then the
+     * current GtidSet will become A:1-150.
+     * @param gtid GTID ("source_id:transaction_id")
+     * @return whether or not gtid was added to the set (false if it was already there)
+     */
+    public boolean overwriteThenAdd(String gtid) {
+        String[] split = gtid.split(":");
+        String sourceId = split[0];
+        long transactionId = Long.parseLong(split[1]);
+        UUIDSet uuidSet = map.get(sourceId);
+        if (uuidSet == null) {
+            map.put(sourceId, uuidSet = new UUIDSet(sourceId, new ArrayList<>()));
+        }
+        if (!uuidSet.isOverwrited) {
+            uuidSet.overwrite(transactionId);
+            return true;
+        } else {
+            return uuidSet.add(transactionId);
+        }
+    }
+
+    /**
+     * Determine if the GTIDs represented by this object are contained completely within the supplied set of GTIDs.
+     * Note that if two {@link GtidSet}s are equal, then they both are subsets of the other.
+     * @param other the other set of GTIDs; may be null
+     * @return {@code true} if all of the GTIDs in this set are equal to or completely contained within the supplied
+     * set of GTIDs, or {@code false} otherwise
+     */
+    public boolean isContainedWithin(GtidSet other) {
+        if (other == null) {
+            return false;
+        }
+        if (this == other) {
+            return true;
+        }
+        if (this.equals(other)) {
+            return true;
+        }
+        for (UUIDSet uuidSet : map.values()) {
+            UUIDSet thatSet = other.getUUIDSet(uuidSet.getUUID());
+            if (!uuidSet.isContainedWithin(thatSet)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        return map.keySet().hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+        if (obj instanceof GtidSet) {
+            GtidSet that = (GtidSet) obj;
+            return this.map.equals(that.map);
+        }
+        return false;
+    }
+
+    @Override
+    public String toString() {
+        List<String> gtids = new ArrayList<String>();
+        for (UUIDSet uuidSet : map.values()) {
+            gtids.add(uuidSet.getUUID() + ":" + join(uuidSet.intervals, ":"));
+        }
+        return join(gtids, ",");
+    }
+
+    private static String join(Collection<?> o, String delimiter) {
+        if (o.isEmpty()) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        for (Object o1 : o) {
+            sb.append(o1).append(delimiter);
+        }
+        return sb.substring(0, sb.length() - delimiter.length());
+    }
+
+    /**
+     * A range of GTIDs for a single server with a specific UUID.
+     * @see GtidSet
+     */
+    public static final class UUIDSet {
+
+        private String uuid;
+        private List<Interval> intervals;
+        private boolean isOverwrited = false;
+
+        public UUIDSet(String uuid, List<Interval> intervals) {
+            this.uuid = uuid;
+            this.intervals = intervals;
+            if (intervals.size() > 1) {
+                joinAdjacentIntervals(0);
+            }
+        }
+
+        private void overwrite(long transactionId) {
+            intervals = new ArrayList<>(Collections.singletonList(new Interval(1, transactionId)));
+            isOverwrited = true;
+        }
+
+        private boolean add(long transactionId) {
+            int index = findInterval(transactionId);
+            boolean addedToExisting = false;
+            if (index < intervals.size()) {
+                Interval interval = intervals.get(index);
+                if (interval.start == transactionId + 1) {
+                    interval.start = transactionId;
+                    addedToExisting = true;
+                } else
+                if (interval.end + 1 == transactionId) {
+                    interval.end = transactionId;
+                    addedToExisting = true;
+                } else
+                if (interval.start <= transactionId && transactionId <= interval.end) {
+                    return false;
+                }
+            }
+            if (!addedToExisting) {
+                intervals.add(index, new Interval(transactionId, transactionId));
+            }
+            if (intervals.size() > 1) {
+                joinAdjacentIntervals(index);
+            }
+            return true;
+        }
+
+        /**
+         * Collapses intervals like a-(b-1):b-c into a-c (only in index+-1 range).
+         */
+        private void joinAdjacentIntervals(int index) {
+            for (int i = Math.min(index + 1, intervals.size() - 1), e = Math.max(index - 1, 0); i > e; i--) {
+                Interval a = intervals.get(i - 1), b = intervals.get(i);
+                if (a.end + 1 == b.start) {
+                    a.end = b.end;
+                    intervals.remove(i);
+                }
+            }
+        }
+
+        /**
+         * @return index which is either a pointer to the interval containing v or a position at which v can be added
+         */
+        private int findInterval(long v) {
+            int l = 0, p = 0, r = intervals.size();
+            while (l < r) {
+                p = (l + r) / 2;
+                Interval i = intervals.get(p);
+                if (i.end < v) {
+                    l = p + 1;
+                } else
+                if (v < i.start) {
+                    r = p;
+                } else {
+                    return p;
+                }
+            }
+            if (!intervals.isEmpty() && intervals.get(p).end < v) {
+                p++;
+            }
+            return p;
+        }
+
+        /**
+         * Get the UUID for the server that generated the GTIDs.
+         * @return the server's UUID; never null
+         */
+        public String getUUID() {
+            return uuid;
+        }
+
+        /**
+         * Get the intervals of transaction numbers.
+         * @return the immutable transaction intervals; never null
+         */
+        public List<Interval> getIntervals() {
+            return Collections.unmodifiableList(intervals);
+        }
+
+        /**
+         * Determine if the set of transaction numbers from this server is completely within the set of transaction
+         * numbers from the set of transaction numbers in the supplied set.
+         * @param other the set to compare with this set
+         * @return {@code true} if this server's transaction numbers are equal to or a subset of the transaction
+         * numbers of the supplied set, or false otherwise
+         */
+        public boolean isContainedWithin(UUIDSet other) {
+            if (other == null) {
+                return false;
+            }
+            if (!this.getUUID().equalsIgnoreCase(other.getUUID())) {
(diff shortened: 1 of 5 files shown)


ruanhang1993:
Hi, @wallkop.
Thanks for your contribution.
IMO, I think what we should do is to modify the `checkGtidSet` instead of other parts.

The new startup modes in 2.3.0 make the gtid set in checkpoint start from a middle position instead of 1. This behavior is right because it records the gtid set that it process. The `checkGtidSet` method does not think about this situation.The `checkGtidSet` method should check that the gtid set behind the checkpoint/savepoint is still available in the server. 

For example, the available gtid(from SHOW MASTER STATUS) set is 'uuidA:1-200,uuidB:1-300'. The purged gtid set(from SELECT @@global.gtid_purged) is 'uuidA:1-100,uuidB:1-200'. So the non-purged gtid set is 'uuidA:101-200,uuidB:201-300'. `checkGtidSet` method should make sure that the next gtid is in non-purged gtid set 'uuidA:101-200,uuidB:201-300'. 

If the gtid set in the checkpoint/savepoint is 'uuidA:50-90,uuidB:210-260'. We could use the end of interval from each uuidSet. The next gtid will be 'uuidA:91,uuidB:261'. It is not contained in non-purged gtid set. `checkGtidSet` method should return `false`.

If the gtid set in the checkpoint/savepoint is 'uuidA:190-195,uuidB:210-280'. The next gtid will be 'uuidA:196,uuidB:281'.  It is contained in non-purged gtid set. `checkGtidSet` method should return `true`.

The classes `GtidSet`/`UuidSet` and MySQL provide many useful method for us.

wallkop:
> Hi, @wallkop. Thanks for your contribution. IMO, I think what we should do is to modify the `checkGtidSet` instead of other parts.
> 
> The new startup modes in 2.3.0 make the gtid set in checkpoint start from a middle position instead of 1. This behavior is right because it records the gtid set that it process. The `checkGtidSet` method does not think about this situation.The `checkGtidSet` method should check that the gtid set behind the checkpoint/savepoint is still available in the server.
> 
> For example, the available gtid(from SHOW MASTER STATUS) set is 'uuidA:1-200,uuidB:1-300'. The purged gtid set(from SELECT @@global.gtid_purged) is 'uuidA:1-100,uuidB:1-200'. So the non-purged gtid set is 'uuidA:101-200,uuidB:201-300'. `checkGtidSet` method should make sure that the next gtid is in non-purged gtid set 'uuidA:101-200,uuidB:201-300'.
> 
> If the gtid set in the checkpoint/savepoint is 'uuidA:50-90,uuidB:210-260'. We could use the end of interval from each uuidSet. The next gtid will be 'uuidA:91,uuidB:261'. It is not contained in non-purged gtid set. `checkGtidSet` method should return `false`.
> 
> If the gtid set in the checkpoint/savepoint is 'uuidA:190-195,uuidB:210-280'. The next gtid will be 'uuidA:196,uuidB:281'. It is contained in non-purged gtid set. `checkGtidSet` method should return `true`.
> 
> The classes `GtidSet`/`UuidSet` and MySQL provide many useful methods for us.


@ruanhang1993 , Thank you very much for your reply.

In fact, I have been considering modifying only the checkGtidSet method to solve the problem, but then I found that it was not feasible. Let me give you an example:
```
A MySQL instance:
Executed Gtids: 9cb57eca-8662-11ed-bf33-00163e006d32:1-147615502
Purged Gtids: 9cb57eca-8662-11ed-bf33-00163e006d32:1-117083571
```

If we start in the earliest mode, it will start consuming data from `9cb57eca-8662-11ed-bf33-00163e006d32:117083572` and record a checkpoint, for example:
`Binlog offset on checkpoint 14: {transaction_id=null, ts_sec=1679913081, file=mysql-bin.000540, pos=291722698, kind=SPECIFIC, gtids=9cb57eca-8662-11ed-bf33-00163e006d32:117083572-117134093, row=1, event=2, server_id=782376004}`

At this point, if we make checkGtidSet return true and recover from the checkpoint, the following error will be prompted:
```
Caused by: com.github.shyiko.mysql.binlog.network.ServerException: The slave is connecting using CHANGE MASTER TO MASTER_AUTO_POSITION = 1, but the master has purged binary logs containing GTIDs that the slave requires.
	at com.github.shyiko.mysql.binlog.BinaryLogClient.listenForEventPackets(BinaryLogClient.java:944) ~[zyb_flink_cdc-1.0-SNAPSHOT.jar:?]
	at com.github.shyiko.mysql.binlog.BinaryLogClient.connect(BinaryLogClient.java:599) ~[zyb_flink_cdc-1.0-SNAPSHOT.jar:?]
	at com.github.shyiko.mysql.binlog.BinaryLogClient$7.run(BinaryLogClient.java:857) ~[zyb_flink_cdc-1.0-SNAPSHOT.jar:?]
	... 1 more
```

Therefore, I tried to modify the GTIDs stored in offsetContext and developed the solution in this PR.

However, I strongly agree with your suggestion that making significant modifications to the GTIDs stored in offsetContext will make future maintenance very difficult, especially when upgrading Debezium versions, and may also cause other unforeseen problems.

To this end, I have re-examined this issue and found the real problem:
In fact, when Debezium sets GTID, it has already considered the issue of MySQL multiple source failover and purged GTID, as can be seen in the code: `Line 1052 and Lines 1293~1336 of the MySqlStreamingChangeEventSource class`.

It implements a `filterGtidSet(MySqlOffsetContext offsetContext, GtidSet availableServerGtidSet, GtidSet purgedServerGtid)` method to merge purgedGtids to avoid Debezium reading expired GTIDs. However, this method has a defect, as seen in `Lines 1325~1329`:
```
mergedGtidSet =
        relevantAvailableServerGtidSet
                .retainAll(uuid -> knownGtidSet.forServerWithId(uuid) != null)
                .with(purgedServerGtid)
                .with(filteredGtidSet);
```
This code tries to implement three things in order:

* 1.Filter uuid from `relevantAvailableServerGtidSet` (equivalent to the gtid obtained from show master status if `gtidSourceFilter` is not enabled) and keep only the uuid contained in `offsetContext's gtid`.

* 2.Use `purgedServerGtid` to override the previous result. If a uuid in `purgedServerGtid` already exists in the previous result, it will be overwritten with the uuid in `purgedServerGtid`.

* 3.Use `filteredGtidSet` to override the previous result. If a uuid in `filteredGtidSet` already exists in the previous result, it will be overwritten with the uuid in `filteredGtidSet`.

If we use the example I mentioned earlier, where recovering from checkpoint fails, and try to execute this function, we will find that the execution process is as follows:
```
filteredGtidSet='9cb57eca-8662-11ed-bf33-00163e006d32:117083572-117134093' //(from checkpoint)

knownGtidSet=filteredGtidSet

relevantAvailableServerGtidSet='9cb57eca-8662-11ed-bf33-00163e006d32:1-147615502' //(from show master status)

purgedServerGtid='9cb57eca-8662-11ed-bf33-00163e006d32:1-117083571' //(from SELECT @@global.gtid_purged)

relevantAvailableServerGtidSet.retainAll(uuid -> knownGtidSet.forServerWithId(uuid) != null)
【RESULT】> [9cb57eca-8662-11ed-bf33-00163e006d32:1-147615502]

.with(purgedServerGtid)
【RESULT】> [9cb57eca-8662-11ed-bf33-00163e006d32:1-117083571]

.with(filteredGtidSet);
【RESULT】> [9cb57eca-8662-11ed-bf33-00163e006d32:117083572-117134093]
```

We can see that the final result does not match the expected result, as `purgedServerGtid` is not merged correctly, but is overwritten by `filteredGtidSet` because they have the same uuid. 

In fact, if there are multiple uuids due to a master-slave switch, this method will merge the expired uuids from the slave into the result (the slave's uuid is in `purgedServerGtid` but may not be in `filteredGtidSet`), and eventually work properly. However, all of this assumes that `filteredGtidSet` is set correctly: it needs to consider the case of gtid expiration and include that part, but clearly the gtid recorded in our checkpoint does not include `purgedGtids`, resulting in an error.

Therefore, making a slight adjustment to this method to merge the same uuid in `purgedServerGtid` and `filteredGtidSet` will solve the problem, and this is what my new solution plans to do:

* 1.Modify the checkGtidSet method to make its judgment logic more robust and work like the example you gave in your comment.
* 2.Modify the filterGtidSet method to merge purgedGtids in the same uuid.
* _If possible, I will also publish an issue to fix this problem in the Debezium community, so that we can smoothly upgrade Debezium version in the future._

<b>Thank you again for your comment, as it has further guided me. If you agree with my solution, I will modify this PR and improve the related tests.</b>

wallkop:
Sorry, there are some omissions in the new solution I gave above, it can only solve the problem when `StartupMode=earliest`.

If the StartupMode is set to` BinlogFile&Pos` or `Timestamp`, although the trunk can be restored from the checkpoint, it will generate some data that the user does not want to see, as shown in the figure below:

offset

So my new solution above will not be able to solve all the problems. 
For the time being, I have to keep the original implementation in this PR: `to solve the problem based on maintaining the complete Gtids in offsetContext`.

If you have other better suggestions, please let me know, thank you very much.

ruanhang1993:
Hi, @wallkop. Thanks for the quick reply. I will read these details and response later.

ruanhang1993:
Hi, @wallkop ,
Sorry for my late reply. I read the details and I wonder if it is possible to complete this task by changing the starting offset of the BinlogSplit for these startup modes.

I am not sure if it is right in the following way:
`MySqlBinlogSplitAssigner#createBinlogSplit` defines the BinlogSplit. If the startup mode is earliest, timestamp or specific offset, we should put the purged gtid set into the starting offset. Then everything will be ok.

Please correct me if I miss something. Thanks ~

wallkop:
Hi, @ruanhang1993, thank you very much for your reply. Let me try to explain the reason.

With GTID enabled in MySQL, only defining the purged GTID set in `MySqlBinlogSplitAssigner#createBinlogSplit` can allow `earliest` to recover from the checkpoint normally, but it cannot enable recovery of binlog file&pos and timestamp from the checkpoint normally.

Here is my analysis:

> Binlog file&pos:

When adding the purged GTID set in createBinlogSplit, the reader will start from the purged GTID set rather than the `binlog file&pos` when starting the reader, which is not expected.
The reason is that if the purged GTID set is set, the offsetContext.gtidSet will not return null in the `MySqlStreamingChangeEventSource#filterGtidSet` method, which will eventually result in the `client.setGtidSet(filteredGtidSetStr)` being set in `MySqlStreamingChangeEventSource#execute`, causing the reader to not start from the binlog file&pos.

> Timestamp:

The reader can start normally, but it cannot recover normally from the checkpoint using the timestamp. The reason is that when starting with a timestamp, Debezium's client will be set to start from the purged GTID set. However, in `MySqlBinlogSplitReadTask#handleEvent`, events that do not match the timestamp will be skipped, resulting in the offset recording being disjointed, such as `sourceId: 1-150:300-500`, which will eventually cause the recovery from the checkpoint to either re-consume the `sourceId: 150-300` data or fail to recover because the binlog has been purged. I don't think this is what we expect.

ruanhang1993:
@PatrickRen Do you have time to review this solution?

wallkop:
> @PatrickRen Do you have time to review this solution?

Hi @ruanhang1993 @PatrickRen ~

What is the progress on this PR?
If there are any concerns about the proposed solution, I would like to discuss them together, and I am more than willing to try a better approach.

PatrickRen:
Hi @wallkop and quite sorry for my late response. Instead of rewriting the GTID after the source starts to read the binlog, I'm still thinking that if it is possible to"cheat" the MySQL with a fake GTID right after the job restarts from a checkpoint. 

Considering the semantic of the checkpoint, the source only need to get an exact point or offset that it should start to read from after failover. That means although we store a set of GTID (let's say A:1-100) into the state, we only care about the ending GTID (A:100) for recovering and let the server send binlog events right after the ending GTID, so I'm thinking a potential solution:

Before the source do checkGtidSet() and setup the starting offset of the binlog client, we get the current available GTID set  on master (let's say A:1-100), and truncate it with the ending GTID restored from checkpoint (For GTID set A:40-50 in checkpoint, we use A:50) to only keep GTID set before it (the result or truncate is A:1-50). We use this truncated GTID set to do the GTID check and setup starting offset of binlog. This could "cheat" the MySQL server to pass the GTID validation and MySQL server can still respond with correct binlog events after the checkpointed offset. What do you think?

This is truly a tricky case and my explanation might be confusing. I'll try to give a PoC later.

wallkop:
Hello @PatrickRen ~ Thank you very much for your suggestion!

I believe we are aligned in our approach to the solution, where we only need to focus on the ending GTID position for recovery. When a job recovers from a checkpoint, we can first correct the GTID in the status, for example, modify `A:40-50` to `A:1-50`, to ensure successful deception of the check mechanism. Of course, we need to fully consider all other scenarios to deal with some unexpected boundary issues. I will later compile a list of cases to enumerate all possibilities as much as possible.

I will modify this PR later to implement the solution mentioned above, and the related test cases will be preserved.

PatrickRen:
@wallkop Thanks for your reply! Here's the PoC I mentioned: https://github.com/PatrickRen/flink-cdc-connectors/commit/2189da35a32eaafaf4c72dacdd439b588adb047c

We are planning to get this PR done in 2.4.0. As we are approaching the code freeze of version 2.4.0, we can cooperate more tightly on this issue. Feel free to take my commit directly into your PR if you think the patch looks good to you. It'll be nice that we can fix the issue within this week. Sorry for rushing you!

wallkop:
> @wallkop Thanks for your reply! Here's the PoC I mentioned: [PatrickRen@2189da3](https://github.com/PatrickRen/flink-cdc-connectors/commit/2189da35a32eaafaf4c72dacdd439b588adb047c)
> 
> We are planning to get this PR done in 2.4.0. As we are approaching the code freeze of version 2.4.0, we can cooperate more tightly on this issue. Feel free to take my commit directly into your PR if you think the patch looks good to you. It'll be nice that we can fix the issue within this week. Sorry for rushing you!

Hi @PatrickRen, Thank you very much, your POC implementation is very elegant, I will try my best to improve it on the basis of it, the transformation work has almost been completed, and a complete test is in progress. Since I accidentally closed this PR just now, I will launch a new PR to fix `issue:1944` issue as soon as this Saturday night.

PR #2099, opened 2023-04-22
[core] Support print options for flink cdc sql connectors

### LINKED ISSUE #1742: [core] Format print connector config to log.
**Is your feature request related to a problem? Please describe.**
We need to print connector config to log just like debezium does (formatted and sensetive info hidden).

TyrantLucifer:
close #1742


TyrantLucifer:
Hi @GOODBOY008 , in this task I have changed the following things:

1. Add dependency `flink-cdc-base` into connectors modules in order to code reused.
2. In `connector-db2` module I try to add `flink-cdc-base` but this will effact the compile `flink-sql-connector-db2-cdc`, the error message as the following shown:

```log
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.1.1:shade (shade-flink) on project flink-sql-connector-db2-cdc: Error creating shaded jar: Problem shading JAR /home/tyrantlucifer/.m2/repository/com/zaxxer/HikariCP/4.0.3/HikariCP-4.0.3.jar entry META-INF/versions/11/module-info.class: java.lang.IllegalArgumentException -> [Help 1]
```

So I add a same class into module, for this problem do you have any good advice?

flink-connector-db2-cdc/src/main/java/com/ververica/cdc/connectors/db2/table/Db2TableSourceFactory.java
```diff
@@ -113,6 +115,18 @@ public DynamicTableSource createDynamicTableSource(Context context) {
         ResolvedSchema physicalSchema = context.getCatalogTable().getResolvedSchema();
         StartupOptions startupOptions = getStartupOptions(config);
 
+        OptionUtils.printOptions(
```
ruanhang1993, line 118: The `config` from `helper.getOptions()` is actually an instance of `Configuration`. We could cast it to `Configuration` then get a map.

ruanhang1993 (commented):
@TyrantLucifer Thanks for your contribution. I left some comments.

flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/utils/OptionUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+/** A utility class to print configuration of connectors. */
+public class OptionUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(OptionUtils.class);
+
+    /** Utility class can not be instantiated. */
+    private OptionUtils() {}
+
+    public static void printOptions(Map<String, String> config, ConfigOption<?>... options) {
```
ruanhang1993, line 35: This method should be like this :
```
public static void printOptions(String identifier, Map<String, String> config) {
        Map<String, String> hideMap = ConfigurationUtils.hideSensitiveValues(config);
        LOG.info("Print {} connector configuration:", identifier);
        for (String key : hideMap.keySet()) {
            LOG.info("{} = {}", key, hideMap.get(key));
        }
    }
```

flink-connector-db2-cdc/src/main/java/com/ververica/cdc/connectors/db2/utils/OptionUtils.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.flink.configuration.ReadableConfig;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.List;
+
+/** A utility class to print configuration of connectors. */
+public class OptionUtils {
```
ruanhang1993, line 29: Why not invoke the class in base module?

ruanhang1993 (commented):
Maybe each module invoke the base module is not a good idea. Serval connectors should not rely on the base module before they are moved to the new incremental snapshot framework.

I think we should introduce separate OptionUtils for them.

ruanhang1993 (approved):
LGTM.

PR #2104, opened 2023-04-25
[Oracle] Capture changes made by connector user & document that SYS/SYSTEM changes are not captured

e-mhui:
[Oracle] Capture changes made by connector user & document that SYS/SYSTEM changes are not captured


e-mhui:
It has been fixed in https://github.com/ververica/flink-cdc-connectors/pull/2156

Issue #2106, opened 2023-05-01
[Bug] NullPointerException when executing a rename collection operation in MongoDB


e-mhui:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.13.5

### Flink CDC version

2.3, master

### Database and its version

MongoDB 4.4

### Minimal reproduce step

 NullPointerException when executing a rename collection operation in MongoDB.

```
57388 [flink-akka.actor.default-dispatcher-11] INFO  org.apache.flink.runtime.executiongraph.ExecutionGraph - Source: table_1[1] -> Calc[2] -> ConstraintEnforcer[3] -> Sink: Collect table sink (1/1) (445d32c31dfc855a98170971e52a43c8_cbc357ccb763df2852fee8c4fc7d55f2_0_0) switched from RUNNING to FAILED on 980da0af-6746-4a39-a94e-cc529e2ce51d @ localhost (dataPort=-1).
java.lang.NullPointerException: null
	at org.apache.flink.util.Preconditions.checkNotNull(Preconditions.java:59) ~[flink-core-1.16.0.jar:1.16.0]
	at com.ververica.cdc.connectors.mongodb.table.MongoDBConnectorDeserializationSchema.deserialize(MongoDBConnectorDeserializationSchema.java:128) ~[classes/:?]
	at com.ververica.cdc.debezium.internal.DebeziumChangeFetcher.handleBatch(DebeziumChangeFetcher.java:236) ~[classes/:?]
	at com.ververica.cdc.debezium.internal.DebeziumChangeFetcher.runFetchLoop(DebeziumChangeFetcher.java:162) ~[classes/:?]
	at com.ververica.cdc.debezium.DebeziumSourceFunction.run(DebeziumSourceFunction.java:445) ~[classes/:?]
	at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:110) ~[flink-streaming-java-1.16.0.jar:1.16.0]
	at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:67) ~[flink-streaming-java-1.16.0.jar:1.16.0]
	at org.apache.flink.streaming.runtime.tasks.SourceStreamTask$LegacySourceFunctionThread.run(SourceStreamTask.java:333) ~[flink-streaming-java-1.16.0.jar:1.16.0]
```

### What did you expect to see?

The task is running normally.

[...]

PR #2107, opened 2023-05-01
[MongoDB] Fix NullPointerException when executing ddl operation in MongoDB

e-mhui:
[MongoDB] Fix NullPointerException when executing ddl operation in MongoDB

Fix: #2106 

When performing DDL operations in MongoDB, documentKey and fullDocument are null, so we need to check if it is a DML operation and retrieve documentKey and fullDocument from the record.

```java
    if (isDMLOperation(op)) {
        documentKey =
                checkNotNull(
                        extractBsonDocument(
                                value, valueSchema, MongoDBEnvelope.DOCUMENT_KEY_FIELD));
        fullDocument =
                extractBsonDocument(value, valueSchema, MongoDBEnvelope.FULL_DOCUMENT_FIELD);
    }
```


Jiabao-Sun (approved):
Thanks @e-mhui for report this problem.
I think delaying the empty check can also solve this problem.

PR #2142, opened 2023-05-17
[docs] fix incorrect version to flink-1.16.0

sywmomoda:
Fix an incorrect flink version in the documentation


ruanhang1993:
@sywmomoda Thanks.

We will update in version 2.4.0. Would you like to check in 2.4.0?

Close this for 2.3 now.

PR #2145, opened 2023-05-19
using tikv limited scan  to avoid oom

LiPL2017:
Currently, tikv loads all the records in a huge table into memory which will cause a memory overflow


yuxiqian:
Hi @LiPL2017, thanks for your great work! We're planning a general rate-limit framework in the following Flink 3.x releases, so closing this PR for now.

PR #1854, opened 2023-01-04
[oceanbase] add jdbc options and support oracle mode

whhe:
Use oceanbase official jdbc driver to support oracle mode and add jdbc properties, close #983, close #1777.

PR #2146, opened 2023-05-19
fix: timestamp incompatibility with OceanBase 4.x

whhe:
Add the logic to convert data from LocalDateTime to TimestampData, as the latest version of OceanBase may get LocalDateTime objects from a Timestamp or Datetime field by the JDBC driver.


whhe:
CI failed with `System.IO.IOException: No space left on device`, maybe we need to check the status of host machine. @leonardBang

whhe:
Impl in #1854

Issue #1731, opened 2022-11-10
[cdc-base] Close idle readers when snapshot phase ended.


Jiabao-Sun:
**Is your feature request related to a problem? Please describe.**
Before in Flink 1.14, checkpoint on finished source feature is not supported.
We can not close the idle readers when parallel snapshot finished.

**Describe the solution you'd like**
If the runtime Flink version is over 1.14, close idle readers when snapshot phase ended to reduce resource usage.

PR #2162, opened 2023-05-29
[cdc-base] Close idle readers when snapshot finished

Jiabao-Sun:
Close idle readers when snapshot finished #1731


flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/options/SourceOptions.java
```diff
@@ -111,4 +112,15 @@ public class SourceOptions {
                                     + " The table chunks would use evenly calculation optimization when the data distribution is even,"
                                     + " and the query for splitting would happen when it is uneven."
                                     + " The distribution factor could be calculated by (MAX(id) - MIN(id) + 1) / rowCount.");
+
+    @Experimental
+    public static final ConfigOption<Boolean> SCAN_INCREMENTAL_CLOSE_IDLE_READER_ENABLED =
+            ConfigOptions.key("scan.incremental.close-idle-reader.enabled")
+                    .booleanType()
+                    .defaultValue(false)
```
GOODBOY008, line 120: I think this option's default value should be `true`  , for the old flink version which don't support should be setted `false` by user.

GOODBOY008: @Jiabao-Sun As you said , flink disabled by default. So ignore this comment. :)

PR #2171, opened 2023-06-01
[tool] add versions-maven-plugin

gtk96:
Our new version is about to be released.I want to know if we are currently manually modified version numbers? If so, then this plug-in can effectively reduce our labor.


gaotingkai:
> Our new version is about to be released.I want to know if we are currently manually modified version numbers? If so, then this plug-in can effectively reduce our labor.

I found that maven has relevant commands. This plug-in only provides richer functions. 😥

Issue #2175, opened 2023-06-03
[Bug] old change data that will be captured when the latest mode starts


fuyun2024:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

flink 1.13.5

### Flink CDC version

master

### Database and its version

sqlserver2012

### Minimal reproduce step

1. Configure from ` LATEST_ OFFSET ` Start CDC task
2. Observe that the captured data contains old change data.

### What did you expect to see?

Change data from now on

### What did you see instead?

Old change data

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

GOODBOY008:
https://github.com/ververica/flink-cdc-connectors/pull/2176 merged.

PR #2176, opened 2023-06-03
[sqlserver] Fix old change data that will be captured when the latest…

fuyun2024:
Fix this pr : #2175


GOODBOY008 (approved):
LGTM , @fuyun2024 Thanks for your contributions.

PR #2189, opened 2023-06-07
[oceanbase] always cleanup connections before exit

whhe:
Reconnect to oblogproxy with the same client id while the previous connection is not closed will cause exception 'Duplication exist clientId' at oblogproxy server side. So the process should always cleanup the connections before exit the source function.

Related: https://github.com/oceanbase/oblogproxy/issues/46


flink-connector-oceanbase-cdc/src/main/java/com/ververica/cdc/connectors/oceanbase/source/OceanBaseRichSourceFunction.java
```diff
@@ -141,22 +141,25 @@ public void open(final Configuration config) throws Exception {
...
-
-        if (shouldReadSnapshot()) {
-            LOG.info("Snapshot reading started");
-            readSnapshotRecords();
-            LOG.info("Snapshot reading finished");
-        } else {
-            LOG.info("Snapshot reading skipped");
+            logProxyClient.join();
+        } finally {
+            cancel();
```
ruanhang1993, line 161: This method will stop `logProxyClient` too. Is it a right behavior?

ruanhang1993 (approved):
LGTM.

PR #2213, opened 2023-06-14
[tests][mysql] Shutdown mysql container timely for legacy mysql tests

leonardBang:
Shutdown mysql container timely for legacy mysql tests


PatrickRen (approved):
@leonardBang Thanks for the patch! LGTM.

PR #2219, opened 2023-06-16
[postgres] Create slot for backfill task before snapshot reading

leonardBang:


flink-connector-postgres-cdc/src/main/java/com/ververica/cdc/connectors/postgres/source/fetch/PostgresScanFetchTask.java
```diff
@@ -92,12 +94,15 @@ public void execute(Context context) throws Exception {
...
+                        (PostgresReplicationConnection) ctx.getReplicationConnection(),
                         ctx.getDbzConnectorConfig(),
                         ctx.getDatabaseSchema(),
                         ctx.getOffsetContext(),
                         ctx.getDispatcher(),
                         ctx.getSnapshotChangeEventSourceMetrics(),
-                        split);
+                        split,
+                        "a",
+                        "b");
```
ruanhang1993, line 105: The slotName and pluginName will always be 'a' and 'b'. Do we really need these parameters?

leonardBang: 😭, I forgot update here

PR #2246, opened 2023-06-28
Fix  the  PostgresSourceBuilder's class can't be init with datastream api

syyfffy:
if the build function is  private ,user can not init the PostgresSourceBuilder's class.


syyfffy:
with pgsql-cdc datasream api , user can not new the class
image

syyfffy:
Please help me review it when you are free @ruanhang1993 @leonardBang

PR #2229, opened 2023-06-21
[mongodb] supports pre- & post-image feature in mongo 6.0 (#2062)

yuxiqian:
This PR comes with the following changes:

* Updates `org.mongodb.kafka` and `mongodb-driver-sync` to fit MongoDB 6, and did minor fixes to handle breaking changes of dependencies.

* Adds a MongoDB CDC config `full.document.pre.post.image` to switch pre- and post-image retrieving feature on / off. Requires `scan.incremental.snapshot.enabled = true`.

* Generates normal `UPDATE BEFORE` and `DELETE` events when `full.document.pre.post.image = true`.

* Provides `UPDATE AFTER` events by post-image atomically when `full.document.pre.post.image = true`.

* Creates `MongoDBFullDocumentBeforeChangeITCase` to test its functionality.

> Declared a new dependency (`org.testcontainers:mongodb`) to implement `MongoDBContainer` that works with MongoDB 6. Original version was moved to `LegacyMongoDBContainer` and still used by `Legacy-` testcases.

PR #2253, opened 2023-06-30
Deprecate `copyExisting` option of legacy MongoDBSource

yuxiqian:
This PR requires #2229 being merged first.

Changes were made to:

* deprecate `copyExisting` config option of legacy `MongoDBSource` and suggest migrating to `startupOptions`
* added support of `TIMESTAMP` startup mode in legacy `MongoDBSource`


yuxiqian:
Closed, merged into #2229.

PR #2257, opened 2023-06-30
[Mysql]  Mysql connector can't handle the case sensitive of rename/change column statement (#2144)

### LINKED ISSUE #2144: [Bug] DBZ-5589 Mysql connector can't handle the case sensitive of rename/change column statement
### Search before asking

### Flink version

ALL

### Flink CDC version

ALL

### Database and its version

MySQL, ALL version

### Minimal reproduce step

1. create a table and write some data；
2. listen it‘s binlog Source:mysql , Sink:print；
3. alter a column case and update data.

when i modify a clumn case， the filed name in the binlog record should be changed.

### What did you see instead?

I changed a field name to lowercase, but its field name in binlog is still uppercase.

### Anything else?

its a bug of debezium. 
source code in 
debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/AlterTableParserListener.java:AlterTableParserListener.java
Debezium 2.0.0 fix it by modify line 187 and line 349.
Use equals insdead of equalsIgnoreCase method.
alter_clumn_test

- [x] I'm willing to submit a PR!

dua01:
DBZ-5589 Mysql connector can't handle the case sensitive of rename/change column statement 
The bug is only fixed in debezium2.0 version, and I have copied a copy of the solution.

### DIFF
--- a/flink-connector-debezium/src/main/java/io/debezium/relational/TableEditorImpl.java
+++ b/flink-connector-debezium/src/main/java/io/debezium/relational/TableEditorImpl.java
@@ -1,0 +1,275 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.relational;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


yuxiqian:
Hi @dua01, thanks for you contribution! Could you please rebase this PR with latest `master` branch before it could be merged? Also please describe this as a temporary workaround in comments and need to be removed after upgraded Debezium dependency.

PR #2267, opened 2023-07-06
[mongodb] remove duplicate records during snapshot failover

yuxiqian:
This PR requires #2229 being merged first.

PR #2229 allows MongoDB CDC generates full change-log data and eliminates the necessity of `ChangelogNormalize` node.

However, when:

* `parallelismSnapshot` is set to `false` (thus legacy `MongoDBSource` will be used)
* `scanFullChangelog` is set to `true` (thus `ChangelogNormalize` node will not exist)
* crash occurred during initial snapshot (record will be emitted more than once)

As a result, downstream may receive duplicate records.

To fix the problem, `DebeziumDeduplicateSourceFunction` was created to store snapshotted document ID in Flink's state and filter out duplicate records. State will be cleared after snapshot phase ends.

Notice that `DebeziumSourceFunction` class was slightly modified for extension. More review and discussion is necessary since it's widely involved.


yuxiqian:
Done, rebased, marked ready for review. Could @Jiabao-Sun @leonardBang help checking this?

yuxiqian:
To avoid emitting duplicate snapshot records to downstream, I stored **all snapshotted document ID** in Source State, which might be an issue.

Since a) MongoDB's `find()` keeps order by inserting time, and b) legacy source has no parallel race issue, is it possible to store **last snapshotted document ID** only, as a delimiter to determine last snapshot progress?

cc @Jiabao-Sun

yuxiqian:
Since this PR is not handling database updates during failover correctly, and this issue affects legacy source only which might be deprecated in the future, this issue won't be fixed for now.

PR #2304, opened 2023-07-17
[Oracle]修复oracle数据库表主键是NUMBER类型且不指定p,s长度时报校验异常:org.apache.flink.table.api.Val…

nidonglin:
修复oracle数据库表主键是NUMBER类型且不指定p,s长度时报校验异常:org.apache.flink.table.api.ValidationException: Decimal precision must be between 1 and 38 (both inclusive).


yuxiqian:
Could you please also rebase this? Thanks!

Issue #2302, opened 2023-07-14
[Bug]  Tried to use mongodb connector but throwing an error


rubensacoustic:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.15

### Flink CDC version

2.4.0

### Database and its version

Mongo Atlas  6.0.6

### Minimal reproduce step

This is the code I'm trying to run, is taken from example.
```
      MongoDBSource<String> mongoSource =
              MongoDBSource.<String>builder()
                      .connectionOptions("retryWrites=true&w=majority&ssl=true")
                      .scheme("mongodb+srv")
                      .hosts("xxxxxx")
                      .databaseList("dbtest") // set captured database, support regex
                      .collectionList("dbtest.config_test") //set captured collections, support regex
                      .username("xxx")
                      .password("xxx")
                      .deserializer(new JsonDebeziumDeserializationSchema())
                      .build();

      StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
      // enable checkpoint
      env.enableCheckpointing(3000);
      // set the source parallelism to 2
      env.fromSource(mongoSource, WatermarkStrategy.noWatermarks(), "MongoDBIncrementalSource")
              .setParallelism(2)
              .print()
              .setParallelism(1);

      env.execute("Print MongoDB Snapshot + Change Stream");

```

### What did you expect to see?

The records flowing throught the pipeline

### What did you see instead?

An exception like this:

com.mongodb.MongoQueryException: Query failed with error code 8000 and error message 'noTimeout cursors are disallowed in this atlas tier' on server dev-ue1-tealeaf-md-shard-00-02.9zymn.mongodb.net:27017

### Anything else?

I see that `noTimeout` cursors are used here:

https://github.com/ververica/flink-cdc-connectors/blob/62b32713a472f7db24bd40ebed9a7e9c04bb4529/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/reader/fetch/MongoDBScanFetchTask.java#L116

I don't understand well the code base I think this is for  snapshots, Is there a way to avoid using `noTimeout` cursors? As it seems like mongo atlas doesn't support it on my tier.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

yuxiqian:
[...]

PR #2332, opened 2023-07-21
[mongodb] Add `disableCursorTimeout` option, addresses #2302

yuxiqian:
This PR fixes issue #2302.

Currently, MongoDB CDC uses `noCursorTimeout` option when performing incremental snapshot. However, some MongoDB server doesn't support creating cursor with unlimited timeout.

To address this issue, an option `disableCursorTimeout` was added in `mongodb.source.MongoDBSource` (but not its legacy counterpart `mongodb.MongoDBSource` since it isn't affected).


yuxiqian:
Haven't added it into Table API since I noticed that DataSource API seems to have more "hidden" (undocumented) options than Table API. Plus, this option only takes into effect if `scan.incremental.snapshot` is set to `true`.

yuxiqian:
Could @Jiabao-Sun help reviewing it?

Jiabao-Sun:
Thanks @yuxiqian for this enhancement.
Shall we add it as a table option as well?

yuxiqian:
Added. Picked name `scan.cursor.no-timeout` from Flink MongoDB Connector (non-CDC version).

Jiabao-Sun (approved):
Thanks @yuxiqian for the updates. LGTM.

PR #2334, opened 2023-07-21
[mongodb] [tests] Loosen E2E timeout limitations

yuxiqian:
Currently MongoDB E2E test has rather tight limitations. Container startup time limit is 30 seconds, and each E2E test are expected to complete data capturing in 60 seconds. When Azure pipeline is busy, unexpected timeout exception might occur.

This PR loosens container startup timeout to 120 seconds (closer to TiDB, PolarDB, and OceanBase) and capturing timeout to 150 seconds (like DB2).


leonardBang (approved):

PR #2338, opened 2023-07-22
[jdbc] support read snapshot only

zhaomin1423:
Firstly, support oracle, postgres and sqlserver


yuxiqian:
Thanks for your contribution! Closing this since it has been addressed in #2901.

Issue #2350, opened 2023-07-27
[Bug][Oracle] Fix option to not include.schema.changes


rbreejen:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.16.1

### Flink CDC version

2.4.1

### Database and its version

Oracle 19 & 21

### Minimal reproduce step

I was trying to circumvent the option to include schema changes (which results in the error #2315) in the Oracle CDC connector by setting include.schema.changes to false. Example:

```
JdbcIncrementalSource<String> oracleChangeEventSource =
	new OracleSourceBuilder()
  	...                   
	    .includeSchemaChanges(true)
```
to
```
JdbcIncrementalSource<String> oracleChangeEventSource =
	new OracleSourceBuilder()
  	...                   
	    .includeSchemaChanges(false)
```

### What did you expect to see?

Not the bug related to #2315 because it shouldn't produce schema changes at all.

### What did you see instead?

I did see the bug mentioned by #2315 

### Anything else?

The reason for why it didn't work is because the option to not include schema changes is not propogated to the Debezium engine (see fix). The default option in Debezium Engine is to [include](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/relational/RelationalDatabaseConnectorConfig.java#L446) schema changes.

I propose a fix similar to the [Postgres implementation](https://github.com/ververica/flink-cdc-connectors/blob/master/flink-connector-postgres-cdc/src/main/java/com/ververica/cdc/connectors/postgres/source/config/PostgresSourceConfigFactory.java#L83) & the [default SourceFunction behaviour](https://github.com/ververica/flink-cdc-connectors/blob/master/flink-connector-debezium/src/main/java/com/ververica/cdc/debezium/DebeziumSourceFunction.java#L378).

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

rbreejen:
See MR #2351

GOODBOY008:
@rbreejen You should refer oracle connector page to find out, how to pass configurations to debezium.

PR #2351, opened 2023-07-27
[Oracle] Fix option to not include.schema.changes

rbreejen:
With this fix it is possible to not include schema changes in the Oracle CDC connector, as mentioned in issue #2350.


GOODBOY008:
@rbreejen You should refer oracle connector page to find out, how to pass configurations to debezium.

PR #2353, opened 2023-08-01
Remove meaningless code

liugddx:


flink-cdc-base/src/test/java/com/ververica/cdc/connectors/base/MySqlChangeEventSourceExampleTest.java
```diff
@@ -119,8 +119,7 @@ public void testConsumingScanEvents() throws Exception {
                         WatermarkStrategy.noWatermarks(),
                         "MySqlParallelSource")
                 .setParallelism(4)
-                .print()
-                .setParallelism(1);
```
leonardBang, line 123: It aims to  set parallelism of PrintSink operator to 1 , so it should be meaningful.

PR #2357, opened 2023-08-02
Remove `SOURCE_IDLE_TIME` and `CURRENT_EMIT_EVENT_TIME_LAG` metrics

yuxiqian:
[Flink operators](https://github.com/apache/flink/blob/fc3035b40fbdf32694eb90c78c4dd68608c1c9d1/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/groups/InternalSourceReaderMetricGroup.java#L62-L63) have registered these metrics. Connectors don't have to report them correspondingly. This PR removes these metrics from CDC connectors.

Note: This feature was added in Flink 1.14+ ([FLINK-23652](https://issues.apache.org/jira/browse/FLINK-23652)), so with 1.13 these metrics will be missing.


yuxiqian:
Requested @ruanhang1993's review.

PR #2362, opened 2023-08-02
[flink-connector-mysql-cdc] fix snapshot read thread exit due to erro…

TJX2014:
Currently when OutOfMemoryError or other Error not Exception caused in fetcher thread, outside will not know it but the fetcher thread is exit; So this patch change the Exception catch to Throwable which will catch OutOfMemoryError and something else.


TJX2014:
cc @ruanhang1993 @leonardBang

leonardBang (approved):
Thanks @TJX2014 for the contribution, LGTM

Issue #2381, opened 2023-08-10
[Bug] Keyword conflict


Malcolmjian:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.16.0

### Flink CDC version

2.3.0

### Database and its version

mysql 5.7

### Minimal reproduce step

CREATE TABLE `feed_topic_links` (
  `index` int(10) unsigned NOT NULL AUTO_INCREMENT ,
  `topic_id` int(10) unsigned NOT NULL COMMENT 'Topic ID',
  `feed_id` int(10) unsigned NOT NULL COMMENT 'Feed ID',
  PRIMARY KEY (`index`),
  KEY `feed_topic_links_topic_id_index` (`topic_id`),
  KEY `feed_topic_links_feed_id_index` (`feed_id`)
) ENGINE=InnoDB AUTO_INCREMENT=17028 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;

![image](https://github.com/ververica/flink-cdc-connectors/assets/19792174/9cdb83d5-b444-43a2-855d-f35e7418143b)

When the primary key of a table is a MySQL keyword, it will cause an SQL exception. My solution is to add backticks (`) around the keyword. Please help me translate this into English and optimize the statement.

When the primary key of a table coincides with a MySQL keyword, it triggers an SQL exception. To resolve this issue, simply place backticks (`) around the keyword.

### What did you expect to see?
```
select * from xxx whre `index` >=100 and not(`index` = 100)
```
### What did you see instead?

![image](https://github.com/ververica/flink-cdc-connectors/assets/19792174/9cdb83d5-b444-43a2-855d-f35e7418143b)

### Are you willing to submit a PR?

I'm willing to submit a PR!

GOODBOY008:
Add ` around `index` field name. @Malcolmjian

PR #2387, opened 2023-08-14
fixbug mysql keyword conflict issues(#2381)

Malcolmjian:


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/StatementUtils.java
```diff
@@ -321,4 +321,10 @@ private static String buildSelectWithBoundaryRowLimits(
     private static String quotedTableIdString(TableId tableId) {
         return tableId.toQuotedString('`');
     }
+
+    private static String quotedFieldString(String field) {
```
ruanhang1993, line 325: This method is duplicated as `StatementUtils#quote`. We could use `quote` directly.

ruanhang1993 (commented):
@Malcolmjian Thanks for the PR. A little comment is left.

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/utils/StatementUtils.java
Malcolmjian: Okay, let me reprocess this.

PR #2389, opened 2023-08-14
[postgres]  PostgresSql snapshot sql 

linzhicong1996:
the column will be case-sensitive,so pk should added double quotation.And my postgresql version 13.2


yuxiqian:
Thanks @linzhicong1996 for the great work! Before this PR could be merged, could you please rebase it with latest `master` branch?

PR #2399, opened 2023-08-16
[hotfix][mysql] Add finished unack splits to state for the MysqlSourceReader

ruanhang1993:


leonardBang (approved):
Thanks @ruanhang1993 for the fix, LGTM

PR #2416, opened 2023-08-21
[sqlserver] Add support bigint type as table split column

GOODBOY008:
Add support bigint type as table split column


GOODBOY008:
@ruanhang1993 PTAL

Jiabao-Sun (approved):
Thanks @GOODBOY008.
LGTM.

PR #2432, opened 2023-08-28
[postgres] Reuse pool connection based on database in postgres cdc

loserwang1024:
Fix https://github.com/ververica/flink-cdc-connectors/issues/2428


loserwang1024:
@TyrantLucifer , CC

loserwang1024:
@ruanhang1993 @whhe , CC, WDYT?

whhe (approved):
LGTM

cobolbaby:
With the connection pool enabled, which phase of the data synchronization will experience optimization in terms of connection count? Is it during the full synchronization phase? Has there been an estimation of the optimization effect?

PR #2437, opened 2023-08-29
[cdc-base] refactor param of JdbcConnectionPoolFactory

loserwang1024:
This is not a bug, but an improving change for now JdbcConnectionPoolFactory.

Current JdbcConnectionPoolFactory  produce new datasources based on initial params JdbcSourceConfig, rather than JdbcConfiguration in real JDBC, which occur will problems.
For example: add a new test in MySqlChangeEventSourceExampleTest:
```java
    @Test
    public void testJdbcConnectionTest() throws SQLException {

        JdbcConfiguration jdbcConfiguration =
                JdbcConfiguration.create()
                        .withHostname("testhost1")
                        .withPort(8080)
                        .withDatabase("test-database")
                        .withUser("test-user")
                        .withPassword("test-passwotd")
                        .build();

        // if this two configurations is not same, what will happens
        MySqlSourceConfig mySqlSourceConfig =
                new MySqlSourceConfig(
                        StartupOptions.latest(),
                        Lists.newArrayList(inventoryDatabase.getDatabaseName()),
                        Lists.newArrayList(inventoryDatabase.getDatabaseName() + ".products"),
                        1,
                        1,
                        1.00,
                        2.00,
                        false,
                        false,
                        new Properties(),
                        jdbcConfiguration,
                        "com.mysql.cj.jdbc.Driver",
                        MYSQL_CONTAINER.getHost(),
                        MYSQL_CONTAINER.getDatabasePort(),
                        inventoryDatabase.getUsername(),
                        inventoryDatabase.getPassword(),
                        100,
                        "UTC+8",
                        Duration.ofSeconds(5),
                        1,
                        10);

        JdbcConnection jdbc =
                new JdbcConnection(
                        jdbcConfiguration,
                        new JdbcConnectionFactory(
                                mySqlSourceConfig, new MysqlPooledDataSourceFactory()),
                        "`",
                        "`");

        jdbc.connect();
        Assert.assertTrue(jdbc.isConnected());
        Assert.assertFalse(jdbc.username().equals(inventoryDatabase.getDatabaseName()));
        Assert.assertFalse(jdbc.config().getHostname().equals(inventoryDatabase.getHost()));
    }
```
It turns out that Jdbc 's username and hostname are different from real connection.

## Brief Change Log
JdbcSourceConfig contains too much informations for JdbcConnectionPoolFactory. So extract a new param DataSourcePoolConfig, which contains the informations needed for thread pool. Combine DataSourcePoolConfig and JDBC's JdbcConfiguration to produce a new datasource.

### DIFF
--- a/flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/config/DataSourcePoolConfig.java
+++ b/flink-cdc-base/src/main/java/com/ververica/cdc/connectors/base/config/DataSourcePoolConfig.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2023 Ververica Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ververica.cdc.connectors.base.config;
+
+import java.io.Serializable;
+import java.time.Duration;
+
+/** A config to create HikariDataSource. */
+public class DataSourcePoolConfig implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private final Duration connectTimeout;
+    private final int connectMaxRetries;
+    private final int connectionPoolSize;
+
+    private final String driverClassName;
+
+    private final String serverTimeZone;
+
+    public String getDriverClassName() {
+        return driverClassName;
+    }
+
+    public DataSourcePoolConfig(
+            Duration connectTimeout,
+            int connectMaxRetries,
+            int connectionPoolSize,
+            String driverClassName,
+            String serverTimeZone) {
+        this.connectTimeout = connectTimeout;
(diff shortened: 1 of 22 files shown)


whhe:
AFAIK the `JdbcSourceConfig` is a wrap of Debezium class, and both of them are generated from the same `Configuration` object of connector options. So I think it's appropriate to use `JdbcSourceConfig` in connection pool. 

In your implementation, we assume that users may provide different configurations in one task, which is very confusing, and the other options which are not included in `DataSourcePoolConfig` like connectTimeout still can be inconsistent.

loserwang1024:
Hi, @whhe , thanks for your advice. I know you mean that it's ok for the whole system.But as I showed in testJdbcConnectionTest , from the perspective of JdbcConnectionPoolFactory which may cause inconsistence between JDBCConnection and real connection.

> the JdbcSourceConfig is a wrap of Debezium class

JdbcSourceConfig  contains all the params of a source, which is more than a pool factory need. So I want to  minimize the params for pool factory.


> DataSourcePoolConfig like connectTimeout still can be inconsistent.

ConnectTimeout is bound to JdbcConfiguration. Different database's connection can use different timeout setting, which depends on user. I think more flexible。 

WDYT?

whhe:
You're right, our connection pool implementation should be able to create different kinds of data sources. 

I think this feature can be added in `JdbcConnectionPools`, and we should keep the `JdbcConnectionPoolFactory` to be bound to one database type. I noticed you caught a bug https://github.com/ververica/flink-cdc-connectors/issues/2449, may be you can have a try to fix it at the same time.

loserwang1024:
@whhe , I agree with you. I will close this pr and fix the problem in https://github.com/ververica/flink-cdc-connectors/issues/2449,

PR #2453, opened 2023-09-02
[mysql] Add notification after snapshot phase ends, before binlog stream starts with GTIDs via Slack 

SML0127:
First, let me say thank you to everyone who contributes to Flink CDC.
Our team use Flink CDC to perform MySql CDC.

But since [there is no Snapshot Only mode for MySql yet](https://github.com/ververica/flink-cdc-connectors/issues/1687), we had a need to be notified when a snapshot is completed and when the binlog stream is started.

To accomplish this, we **implemented a notification when a snapshot ends and when a binlog stream starts with GTIDs.**

--- 

Here's the team's use case in more detail 
1. We set parallelism to 2 or more for large tables.
2. And we send change event log to kafka to use Debezium's JDBC Sink Connector, which supports [Schema Evolution](https://debezium.io/documentation/reference/stable/connectors/jdbc.html#jdbc-schema-evolution).
3. Sinking to Kafka is slower than MySqlSource operator, so we give more paralleisms to sink operator more parallelism than MySqlSource.
4. In this case, the transfer is done in rebalance mode from source operator to sink operator, so the order for the same PK is not guaranteed when transferring binlogs.
5. So we restart the job based on GTIDs with parallelism equal to 1 at the end of the snapshot phase .

To do this, we needed (1) to be notified that the snapshot ended and the binlog stream started, and (2) to know from which GTIDs the binlog stream started.

---

This is portion of our code. **We assumed that we only capture 1 table per flink cdc job**
```scala
  def getMySQLSourceOperator(): MySqlSource[String] = {
    MySqlSource.builder[String]()
      .hostname(mySqlConfig.host)
      .port(mySqlConfig.port)
      .serverTimeZone(mySqlConfig.timeZone)
      .databaseList(mySqlConfig.database)
      .tableList(mySqlConfig.table)
      .username(mySqlConfig.user)
      .serverId(mySqlConfig.serverIdRange)
      .password(mySqlConfig.password)
      .startupOptions(mySqlConfig.startupMode)
      .fetchSize(mySqlConfig.fetchSize)
      .splitSize(mySqlConfig.splitSize)
      .chunkKeyColumn(new ObjectPath(mySqlConfig.database, mySqlConfig.table), mySqlConfig.chunkKeyColumn)
      .connectionPoolSize(mySqlConfig.poolSize)
      .scanNewlyAddedTableEnabled(false)
      .includeSchemaChanges(false)
      .debeziumProperties(mySqlConfig.dbzProps)
      .closeIdleReaders(true)
      .notifySnapshotToBinlogSwitch("slack-hook-url") // here what we implemented
      .deserializer(new JsonDebeziumDeserializationSchema(true, mySqlConfig.jsonConverterProps))
      .build()
  }
```

---

We can't share a real picture of the notification, because our company recommends using in-house tools rather than Slack(🥲🥲🥲) and has some security policy..

But it looks something like the format below!
- Snapshot finished notifiaction.
```
[SNAPSHOT FINISHED]
 Database: test_database
 Table: test_table
```
- Binlog stream start notification.
```
[BINLOG STREAM START]
 Database: test_database
 Table: test_table
 GTIDs: 3bda59bb-2fc8-11eb-855f-fa163e2550e3:1-128377129,3c3a6a1b-c931-11ed-b0db-b4055dec129e:1-273703345,901d637c-8add-11eb-8e3f-b4055d3355a6:1-3641352422,b46b8251-5254-11ed-a648-d0946637df48:1-1069556331,db449f07-c53e-11e8-b8c2-d094663d3d1d:1-3543229125,e8d62f95-c77a-11e8-9270-d0946637df48:1-5715021533
```

---

If you are interested in this feature, please feel free to check it out.  Any comments or feedback would also be appreciated.
Have a great day all Flink users 😄

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
@@ -124,6 +124,8 @@ public class SnapshotSplitReader implements DebeziumReader<SourceRecords, MySqlS
                         statefulTaskContext.getTopicSelector(),
                         statefulTaskContext.getSnapshotReceiver(),
                         StatefulTaskContext.getClock(),
+                        statefulTaskContext.getSourceConfig().getWatchTowerId(),
+                        statefulTaskContext.getSourceConfig().getTableList().get(0),
                         currentSnapshotSplit);
         executorService.execute(
                 () -> {
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/debezium/reader/SnapshotSplitReader.java
@@ -124,7 +124,6 @@ public class SnapshotSplitReader implements DebeziumReader<SourceRecords, MySqlS
                         statefulTaskContext.getTopicSelector(),
                         statefulTaskContext.getSnapshotReceiver(),
                         StatefulTaskContext.getClock(),
-                        statefulTaskContext.getSourceConfig().getWatchTowerId(),
                         statefulTaskContext.getSourceConfig().getTableList().get(0),
                         currentSnapshotSplit);
         executorService.execute(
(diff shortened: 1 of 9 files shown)


ruanhang1993:
@SML0127 Thanks for contributing. 

I think the main feature you need is the way to send events to the other external system from the split enumerator.
This PR will only support this one situation which need send events to slack. 

If many users need this feature, I think the design will be better as follows.

1. Add new interface `ExternalSystemEvent` to define the sent events.
2. Add new interface `ExternalSystemListener` to listen the  `ExternalSystemEvent`s. And they will apply different operations when receiving different events.
3. Implement some events like `SnapshotPhaseFinishedEvent`. And send them to `ExternalSystemListener` in the split enumerator.
4. Add some table options for sql users to provide their `ExternalSystemListener`. For example:
   - `external-system.listener.class`: the external system listener
   - `external-system.*`: provide some other parameters for creating the listener

In this design, the users will be able to implement their own listener. And we could provide more kinds of event besides `SnapshotPhaseFinishedEvent`.

SML0127:
@ruanhang1993 
First of all, we are using Flink CDC really well, and thank you for making it.

We've been using this notification feature and found it to be more useful than we expected, so we created thisPR.

We will consider your comments and improve the code further.😄

qinxiang01:
请问下同步多张表的时候，能不能每张表都同步呢

qinxiang01:
请问下同步多张表的时候，能不能每张表都通知呢

PR #2464, opened 2023-09-06
[tests][e2e] add log properties file to e2e tests

loserwang1024:
There is no `log4j2-test.properties` file in e2e tests, so it's hard to find problem in test.Now add `log4j2-test.properties` flie to e2e tests.

### DIFF
--- a/flink-cdc-e2e-tests/src/test/resources/log4j2-test.properties
+++ b/flink-cdc-e2e-tests/src/test/resources/log4j2-test.properties
@@ -1,0 +1,26 @@
+################################################################################
+#  Copyright 2023 Ververica Inc.
+#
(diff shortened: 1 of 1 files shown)


loserwang1024:
@yuxiqian , CC

yuxiqian:
LGTM

Irrelevant to this PR, but I noticed that `log4j2-test.properties` in `flink-connector-db2-cdc` sets it target as `appender.testlogger.target = SYSTEM_OUT` while all other tests use `SYSTEM_ERR`. Should this be fixed, too?

loserwang1024:
@yuxiqian , it seems that two output just with different color, not a big problem.

yuxiqian:
OK, no problems for me.

whhe (approved):

Issue #1327, opened 2022-06-29
[Mysql cdc] MySqlSourceEnumerator - Received invalid request meta group id 5, the invalid meta group id range is [0, 4]


ancongxue:
**Describe the bug(Please use English)**
遇到报错：
MySqlSourceEnumerator - Received invalid request meta group id 5, the invalid meta group id range is [0, 4]
导致sourceReader不能正常消费binlog。

**Environment :**
 - Flink version : 1.14.4
 - Flink CDC version: 2.2.0
 - Database and version: mysql 5.7.28-31-57.2-log

**To Reproduce**
1. 任务的StartupMode为INITIAL，scanNewlyAddedTableEnabled为true，checkpoint开启，先正常消费表到binlog阶段，停任务；
2. 修改任务配置，新增加需要监听的表，从上次ck启动；
3. 新增表的snapshot阶段做完，往enumerator反馈完FinishedSnapshotSplitsReportEvent，此时又做完一个ck2；
4. 把任务停掉，从ck2启动，出现图下报错，导致source reader不能继续消费binlog。

**Additional Description**
MySqlSourceEnumerator在2022-06-23 14:25:39挂掉后，之后的每一次故障恢复都会报这个错
![image](https://user-images.githubusercontent.com/26142121/176357946-46d5f02d-6a17-463a-99e1-14d313f29d08.png)

MySqlSourceEnumerator报错
<img width="1283" alt="image" src="https://user-images.githubusercontent.com/26142121/176358073-ebbeab69-f943-4af8-a61f-7999cd663d9a.png">

'Mysql Source account_0'这个MysqlSourceReader启动从ck恢复，没能正常继续消费binlog文件
<img width="1285" alt="image" src="https://user-images.githubusercontent.com/26142121/176358346-5b86e5b8-57e6-4fd4-be50-3c2f90093a77.png">

qg-lin:
Environment :

Flink version : 1.16.0
Flink CDC version: 2.4.1
Database and version: 5.7.36-txsql-log

logs:

2023-09-08 10:24:00,838 INFO  com.ververica.cdc.connectors.mysql.source.enumerator.MySqlSourceEnumerator [] - The enumerator assigns split MySqlSnapshotSplit{tableId=new-vehicle.activity_push_record, splitId='DB.TABLE:0', splitKeyType=[`id` BIGINT NOT NULL], splitStart=null, splitEnd=null, highWatermark=null} to subtask 0
[...]

PR #2482, opened 2023-09-12
[mysql] Fix newly added table invalid request meta group id error(#1327)

qg-lin:
try to fix https://github.com/ververica/flink-cdc-connectors/issues/2421

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
@@ -153,6 +153,8 @@ public class MySqlSnapshotSplitAssigner implements MySqlSplitAssigner {
         this.currentParallelism = currentParallelism;
         this.alreadyProcessedTables = alreadyProcessedTables;
         this.remainingSplits = new CopyOnWriteArrayList<>(remainingSplits);
+        // When job restore from savepoint, sort the existing tables and newly added tables
+        // to let enumerator only send newly added tables's BinlogSplitMetaEvent
         this.assignedSplits =
                 assignedSplits.entrySet().stream()
                         .sorted(Comparator.comparing(Map.Entry::getKey))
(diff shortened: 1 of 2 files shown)


flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
```diff
@@ -152,7 +153,15 @@ private MySqlSnapshotSplitAssigner(
...
-        this.assignedSplits = assignedSplits;
+        this.assignedSplits =
+                assignedSplits.entrySet().stream()
+                        .sorted(Comparator.comparing(Map.Entry::getKey))
+                        .collect(
+                                Collectors.toMap(
+                                        Map.Entry::getKey,
+                                        Map.Entry::getValue,
+                                        (o, o2) -> o,
+                                        LinkedHashMap::new));
```
ruanhang1993, line 164: I do not think we should sort `assignedSplits` here. `assignedSplits` will be changed at other methods and we cannot keep the order.
I think the problem is that the finished infos in `BinlogSplit` are not sorted. Maybe it should be sorted in `getFinishedSnapshotSplitInfos`.

ruanhang1993: I review again and understand the changes. Please ignore this comment.

qg-lin: When job restore from savepoint, if sort in getFinishedSnapshotSplitInfos by splitId and newly added tables are before exist tables that will casue newly added table(eg: eixst table list:[t1, t2], newly add table list:[a1, a2]) can't receive binlog data because of MySqlSourceEnumerator#sendBinlogMeta get wrong split partition by requestMetaGroupId

leonardBang (approved):
Thanks @qg-lin for the contribution, LGTM

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
loserwang1024: @qg-lin, @ruanhang1993 , Hi, I understand why not sorted  in `getFinishedSnapshotSplitInfos` by splitId. But whether  need to sort `assignedSplits`? If not sort, any problem will occur?

PR #2488, opened 2023-09-13
[hotfix] Mongo CDC fails to capture collections with `.` in names

yuxiqian:
For now, [MongoDB allows](https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Restriction-on-Collection-Names) collection names containing dots (`.`). According to [Mongo CDC docs](https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mongodb-cdc.html#connector-options), it should be ok to match them with fully-qualified regex like this:

```javascript
db[.]coll[.]name[.]with[.]dots // matches collection "coll.name.with.dots" in "db" database
```

However, it doesn't work when incremental snapshot option is enabled. Here's the minimum POC:

```java
@Test
public void testMatchCollectionWithDots() throws Exception {
    // 1. Given colllections:
    // db: [coll.name]
    String db = CONTAINER.executeCommandFileInSeparateDatabase("ns-dotted");

    TableResult result = submitTestCase(db, db + "[.]coll[.]name");

    // 2. Wait change stream records come
    waitForSinkSize("mongodb_sink", 3);

    // 3. Check results
    String[] expected =
            new String[] {
                String.format("+I[%s, coll.name, A101]", db),
                String.format("+I[%s, coll.name, A102]", db),
                String.format("+I[%s, coll.name, A103]", db)
            };

    List<String> actual = TestValuesTableFactory.getResults("mongodb_sink");
    assertThat(actual, containsInAnyOrder(expected));

    result.getJobClient().get().cancel().get();
}
```

It might be caused by a glitch in [Debezium framework](https://github.com/debezium/debezium/blob/5e243e2480f1bd2525c7de27672f98fe7649d01b/debezium-core/src/main/java/io/debezium/relational/TableId.java#L104C5-L118C6) which simply regards `.` as a separator between catalog, schema, and table (instead of a valid character in table name).

Currently, Mongo CDC never requests schema field (`useCatalogBeforeSchema` is always true), so a weaker version of `splitTableId` function was implemented, and hopefully it could correctly handle dots presenting in collection names.

### DIFF
--- a/flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/tests/MongoE2eITCase.java
+++ b/flink-cdc-e2e-tests/src/test/java/com/ververica/cdc/connectors/tests/MongoE2eITCase.java
@@ -89,7 +89,7 @@ public class MongoE2eITCase extends FlinkContainerTestEnvironment {
         super.before();
 
         container =
-                new MongoDBContainer("mongo:6.0.6")
+                new MongoDBContainer("mongo:6.0.9")
                         .withSharding()
                         .withNetwork(NETWORK)
                         .withNetworkAliases(INTER_CONTAINER_MONGO_ALIAS)
--- a/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/dialect/MongoDBDialect.java
+++ b/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/dialect/MongoDBDialect.java
@@ -70,11 +70,29 @@ public class MongoDBDialect implements DataSourceDialect<MongoDBSourceConfig> {
         return "MongoDB";
     }
 
+    private static TableId parseTableId(String str) {
+        return parseTableId(str, true);
+    }
+
+    private static TableId parseTableId(String str, boolean useCatalogBeforeSchema) {
+        String[] parts = str.split("[.]", 2);
+        int numParts = parts.length;
+        if (numParts == 1) {
+            return new TableId(null, null, parts[0]);
+        } else if (numParts == 2) {
+            return useCatalogBeforeSchema
+                    ? new TableId(parts[0], null, parts[1])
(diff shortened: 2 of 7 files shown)


yuxiqian:
PTAL @Jiabao-Sun @leonardBang

Jiabao-Sun:
Thanks @yuxiqian for this great work.

The root cause of this error is that the collection configuration option contains the '.' character, which is a reserved character in regular expressions. As a result, the collection configuration option is mistakenly inferred as a regular expression. 
We have encountered an issue where the '-' character cannot be matched before and we have resolved this problem by implementing some special conditional logic.

The reason for distinguishing between regular and non-regular database and collection configuration options is because we have pushed the filtering of databases and collections down to the ChangeStream level. This allows us to save the additional query overhead of fullDocumentLookup. When specifying explicit databases and collections, the filtering overhead on ChangeStream is minimal. However, if we were to treat all database and collection configuration options as regular expressions, it would increase the computational cost and response time of MongoDB.

Unfortunately, we cannot accurately determine whether a string is a regular expression or a non-regular expression because all strings can be considered as regular expressions. This has also led to these issues where accurate matching cannot be achieved. 

Additionally, in order to handle whether a configuration option is a regular expression or a non-regular expression, a lot of complex conditional logic has been implemented in the code. I believe it would be beneficial to explicitly distinguish between configuration options that belong to regular expressions and those that are explicit database and collection configurations. For example, we could introduce separate options such as `databaseRegex` and `collectionRegex`.

cc @leonardBang

yuxiqian:
> The root cause of this error is that the collection configuration option contains the '.' character, which is a reserved character in regular expressions

Actually regex matching here works as expected (`db[.]coll[.]name` correctly matches `coll.name` in `db`). The problem is when parallelism snapshot is enabled, Debezium can't correctly parse TableId from names like `db.coll.name` due to unexpected `.` in table names.

+1 for explicit regex options, though it's irrelevant to this bug report.

PR #2503, opened 2023-09-18
fix: mongo collection regex match logic

yuxiqian:
Commit https://github.com/ververica/flink-cdc-connectors/commit/44cd46e227a5e772a6f740b3a8664b8e8afd2c77 introduces a change that using [.] as delimiter between database name and collection name to avoid conflicts between dot (`.`) and regex meta character. However this turns out to be unnecessary since in Regex mode, full qualified names will always be used and no pasting will be performed then, and such change will deteriorate performance.

This PR reverts the change above.


yuxiqian:
cc @Jiabao-Sun @leonardBang @ruanhang1993

Jiabao-Sun (approved):
Thanks @yuxiqian for this contribution.
LGTM.

PR #2508, opened 2023-09-20
[feature] Add remove db docker image after test and revert e2e-2

GOODBOY008:


whhe:
I'm wondering if it would be better to keep e2e splitted into several pipelines, which can reduce overall execution time.

GOODBOY008:
> I'm wondering if it would be better to keep e2e splitted into several pipelines, which can reduce overall execution time.

Add new pipeline just want to solove disk with no space case e2e failed.

whhe (approved):

PR #2523, opened 2023-09-26
[WIP][mysql] Add ExternalSystemListener to notify user about the status of connector.

lvyanquan:
this closes https://github.com/ververica/flink-cdc-connectors/issues/2520

Try to support notification about the status of connector, since it may be a common demand from community.
However, I met some problem when add a preliminary implementation, want to get some advice from yours.
1. When I try to define an enum of ExternalSystemEvent, I found that it's similar to [AssignerStatus](https://github.com/ververica/flink-cdc-connectors/blob/6ca87c4d57a790c0e51247b275f82c03d8a8f9f8/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/assigners/AssignerStatus.java#L117), can we reuse it?
2. Should we send information about finished splits to listener? then it would be very similar to checkpoint.
3. Should we add a default implement like KafkaListener?

### DIFF
--- a/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/listener/AbstractListenerMessage.java
+++ b/flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/listener/AbstractListenerMessage.java
@@ -16,13 +16,16 @@
 
 package com.ververica.cdc.connectors.mysql.source.listener;
 
-/** The information that will be sent with the notification. */
-public class ListenerMessageInformation {
-    private final String startupMode;
-    private final String gtids;
+import java.util.Properties;
 
-    public ListenerMessageInformation(String startupMode, String gtids) {
-        this.startupMode = startupMode;
-        this.gtids = gtids;
+/** The information that will be sent with the notification. */
+public abstract class AbstractListenerMessage {
+
+    public Properties getProperties() {
+        return properties;
     }
+
+    Properties properties = new Properties();
+
+    abstract String getName();
 }
--- a/docs/content/connectors/mysql-cdc.md
+++ b/docs/content/connectors/mysql-cdc.md
@@ -329,6 +329,20 @@ During a snapshot operation, the connector will query each included table to pro
       <td>Boolean</td>
       <td>Whether to close idle readers at the end of the snapshot phase. The flink version is required to be greater than or equal to 1.14 when 'execution.checkpointing.checkpoints-after-tasks-finish.enabled' is set to true.</td>
     </tr>
+    <tr>
+      <td>external-system.listener.class</td>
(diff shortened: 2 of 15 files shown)


lvyanquan:
Hi, @ruanhang1993 
Not sure if my design and confusion are reasonable, Looking forward to receiving your opinion.

SML0127:
@lvyanquan
Wow..
First of all, Thank you for implementing this function I was hoping for.

I've been so busy at work and was just about to implement it, but you've already implemented it! 🤣🤣🤣

Can I run this pr with our company in-house notification tools and give you some feedback??🙇🏻‍♂️🙇🏻‍♂️

SML0127:
> @lvyanquan Wow.. First of all, Thank you for implementing this function I was hoping for.
> 
> I've been so busy at work and was just about to implement it, but you've already implemented it! 🤣🤣🤣
> 
> Can I run this pr with our company in-house notification tools and give you some comment??🙇🏻‍♂️🙇🏻‍♂️

I added some information (e.g. `startupMode`, `gtids`) for notification message and created @[PR](https://github.com/lvyanquan/flink-cdc-connectors/pull/2).

As our team implemented and used this notification feature, we realized that the `startupMode` and `gtids` information was really useful. So I added it.

I hope you give it a try, and thanks for any comments!! 😄🙇🏻‍♂️

lvyanquan:
It makes sense to me, and I create an abstract class [AbstractListenerMessage](https://github.com/ververica/flink-cdc-connectors/pull/2523/files#diff-698c27eee7807422a2eee230a778860d120669f04191d4dc28432cf0614190f4) since there are other datasources like oracle、postgreSQL.

SML0127:
> It makes sense to me, and I create an abstract class [AbstractListenerMessage](https://github.com/ververica/flink-cdc-connectors/pull/2523/files#diff-698c27eee7807422a2eee230a778860d120669f04191d4dc28432cf0614190f4) since there are other datasources like oracle、postgreSQL.

LGTM!
I learn a lot from your implementation😃

flink-connector-mysql-cdc/src/main/java/com/ververica/cdc/connectors/mysql/source/listener/AbstractListenerMessage.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Properties;
+
+/** The information that will be sent with the notification. */
+public abstract class AbstractListenerMessage {
+
+    public Properties getProperties() {
+        return properties;
+    }
+
+    Properties properties = new Properties();
```
whhe, line 28: This field 'properties' is for?

yuxiqian:
Is there any progress on this PR? cc @lvyanquan

PR #2530, opened 2023-09-27
[Improve]Support customer debezium properties in DebeziumSourceFunction

JNSimba:
Currently, in DebeziumSourceFunction, some properties are hardcoded, such as `include.schema.changes`, so that even opening this property peripherally will not work. 
Therefore, the modification is that if the key exists in the `properties`, it will not be overwritten.


gtk96:
cc @GOODBOY008 @ruanhang1993

ruanhang1993 (approved):
LGTM.

PR #2542, opened 2023-10-09
[mongodb] Disable hashed index for sharded split strategy

Jiabao-Sun:
The value of the Hash indexing boundary is the hash code of the shard keys, so it is difficult to determine whether the change record belongs to the chunk range.

Therefore, we do not support hashed index in the sharded split strategy.


Jiabao-Sun:
Hi @ruanhang1993, could you have review this when you have time?

PR #2557, opened 2023-10-17
[mysql] Fix a potential npe because "currentSplittingTableId" is nullable.

FlechazoW:
Using "currentSplittingTableId.equals" may cause a npe, because currentSplittingTableId is nullable.


ruanhang1993:
@FlechazoW Thanks for contributing.
The method `hasNextChunk` has already checked the `currentSplittingTableId` is not `null`.

FlechazoW:
Ok. I will close this pr.

Issue #2556, opened 2023-10-17
[Bug] table column name is "rowid", oceanbase-cdc read null


zane-rl:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

flink V1.15.3

### Flink CDC version

flink-connector-oceanbase-cdc  V2.4.1

### Database and its version

oceanbase-ce:4.1.0.0

### Minimal reproduce step

1、create table products and insert datas:
```
CREATE TABLE products (
  rowid INTEGER NOT NULL AUTO_INCREMENT PRIMARY KEY,
  name VARCHAR(255) NOT NULL,
  description VARCHAR(512)
);
ALTER TABLE products AUTO_INCREMENT = 101;
INSERT INTO products
VALUES (default,"scooter","Small 2-wheel scooter"),
       (default,"car battery","12V car battery"),
       (default,"12-pack drill bits","12-pack of drill bits with sizes ranging from #40 to #3"),
       (default,"hammer","12oz carpenter's hammer"),
       (default,"hammer","14oz carpenter's hammer"),
       (default,"hammer","16oz carpenter's hammer"),
       (default,"rocks","box of assorted rocks"),
       (default,"jacket","water resistent black wind breaker"),
       (default,"spare tire","24 inch spare tire");
```

2、create java file:
```
public class OceanBaseCdcConnector {

    public static void main(String[] args) {

        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
        env.enableCheckpointing(1000, CheckpointingMode.EXACTLY_ONCE);
        //env.getConfig()
        //env.getCheckpointConfig().setCheckpointStorage("file:///d:/checkpoint");
        env.getCheckpointConfig().setCheckpointStorage("file:///Users/boleyn/checkpoint");

        StreamTableEnvironment tenv = StreamTableEnvironment.create(env);
        //tenv.getConfig().getConfiguration().setString("table.exec.sink.not-null-enforcer", "DROP");

        tenv.executeSql("CREATE TABLE products (\n" +
                "    rowid INT,\n" +
                "    name STRING,\n" +
                "    description STRING,\n" +
                "    PRIMARY KEY (rowid) NOT ENFORCED\n" +
                "  ) WITH (\n" +
                "    'connector' = 'oceanbase-cdc',\n" +
                "    'scan.startup.mode' = 'initial',\n" +
[...]

PR #2560, opened 2023-10-17
[oceanbase] support case-insensitive column name

whhe:
Try to bypass  issue #2556.

### DIFF
--- a/docs/content/connectors/oceanbase-cdc(ZH).md
+++ b/docs/content/connectors/oceanbase-cdc(ZH).md
@@ -316,7 +316,7 @@ OceanBase CDC 连接器包括用于 SQL 和 DataStream API 的选项，如下表
                 <td>否</td>
                 <td style="word-wrap: break-word;">true</td>
                 <td>Boolean</td>
-                <td>列名是否大小写敏感，默认为 true，当置为 false 时，将会把列名转为小写使用。</td>
+                <td>列名是否大小写敏感，默认为 true。</td>
             </tr>
         </tbody>
     </table>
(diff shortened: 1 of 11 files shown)


docs/content/connectors/oceanbase-cdc(ZH).md
```diff
@@ -311,6 +311,13 @@ OceanBase CDC 连接器包括用于 SQL 和 DataStream API 的选项，如下表
                 <td>String</td>
                 <td>传递自定义 JDBC URL 属性的选项。用户可以传递自定义属性，如 'jdbc.properties.useSSL' = 'false'。</td>
             </tr>
+            <tr>
+                <td>column.case-sensitive</td>
+                <td>否</td>
+                <td style="word-wrap: break-word;">true</td>
+                <td>String</td>
```
lvyanquan, line 318: should be `Boolean` type?

lvyanquan (approved):

zane-rl:
if set column.case-sensitive=false config, the column name cotains capital letter also read null, for example a column name is 'productTitle'
@whhe

whhe:
> if set column.case-sensitive=false config, the column name cotains capital letter also read null, for example a column name is 'productTitle' @whhe

You are right, so when it is set to be false, user shoud use lowercase fileds in flink sql.

whhe:
> if set column.case-sensitive=false config, the column name cotains capital letter also read null, for example a column name is 'productTitle' @whhe

I thought about it again and realized that the insensitive fields in flink sql should also be supported at the same time. The latest commit is made for it.

leonardBang:
We may need consider more case-sensitive with Flink SQL‘s case-sensitive

PR #2564, opened 2023-10-19
[github] Add labeler for pull-request

lvyanquan:
Add an action to auto label pr with connector type based on the files it changed for user searching.
For example:

截屏2023-10-19 上午11 55 48

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,74 @@
+#
+# Copyright 2023 Ververica Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Pull Request Labeler Github Action Configuration: https://github.com/marketplace/actions/labeler
+
+build:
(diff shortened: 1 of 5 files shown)


FlechazoW:
I think it would be better to use lowercase.

lvyanquan:
address it.

FlechazoW (approved):
Nice job!

gtk96 (approved):
+1

.github/labeler.yml
```diff
@@ -0,0 +1,63 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
```
leonardBang, line 2: Could you add copyright info when you add header?  and you can also add header for other `yml` files

.github/labeler.yml
```diff
@@ -0,0 +1,63 @@
...
+test:
+  - flink-cdc-e2e-tests/**/*
+  - flink-connector-test-util/**/*
+db2:
+  - flink-connector-db2-cdc/**/*
+  - flink-sql-connector-db2-cdc/**/*
+mongodb:
+  - flink-connector-mongodb-cdc/**/*
+  - flink-sql-connector-mongodb-cdc/**/*
+mysql:
```
leonardBang, line 43: Could you refer the latest module structure? I think we should have `mysql-cdc-connector` and `mysql-pipeline-connector`

leonardBang (commented):
Thanks @lvyanquan for the contribution, sorry for the late review. I left some comments.

.github/labeler.yml
lvyanquan: mysql-pipeline-connector is not added yet, do we need to add label for this and other connectors in advance?

leonardBang: Good catch, we can add them later

leonardBang (approved):
Thanks @lvyanquan for the update, LGTM

PR #2565, opened 2023-10-19
[oceanbase] Use global timestamp to start log client

whhe:
This pr changed the startup order of snapshot and change events reading, and it will reduce the risk of change events buffer OOM.


leonardBang (approved):
Thanks @whhe for the contribution, LGTM, feel free to merge

xinwen-cs:
In oceanbase-ce:3.1.4, SELECT TS_VALUE FROM oceanbase.V$OB_TIMESTAMP_SERVICE return microsecond rather than nanosecond

In oceanbase-ce:4.0.0.0, oceanbase.V$OB_TIMESTAMP_SERVICE or SYS.V$OB_TIMESTAMP_SERVICE does not exist

Is this code only test on enterprise edition?

xinwen-cs:
@whhe

whhe:
> In oceanbase-ce:3.1.4, SELECT TS_VALUE FROM oceanbase.V$OB_TIMESTAMP_SERVICE return microsecond rather than nanosecond
> 
> In oceanbase-ce:4.0.0.0, oceanbase.V$OB_TIMESTAMP_SERVICE or SYS.V$OB_TIMESTAMP_SERVICE does not exist
> 
> Is this code only test on enterprise edition?

Thanks for your reply. The target version is 4.2.1, and you are right that 3.x and 4.x use different precision in clock value. I will add a fallback process for it as soon as possible.

PR #2578, opened 2023-10-23
MysqlTableSource return metadata in a fixed order

### LINKED ISSUE #2576: Make the mysql readable metadata get a fixed order
### Search before asking

### Motivation

Now `listReadableMetadata` in `MysqlTableSource` return a `HashMap`. So the returned metadata is orderless. Its order will be different between jdk 8 and jdk 11.

We should return a `LinkedHashMap` to use a fixed order.

### Solution

_No response_

### Alternatives

_No response_

### Anything else?

_No response_

- [X] I'm willing to submit a PR!

ruanhang1993:
This fixes #2576


leonardBang (approved):

PR #2595, opened 2023-11-01
Use SampleBucketSplitStrategy for shard collection with hashed keys #2504

### LINKED ISSUE #2504: [Bug] [mongodb] pull big collection (>20 billion records) lead to checkpoint hang or OOM
### Search before asking

### Flink version

1.17.0

### Flink CDC version

2.4.1

### Database and its version

Using MongoDB:          4.2.11

### Minimal reproduce step

1. prepare a mongodb collection with more than 10 billion records
2. pull records from mongodb, sync to paimon(parquet format on hdfs) or map record to integer 1 and set  up a minute window to count. 
3.  when the application runs for some time (30 minutes ~ a few hours)，it hangs or run into OOM exception and can recover.

the application runs without error and pull all records from mongodb.

### What did you see instead?

the application runs for server minutes and fail, only 5% less records are pulled from mongodb.

### Anything else?

this is the metrics without sink, only count pulled records.

cdc_topo

cdc_checkpoint

cdc_metric1

cdc_metric2

vanliu-tx:
Use SampleBucketSplitStrategy for shard collection which contain hashed key, this would avoid OOM exception if some chunk contains huge records which could not be hold in a local HashMap.

Close #2504

### DIFF
--- a/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/assigners/splitters/SampleBucketSplitStrategy.java
+++ b/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/assigners/splitters/SampleBucketSplitStrategy.java
@@ -26,6 +26,8 @@ import io.debezium.relational.history.TableChanges;
 import org.bson.BsonDocument;
 import org.bson.BsonValue;
 import org.bson.conversions.Bson;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -66,9 +68,13 @@ import static com.ververica.cdc.connectors.mongodb.source.utils.MongoUtils.colle
 @Internal
 public class SampleBucketSplitStrategy implements SplitStrategy {
 
+    private static final Logger LOG = LoggerFactory.getLogger(SampleBucketSplitStrategy.class);
+
     public static final SampleBucketSplitStrategy INSTANCE = new SampleBucketSplitStrategy();
     private static final int DEFAULT_SAMPLING_THRESHOLD = 102400;
     private static final double DEFAULT_SAMPLING_RATE = 0.05;
+    private static final int SAMPLING_COUNT_BY_RATE_THRESHOLD = 1000000;
+    private static final int MAX_SAMPLE_PER_CHUNK = 20;
 
     private SampleBucketSplitStrategy() {}
 
@@ -93,6 +99,10 @@ public class SampleBucketSplitStrategy implements SplitStrategy {
         } else {
             // sampled using sample rate.
             numberOfSamples = (int) Math.floor(count * DEFAULT_SAMPLING_RATE);
+            // avoid sampling too much records on a huge collection
+            if (numberOfSamples > SAMPLING_COUNT_BY_RATE_THRESHOLD) {
+                numberOfSamples = numChunks * MAX_SAMPLE_PER_CHUNK;
+            }
         }
 
         TableId collectionId = splitContext.getCollectionId();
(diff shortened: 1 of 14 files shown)


flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/assigners/splitters/SampleBucketSplitStrategy.java
```diff
@@ -66,9 +68,13 @@
 @Internal
 public class SampleBucketSplitStrategy implements SplitStrategy {
 
+    private static final Logger LOG = LoggerFactory.getLogger(SampleBucketSplitStrategy.class);
+
     public static final SampleBucketSplitStrategy INSTANCE = new SampleBucketSplitStrategy();
     private static final int DEFAULT_SAMPLING_THRESHOLD = 102400;
     private static final double DEFAULT_SAMPLING_RATE = 0.05;
+    private static final int SAMPLING_COUNT_BY_RATE_THRESHOLD = 1000000;
+    private static final int MAX_SAMPLE_PER_CHUNK = 20;
```
Jiabao-Sun, line 77: Thanks @vanliu-tx for the contribution.
I think we can add a config option `scan.incremental.snapshot.chunk.samples`: The number of samples to take per chunk. 

Refers to https://github.com/apache/flink-connector-mongodb/blob/main/flink-connector-mongodb/src/main/java/org/apache/flink/connector/mongodb/source/enumerator/splitter/MongoSampleSplitter.java.

vanliu-tx: Config option added. Please help to review ;-)

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/assigners/splitters/SampleBucketSplitStrategy.java
```diff
@@ -91,8 +94,8 @@ public Collection<SnapshotSplit> split(SplitContext splitContext) {
             // full sampling if document count less than sampling size threshold.
             numberOfSamples = (int) count;
         } else {
-            // sampled using sample rate.
-            numberOfSamples = (int) Math.floor(count * DEFAULT_SAMPLING_RATE);
+            // avoid sampling too much records on a huge collection
```
Jiabao-Sun, line 97:
```suggestion
```

flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/assigners/splitters/SampleBucketSplitStrategy.java
```diff
@@ -105,9 +108,18 @@ public Collection<SnapshotSplit> split(SplitContext splitContext) {
             pipeline.add(sample(numberOfSamples));
         }
```
Jiabao-Sun, line 109:
```java
        if (count < DEFAULT_SAMPLING_THRESHOLD) {
            // full sampling if document count less than sampling size threshold.
            numberOfSamples = (int) count;
        } else {
            // sampled using sample rate.
            numberOfSamples = (int) Math.floor(count * DEFAULT_SAMPLING_RATE);
        }
```
As the old code always `numberOfSamples <= count`.
If the user configuration is open,  there's the possibility of `numberOfSamples > count` .

So, I think `if (numberOfSamples < count)` is safe here.

Jiabao-Sun (commented):
Thanks @vanliu-tx for the great work.
I left some comments.
By the way, can we add the options test in `MongoDBTableFactoryTest`?
Should the `samplesPerChunk` method be added to `MongoDBSourceBuilder` as well?

vanliu-tx:
> Thanks @vanliu-tx for the great work. I left some comments. By the way, can we add the options test in `MongoDBTableFactoryTest`? Should the `samplesPerChunk` method be added to `MongoDBSourceBuilder` as well?

ok.

Jiabao-Sun (approved):
Thanks @vanliu-tx, LGTM

edmond-kk:
Is this judgment correct?
`            if (!(isValidShardedCollection(collectionMetadata)
                    && isNotShardedByHash(collectionMetadata))) {
                LOG.warn(
                        "Collection {} does not appear to be sharded, or shared by hash, fallback to SampleSplitter.",
                        collectionId);
                return SampleBucketSplitStrategy.INSTANCE.split(splitContext);
            }`
if collection is shard by range,the code will not enter this branch.

PR #2599, opened 2023-11-02
[hotfix] [docs] Fix typo 

YuanHanzhong:


README.md
```diff
@@ -130,7 +130,7 @@ CDC Connectors for Apache Flink<sup>®</sup> welcomes anyone that wants to help
 
 ### Code Contribute
 
-1. Left comment under the issue that you want to take
+1. Leave a comment under the issue that you want to take
```
gtk96, line 133: left a comment ,In fact, both ways of writing are correct.

YuanHanzhong: Yes, actually it does not affect user get the meaning. What about others?

PR #2623, opened 2023-11-03
[pom] Use ${revision} replace specific version.

joyCurry30:
As we use specific version like "3.0-SNAPSHOT", we need to repeat modifications between different modules. So why not use a placeholder like ${revision}.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@ under the License.
 
     <groupId>com.ververica</groupId>
     <artifactId>flink-cdc-connectors</artifactId>
-    <version>3.0-SNAPSHOT</version>
+    <version>${revision}</version>
     <packaging>pom</packaging>
 
     <modules>
@@ -60,6 +60,7 @@ under the License.
     </distributionManagement>
 
     <properties>
+        <revision>3.0-SNAPSHOT</revision>
         <java.version>1.8</java.version>
         <scala.binary.version>2.12</scala.binary.version>
         <maven.compiler.source>${java.version}</maven.compiler.source>
@@ -84,6 +85,7 @@ under the License.
         <commons-lang3.version>3.12.0</commons-lang3.version>
         <json-path.version>2.7.0</json-path.version>
         <markBundledAsOptional>true</markBundledAsOptional>
+        <flatten-maven-plugin.version>1.5.0</flatten-maven-plugin.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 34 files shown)


whhe (commented):
If 'revision' property is used, we should also add the `flatten-maven-plugin`, refer to https://maven.apache.org/maven-ci-friendly.html#install-deploy

joyCurry30:
@whhe I add the flatten-maven-plugin. cc plz.

pom.xml
```diff
@@ -465,6 +467,33 @@ under the License.
                     <doclint>none</doclint>
                 </configuration>
             </plugin>
+
+            <!-- make sure that flatten runs after shaded -->
```
lvyanquan, line 471: LGTM. minor suggestion: add comment to warn that this plugin takes effect for Maven version upper than 3.5.0.

lvyanquan (approved):

pom.xml
whhe: +1

joyCurry30: Done.

pom.xml
```diff
@@ -465,6 +467,34 @@ under the License.
...
+            <!-- This plugin takes effect for Maven version upper more than 3.5.0 -->
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>flatten-maven-plugin</artifactId>
+                <version>${flatten-maven-plugin.version}</version>
+                <configuration>
+                    <updatePomFile>true</updatePomFile>
```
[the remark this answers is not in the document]
joyCurry30: Ok. I will add it.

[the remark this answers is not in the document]
joyCurry30: Ok. I will add it.

whhe (approved):
LGTM, wait for ci to pass.

leonardBang:
Thanks @joyCurry30 and @whhe for the contribution and review, I've one minor question: What's the meaning `revision`? could we use maven default properties `${project.version}` in this case ?

whhe:
> Thanks @joyCurry30 and @whhe for the contribution and review, I've one minor question: What's the meaning `revision`? could we use maven default properties `${project.version}` in this case ?

As we need to declare `${project.version}` explicitly in each pom file (the parent part), we need a new property to set a original value for it, and `revision` is a common option to be used here.

leonardBang:
Thanks @whhe for the explanation, got it.

Issue #2622, opened 2023-11-03
[flink-cdc-dist] Collect all distro resources into a TAR


PatrickRen:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Motivation

Use maven-assembly-plugin to collect all resources into a TAR file, including:

- bin/flink-cdc.sh
- lib/flink-cdc-${version}.jar
- conf/flink-cdc.yaml

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PatrickRen:
Implemented by #2636

PR #2636, opened 2023-11-06
[3.0] [flink-cdc-dist] Collect all distro resources into a TAR

### LINKED ISSUE #2621: [flink-cdc-dist] Build a uber JAR for all flink-cdc modules
### Search before asking

### Motivation

Use maven-shade-plugin to create a uber JAR that contains all modules

### Solution

_No response_

### Alternatives

_No response_

### Anything else?

_No response_

GOODBOY008:
Close #2621 #2622


flink-cdc-dist/pom.xml
```diff
@@ -62,4 +61,154 @@ under the License.
...
+                        <executions>
+                            <execution>
+                                <id>dist</id>
+                                <goals>
+                                    <goal>run</goal>
+                                </goals>
+                                <phase>package</phase>
+                                <configuration>
+                                    <target>
+                                        <copy file="${project.build.directory}/flink-cdc-connectors-${project.version}-bin.tar.gz" tofile="${project.basedir}/../dist/flink-cdc-connectors-${project.version}-bin.tar.gz" overwrite="true" />
```
PatrickRen, line 202: Maybe we could skip the step of creating `dist` under the project root, and just leave the tar under `flink-cdc-dist/target`. The `dist` contains just one tarball, and we have to maintain a lot of things for it, like `.gitignore`, and one missing issue in your PR is that we need to remove it during `clean` phase.

flink-cdc-dist/src/main/assembly/conf/.gitkeep
PatrickRen, line 1: Is there any specific reason to keep `conf`, `lib` and `log` in the source code?

flink-cdc-dist/src/main/assembly/assembly.xml
```diff
@@ -0,0 +1,62 @@
...
+    <formats>
+        <format>tar.gz</format>
+    </formats>
+    <includeBaseDirectory>true</includeBaseDirectory>
+    <baseDirectory>flink-cdc-connectors-${version}-bin</baseDirectory>
+
+    <dependencySets>
+        <dependencySet>
+            <unpack>false</unpack>
+            <scope>runtime</scope>
```
PatrickRen, line 26: I found some suspicious JARs in the final tarball:
- Awaitility: This should be a test dependency, and should not be a part of the distribution. We need to change its scope to `test` in pom. 
- Hamcrest: This is defined as test scope in out project, but brought into the project as `compile` by Awaitility. I guess it will disappear once we change Awaitility to test scope.
- flink-cdc-dist: This is a JAR without any class file inside. We can just skip it when assembling. 
- commons-cli: What about packaging it directly into `flink-cdc-cli`?

Things missing:
- Log4j JAR and config file: Users cannot have logging locally without them

tools/ci/compile.sh
```diff
@@ -40,7 +40,7 @@ echo "==========================================================================
 
 EXIT_CODE=0
 
-run_mvn clean package -Dmaven.javadoc.skip=true -U -DskipTests | tee $MVN_CLEAN_COMPILE_OUT
+run_mvn clean install -Dmaven.javadoc.skip=true -U -DskipTests | tee $MVN_CLEAN_COMPILE_OUT
```
PatrickRen, line 43: What's the purpose of this one?

PatrickRen (commented):
@GOODBOY008 Thanks for the PR! I left some comments.

flink-cdc-dist/pom.xml
```diff
@@ -62,4 +61,154 @@ under the License.
...
+                            </execution>
+                        </executions>
+                        <configuration>
+                            <artifactItems>
+
+                                <artifactItem>
+                                    <groupId>com.ververica</groupId>
+                                    <artifactId>flink-sql-connector-mysql-cdc</artifactId>
+                                    <version>${project.version}</version>
+                                    <destFileName>mysql-cdc-connector.jar</destFileName>
```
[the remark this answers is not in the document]
leonardBang: +1, our connector jar is fat jar which will grow quickly in next development, we can offer some utils to help user to pull connector jar in the future.

flink-cdc-dist/src/main/assembly/assembly.xml
PatrickRen: About log4j I made a second thought, that maybe it's better to use log4j JARs under FLINK_HOME directly. We just need to provide a log4j configuration file for printing Flink CDC CLI and composer logs to a local file. Sorry for that!

PatrickRen:
I pushed 4 commits to the PR just now:

1. Re-arrange pom sttructure
- flink-cdc-dist should depends on all Flink CDC modules to enforce the building order
- Package a uber JAR with all Flink CDC classes and dependencies
- Remove the profile to make packaging flink-cdc-dist as the default behavior

2. Move `cdc-bin` to `flink-cdc-bin`

3. Improve assembly files
- Keep a directory under target for easier validation
- Add logic for copying flink-cdc-dist uber JAR to lib

4. Provide a log4j configuration for CLI running

@GOODBOY008 Could you take a look at the PR? I checked the packaged tarball and it should work as expected now.

GOODBOY008:
@PatrickRen I tested on my local, works as we want. Thank you for complete my rest work and address issue. 👍

leonardBang (approved):
Thanks @GOODBOY008 and @PatrickRen for the great work, I went through the files, LGTM

PR #2641, opened 2023-11-06
[3.0][cdc-common] Introduce Schema API for CDC 3.0

### LINKED ISSUE #2628: [flink-cdc-common] Introduce Schema to represent schema of table or data collection
### Search before asking

### Motivation

Introduce Schema to represent schema of table or data collection

### Solution

_No response_

### Alternatives

_No response_

### Anything else?

_No response_

leonardBang:
This PR aims to resolve #2628

### DIFF
--- a/flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
+++ b/flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
@@ -31,11 +31,9 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
@@ -47,7 +45,7 @@ public class Schema implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
-    private final List<Column> columns;
+    private final Map<String, Column> columns;
 
     private final List<String> primaryKeys;
 
@@ -55,10 +53,8 @@ public class Schema implements Serializable {
 
     private final @Nullable String comment;
 
-    private transient volatile Map<String, Column> nameToColumns;
-
     private Schema(
-            List<Column> columns,
+            Map<String, Column> columns,
             List<String> primaryKeys,
             Map<String, String> options,
             @Nullable String comment) {
@@ -75,19 +71,19 @@ public class Schema implements Serializable {
 
     /** Returns all {@link Column}s of this schema. */
     public List<Column> getColumns() {
-        return columns;
+        return new ArrayList<>(columns.values());
     }
 
     /** Returns all column names. It does not distinguish between different kinds of columns. */
     public List<String> getColumnNames() {
-        return columns.stream().map(Column::getName).collect(Collectors.toList());
+        return new ArrayList<>(columns.keySet());
     }
 
     /**
      * Returns all column data types. It does not distinguish between different kinds of columns.
      */
     public List<DataType> getColumnDataTypes() {
-        return columns.stream().map(Column::getType).collect(Collectors.toList());
+        return columns.values().stream().map(Column::getType).collect(Collectors.toList());
     }
 
     /** Returns the primary keys of the table or data collection. */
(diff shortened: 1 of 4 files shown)


flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * A builder for constructing an immutable but still unresolved {@link
+     * org.apache.flink.table.api.Schema}.
+     */
+    @PublicEvolving
+    public static final class Builder {
+
+        private final List<Column> columns;
+        private final Set<String> columnNames;
```
Jiabao-Sun, line 165: How about use `private final Map<String, Column>` here?
Then we can change the `Schema`'s constructor as  `Schema(Map<String, Column> columns, ...)`

leonardBang: It's okay to introduce the map in Builder, but keeping `Map<String, Column>` in `Schema` is a little wired it's implementation need instead of API need. Btw, `Column` contains its `name` field which may be error-prone (maybe different with the key in the map if we only maintain all columns of `schema`)

flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
```diff
@@ -0,0 +1,314 @@
...
+@PublicEvolving
+public class Schema implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private final List<Column> columns;
+
+    private final List<String> primaryKeys;
+
+    private final Map<String, String> options;
```
ruanhang1993, line 54: What will be stored in the `options`? It is strange that there is a `options` in `Schema`.

GOODBOY008: I think this option for store like `Charset`  name ?

flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
```diff
@@ -0,0 +1,314 @@
...
+    // -----------------------------------------------------------------------------------
+    private void initializeNameToColumns() {
+        if (nameToColumns == null) {
+            // make the method thread-safe
+            synchronized (this) {
+                // we need to check nullability again here
+                if (nameToColumns == null) {
+                    nameToColumns = new HashMap<>();
+                    for (Column col : columns) {
+                        nameToColumns.put(col.getName(), col);
```
GOODBOY008, line 139: It's fine to use `col.getName().toLowerCase()`.

flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
GOODBOY008: > It's okay to introduce the map in Builder, but keeping `Map<String, Column>` in `Schema` is a little wired it's implementation need instead of API need. Btw, `Column` contains its `name` field which may be error-prone (maybe different with the key in the map if we only maintain all columns of `schema`)

+1 , Build `Map<String, Column>` inner will be better.

flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
leonardBang: Each table may have its own chunkKey, its own chunkSize

flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
leonardBang: The column name is case-sensitive by default in flink, so I do not tend to break this, @zhongqiang

Jiabao-Sun (approved):

GOODBOY008 (approved):

leonardBang:
Thanks @GOODBOY008 @Jiabao-Sun for the review work, merging...

flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /** Returns the number of columns of this schema. */
+    public int getColumnCount() {
+        return columns.size();
+    }
+
+    /** Returns all {@link Column}s of this schema. */
+    public List<Column> getColumns() {
+        return new ArrayList<>(columns.values());
```
lvyanquan, line 74: Hi @leonardBang, this method may return a different order with the order column was added, should we add a list to maintain the order? it may lead to inconsistent order when creating sink table.

leonardBang: I will open a PR to address this one

PR #2649, opened 2023-11-07
[3.0][cdc-runtime] add DataSinkWriterOperator to process Event

lvyanquan:
close https://github.com/ververica/flink-cdc-connectors/issues/2605

points to be discussed:
- [x] should we add a new interface [SupportSchemaEvolutionWriting](https://github.com/ververica/flink-cdc-connectors/pull/2649/files#diff-68f4e50b509259e3895e1566ba6b5a4fecff0ecb92ab91dd2adf087ade556d33), or let writer deal with it in write method？
- Unnecessary.
- [x] is there a better way to [access SinkWriter](https://github.com/ververica/flink-cdc-connectors/pull/2649/files#diff-a203655b5f9025aaba19f82f8ed58e6e67a658a45f01a3f0de581e108899a7b7R73) rather than reflection, as sinkWriter is `private` scope？
- Reflection is acceptable.
- [x] [SCHEMA_EVOLUTION_OPERATOR_ID](https://github.com/ververica/flink-cdc-connectors/pull/2649/files#diff-fcc6da8c2ff9a2d029d78af4d61216271294b3edaf9452ef3f3fd4398cd046faR49) is an identification to get SchemaOperatorCoordinator, does it meet the expectations of SchemaOperator？
- OperatorID of SchemaOperator is created by Composer.

@PatrickRen @ruanhang1993 cc and Looking forward to your suggestions.


PatrickRen:
@lvyanquan Thanks for pull requests!

1. I don't quite get the usage of the interface `SupportSchemaEvolutionWriting`. Which component should this interface be applied to? 
2. Using reflections is acceptable to me. I couldn't find better solution for now. Otherwise we have to copy-paste and rewrite the entire `SinkWriterOperator` from Flink.
3. I'm not sure if it is possible to build the operator ID in composing phase, instead of hard-coding in schema operator coordinator. It looks like the composer's responsibility to generate a ID and use it to construct the schema operator and the sink operator.

flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/schema/event/SinkWriterRegisterEvent.java
```diff
@@ -23,4 +23,14 @@
 /** A {@link OperatorEvent} that register sink writer to {@link SchemaOperatorCoordinator}. */
 public class SinkWriterRegisterEvent implements OperatorEvent {
     private static final long serialVersionUID = 1L;
+
+    private final int taskId;
```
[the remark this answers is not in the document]
lvyanquan: All SinkWriter will be registered during setting up, is TableId necessary?

lvyanquan:
@PatrickRen 

> 1. I don't quite get the usage of the interface `SupportSchemaEvolutionWriting`. Which component should this interface be applied to?   

SupportSchemaEvolutionWriting is designed for SinkWriter, just to separate the processing logic of SchemaChangeEvent and 
DataChangeEvent. But it's somewhat redundant, and was deleted now.

> 3. I'm not sure if it is possible to build the operator ID in composing phase, instead of hard-coding in schema operator coordinator. It looks like the composer's responsibility to generate a ID and use it to construct the schema operator and the sink operator.   

Keep this code currently until another implementation is available

ruanhang1993 (approved):
LGTM.

flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/sink/SchemaEvolutionClient.java
```diff
@@ -0,0 +1,62 @@
...
+import com.ververica.cdc.runtime.operators.schema.event.FlushSuccessEvent;
+import com.ververica.cdc.runtime.operators.schema.event.SinkWriterRegisterEvent;
+
+import java.io.IOException;
+
+/**
+ * Client for {@link DataSinkWriterOperator} interact with {@link SchemaOperatorCoordinator} when
+ * table schema evolution happened.
+ */
+@PublicEvolving
```
PatrickRen, line 37: Is there any specific reason to expose `SchemaEvolutionClient` as public API? If we expose this one, we need to reconsider the constructor of it because `TaskOperatorEventGateway` might be confusing to developers.

flink-cdc-runtime/src/main/java/org/apache/flink/streaming/runtime/operators/sink/DataSinkWriterOperator.java
```diff
@@ -0,0 +1,98 @@
...
+                        schemaOperatorID);
+    }
+
+    @Override
+    public void open() throws Exception {
+        super.open();
+        Class<? extends SinkWriterOperator> cls = this.getClass();
+        Field field = cls.getDeclaredField("sinkWriter");
+        field.setAccessible(true);
+        copySinkWriter = (SinkWriter) field.get(this);
```
PatrickRen, line 78: What about wrapping a helper function here?

lvyanquan: Address it.

flink-cdc-runtime/src/main/java/org/apache/flink/streaming/runtime/operators/sink/DataSinkWriterOperator.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * <p>The operator is always part of a sink pipeline and is the first operator.
+ *
+ * @param <CommT> the type of the committable (to send to downstream operators)
+ */
+public class DataSinkWriterOperator<CommT> extends SinkWriterOperator<Event, CommT> {
+
+    SchemaEvolutionClient schemaEvolutionClient;
+
+    private SinkWriter copySinkWriter;
```
PatrickRen, line 50: `SinkWriter` -> `SinkWriter<Event>`

flink-cdc-runtime/src/main/java/org/apache/flink/streaming/runtime/operators/sink/DataSinkWriterOperator.java
```diff
@@ -0,0 +1,116 @@
...
+ * same parallelism or send them downstream to a {@link CommitterOperator} with a different
+ * parallelism.
+ *
+ * <p>The operator is always part of a sink pipeline and is the first operator.
+ *
+ * @param <CommT> the type of the committable (to send to downstream operators)
+ */
+public class DataSinkWriterOperator<CommT> extends SinkWriterOperator<Event, CommT> {
+
+    SchemaEvolutionClient schemaEvolutionClient;
```
PatrickRen, line 48: The visibility should be `private` I guess

PatrickRen (commented):
Two tiny comments for the operator

flink-cdc-runtime/src/main/java/org/apache/flink/streaming/runtime/operators/sink/DataSinkWriterOperator.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private SinkWriter copySinkWriter;
+
+    private final OperatorID schemaOperatorID;
+
+    DataSinkWriterOperator(
```
PatrickRen, line 54: What about using `public` for the constructor? This operator is located under package `org.apache.flink.xxx`, but callers of the constructor like composer are located under `com.ververica.cdc.xxx`

PatrickRen (commented):
@lvyanquan Sorry for flushing your thread! Found another issue.

flink-cdc-runtime/src/main/java/org/apache/flink/streaming/runtime/operators/sink/DataSinkWriterOperator.java
lvyanquan: Done. 
BTW, do you know how to create ProcessingTimeService and MailboxExecutor correctly, I couldn't find a suitable method. Setting them to null would have no effect if these components were only used in [Sink.InitContext](https://github.com/apache/flink/blob/5056b0d5dfec25ef90cd0f0e860816ade90d3869/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/sink/SinkWriterOperator.java#L244), but I'm not sure what it would be used somewhere.

PatrickRen (approved):
@lvyanquan Thanks for the update! LGTM.

Issue #2608, opened 2023-11-02
[flink-cdc-composer] Core interfaces exposed to CLI frontend


PatrickRen:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Motivation

Core composer interfaces exposed to the client (CLI frontend)

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

Jiabao-Sun:
Closed by #2656

PR #2656, opened 2023-11-07
[flink-cdc-composer] Base interfaces and models for composer and pipeline definition

PatrickRen:
This pull request implements #2608, which introduces base interfaces exposed to upper-level CLI, including:

- `PipelineComposer`: interface for composer
- `PipelineDef`: Java object abstracting a pipeline definition
- `PipelineExecution`: Abstraction for an execution of the pipeline

And a Flink implementation (just a framework without solid composing logic) is made as a proof of concept.

### DIFF
--- a/flink-cdc-composer/src/main/java/com/ververica/cdc/composer/definition/PipelineDef.java
+++ b/flink-cdc-composer/src/main/java/com/ververica/cdc/composer/definition/PipelineDef.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2023 Ververica Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ververica.cdc.composer.definition;
+
+import org.apache.flink.configuration.Configuration;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Definition of a pipeline.
+ *
+ * <p>A pipeline consists of following components:
+ *
+ * <ul>
+ *   <li>Source: data source of the pipeline. Required in the definition.
+ *   <li>Sink: data destination of the pipeline. Required in the definition.
+ *   <li>Routes: routers specifying the connection between source tables and sink tables. Optional
+ *       in the definition.
+ *   <li>Transforms: transformations for applying modifications to data change events. Optional in
+ *       the definition.
+ *   <li>Config: configurations of the pipeline. Optional in the definition.
+ * </ul>
+ *
+ * <p>This class keeps track of the raw pipeline definition made by users via pipeline definition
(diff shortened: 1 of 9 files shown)


flink-cdc-composer/src/main/java/com/ververica/cdc/composer/flink/FlinkPipelineComposer.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    private FlinkPipelineComposer(StreamExecutionEnvironment env, boolean isBlocking) {
+        this.env = env;
+        this.isBlocking = isBlocking;
+    }
+
+    @Override
+    public PipelineExecution compose(PipelineDef pipelineDef) {
+        return new FlinkPipelineExecution(env, "CDC Job", isBlocking);
```
Jiabao-Sun, line 52: `pipelineDef` seems useless here.
Do we need a hook passing configurations to `StreamExecutionenvironment`?

PatrickRen: Currently `FlinkPipelineComposer` doesn't implement any solid functionality because it depends on common and runtime module, so there's no usage for `pipelineDef` now. The implementation of `compose` would be like:

```java
@Override
public PipelineExecution compose(PipelineDef pipelineDef) {
    // Translate routers and other operators ...
    Source flinkSource = translateToSource(pipelineDef.getSource());
    Sink flinkSink = translateToSink(pipelineDef.getSink())
    
    // Build Flink DataStream job
     env.fromSource(flinkSource)
        .addTransformation(router)
        .sinkTo(flinkSink);

    // Construct execution
    return new FlinkPipelineExecution(env, jobName, isBlocking);
}
```
Then the `pipelineDef` will be used.

I don't quite get the "passing configuration by hook" part. Could you elaborate more about it?

flink-cdc-composer/src/main/java/com/ververica/cdc/composer/definition/PipelineDef.java
```diff
@@ -0,0 +1,120 @@
...
+ * file. A definition will be translated to a {@link com.ververica.cdc.composer.PipelineExecution}
+ * by {@link com.ververica.cdc.composer.PipelineComposer} before being submitted to the computing
+ * engine.
+ */
+public class PipelineDef {
+    private final SourceDef source;
+    private final SinkDef sink;
+    private final List<RouteDef> routes;
+    private final List<TransformDef> transforms;
+    private final Configuration config;
```
GOODBOY008, line 49: `Configuration is optional ` is ok , some fixed common configuration key should add as a constant (default value,desc,type etc.) for easy to access.

GOODBOY008: I got it , `com.ververica.cdc.composer.flink.FlinkPipelineComposer#compose` is not complete.

flink-cdc-composer/src/main/java/com/ververica/cdc/composer/flink/FlinkPipelineComposer.java
GOODBOY008: I got it , @PatrickRen Can u add `todo` to remind the uncomplete logic.

flink-cdc-composer/src/main/java/com/ververica/cdc/composer/definition/PipelineDef.java
PatrickRen: Yeah I can't tell what configurations are needed right now, but will do as we implement the composer later. Thanks!

Issue #2657, opened 2023-11-07
Introduce JUnit 5 and AssertJ dependencies


PatrickRen:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Motivation

In order to align with the test framework used in Apache Flink and improve the experience of writing test code, we introduce:

- JUnit 5 as the framework for all new tests
- AssertJ for all assertions in new tests

Note that introducing JUnit 5 won't break existing JUnit 4 style tests, and JUnit 5 provides compatibility running environment (junit-vintage-engine) for JUnit 4 tests.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #2659, opened 2023-11-07
[test] Introduce JUnit 5 and AssertJ dependencies

PatrickRen:
This pull request implement #2657, which introduces JUnit 5 and AssertJ dependencies in the root pom, including:

- Use `junit-bom` in dependency management to pin JUnit 5 versions for all modules
- Introduce AssertJ as global test dependency
- Bump `maven-surefire-plugin` version to 3.0.0-M5, which has better support for JUnit 5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -374,14 +374,7 @@ under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
-                <version>2.22.0</version>
-                <dependencies>
-                    <dependency>
-                        <groupId>org.apache.maven.surefire</groupId>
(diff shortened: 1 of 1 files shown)


PatrickRen:
Actually the project is already using JUnit 5 (see the import [here](https://github.com/ververica/flink-cdc-connectors/blob/4288cf6cae867c4f08c29fa145a4a68416c3f960/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/com/ververica/cdc/connectors/mysql/debezium/DebeziumUtilsTest.java#L23)), which is a transitive dependency of Flink test utils. This pull request makes the dependency clearer and under managed.

leonardBang (approved):
Thanks @PatrickRen for the contribution, LGTM

PR #2664, opened 2023-11-07
[3.0][cdc-common] Add SchemaChangeEvent implementations

lvyanquan:
this closes https://github.com/ververica/flink-cdc-connectors/issues/2642

@leonardBang @Jiabao-Sun @PatrickRen  PTAL.

### DIFF
--- a/flink-cdc-common/src/main/java/com/ververica/cdc/common/event/AddColumnEvent.java
+++ b/flink-cdc-common/src/main/java/com/ververica/cdc/common/event/AddColumnEvent.java
@@ -20,6 +20,8 @@ import org.apache.flink.annotation.PublicEvolving;
 
 import com.ververica.cdc.common.schema.Column;
 
+import javax.annotation.Nullable;
+
 import java.io.Serializable;
 import java.util.List;
 import java.util.Objects;
@@ -29,7 +31,7 @@ import java.util.Objects;
  * lenient column type changes.
  */
 @PublicEvolving
-public final class AddColumnEvent implements SchemaChangeEvent, Serializable {
+public final class AddColumnEvent implements SchemaChangeEvent {
 
     private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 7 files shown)


flink-cdc-common/src/main/java/com/ververica/cdc/common/event/AddColumnEvent.java
```diff
@@ -0,0 +1,79 @@
...
+ * lenient column type changes.
+ */
+@PublicEvolving
+public final class AddColumnEvent implements SchemaChangeEvent {
+
+    private static final long serialVersionUID = 1L;
+
+    private final TableId tableId;
+
+    private final List<Column> addedColumns;
```
ruanhang1993, line 37: When adding column, the column order may be changed. I don't see this information in this class. 
For example, alter column in Mysql.
```sql
ALTER TABLE table_name
ADD [COLUMN] column_definition [FIRST|AFTER existing_column]
```

lvyanquan: Add `ColumnWithPosition` to address it.

lvyanquan: Hi @ruanhang1993, I found that [columns in Schema](https://github.com/ververica/flink-cdc-connectors/blob/99d1ad5e528e9531ed1eb703479e7f064f910930/flink-cdc-common/src/main/java/com/ververica/cdc/common/schema/Schema.java#L48) is unordered. So it would be invalid for CreateTableEvent to set position.

lvyanquan: There will be another pr to solve the right order in Schema. So we can ignore it and continue to complete this pr.

Jiabao-Sun:
Do we need to support `RenameTableEvent` and `DropTableEvent`?

ruanhang1993:
@lvyanquan Because `SchemaChangeEvent` will be put in `CoordinatorRequest` which is `Serializable`, we should make `SchemaChangeEvent` implement `Serializable`.

lvyanquan:
Some commits were add by mistakes, will resubmit it later.

lvyanquan:
@ruanhang1993  CC again.

flink-cdc-common/src/main/java/com/ververica/cdc/common/event/AddColumnEvent.java
```diff
@@ -0,0 +1,157 @@
...
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * A {@link SchemaChangeEvent} that represents an {@code ADD COLUMN} DDL, which may contain the
+ * lenient column type changes.
+ */
+@PublicEvolving
+public final class AddColumnEvent implements SchemaChangeEvent, Serializable {
```
ruanhang1993, line 32: `SchemaChangeEvent` already extends `Serializable`. We do not need to implement `Serializable` here.

flink-cdc-common/src/main/java/com/ververica/cdc/common/event/AddColumnEvent.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /** represents result of an ADD COLUMN DDL that may change column sequence. */
+    public static class ColumnWithPosition implements Serializable {
+
+        final Column column;
+
+        final ColumnPosition position;
+
+        final Column otherColumn;
```
ruanhang1993, line 65:
```
/** The added column lies in the position relative to this column . */
@Nullable final Column existingColumn;
```

flink-cdc-common/src/main/java/com/ververica/cdc/common/event/AddColumnEvent.java
```diff
@@ -0,0 +1,157 @@
...
+        BEFORE,
+        AFTER,
+        FIRST,
+        LAST
+    }
+
+    /** represents result of an ADD COLUMN DDL that may change column sequence. */
+    public static class ColumnWithPosition implements Serializable {
+
+        final Column column;
```
ruanhang1993, line 61: /** The added column. */
final Column addedColumn;

flink-cdc-common/src/main/java/com/ververica/cdc/common/event/AddColumnEvent.java
```diff
@@ -0,0 +1,157 @@
...
+
+        final ColumnPosition position;
```
ruanhang1993, line 63: /** The position of the added column. */

ruanhang1993 (approved):
LGTM

Issue #2665, opened 2023-11-08
[flink-cdc-common] add implementation for DataChangeEvent


lvyanquan:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Motivation

Add implementation for [DataChangeEvent](https://github.com/ververica/flink-cdc-connectors/blob/99d1ad5e528e9531ed1eb703479e7f064f910930/flink-cdc-common/src/main/java/com/ververica/cdc/common/event/DataChangeEvent.java#L28).

you can change `DataChangeEvent` from `interface` to `class`,  as It may have only one implementation class.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

joyCurry30:
I'm willing to do this task, can you assign this to me?

PR #2677, opened 2023-11-09
[3.0][cdc-common] Change DataChangeEvent from interface to class, and add implementation for DataChangeEvent.

joyCurry30:
Refer to #2665 
Change DataChangeEvent from interface to class, and add implementation for DataChangeEvent.
PTAL. @lvyanquan @GOODBOY008

### DIFF
--- a/flink-cdc-common/src/main/java/com/ververica/cdc/common/event/DataChangeEvent.java
+++ b/flink-cdc-common/src/main/java/com/ververica/cdc/common/event/DataChangeEvent.java
@@ -30,7 +30,7 @@ public class DataChangeEvent implements ChangeEvent, Serializable {
 
     private static final long serialVersionUID = 1L;
 
-    public DataChangeEvent(
+    private DataChangeEvent(
             TableId tableId,
             RecordData before,
             RecordData after,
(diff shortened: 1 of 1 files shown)


GOODBOY008 (approved):
Thanks for you contritbutions, I left a minor comment.

flink-cdc-common/src/main/java/com/ververica/cdc/common/event/DataChangeEvent.java
```diff
@@ -18,24 +18,162 @@
...
+    }
+
+    public Map<String, String> meta() {
+        return meta;
+    }
+
+    /**
+     * Creates a {@link DataChangeEvent} instance that describes the insert event.
+     *
+     * @param tableId table id.
```
GOODBOY008, line 84: Remove `@param` will be more clean.

lvyanquan (approved):
+1

flink-cdc-common/src/main/java/com/ververica/cdc/common/event/DataChangeEvent.java
joyCurry30: Yes, I have removed them.

PR #2678, opened 2023-11-09
[build][tests] Enable CI tests for CDC 3.0 modules

GOODBOY008:


tools/azure-pipelines/jobs-template.yml
```diff
@@ -72,6 +72,10 @@ jobs:
       clean: all
     strategy:
       matrix:
+        core:
+          module: core
+        pipeline_connectors:
+          module: pipeline_connectors
```
leonardBang, line 78: We’d better add these here
https://github.com/ververica/flink-cdc-connectors/pull/2678/files#diff-fbc7cebf7845319dcfd1e61c4adbe25fcbdd290cbfbb5b56afdd62b3266db770L150 
as the e2e tests will require pipeline_connectors when we add e2e tests for cdc 3.0 pipeline

GOODBOY008: I think there is no need to add test_2 too. Because before cdc e2e depend other cdc moudle and not added.

leonardBang: Sorry @GOODBOY008 , I mixed the compile and test stage.

leonardBang (approved):
+1

PR #2679, opened 2023-11-10
[hotfix][docs] Change to the correct version

gtk96:
We don't have version 2.5.0, which needs to be modified here.


leonardBang (approved):
Thanks @gtk96 , +1

PR #2716, opened 2023-11-17
[3.0][cdc-common] Use 'ConfigOption' of cdc-common.

joyCurry30:
As we have own ConfigOption, we should use 'ConfigOption' of cdc-common in Factory.

cc @lvyanquan @GOODBOY008


leonardBang (approved):
Thanks @joyCurry30 for the contribution, LGTM

PR #2749, opened 2023-11-24
[cdc-composer] avoid NoPointException while getting option from Context.

lvyanquan:
Configuration provide a convenient way for DataSourceFactory to get ConfigOption.


leonardBang (commented):
Thanks @lvyanquan for the contribution, a test is welcome

lvyanquan:
> a test is welcome

Address it.

PR #2756, opened 2023-11-27
[cdc-common] add PublicEvolving annotation to SchemaUtils.

lvyanquan:
add PublicEvolving annotation to SchemaUtils, add modify code of SchemaManager to reuse the method of SchemaUtils.

@PatrickRen CC.

### DIFF
--- a/flink-cdc-common/src/main/java/com/ververica/cdc/common/utils/SchemaUtils.java
+++ b/flink-cdc-common/src/main/java/com/ververica/cdc/common/utils/SchemaUtils.java
@@ -24,8 +24,6 @@ import com.ververica.cdc.common.event.RenameColumnEvent;
 import com.ververica.cdc.common.event.SchemaChangeEvent;
 import com.ververica.cdc.common.schema.Column;
 import com.ververica.cdc.common.schema.Schema;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.LinkedList;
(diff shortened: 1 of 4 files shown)


flink-cdc-common/src/main/java/com/ververica/cdc/common/utils/SchemaUtils.java
```diff
@@ -32,6 +33,7 @@
 import java.util.stream.Collectors;
 
 /** Utils for {@link Schema} to perform the ability of evolution. */
+@PublicEvolving
```
PatrickRen, line 36: As we are promoting this utility to public, I think we need to be careful about the API, so I have concerns about the method `applySchemaChangeEvent` with parameter `ignoreException`. I prefer to remove this one, as I can't see any use cases for ignoring exceptions. Currently there's no usage of it, and the warning log is empty if the exception is ignored, which confuses me a bit. 

A better behavior would be clearly defining the exception on the method (`throws XXXException`) and letting the caller catch and handle it.

flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/schema/coordinator/SchemaManager.java
```diff
@@ -173,27 +167,8 @@ private void handleAlterColumnTypeEvent(AlterColumnTypeEvent event) {
...
-            }
-        }
-
-        // Dump the rest of information
-        schemaBuilder.primaryKey(oldSchema.primaryKeys());
-        schemaBuilder.comment(oldSchema.comment());
-        schemaBuilder.options(oldSchema.options());
-        Schema newSchema = schemaBuilder.build();
-        registerNewSchema(event.tableId(), newSchema);
+        registerNewSchema(
```
PatrickRen, line 170: We can remove the if-else-if tree in `SchemaManager#applySchemaChange`, as `SchemaUtils#applySchemaChangeEvent` does the `instanceof` logic too.

PatrickRen (commented):
@lvyanquan Thanks for the PR! I left some comments.

lvyanquan:
Address it. And XXXException requires unified processing, so just keep to use IllegalArgumentException here.

PatrickRen (approved):
@lvyanquan Thanks for the update! LGTM.

Issue #2759, opened 2023-11-27
[Bug][mongo] rewriteOutputBuffer should emit +I rather than +U


loserwang1024:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.18

### Flink CDC version

2.4.2

### Database and its version

Whatever

#### Reason
Snapshot split read will read snapshot data, and then backfill log between (low_watermark, high_watermark). The whole read is seen as the snapshot image at high_watermark, which is all +I records to downstream。

<img width="839" alt="image" src="https://github.com/ververica/flink-cdc-connectors/assets/125648852/35321a75-d379-44fb-9784-eadea91dd3ce">
However, rewriteOutputBuffer  method replace record here but not change the type to +I。

#### Minimal reproduce step
Update or Replace value during [snapshot, highwatermark), for example:
[test](https://github.com/ververica/flink-cdc-connectors/commit/ec092e141b8161ac4321206a82038efeeebecc64#r133562390), 
here I want to  get "+I[2000, user_21, Pittsburgh, 123567891234]", rather return +U[2000, user_21, Pittsburgh, 123567891234]

### What did you expect to see?

 get "+I[2000, user_21, Pittsburgh, 123567891234]"

### What did you see instead?

return +U[2000, user_21, Pittsburgh, 123567891234]

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #2760, opened 2023-11-27
[Bug][mongo] Fix MongoDB's rewriteOutputBuffer to emit +I rather than +U(or -U)

loserwang1024:
Fix [#2759](https://github.com/ververica/flink-cdc-connectors/issues/2759)

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/reader/fetch/MongoDBFetchTaskContext.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/reader/fetch/MongoDBFetchTaskContext.java
@@ -162,6 +162,8 @@ public class MongoDBFetchTaskContext implements FetchTask.Context {
                 case INSERT:
                 case UPDATE:
                 case REPLACE:
+                    value.put(
+                            MongoDBEnvelope.OPERATION_TYPE_FIELD, OperationType.INSERT.getValue());
                     outputBuffer.put(key, changeRecord);
                     break;
                 case DELETE:


loserwang1024:
@Jiabao-Sun , @Shawn-Hx , CC

Jiabao-Sun:
+U and +I should have the same processing in the external system. 
Please give an example to explain the specific impact?

loserwang1024:
@Jiabao-Sun , moreover, if `enableFullDocPrePostImage` = true, also will use MongoDBConnectorFullChangelogDeserializationSchema and  emit -U without +I before.

Jiabao-Sun:
> @Jiabao-Sun , moreover, if `enableFullDocPrePostImage` = true, also will use MongoDBConnectorFullChangelogDeserializationSchema and emit -U without +I before.

-U and +U contains the same key, it should be covered by +U ?

loserwang1024:
> -U and +U contains the same key, it should be covered by +U ?


@Jiabao-Sun , `MongoDBFetchTaskContext#rewriteOutputBuffer` method just covered with modify opration(it may contains +U or -U, depending on config `"scan.full-changelog"`).  Then  `MongoDBConnectorDeserializationSchema#deserialize` will only get +U from it , while `MongoDBConnectorFullChangelogDeserializationSchema#deserialize` will get +U and -U from it

Shawn-Hx:
I agree with @loserwang1024. Logically, data before highwatermark are snapshot data, which should be seen as +I. Otherwise the downstream operators may see a +U/-U record without +I before. I'm not sure whether this behavior may cause data incorrectness or not, but using +I is correct with no doubt. @Jiabao-Sun

Jiabao-Sun:
OK, @loserwang1024 could you help add some tests for this?

loserwang1024:
@Jiabao-Sun , thanks a lot. It's not easy to add test now. 
In https://github.com/ververica/flink-cdc-connectors/issues/2759, I add hooks to do database operations during snapshot split code and also add test in it. However, that PR depends on this PR.Therefore, please merge this PR at first.
The detail can see this branch: https://github.com/loserwang1024/flink-cdc-connectors/tree/add-snapshot-hooks.  MongoDBFullChangelogITCase#testEnableBackfillWithPreHighWaterMark、testEnableBackfillWithPostLowWaterMark is what you need.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/com/ververica/cdc/connectors/mongodb/source/reader/fetch/MongoDBFetchTaskContext.java
```diff
@@ -162,6 +163,7 @@ public void rewriteOutputBuffer(
                 case INSERT:
                 case UPDATE:
                 case REPLACE:
+                    value.put(MongoDBEnvelope.OPERATION_TYPE_FIELD, INSERT.getValue());
```
Jiabao-Sun, line 166:
```suggestion
                    value.put(MongoDBEnvelope.OPERATION_TYPE_FIELD, OperationType.INSERT.getValue());
```

loserwang1024: Thanks a lot , I will do it

Jiabao-Sun (approved):

PR #2767, opened 2023-11-28
[cdc-runtime] maintain a set of TableId in DataSinkWriterOperator

lvyanquan:
maintain a set of TableId in DataSinkWriterOperator to make sure that a CreateTableEvent is sent before all DataChangeEvents with the same TableId.


lvyanquan:
@PatrickRen @leonardBang PTAL.

PatrickRen (approved):
@lvyanquan Thanks for the patch! LGTM.

PR #2771, opened 2023-11-28
[cdc-runtime] Improve DataSinkWriterOperator logic that only emit latest schema for events that are not CreateTableEvent

PatrickRen:
This pull request improves DataSinkWriterOperator logic that only emit latest schema for events that are not CreateTableEvent. Otherwise, sink will receive multiple CreateTableEvent for one table, which is unexpected.


leonardBang (approved):
+1

PR #2774, opened 2023-11-28
[cdc-pipeline-connector][values] Avoid using `default` as name of namespace and schema.

lvyanquan:
Because `default` may be keyword  for some connectors, and could not be a name of database like doris.


lvyanquan:
@leonardBang PTAL.

PatrickRen (approved):
@lvyanquan Thanks for the PR! LGTM.

PR #2778, opened 2023-11-29
[cdc-common] Add helper method for creating ConfigOption directly from Flink option

PatrickRen:
This pull request adds a helper function for `ConfigOptions`, that can directly reuse Flink's ConfigOption.


leonardBang:
As we discussed offline, we can close this one ? @PatrickRen

PR #2781, opened 2023-11-30
[cdc-composer] DataSinkTranslator add sink function support

GOODBOY008:


GOODBOY008:
@PatrickRen @leonardBang PTAL

PR #2786, opened 2023-11-30
[cdc-runtime][cdc-composer] Support routing in Flink CDC pipeline

PatrickRen:
This pull request implements routing in Flink CDC pipeline, including:

- Introduce the new RouteFunction for actually executing the routing logic in Flink job
- Chaining router into the DataStream by Flink composer

This PR also introduces Assertj-style assertion systems for Flink CDC specific data structure, in order to make the test logic easier to be written.

### DIFF
--- a/flink-cdc-cli/src/main/java/com/ververica/cdc/cli/parser/YamlPipelineDefinitionParser.java
+++ b/flink-cdc-cli/src/main/java/com/ververica/cdc/cli/parser/YamlPipelineDefinitionParser.java
@@ -32,7 +32,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.regex.Pattern;
 
 import static com.ververica.cdc.common.utils.Preconditions.checkNotNull;
 
@@ -50,9 +49,9 @@ public class YamlPipelineDefinitionParser implements PipelineDefinitionParser {
     private static final String NAME_KEY = "name";
 
     // Route keys
-    private static final String MATCHER_KEY = "matcher";
-    private static final String REPLACE_KEY = "replace";
-    private static final String DESCRIPTION_KEY = "description";
+    private static final String ROUTE_SOURCE_TABLE_KEY = "sourceTable";
+    private static final String ROUTE_SINK_TABLE_KEY = "sinkTable";
+    private static final String ROUTE_DESCRIPTION_KEY = "description";
 
     private final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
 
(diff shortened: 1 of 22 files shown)


flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/route/RouteFunction.java
```diff
@@ -0,0 +1,171 @@
...
+            Selectors selectors = route.f0;
+            TableId replaceBy = route.f1;
+            if (selectors.isMatch(tableId)) {
+                return recreateChangeEvent(changeEvent, replaceBy);
+            }
+        }
+        return event;
+    }
+
+    private ChangeEvent recreateChangeEvent(ChangeEvent event, TableId tableId) {
```
lvyanquan, line 105: Can we add a `copy` method to ChangeEvent to avoid these complex classifications？

flink-cdc-cli/src/main/java/com/ververica/cdc/cli/parser/YamlPipelineDefinitionParser.java
```diff
@@ -50,9 +49,9 @@ public class YamlPipelineDefinitionParser implements PipelineDefinitionParser {
     private static final String NAME_KEY = "name";
 
     // Route keys
-    private static final String MATCHER_KEY = "matcher";
-    private static final String REPLACE_KEY = "replace";
-    private static final String DESCRIPTION_KEY = "description";
+    private static final String ROUTE_SOURCE_TABLE_KEY = "sourceTable";
```
lvyanquan, line 52: Will `sourceTable` be better?

lvyanquan (commented):
Thanks @PatrickRen for this contribution, LGTM, left two discussion.

flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/route/RouteFunction.java
PatrickRen: Hmmm my concern is that it might be confusing if we introduce this `copy` in the `ChangeEvent`, which is now a clean interface. 

`copy` sounds like to create a new event which is exactly the same, but here we need to rewrite the table ID field, that means something like `replaceTableId(TableId)`, but it's quite confusing to implementations that why they have to provide an ability to change table id only but not for other fields.

flink-cdc-cli/src/main/java/com/ververica/cdc/cli/parser/YamlPipelineDefinitionParser.java
PatrickRen: The best naming in my mind is:
`sourceTable` -> `matcher`: this gives a hint to user that a regex is required here
`sinkTable` -> `replaceBy`: this will be used to replace the original table ID. 
which is like the replacing functionality in text editor.

leonardBang: Let's keep the simple name as we designed, source/sink concepts is better as flink users are familiar with them

PR #2796, opened 2023-12-02
[cdc-composer] Add configuration from flink conf to remote cluster.

lvyanquan:
To add configuration of checkpoint or restart strategy or others.


lvyanquan:
@PatrickRen PTAL.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/main/java/com/ververica/cdc/connectors/values/source/ValuesDataSource.java
```diff
@@ -112,7 +112,7 @@ public ValuesSource(
 
         @Override
         public Boundedness getBoundedness() {
-            return Boundedness.BOUNDED;
+            return isInSnapshotPhase ? Boundedness.BOUNDED : Boundedness.CONTINUOUS_UNBOUNDED;
```
lvyanquan, line 115: Need to add an option for Source to control whether to support Boundedness.CONTINUOUS_UNBOUNDED, will do later.

PR #2799, opened 2023-12-02
[cdc-connector][sqlserver][tests] Fix UT errors by increasing the wait time after committing SQL

gong:
Fix UT errors by increasing the wait time after committing SQL.
 commit() can't be  not effective when wait 1s. So change 1s to 5s.
image


leonardBang (commented):
Thanks @gong for the troubleshooting, let's keep an eye on the CI result

leonardBang (approved):

PR #2801, opened 2023-12-02
[pipeline-common] use column name to judge whether a column is existed in a specific schema.

lvyanquan:
Will use name instead of Column for some SchemaChangeEvent in the future.


lvyanquan:
@ruanhang1993 CC.

leonardBang (approved):
+1

PR #2805, opened 2023-12-03
[cdc-runtime] Setup waitFlushSuccess before responding operator to flush

PatrickRen:
This pull request fixes a bug in `SchemaRegistryRequestHandler`, which is caught [here](https://dev.azure.com/ververica/flink-cdc-connectors/_build/results?buildId=4280&view=logs&j=efbdbacd-38c9-5b2e-0c3e-e76049e31803&t=f16af5b3-8c31-57fc-a5cb-048f335759b4&l=4627).

There might be a case that sink flushes faster than schema operator requesting release upstream. In that case waitFlushSuccess has not been set when sink has flushed successfully, which will lead to a NullPointerException.


ruanhang1993 (commented):
`startNextSchemaChangeRequest` also need to be changed.

ruanhang1993 (approved):
LGTM.

PR #2807, opened 2023-12-04
[cdc-cli][cdc-composer][cdc-dist] Support submitting job to general Flink environment

PatrickRen:
This pull request adds support to submit Flink CDC pipeline job to more Flink deployments, such as YARN. 

Please note that there are still many things that can be improved comparing to the native `flink run` or SQL client. For example to submit to a YARN session cluster, user needs to set `execution.target: yarn-session` and `yarn-application-id: {flink-app-id}` in order to make everything works.


leonardBang (approved):
Thanks @PatrickRen for the contribution, LGTM

PR #2808, opened 2023-12-04
[mysql-cdc] Current binlog offset add serverId, fix incremental-snapshot scan data lose

EchoLee5:
Current binlog offset add serverId, fix incremental-snapshot scan data lose
link https://github.com/ververica/flink-cdc-connectors/issues/2794


ruanhang1993 (commented):
@EchoLee5 Thanks for the fix. Could you add some tests to cover this bug? 
I think we could start a mysql container without `gtid-mode = on` to reproduce this bug.

ruanhang1993:
The problem will occur when the gtid mode is closed and some data changes happen in the snapshot phase.

EchoLee5:
> The problem will occur when the gtid mode is closed and some data changes happen in the snapshot phase.

Sorry for being late, I have added unit tests, please review them for me, thank you!

yuxiqian:
Hi @EchoLee5, could you please rebase this PR with latest `master` branch before it could be merged?

cc @ruanhang1993

PR #2813, opened 2023-12-05
[cdc-cli] When using miniCluster, don't check 'FLINK_HOME'.

joyCurry30:
If 'FLINK_HOME' is not set, we can use empty configuration, rather than throwing exception.
cc @lvyanquan

### DIFF
--- a/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontend.java
+++ b/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontend.java
@@ -21,8 +21,9 @@ import org.apache.flink.cdc.cli.utils.ConfigurationUtils;
 import org.apache.flink.cdc.cli.utils.FlinkEnvironmentUtils;
 import org.apache.flink.cdc.common.annotation.VisibleForTesting;
 import org.apache.flink.cdc.common.configuration.Configuration;
+import org.apache.flink.cdc.composer.PipelineComposer;
 import org.apache.flink.cdc.composer.PipelineExecution;
-
+import org.apache.flink.cdc.composer.flink.FlinkPipelineComposer;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
 import org.apache.commons.cli.DefaultParser;
@@ -86,6 +87,18 @@ public class CliFrontend {
         // Global pipeline configuration
         Configuration globalPipelineConfig = getGlobalConfig(commandLine);
 
+        // Create pipeline composer
+        boolean useMiniCluster = commandLine.hasOption(CliFrontendOptions.USE_MINI_CLUSTER);
+        PipelineComposer composer =
+                useMiniCluster
+                        ? FlinkPipelineComposer.ofMiniCluster()
+                        : createRemoteComposer(commandLine);
+
+        // Build executor
+        return new CliExecutor(pipelineDefPath, globalPipelineConfig, composer);
+    }
+
+    private static PipelineComposer createRemoteComposer(CommandLine commandLine) throws Exception {
         // Load Flink environment
         Path flinkHome = getFlinkHome(commandLine);
         Configuration flinkConfig = FlinkEnvironmentUtils.loadFlinkConfiguration(flinkHome);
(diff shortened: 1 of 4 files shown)


flink-cdc-cli/src/main/java/com/ververica/cdc/cli/CliFrontend.java
```diff
@@ -105,25 +104,34 @@ static CliExecutor createExecutor(CommandLine commandLine) throws Exception {
                 additionalJars);
     }
 
-    private static Path getFlinkHome(CommandLine commandLine) {
+    private static Configuration loadFlinkConfiguration(CommandLine commandLine) throws Exception {
+        Optional<Path> flinkHome = getFlinkHome(commandLine);
+        if (flinkHome.isPresent()) {
+            return FlinkEnvironmentUtils.loadFlinkConfiguration(flinkHome.get());
+        }
+        return new Configuration();
```
lvyanquan, line 112: We use Configuration from Flink conf to open checkpoint and submit job to rest client, if Configuration is empty, job could not be submitted to remote cluster, see https://github.com/ververica/flink-cdc-connectors/blob/33b7c85093a69382d23c309a9b209ddae27a2968/flink-cdc-composer/src/main/java/com/ververica/cdc/composer/flink/FlinkPipelineComposer.java#L60.

flink-cdc-cli/src/main/java/com/ververica/cdc/cli/CliFrontend.java
```diff
@@ -105,25 +104,34 @@ static CliExecutor createExecutor(CommandLine commandLine) throws Exception {
...
+    }
+
+    private static Optional<Path> getFlinkHome(CommandLine commandLine) {
         // Check command line arguments first
         String flinkHomeFromArgs = commandLine.getOptionValue(CliFrontendOptions.FLINK_HOME);
         if (flinkHomeFromArgs != null) {
             LOG.debug("Flink home is loaded by command-line argument: {}", flinkHomeFromArgs);
-            return Paths.get(flinkHomeFromArgs);
+            return Optional.of(Paths.get(flinkHomeFromArgs));
```
lvyanquan, line 120: Paths#get will never return null, it's better to use Files#exists judgement.

flink-cdc-cli/src/test/java/com/ververica/cdc/cli/CliFrontendTest.java
```diff
@@ -83,16 +83,6 @@ void testGlobalPipelineConfigParsing() throws Exception {
...
     }
 
-    @Test
-    void testAdditionalJar() throws Exception {
-        String aJar = "/foo/jar/a.jar";
-        String bJar = "/foo/jar/b.jar";
-        CliExecutor executor =
-                createExecutor(
-                        pipelineDef(), "--flink-home", flinkHome(), "--jar", aJar, "--jar", bJar);
-        assertThat(executor.getAdditionalJars()).contains(Paths.get(aJar), Paths.get(bJar));
```
lvyanquan, line 93: This test is use to verify the correctness of `--jar` option, it's better to remain it.

lvyanquan:
Overall LGTM. Can you add a test case like [CliFrontendTest#testMissingFlinkHome](https://github.com/ververica/flink-cdc-connectors/blob/08d6b8178ffe89f787067e9e0cc7324b0e66fb97/flink-cdc-cli/src/test/java/com/ververica/cdc/cli/CliFrontendTest.java#L64)?

lvyanquan (approved):

lvyanquan:
@joyCurry30 To go a step further, can you add a section in [tutorial](https://ververica.github.io/flink-cdc-connectors/master/content/quickstart/mysql-doris-pipeline-tutorial.html) to introduce this `--use-mini-cluster` parameter?

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontend.java
```diff
@@ -86,6 +87,18 @@ static CliExecutor createExecutor(CommandLine commandLine) throws Exception {
         // Global pipeline configuration
         Configuration globalPipelineConfig = getGlobalConfig(commandLine);
 
+        // Create pipeline composer
+        boolean useMiniCluster = commandLine.hasOption(CliFrontendOptions.USE_MINI_CLUSTER);
```
PatrickRen, line 91: How could the MiniCluster run without specifying `FLINK_HOME`? We don't package `flink-dist` into our distribution, so I assume that the MiniCluster cannot even be started without specifying the place where Flink related classes are.

PatrickRen (commented):
@joyCurry30 Thanks for the contribution. I left a comment.

PR #2815, opened 2023-12-05
[pipeline-connector][mysql] Enable send schema change by default

leonardBang:
The pipeline framework supports schema evolution by default, we should enable send schema change by default for mysql pipeline source


Jiabao-Sun (approved):
+1

ruanhang1993 (approved):
LGTM.

PR #2818, opened 2023-12-05
[build] Fix some configurations in pom.xml

PatrickRen:
This pull request fixes some configurations in pom.xml, including:

- Skip maven-deploy-plugin for test modules and flink-cdc-dist
- Use `${project.version}` instead of `${revision}` for dependency version, as documented [here](https://maven.apache.org/maven-ci-friendly.html#dependencies)
- Use flink-shaded-force-shading to force all submodules to generate dependency-reduced-pom.xml


leonardBang (approved):
Thanks @PatrickRen for the fix, LGTM

leonardBang:
@PatrickRen could you fix the compile issue ?

PR #2820, opened 2023-12-05
[cdc-pipeline-connector][mysql] Fix precision problem of BIT type conversion

Jiabao-Sun:
Fix precision problem of BIT type conversion.


leonardBang (approved):
Thanks @Jiabao-Sun for troubleshooting and fixing this corner case , LGTM, let's wait the ci green.

Jiabao-Sun:
[DBZ-7230](https://issues.redhat.com/browse/DBZ-7230)

https://github.com/debezium/debezium/pull/5076

Sukang1002:
A similar situation is found if a new column of bit type is added to the mysql table, an error "org.apache.flink.table.api.ValidationException: Binary string length must be between 1 and 2147483647 (both inclusive) " 
 may be reported

yuxiqian:
Hi @Sukang1002, I've tried both pipeline connector and source connector and could not reproduce this issue. Could you please kindly share your configuration so we can investigate it further?

PR #2823, opened 2023-12-06
[cdc-cli][cdc-composer] Use "source-table" and "sink-table" as keywords of route

PatrickRen:
This pull request fixes keywords in route definitions, which should be "source-table" and "sink-table".

Please note that in `RouteDef` these two keywords are still "sourceTable" and "sinkTable", because they are Java object names. Only keywords in YAML need to be adjusted.


PatrickRen:
Looks like the CI is linked to a wrong pull request. 
This should be the real CI run: https://dev.azure.com/ververica/flink-cdc-connectors/_build/results?buildId=4379

ruanhang1993 (approved):
LGTM

PR #2829, opened 2023-12-06
[pipeline-connector][starrocks] Fix char/varchar length inconsistency between cdc and starrocks

banmoy:
In CDC framework, the length of char/varchar type represents the number of characters, but in StarRocks it represent the length of bytes. When mapping CDC char/varchar to that of StarRocks, we should multiply 3 for the length because a UTF-8 chinese character will use 3 bytes


lvyanquan (approved):
LGTM.

lvyanquan:
The CI failure is not releted to this pr, CC @PatrickRen.

PatrickRen (approved):
@banmoy @lvyanquan Thanks for the contribution! LGTM

PR #2831, opened 2023-12-07
[cdc-common] Add default value of pipeline parallelism.

joyCurry30:
Refer to #2855
Add default value of pipeline parallelism. cc @lvyanquan

### DIFF
--- a/flink-cdc-common/src/main/java/com/ververica/cdc/common/pipeline/PipelineOptions.java
+++ b/flink-cdc-common/src/main/java/com/ververica/cdc/common/pipeline/PipelineOptions.java
@@ -37,7 +37,7 @@ public class PipelineOptions {
     public static final ConfigOption<Integer> PIPELINE_PARALLELISM =
             ConfigOptions.key("parallelism")
                     .intType()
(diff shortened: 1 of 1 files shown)


lvyanquan (approved):
LGTM

flink-cdc-common/src/main/java/com/ververica/cdc/common/pipeline/PipelineOptions.java
```diff
@@ -37,7 +37,7 @@ public class PipelineOptions {
     public static final ConfigOption<Integer> PIPELINE_PARALLELISM =
             ConfigOptions.key("parallelism")
                     .intType()
-                    .noDefaultValue()
+                    .defaultValue(1)
```
leonardBang, line 40: This is a public API change, we only introduce this in 3.1 version, we need some discussion in dev group

joyCurry30: @leonardBang Thank you for review. I have supplemented the relevant issues accordingly #2855 .

PR #2837, opened 2023-12-07
[cdc-cli][cdc-composer] Add 'flink-config' for pipeline yaml

joyCurry30:
Refer to #2854 
Add 'flink-config' for pipeline yaml, which will overwrite the config from 'flink-conf.yaml'. cc @lvyanquan @PatrickRen


README.md
```diff
@@ -56,6 +56,10 @@ sink:
 pipeline:
   name: mysql-sync-doris
   parallelism: 4
+
+flink-config:
+  execution.checkpointing.interval: 30000
+  execution.checkpointing.mode: EXACTLY_ONC
```
leonardBang, line 62: We need discuss the user facing API in dev group first and then open a PR, we should be careful when introduce new API for end users.

joyCurry30: @leonardBang Thank you for review. I have supplemented the relevant issues accordingly #2854.

lvyanquan (approved):
LGTM.

flink-cdc-cli/src/main/java/com/ververica/cdc/cli/parser/YamlPipelineDefinitionParser.java
```diff
@@ -43,6 +43,7 @@ public class YamlPipelineDefinitionParser implements PipelineDefinitionParser {
     private static final String SINK_KEY = "sink";
     private static final String ROUTE_KEY = "route";
     private static final String PIPELINE_KEY = "pipeline";
+    private static final String FLINK_CONFIG_KEY = "flink-config";
```
leonardBang, line 46: I'm not tend to introduce another top level concept(section) to users, less is more, could we make user defined flink config as part fo pipeline section?

leonardBang (commented):
Thanks @joyCurry30 for the contribution, I left one comment

flink-cdc-cli/src/main/java/com/ververica/cdc/cli/parser/YamlPipelineDefinitionParser.java
joyCurry30: Sorry for the late reply. Yes, I agreed with you, we should not give users too much concept. But how do we distinguish which configs belong to flink config? Use "flink.config" as the prefix to distinguish?

flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml
```diff
@@ -53,3 +53,7 @@ pipeline:
   name: source-database-sync-pipe
   parallelism: 4
   enable-schema-evolution: false
+
+flink-config:
```
czy006, line 57: flink-config parallelism and cdc-confg parallelism, how is the priority divided here?

joyCurry30: Currently, if parallelism exists both in the Flink config and CDC config, the Flink config will override the CDC config.

czy006: OK

PR #2839, opened 2023-12-07
[postgres] Add commits of PostgresConnection back to distinguish with debezium's origin class

loserwang1024:
In https://github.com/ververica/flink-cdc-connectors/pull/2156, debezium version is bump to 1.9.7.Final, including PostgresConnection. It's a good thing, however some comments of PostgresConnection to distinguish with debezium's origin class are removed. 

Two additional methods of our PostgresConnection is not included in 1.9.7.Final Debezium's PostgresConnection., so these comment should not be removed.

These comments are important for us for future to track the changes of Debezium's  PostgresConnection. If these two PostgresConnection become same, we will remove it from cdc. Without it, it hard for us to distinguish the changes we did.


GOODBOY008 (approved):
LGTM

PR #2848, opened 2023-12-11
[pipeline-connector][doris] Extract type mapping logic

JNSimba:
Currently in DorisSink, the type mapping logic directly reuses the Flink Doris Connector. Although it is more convenient, it forms a strong binding relationship with the connector in terms of freedom, which greatly reduces the degree of freedom.
So the Doris type mapping is extracted here.


vinlee19:
Maybe ,we can add TimeType in DorisMapper to fix https://github.com/apache/flink-cdc/issues/3099.

yuxiqian:
Thanks @JNSimba for the great work, could you please rebase this PR with latest `master` branch before it could be merged?

PR #2864, opened 2023-12-13
[cdc-connector][sqlserver][tests] Fix UT errors by correcting right output

loserwang1024:
As shown in https://github.com/ververica/flink-cdc-connectors/issues/2853, SqlServer incremental source cannot support exactly-once now. 

Between the time that a change is committed in the source table, and the time that the change appears in the corresponding change table, there is always a small latency interval.

Because the time is random, the test com.ververica.cdc.connectors.sqlserver.source.SqlServerSourceITCase#testEnableBackfillWithDMLPreHighWaterMark will sometime fail even sleep for 5 minute.

however, it's rather than tricky,  may not be fixed recently. In oder to not influence other PR's CI , before https://github.com/ververica/flink-cdc-connectors/issues/2853 is fixed,  fix UT errors by correcting right output which includes duplicate data.


loserwang1024:
@gong @leonardBang @GOODBOY008 , before we fix https://github.com/ververica/flink-cdc-connectors/issues/2853, fix the CI not to block other PR.

gong (approved):
LGTM

loserwang1024:
@leonardBang , @GOODBOY008 , CC

leonardBang (approved):
+1

leonardBang:
Thanks @loserwang1024 and @gong for the contribution, LGTM. Let's track this issue in https://github.com/ververica/flink-cdc-connectors/issues/2853

PR #2868, opened 2023-12-13
[cdc-source-connector][oceanbase] use current timestamp when query gts failure

whhe:
Fix compatibility issue introduced by #2565


whhe:
PTAL @leonardBang

leonardBang (approved):
+1, thanks @whhe

PR #2908, opened 2023-12-21
[pipeline-connector] Ability to provide a schema replacement

WholeWorld-Timothy:
In our usage scenario, the source schema of mysql and the target schema of starrocks are often different, but the table names and column names are consistent, and this pr provides this capability.
Maintain a route configuration as follows:
```
route:
  - source-table: test.[\S]*
    sink-table: test02.<>
```
Represents the table below the test will be transferred to the test02, where the symbol <> the representative table name is unchanged, just change the schema. For example, the test.table would be replaced with the test02.table.
And maintain a route configuration as follows:
```
route:
  - source-table: test.[\S]*
    sink-table: test02.s02_<>
```
Represents the table below the test will be transferred to the test02, Replace the original table name with the <> symbol. For example, the test.table would be replaced with the test02.s02_table.
And the user can define the replacement symbols by themselves as follows:
```
route:
  - source-table: mrtdb.[\S]*
    sink-table: mrtdb_sd.sd_s<>
    replace-symbol: s<>
```
Represents the table below the test will be transferred to the test02, Replace the original table name with the s<> symbol.

### DIFF
--- a/flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/route/RouteFunction.java
+++ b/flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/route/RouteFunction.java
@@ -96,6 +96,15 @@ public class RouteFunction extends RichMapFunction<Event, Event> {
             Selectors selectors = route.f0;
             TableId replaceBy = route.f1;
             if (selectors.isMatch(tableId)) {
+                // Add a rule that when configuring * in tablename,
+                // only the namespace name and schemaName name needs to be changed
+                if (replaceBy.getTableName().equalsIgnoreCase("<>")) {
+                    replaceBy =
+                            TableId.parse(
+                                    replaceBy.getNamespace(),
+                                    replaceBy.getSchemaName(),
+                                    tableId.getTableName());
+                }
                 return recreateChangeEvent(changeEvent, replaceBy);
             }
         }
--- a/flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/route/RouteFunction.java
+++ b/flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/route/RouteFunction.java
@@ -46,6 +46,7 @@ public class RouteFunction extends RichMapFunction<Event, Event> {
     public static Builder newBuilder() {
         return new Builder();
     }
+    private final String tableNameReplaceSymbol = "<>";
 
     /** Builder of {@link RouteFunction}. */
     public static class Builder {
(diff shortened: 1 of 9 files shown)


WholeWorld-Timothy:
PTAL @banmoy

lvyanquan:
Hi, @WholeWorld-Timothy this idea looks good to me. 
Can we support adding table name prefix and table name suffix based on this placeholder, And can we expand this to modifying the schema?

WholeWorld-Timothy:
The current mode is a support for modification mode, and I think it is a good idea to add a table name prefix or a table name suffix based on this placeholder, so I can try it out.

WholeWorld-Timothy:
PTAL @lvyanquan

flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/route/RouteFunction.java
```diff
@@ -96,6 +96,16 @@ public Event map(Event event) throws Exception {
...
             if (selectors.isMatch(tableId)) {
+                // Add a rule that when configuring <> symbol in tablename,
+                // the namespace name and schemaName name needs to be changed
+                // the <> symbol needs to be replaced by the table name of the event
+                if (replaceBy.getTableName().contains("<>")) {
+                    replaceBy =
+                            TableId.parse(
+                                    replaceBy.getNamespace(),
+                                    replaceBy.getSchemaName(),
+                                    replaceBy.getTableName().replace("<>", tableId.getTableName()));
```
lvyanquan, line 107: We need to extract this literal into a static constant, and determine a value that will not be accidentally used in the table name.

WholeWorld-Timothy: I changed a version and put the definition of the <> symbol into the configuration file, just like this form:
```
route:
  - source-table: mrtdb.[\S]*
    sink-table: mrtdb_sd.sd_s<>
    replace-symbol: s<>
```
In this way, the user can define the replacement symbols by themselves. Is it any better?

lvyanquan: +1. We just need to provide a default value.
Please add some tests for these scenes.

WholeWorld-Timothy: Two methods,  RouteFunctionTest#testSchemaChangeRouting And RouteFunctionTest#testTableNameChangeRouting were modified to test this scenes.

lvyanquan: To expose this parameter to the users, please add description to [cdc-pipeline.md](https://github.com/ververica/flink-cdc-connectors/blob/15a734efa09343228dc7e69e2074a16bfec40f48/docs/content/overview/cdc-pipeline.md#route).
And add this configuration to [testParsingFullDefinition](https://github.com/ververica/flink-cdc-connectors/blob/15a734efa09343228dc7e69e2074a16bfec40f48/flink-cdc-cli/src/test/java/com/ververica/cdc/cli/parser/YamlPipelineDefinitionParserTest.java#L42).

lvyanquan (approved):
LGTM.

everhopingandwaiting (commented):
LGTM

everhopingandwaiting (approved):
LGTM

WholeWorld-Timothy:
My local compilation and testing are correct, so the error in check may not be caused by the code? How can rerun these check ? I click rerun in github desktop doesn t work

whhe:
Thanks for your contribution, please rebase the master branch and I'll take a look on the CI failure later.

Besides, there are some doc modifications which seems to be unnecessary, could you please revert them back?

yuxiqian:
Hi @WholeWorld-Timothy, could you please rebase this PR with latest `master` branch and address the comments above?

cc @whhe

leonardBang:
Closed by https://github.com/apache/flink-cdc/pull/3428

PR #2927, opened 2023-12-26
[hotfix][cdc-connector][mysql] Expose row_kind metadata column for mysql cdc source.

leonardBang:
We use 'row_kind' name instead of 'op' or 'op_type', the operation type is a concept of database and database only has 'update' type and hasn't 'update_before' and 'update_after' type.


leonardBang:
@whhe Could you help review this PR？

whhe (approved):
LGTM

PR #2944, opened 2023-12-29
[FLINK-35893][cdc-runtime] Write CURRENT_VERSION of TableChangeInfo to state

lvyanquan:
This closes https://github.com/ververica/flink-cdc-connectors/issues/2936.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TableChangeInfo.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TableChangeInfo.java
@@ -109,7 +109,13 @@ public class TableChangeInfo {
     /** Serializer for {@link TableChangeInfo}. */
     public static class Serializer implements SimpleVersionedSerializer<TableChangeInfo> {
 
+        /** The latest version before change of state compatibility. */
+        public static final int VERSION_BEFORE_STATE_COMPATIBILITY = 1;
+
         public static final int CURRENT_VERSION = 2;
+
+        /** Used to distinguish with the state which CURRENT_VERSION was not written. */
+        public static final TableId MAGIC_TABLE_ID = TableId.tableId("__magic_table__");
 
         @Override
         public int getVersion() {
@@ -122,6 +128,8 @@ public class TableChangeInfo {
             SchemaSerializer schemaSerializer = SchemaSerializer.INSTANCE;
             try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
                     DataOutputStream out = new DataOutputStream(baos)) {
+                tableIdSerializer.serialize(MAGIC_TABLE_ID, new DataOutputViewStreamWrapper(out));
+                out.writeInt(CURRENT_VERSION);
                 tableIdSerializer.serialize(
                         tableChangeInfo.getTableId(), new DataOutputViewStreamWrapper(out));
                 schemaSerializer.serialize(
@@ -139,6 +147,12 @@ public class TableChangeInfo {
             try (ByteArrayInputStream bais = new ByteArrayInputStream(serialized);
                     DataInputStream in = new DataInputStream(bais)) {
                 TableId tableId = tableIdSerializer.deserialize(new DataInputViewStreamWrapper(in));
+                if (tableId.equals(MAGIC_TABLE_ID)) {
+                    version = in.readInt();
+                    tableId = tableIdSerializer.deserialize(new DataInputViewStreamWrapper(in));
+                } else {
+                    version = VERSION_BEFORE_STATE_COMPATIBILITY;
+                }
                 Schema originalSchema =
                         schemaSerializer.deserialize(version, new DataInputViewStreamWrapper(in));
                 Schema transformedSchema =
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TableChangeInfo.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TableChangeInfo.java
@@ -109,7 +109,7 @@ public class TableChangeInfo {
     /** Serializer for {@link TableChangeInfo}. */
     public static class Serializer implements SimpleVersionedSerializer<TableChangeInfo> {
 
-        public static final int CURRENT_VERSION = 1;
+        public static final int CURRENT_VERSION = 2;
 
         @Override
         public int getVersion() {
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/ColumnSerializer.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/ColumnSerializer.java
@@ -46,6 +46,13 @@ public class ColumnSerializer extends TypeSerializerSingleton<Column> {
     private final MetadataColumnSerializer metadataColumnSerializer =
             MetadataColumnSerializer.INSTANCE;
 
+    private static int currentVersion = 2;
+
+    /** Update {@link #currentVersion} as We did not directly include this version in the file. */
+    public static void updateVersion(int version) {
+        currentVersion = version;
+    }
+
     @Override
     public boolean isImmutableType() {
         return false;
@@ -92,12 +99,16 @@ public class ColumnSerializer extends TypeSerializerSingleton<Column> {
 
     @Override
     public Column deserialize(DataInputView source) throws IOException {
+        return deserialize(currentVersion, source);
+    }
+
+    public Column deserialize(int version, DataInputView source) throws IOException {
         ColumnType columnType = enumSerializer.deserialize(source);
         switch (columnType) {
             case METADATA:
                 return metadataColumnSerializer.deserialize(source);
             case PHYSICAL:
-                return physicalColumnSerializer.deserialize(source);
+                return physicalColumnSerializer.deserialize(version, source);
             default:
                 throw new IOException("Unknown column type: " + columnType);
         }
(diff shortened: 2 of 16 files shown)


lvyanquan:
To support column comment, we need to add a configuration like [include-schema-comments](https://debezium.io/documentation/reference/2.4/connectors/mysql.html#mysql-property-include-schema-comments)

lvyanquan:
PTAL @ruanhang1993 @Jiabao-Sun.

Jiabao-Sun:
Thanks @lvyanquan for this contribution.

The `defaultValueExpression` of debezium is an unparsed default value expression, which may even include functions like `ow()`. I want to know what is the purpose of `defaultValue` and if there could be some issues if it is used to populate the value of the sink.

And it is named `defaultValueExpression` instead of `defaultValue` because some certain values are not easily serialized. Therefore, the original expression is preserved during serialization and restored to the `Schema`'s defaultValue using `defaultValueConverter`.

Refers to https://github.com/debezium/debezium/pull/2671.

lvyanquan:
Thanks @Jiabao-Sun for pointing out this, We will indeed encounter unexpected situations when using NOW() or CURRENT_TIMESTAMP. I will close this PR later.

lvyanquan:
Hi, @Jiabao-Sun. 
I have checked the [documentation](https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html) of MYSQL and find that: 
> The default value specified in a DEFAULT clause can be a literal constant or an expression. With one exception, enclose expression default values within parentheses to distinguish them from literal constant default values.

And based this code [DefaultValueParserListener](https://github.com/ververica/flink-cdc-connectors/blob/15a734efa09343228dc7e69e2074a16bfec40f48/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/antlr/listener/DefaultValueParserListener.java#L46), we can distinguish between these two situations, the defaultValueExpression from a `expression` should be null.
 so we can still maintain consistency in the `expression` situation and add support `literal constant` using this defaultValueExpression of ColumnImpl, and it‘s actually a literal string.

lvyanquan:
There are three conditions:
1. Default value is literal, the value of column.defaultValueExpression() will be the same as literal.
2. Default value is function, the value of column.defaultValueExpression() will be `null`.
3. Default value is NOW() or CURRENT_TIMESTAMP, the value of column.defaultValueExpression() will be `1970-01-01 00:00:00`.
So we will not pass unresolved values to downstream.

lvyanquan:
@Jiabao-Sun PTAL.

Jiabao-Sun (approved):
LGTM

lvyanquan:
Rebased to master.

yuxiqian:
Hi @lvyanquan could you please rebase this with `master` again? Also cc @Jiabao-Sun

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/PhysicalColumnSerializer.java
yuxiqian, line 1: Seems this is a breaking change of current serialization format. Should we bump serializer version here to keep backwards compatibility?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/ColumnSerializer.java
```diff
@@ -46,6 +46,19 @@ public class ColumnSerializer extends TypeSerializerSingleton<Column> {
     private final MetadataColumnSerializer metadataColumnSerializer =
             MetadataColumnSerializer.INSTANCE;
 
+    /**
+     * Identify for different major version:
+     * for Version 3.0/3.1 is 0,
+     * for version 3.2 is 1
+     */
```
yuxiqian, line 53: This seems not quite accurate, since 3.1.1 has bumped `SchemaSerializer` version from 0 to 1 and adds `partitionKey` options. Since default value will not be shipped until 3.2, should we choose schema-related serialization version?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/PhysicalColumnSerializer.java
lvyanquan: Thanks for pointing out this.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/ColumnSerializer.java
lvyanquan: done.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/SchemaSerializer.java
```diff
@@ -39,8 +39,7 @@ public class SchemaSerializer extends TypeSerializerSingleton<Schema> {
     /** Sharable instance of the TableIdSerializer. */
     public static final SchemaSerializer INSTANCE = new SchemaSerializer();
 
-    private final ListSerializer<Column> columnsSerializer =
-            new ListSerializer<>(ColumnSerializer.INSTANCE);
+    private ListSerializer<Column> columnsSerializer;
```
yuxiqian, line 42: This could be uninitialized when `SchemaSerializer#copy` was called.

lvyanquan: Thanks, addressed it and done state compatibility testing using https://github.com/apache/flink-cdc/pull/3447.

yuxiqian (approved):
Thanks for @lvyanquan's quick fix! Pulled, tested locally and it works fine.

leonardBang:
@lvyanquan Could you check the failed tests? and it's recommended to rebase master.

lvyanquan:
Done rebase.
Add the cause of failed case is that version info were not included in [TransformSchemaOperator](https://github.com/apache/flink-cdc/blob/ea71b2302ddc5f9b7be65843dbf3f5bed4ca9d8e/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformSchemaOperator.java#L127), which will cause incompatible state, I would like to  start a discussion with @aiwenmo @yuxiqian to solve it.

yuxiqian:
> Done rebase.
> 
> Add the cause of failed case is that version info of SERIALIZER were not included in [TransformSchemaOperator](https://github.com/apache/flink-cdc/blob/ea71b2302ddc5f9b7be65843dbf3f5bed4ca9d8e/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformSchemaOperator.java#L127), which will cause incompatible state, I would like to start a discussion with @aiwenmo @yuxiqian to solve it.

My 20 cents: Since it's a legacy problem that we didn't serialize version number, can we add magic header bytes to distinguish newer version (with serializer version) and legacy ones?

yuxiqian:
> Done rebase.
> 
> Add the cause of failed case is that version info of SERIALIZER were not included in [TransformSchemaOperator](https://github.com/apache/flink-cdc/blob/ea71b2302ddc5f9b7be65843dbf3f5bed4ca9d8e/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformSchemaOperator.java#L127), which will cause incompatible state, I would like to start a discussion with @aiwenmo @yuxiqian to solve it.

Can we add some magic header bytes in newer serialization formats (with serializer version of course) to distinguish it with legacy ones?

lvyanquan:
Fixed the test case.

And the wrong use of SERIALIZER in [TransformSchemaOperator](https://github.com/apache/flink-cdc/blob/ea71b2302ddc5f9b7be65843dbf3f5bed4ca9d8e/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformSchemaOperator.java#L127) is actually another problem that could be traced in another jira.

lvyanquan:
@yuxiqian  I've add magic header in  SERIALIZER of TableChangeInfo to distinguish with the state in 3.1.1 and after, can you help to check again?
And this change could be verified after https://github.com/apache/flink-cdc/pull/3426 and this pr, so I tend to add it in current pr.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TableChangeInfo.java
```diff
@@ -139,6 +144,10 @@ public TableChangeInfo deserialize(int version, byte[] serialized) throws IOExce
             try (ByteArrayInputStream bais = new ByteArrayInputStream(serialized);
                     DataInputStream in = new DataInputStream(bais)) {
                 TableId tableId = tableIdSerializer.deserialize(new DataInputViewStreamWrapper(in));
+                if (tableId.equals(magicTableId)) {
+                    version = in.readInt();
+                    tableId = tableIdSerializer.deserialize(new DataInputViewStreamWrapper(in));
+                }
```
yuxiqian, line 150: If no `magicTableId` was present, `version` will not be overwritten and defaults to be latest version (`TableChangeInfo.SERIALIZER.getVersion()`). Is this intended?

lvyanquan: addressed it.

yuxiqian:
Thanks for @lvyanquan's rapid response! Ran migration tests and it works as expected.

* `TableChangeInfo#Serializer#magicTableId` should be named in screaming case `MAGIC_TABLE_ID` as a `static final` constant, or Spotless will complain.

* `MySqlPipelineITCase` needs update to reflect changes in default value:

```
Error:    MySqlPipelineITCase.testInitialStartupMode:228 
expected: CreateTableEvent{tableId=inventory_kohdit.products, schema=columns={`id` INT NOT NULL,`name` VARCHAR(255) NOT NULL,`description` VARCHAR(512),`weight` FLOAT}, primaryKeys=id, options=()}
 but was: CreateTableEvent{tableId=inventory_kohdit.products, schema=columns={`id` INT NOT NULL,`name` VARCHAR(255) NOT NULL 'flink',`description` VARCHAR(512),`weight` FLOAT}, primaryKeys=id, options=()}
...

yuxiqian (approved):
Looks good!

lvyanquan:
Rebased to master.

lvyanquan:
Done reorganize commit message.

Issue #2933, opened 2023-12-27
[Feature][Pipeline] Flink-CDC-Pipeline support database mapper and table add unified prefix or suffix


icchux:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Motivation

  When synchronize the entire MySQL database or multiple databases to Doris, In the current configuration, if I want to implement library name mapping, I need to specify all tables in the route parameter, which is quite cumbersome. So i want to specify a unified database name mapping and add a unified prefix or suffix to the table name.

### Solution

Add the mapper parameter and prioritize using the Route rule for replacement. If a table does not specify a Route rule, use the Mapper rule for replacement
```yaml
source:
  type: mysql
  hostname: localhost
  port: 3306
  username: root
  password: 123456
  tables: app_db.\.*
  server-id: 5400-5404
  server-time-zone: UTC

sink:
  type: doris
  fenodes: 127.0.0.1:8030
  username: root
  password: ""
  table.create.properties.light_schema_change: true
  table.create.properties.replication_num: 1
  
route:
  - source-table: app_db.orders
    sink-table: ods_db.ods_orders
  - source-table: app_db.shipments
    sink-table: ods_db.ods_shipments
  - source-table: app_db.products
    sink-table: ods_db.ods_products
  - source-database: app_db
    sink-database: ods_db
    table-prefix: ods_mapper_
    table-suffix: _di

pipeline:
  name: Sync MySQL Database to Doris
  parallelism: 2
```

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

tercel53:
I also have the same requirement, and I really appreciate this feature.

PatrickRen:
Closing this issue as it has been migrated to [Apache Jira](https://issues.apache.org/jira).

PR #2947, opened 2023-12-31
[feature][pipeline] Support pipeline route params add database routing feature #2933

icchux:
Add the mapper parameter and prioritize using the Route rule for replacement. If a table does not specify a table rule, use the database rule for replacement

### DIFF
--- a/flink-cdc-composer/src/main/java/com/ververica/cdc/composer/definition/SchemaRouteDef.java
+++ b/flink-cdc-composer/src/main/java/com/ververica/cdc/composer/definition/SchemaRouteDef.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2023 Ververica Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ververica.cdc.composer.definition;
+
+import javax.annotation.Nullable;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Definition of a router.
+ *
+ * <p>A mapper definition contains:
+ *
+ * <ul>
+ *   <li>sourceDataBase: sourceDataBase.
+ *   <li>sinkDataBase: sinkDataBase.
+ *   <li>tablePrefix: tablePrefix.
+ *   <li>tableSuffix: tableSuffix.
+ * </ul>
+ */
+public class SchemaRouteDef {
+    private final String sourceDatabase;
+    private final String sinkDatabase;
+    @Nullable private final String tablePrefix;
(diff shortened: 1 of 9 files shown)


flink-cdc-runtime/src/main/java/com/ververica/cdc/runtime/operators/route/RouteFunction.java
```diff
@@ -91,14 +107,36 @@ public Event map(Event event) throws Exception {
...
+                                + Optional.ofNullable(schemaRoute.f3).orElse("");
+                if (tableId.getNamespace() != null) {
+                    replaceIdentifier =
+                            String.join(
+                                    ".", tableId.getNamespace(), schemaRoute.f1, replaceTableName);
+                } else {
+                    replaceIdentifier = String.join(".", schemaRoute.f1, replaceTableName);
+                }
+
+                return recreateChangeEvent(changeEvent, TableId.parse(replaceIdentifier));
```
lvyanquan, line 136: We can use `new TableId(tableId.getSchemaName(), schemaRoute.f1, replaceTableName)` to simplify the `join` and `parse` code.

icchux: get. i will process it，but there only hava TableId.tableId() function can be use，it is ok?
```
// then schema route
        for (Tuple4<String, String, String, String> schemaRoute : schemaRoutes) {
            if (schemaRoute.f0.equals(tableId.getSchemaName())) {
                // database routing, table add prefix and suffix mapping
                TableId replaceBy;
                String replaceTableName =
                        Optional.ofNullable(schemaRoute.f2).orElse("")
                                + tableId.getTableName()
                                + Optional.ofNullable(schemaRoute.f3).orElse("");
                if (tableId.getNamespace() != null) {
                    replaceBy = TableId.tableId(tableId.getNamespace(),schemaRoute.f1, replaceTableName);
                } else {
                    replaceBy = TableId.tableId(schemaRoute.f1, replaceTableName);
                }

                return recreateChangeEvent(changeEvent, replaceBy);
            }
        }
```

lvyanquan: > but there only hava TableId.tableId() function can be use，it is ok?
Yeah, This will be more concise.

flink-cdc-composer/src/main/java/com/ververica/cdc/composer/definition/SchemaRouteDef.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public SchemaRouteDef(
+            String sourceDatabase,
+            String sinkDatabase,
+            @Nullable String tablePrefix,
+            @Nullable String tableSuffix) {
+        this.sourceDatabase = sourceDatabase;
+        this.sinkDatabase = sinkDatabase;
+        this.tablePrefix = tablePrefix;
+        this.tableSuffix = tableSuffix;
```
lvyanquan, line 50: Our pipeline will only contain source/transform/route/sink. Can we move these variables to RouteDef to make PipelineDef clearer. Just determine whether it is a table change or a database change based on a variable, and it can be extended to changes to the namespace.

lvyanquan:
There is another pr https://github.com/ververica/flink-cdc-connectors/pull/2908 provides a new idea, We can use placeholders to represent the original table name and database name, It‘s useful to add suffixes and suffixes while keeping 
 parameters as before.

leonardBang:
Closed by https://github.com/apache/flink-cdc/pull/3428

PR #2957, opened 2024-01-03
[hotfix] Fix NoSuchMethodError of CollectionUtil.newHashMapWithExpectedSize.

lvyanquan:
This closes https://github.com/ververica/flink-cdc-connectors/issues/2943.
It was verified using Flink 1.17.1, but e2e verification is still required.

### DIFF
--- a/flink-cdc-common/src/main/java/com/ververica/cdc/common/utils/InstantiationUtil.java
+++ b/flink-cdc-common/src/main/java/com/ververica/cdc/common/utils/InstantiationUtil.java
@@ -19,7 +19,6 @@ package com.ververica.cdc.common.utils;
 import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.core.memory.DataInputViewStreamWrapper;
 import org.apache.flink.core.memory.DataOutputViewStreamWrapper;
-import org.apache.flink.util.CollectionUtil;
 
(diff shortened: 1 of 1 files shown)


flink-cdc-common/src/main/java/com/ververica/cdc/common/utils/InstantiationUtil.java
```diff
@@ -19,7 +19,6 @@
 import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.core.memory.DataInputViewStreamWrapper;
 import org.apache.flink.core.memory.DataOutputViewStreamWrapper;
-import org.apache.flink.util.CollectionUtil;
```
leonardBang, line 22: Could you help check other reference usage in our code from flink utils package like `org.apache.flink.util.`? We'd better avoid similar issue.

lvyanquan: I've made a check and compared the changes from git log with the referenced methods, and found no other incompatible changes.

leonardBang: Thanks for the check !

leonardBang (approved):
LGTM

Issue #2910, opened 2023-12-21
Flink CDC3.0 Demo Mysql binlog to Starrocks Exception amount of AddColumnEvent is already existed[Bug] 


biyinggun:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.18

### Flink CDC version

3.0

### Database and its version

Starrocks  3.1.6

### Minimal reproduce step

I Can run the Doris Demo (mysq to doris)，But run Starrocks Demo Exception; For the Step:
1.Mysql Sync  to Starrocks database and table pass.
2. INSERT INTO app_db.orders (id, price) VALUES (3, 100.00); pass.
3. ALTER TABLE app_db.orders ADD amount varchar(100) NULL;  Flink CDC Exception :2023-12-21 11:40:16
java.lang.IllegalStateException: Failed to send request to coordinator: com.ververica.cdc.runtime.operators.schema.event.SchemaChangeRequest@ecbe495a .......Caused by: java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: amount of AddColumnEvent is already existed.
But Starrocks table: order.amount is added; 

### What did you expect to see?

2023-12-21 11:40:16
java.lang.IllegalStateException: Failed to send request to coordinator: com.ververica.cdc.runtime.operators.schema.event.SchemaChangeRequest@ecbe495a
	at com.ververica.cdc.runtime.operators.schema.SchemaOperator.sendRequestToCoordinator(SchemaOperator.java:126)
	at com.ververica.cdc.runtime.operators.schema.SchemaOperator.requestSchemaChange(SchemaOperator.java:110)
	at com.ververica.cdc.runtime.operators.schema.SchemaOperator.handleSchemaChangeEvent(SchemaOperator.java:95)
	at com.ververica.cdc.runtime.operators.schema.SchemaOperator.processElement(SchemaOperator.java:85)
	at org.apache.flink.streaming.runtime.tasks.CopyingChainingOutput.pushToOperator(CopyingChainingOutput.java:75)
	at org.apache.flink.streaming.runtime.tasks.CopyingChainingOutput.collect(CopyingChainingOutput.java:50)
	at org.apache.flink.streaming.runtime.tasks.CopyingChainingOutput.collect(CopyingChainingOutput.java:29)
[...]

PR #2982, opened 2024-01-09
[cdc-common] AddColumnEvent is already existed

fanqiejiang8:
Prepare a Pull Request
Fixes[cdc-common] AddColumnEvent is already existed  #2910 
Motivation
Incorrect judgment that column names are equal

Modifications
Iterate over the column information, then compare the columns to see if they are equal and throw an exception if they are


whhe:
@lvyanquan Could you please help to review this pr?

PR #2986, opened 2024-01-10
[hotfix][cdc-connector][mysql] skip SchemaChangeEvents that were not included in capturedTableFilter.

lvyanquan:
This closes https://github.com/ververica/flink-cdc-connectors/issues/2981.


lvyanquan:
@ruanhang1993 @Jiabao-Sun PTAL.

lvyanquan:
Met some failures in CI, will try to fix it.

lvyanquan:
@leonardBang CC.

leonardBang (approved):
Thanks @lvyanquan for the fix, LGTM

steveGuRen:
@joyCurry30 @leonardBang  Logically, it is considered a bug when I have to read all tables in the schema, and the first execution of "alter table sql" is not triggered. However, since TableDiscoveryUtils loads all tables, it works. But in a SaaS environment, with so many tables(five hundred thousand, a tenant has their own tables), this will cause issues. Because we can't load all tables when have to discover the new Table with query sql by "show create sql " for every table.
So I fix the TableDiscoveryUtils so that just load a table in start or ddl change, and shouEmit logic updates too

steveGuRen:
In the design of TiDB, I find that they take into account the scenario of having a large number of tables in a SaaS environment. Therefore, special attention is given to the performance of DDL operations in such contexts.

PR #3028, opened 2024-01-23
[FLINK-34180] Migrate doc website from ververica to flink

GOODBOY008:
[FLINK-34180] Migrate doc website from ververica to flink.


.dlc.json
```diff
@@ -0,0 +1,32 @@
+{
```
PatrickRen, line 1: What's the purpose of this file?

PatrickRen (commented):
@GOODBOY008 Thanks for the awesome work! I left one comment.

.dlc.json
GOODBOY008: This file I would like to check dead link in doc

PatrickRen: What about we introduce this file after we enable the dead link check in the future? I'm not sure links in this file currently are related to Flink CDC's doc.

PatrickRen (approved):
@GOODBOY008 Thanks for the PR! LGTM overall. Let's merge first and see if the doc can be built successfully.

Issue #3035, opened 2024-01-26
[Bug] Meet NoClassDefFoundError when add new table in mysql cdc source


pengmide:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.18

### Flink CDC version

3.0.1

### Database and its version

source: mysql: 5.7
sink: file system

### Minimal reproduce step

1. create mysql cdc flink job
2. add new table in mysql source

 meet exception:

java.lang.RuntimeException: One or more fetchers have encountered exception
	at org.apache.flink.connector.base.source.reader.fetcher.SplitFetcherManager.checkErrors(SplitFetcherManager.java:263)
	at org.apache.flink.connector.base.source.reader.SourceReaderBase.getNextFetch(SourceReaderBase.java:185)
	at org.apache.flink.connector.base.source.reader.SourceReaderBase.pollNext(SourceReaderBase.java:147)
	at org.apache.flink.streaming.api.operators.SourceOperator.emitNext(SourceOperator.java:419)
	at org.apache.flink.streaming.runtime.io.StreamTaskSourceInput.emitNext(StreamTaskSourceInput.java:68)
	at org.apache.flink.streaming.runtime.io.StreamOneInputProcessor.processInput(StreamOneInputProcessor.java:65)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.processInput(StreamTask.java:562)
	at org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.runMailboxLoop(MailboxProcessor.java:231)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.runMailboxLoop(StreamTask.java:858)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:807)
	at org.apache.flink.runtime.taskmanager.Task.runWithSystemExitMonitoring(Task.java:953)
	at org.apache.flink.runtime.taskmanager.Task.restoreAndInvoke(Task.java:932)
	at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:746)
	at org.apache.flink.runtime.taskmanager.Task.run(Task.java:562)
	at java.lang.Thread.run(Thread.java:750)
Caused by: java.lang.NoClassDefFoundError: com/ververica/cdc/common/utils/StringUtils
[...]

PR #3036, opened 2024-01-26
[fix][cdc-connector][mysql] Fix NoClassDefFoundError when create new table in mysql cdc source

pengmide:
### Prepare a Pull Request

Fixes: #3035 

### Motivation
When we package maven jar for flink-cdc-connector-mysql-cdc module, the flink-cdc-common moduel is excluded.
So we can use org.apache.commons.lang3.StringUtils instead of com.ververica.cdc.common.utils.StringUtils and then mysql CDC run correctly when adding new source tables.

Modifications

Change the import of StringUtils from com.ververica.cdc.common.utils to org.apache.commons.lang3


loserwang1024:
Maybe we should includes the flink-cdc-common moduel in flink-cdc-connector-mysql-cdc module? flink-cdc-connector-mysql-cdc also includes some class of  flink-cdc-common, such as annotion?But it also seems that flink-cdc-common is too big for flink-cdc-connector-mysql-cdc.

meicao2999:
I have also encountered Does NoClassDefFoundError. Is there a specific class

yuxiqian:
Hi @pengmide, could you please rebase this PR with latest `master` branch before it could be merged? Thanks!

Issue #2217, opened 2023-06-15
[Bug] Caused by: com.github.shyiko.mysql.binlog.event.deserialization.MissingTableMapEventException and lost data


dylenWu:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.14.6

### Flink CDC version

2.3

### Database and its version

mysql 

### Minimal reproduce step

Caused by: com.github.shyiko.mysql.binlog.event.deserialization.MissingTableMapEventException: No TableMapEventData has been found for table id:15987. Usually that means that you have started reading binary log 'within the logical event group' (e.g. from WRITE_ROWS and not proceeding TABLE_MAP

### What did you expect to see?

flink job can start from checkpoint

### What did you see instead?

no restart and loss data

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

dylenWu:
please aasign this issus to me @leonardBang

dylenWu:
I found the reason, mainly because our binglog client was restarted during the consumption process, but the location of the restart was on the binlog of Write_rows type. As a result, when the binlog was deserialized, the corresponding table of the binlog data could not be found. Information. Because when parsing the data of the binlog type of the Write_rows type, the data of the table_map type must be parsed first.

dylenWu:
我们可以从下面这部分代码看出问题，当出现MissingTableMapEventException这种报错的时候，程序不会抛出异常，而是继续执行，导致过程中Write_rows类型的数据被丢弃，导致数据丢失

leonardBang:
> please aasign this issus to me @leonardBang

Assigned to you @dylenWu, please keep going, looking forward your analysis.

dylenWu:
> 
I think we should add this logic to this method to let the client stop receiving messages, so that the queue poll() method can throw an exception
![image](https://github.com/ververica/flink-cdc-connectors/assets/22993360/64189850-5445-407c-b896-a42b39b5976a)
please review it @leonardBang

qidian99:
@dylenWu The option to just ignore the exception is useful. I encountered a weird case where consuming from a later offset will throw the exception that the binlog file does not exist. I tried to manually specify a earlier offset, and the error showed up. In my case I'm relying on the upsert semanitcs of sink operators so having some data duplication is tolerable.

dylenWu:
[...]

PR #3065, opened 2024-02-13
[MySQL] Fix for DBZ-7037 and 2217 to not clear tableMapEventByTableId for fake Rotates Event

shikai93:
Event log  
`Caused by: com.github.shyiko.mysql.binlog.event.deserialization.MissingTableMapEventException`
would occur as explained in issue #2217  whereby client is restarted during the consumption process, leading to a Rotate Event that clears the tableMapEventByTableId. 

This has been fixed in [DBZ](https://github.com/debezium/debezium/pull/4959) discussing the same issue. This PR tries to port that fix over since Flink CDC clones the MySqlStreamingChangeEventSource.


PatrickRen (approved):
@shikai93 Thanks for the patch! LGTM.

PR #3083, opened 2024-02-27
[FLINK-35071][cdc-connector][cdc-base] Shade guava31 to avoid dependency conflict with flink below 1.18

loserwang1024:
As shown in https://github.com/ververica/flink-cdc-connectors/issues/3071, 
Unlike flink-table-common、flink-connector-base 、flink-core whose maven scope is provided , flink-shaded-guava and flink-shaded-force-shading will be included in the final jar package,  so maybe cause dependency conflict.

Now we can see why dependency conflict occurs:

- Since Flink 1.18, version of guava is upgrade from 30 to 31 (using 31.1-jre-17.0)

image

- Since CDC 3.0.1, version of guava is also upgrade from 30 to 31 (using 31.1-jre-17.0)

- So CDC 3.0.1 is compatible with Flink 1.18, CDC 2.x is is compatible with Flink 1.13-1.17. CDC 2.x and Flink 1.18 dependency conflict, CDC 3.0.1 and Flink 1.17(or earlier version) dependency conflict.
(https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#relocations) in pom.xml.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -284,11 +284,6 @@ limitations under the License.
             <version>${slf4j.version}</version>
         </dependency>
 
-        <dependency>
-            <groupId>org.apache.flink</groupId>
-            <artifactId>flink-shaded-guava</artifactId>
-            <version>31.1-jre-${flink.shaded.version}</version>
-        </dependency>
 
         <dependency>
             <groupId>org.apache.flink</groupId>
--- a/docs/content.zh/docs/connectors/flink-sources/datastream-api-package-guidance.md
+++ b/docs/content.zh/docs/connectors/flink-sources/datastream-api-package-guidance.md
@@ -104,18 +104,6 @@ This guide provides a simple `pom.xml` example for packaging DataStream job JARs
             <version>${flink.version}</version>
             <scope>provided</scope>
         </dependency>
-        <!-- Checked the dependencies of the Flink project and below is a feasible reference. -->
-        <!--  Use flink shaded guava  18.0-13.0 for flink 1.13   -->
-        <!--  Use flink shaded guava  30.1.1-jre-14.0 for flink-1.14  -->
-        <!--  Use flink shaded guava  30.1.1-jre-15.0 for flink-1.15  -->
-        <!--  Use flink shaded guava  30.1.1-jre-15.0 for flink-1.16  -->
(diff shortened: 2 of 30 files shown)


whhe (approved):

pom.xml
```diff
@@ -462,8 +462,15 @@ under the License.
                                         submodules, ${flink.version} will be resolved as the actual Flink version.
                                     -->
                                     <include>org.apache.flink:flink-shaded-force-shading</include>
+                                    <include>org.apache.flink:flink-shaded-guava</include>
                                 </includes>
                             </artifactSet>
+                            <relocations>
+                                <relocation>
+                                    <pattern>flink.shaded.guava</pattern>
```
PatrickRen, line 470: I'm not an expert of `maven-shade-plugin`, but I assume that the pattern should start with `org.apache.flink`. Please correct me if I'm wrong.

PatrickRen (commented):
@loserwang1024 Thanks for the patch! I left a comment.

pom.xml
loserwang1024: Done it.(By the way, I used to use `com.google.guava` rather than flink's shaded guava, but there are lots of limit in `checkstyle.xml`):
```xml
<module name="Regexp">
		<property name="format" value="import com\.google\.common\.base\.Preconditions"/>
		<property name="illegalPattern" value="true"/>
		<property name="message"
			value="Use Flink's Preconditions instead of Guava's Preconditions"/>
</module>
```
or

```xml
		<module name="IllegalImport">
			<property name="illegalPkgs" value="com.google.common"/>
			<message key="import.illegal" value="{0}; Use flink-shaded-guava instead."/>
		</module>

```

PatrickRen: Thanks for the update! You are correct. Depending on Guava directly is prohibited in checkstyle to avoid using non-relocated guava by mistake.

loserwang1024:
@PatrickRen , CC?

yuxiqian:
Seems there's some conflicts between `master` branch, could @loserwang1024 please rebase it when you're available?

leonardBang:
The CI failed, could you fix it ? @loserwang1024

PR #3098, opened 2024-03-05
[MYSQL] support map long type in mysql ddl

TJX2014:


PatrickRen:
@TJX2014 I can't find the LONG type definition in MySQL documentation, so I'm closing this for now. If you could provide more information for this one, please create an issue on Apache Jira and we can discuss it further there.

PR #3100, opened 2024-03-05
[Bug][pipeline]Set default values for the parallelism.

lightzhao:
Set default values for the parallelism. parameter to prevent NPE caused by not setting it.
https://github.com/apache/flink-cdc/blob/96888b2ce0a7981ebe5917b6c27deb4015d845d2/flink-cdc-composer/src/main/java/com/ververica/cdc/composer/flink/FlinkPipelineComposer.java#L95
https://github.com/apache/flink-cdc/blob/96888b2ce0a7981ebe5917b6c27deb4015d845d2/flink-cdc-composer/src/main/java/com/ververica/cdc/composer/flink/translator/DataSourceTranslator.java#L65
image


lightzhao:
@leonardBang PTAL.

PatrickRen:
@lightzhao The default parallelism should have been set to 4 in `flink-cdc-bin/conf/flink-cdc.yaml`

PR #3101, opened 2024-03-05
[Improve][pipeline]Support set checkpoint.

lightzhao:
Support set checkpoint.


lightzhao:
@leonardBang PTAL.

PatrickRen:
#3187 will provide ability for setting flink configs in pipeline YAML, which should already cover this one

Issue #3105, opened 2024-03-06
The Document URL in README.md file  was invalid


hnbian:
<img width="957" alt="image" src="https://github.com/apache/flink-cdc/assets/12964828/fde93c8c-76c0-4fa0-9f31-7d3fd9977c2b">
<img width="1058" alt="image" src="https://github.com/apache/flink-cdc/assets/12964828/e56f3719-1877-4939-85f3-c5c4f489af43">
<img width="903" alt="image" src="https://github.com/apache/flink-cdc/assets/12964828/3811f690-a815-4004-8824-b7cce3a75bf2">

hnbian:
https://github.com/apache/flink-cdc/pull/3106

PatrickRen:
New documentation website is available at https://nightlies.apache.org/flink/flink-cdc-docs-stable/

PR #3106, opened 2024-03-06
[docs]Solve the problem of invalid document URL address in README.md file

hnbian:
https://github.com/apache/flink-cdc/issues/3105

Solve the problem of invalid document URL address in README.md file
Modify the URL in the file that points to the document address


PatrickRen:
@hnbian Thanks for the PR! We've refactored the entire README so the patch won't be valid anymore. Feel free to create issue and PRs if you find more

PR #3118, opened 2024-03-09
[hotfix][cdc][build] Include cdc-common module when shaded for module flink-sql-connector-mysql-cdc

lvyanquan:
`flink-cdc-source-connectors` module is depended on `flink-cdc-common`, but `flink-sql-connector-mysql-cdc` is not included `flink-cdc-common` when packaging.


lvyanquan:
@leonardBang @Jiabao-Sun PTAL.

Jiabao-Sun (approved):
Thanks @lvyanquan, LGTM.
We should change the groupId as well when the project's groupId changed from `com.ververica` to `org.apache.flink`

PR #2315, opened 2023-07-18
[Oracle][MySQL][SqlServer][PostgresSQL] Fix Oracle/MySQL/SqlServer/PostgresSQL CDC parser schema change event failed

e-mhui:
Fix: #2055 #1792 #1379 #1807 #1810 #1803 #1804 #2311 

When a SchemaChangeEvent is triggered, the `schemaChangeRecordValue` will obtain binlog information from **sourceInfo**. However, only MySQL CDC has binlog information, and other CDC will report an error after triggering a SchemaChangeEvent event.

image

So I provided an interface for each connector to implement their own sourceinfo parsing.

```java
/** SchemaChangeEvent Handler. */
public interface SchemaChangeEventHandler {
    Map<String, Object> parseSource(SchemaChangeEvent event);
}
```

Then, it's used in JdbcSourceEventDispatcher#schemaChangeRecordValue.

```java
        private Struct schemaChangeRecordValue(SchemaChangeEvent event) throws IOException {
            Map<String, Object> source = schemaChangeEventHandler.parseSource(event);
            HistoryRecord historyRecord =
                    new HistoryRecord(
                            source,
                            event.getOffset(),
                            event.getDatabase(),
                            null,
                            event.getSchema(),
                            event.getDdl(),
                            event.getTableChanges());
        // ... other code ... 
```

For MySQL CDC

```java
/** MySqlSchemaChangeEventHandler to deal MySql's schema change event. */
public class MySqlSchemaChangeEventHandler implements SchemaChangeEventHandler {

    @Override
    public Map<String, Object> parseSource(SchemaChangeEvent event) {
        Map<String, Object> source = new HashMap<>();
        Struct sourceInfo = event.getSource();
        String fileName = sourceInfo.getString(BINLOG_FILENAME_OFFSET_KEY);
        Long pos = sourceInfo.getInt64(BINLOG_POSITION_OFFSET_KEY);
        Long serverId = sourceInfo.getInt64(SERVER_ID_KEY);
        source.put(SERVER_ID_KEY, serverId);
        source.put(BINLOG_FILENAME_OFFSET_KEY, fileName);
        source.put(BINLOG_POSITION_OFFSET_KEY, pos);
        return source;
    }
}
```

For Oracle CDC

```java
/** OracleSchemaChangeEventHandler to deal Oracle's schema change event. */
public class OracleSchemaChangeEventHandler implements SchemaChangeEventHandler {

    @Override
    public Map<String, Object> parseSource(SchemaChangeEvent event) {
[...]

PR #3151, opened 2024-03-15
Update oracle-cdc.md

stardustman:
fix error the "log.mining.strategy" should be start with prefix "debezium." according to https://nightlies.apache.org/flink/flink-cdc-docs-release-3.0/zh/docs/connectors/cdc-connectors/oracle-cdc/#connector-options and the help from @e-mhui with the link https://github.com/apache/flink-cdc/pull/2315#issuecomment-1999166980


e-mhui:
If you are using Flink SQL, you need to add "debezium" as a prefix. However, when using the DataStream API, you don't need the "debezium" prefix.

PR #3166, opened 2024-03-19
[FLINK-34681][cdc][doc] add standalone.md under deployment directory.

lvyanquan:
add standalone.md

### DIFF
--- a/docs/content/docs/deployment/standalone.md
+++ b/docs/content/docs/deployment/standalone.md
@@ -23,3 +23,109 @@ KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
 under the License.
 -->
+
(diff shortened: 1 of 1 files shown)


loserwang1024 (commented):
Please also add standalone.md which is an empty file in content.zh

lvyanquan:
It has already been added https://github.com/apache/flink-cdc/blob/master/docs/content.zh/docs/deployment/standalone.md

lvyanquan:
> Please also add standalone.md which is an empty file in content.zh

I will create a following PR to support Chinese language with this document and the previously merged documents.

PatrickRen (approved):
@lvyanquan Thanks for the PR! LGTM.

Issue #2854, opened 2023-12-12
[cdc-pipeline] Apply 'flink-confing' config for pipeline.


joyCurry30:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Motivation

Our current task is to submit to the Flink cluster. Currently, the Flink configuration can only be read from 'FLINK_HOME' in the flink-conf.yaml file. This limitation is too restrictive, as it does not allow for flexible configuration of individual pipeline settings.

### Solution

Therefore, I suggest adding a configuration item, "flink-config," to the pipeline YAML configuration file. This will be used to configure Flink settings for individual pipeline.
The values in the "flink-config" configuration item will override those in flink-conf.yaml.

The 'flink-config' configuration items are as follows:
```yaml
flink-config:
  execution.checkpointing.interval: 30000
  execution.checkpointing.mode: EXACTLY_ONCE
```

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PatrickRen:
Closing this issue as it has been migrated to [Apache Jira](https://issues.apache.org/jira).

PR #3187, opened 2024-03-22
[cdc-cli][cdc-composer] Applying 'flink-config' for pipeline yaml

joyCurry30:
Refer to https://github.com/apache/flink-cdc/issues/2854
Applying '[flink-config](https://issues.apache.org/jira/browse/FLINK-config)' for pipeline yaml, which will overwrite the config from '[flink-conf](https://issues.apache.org/jira/browse/FLINK-conf).yaml'.

### DIFF
--- a/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliExecutor.java
+++ b/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliExecutor.java
@@ -60,7 +60,7 @@ public class CliExecutor {
                 pipelineDefinitionParser.parse(pipelineDefPath, globalPipelineConfig);
 
         // Create composer
-        PipelineComposer composer = getComposer(flinkConfig);
+        PipelineComposer composer = getComposer(flinkConfig, pipelineDef.getConfig());
 
         // Compose pipeline
         PipelineExecution execution = composer.compose(pipelineDef);
@@ -69,10 +69,10 @@ public class CliExecutor {
         return execution.execute();
     }
 
-    private PipelineComposer getComposer(Configuration flinkConfig) {
+    private PipelineComposer getComposer(Configuration flinkConfig, Configuration pipelineConfig) {
         if (composer == null) {
             return FlinkEnvironmentUtils.createComposer(
-                    useMiniCluster, flinkConfig, additionalJars);
+                    useMiniCluster, flinkConfig, pipelineConfig, additionalJars);
         }
         return composer;
     }
--- a/flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml
+++ b/flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml
@@ -53,3 +53,8 @@ pipeline:
   name: source-database-sync-pipe
   parallelism: 4
   enable-schema-evolution: false
+  execution.checkpointing.interval: 10000
+  execution.checkpointing.mode: EXACTLY_ONCE
+  # yarn config
+  yarn.staging-directory: /tmp/flink-cdc
+  yarn.application.queue: flink-cdc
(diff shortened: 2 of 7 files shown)


flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliExecutor.java
```diff
@@ -69,10 +69,10 @@ public PipelineExecution.ExecutionInfo run() throws Exception {
         return execution.execute();
     }
 
-    private PipelineComposer getComposer(Configuration flinkConfig) {
+    private PipelineComposer getComposer(Configuration flinkConfig, Configuration pipelineConfig) {
```
lvyanquan, line 72: Can we set flink config in PipelineComposer.compose https://github.com/apache/flink-cdc/blob/1c69a650e3187734962e824e40ed35de0d94ccfc/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java#L96 method to avoid modifying these constructor?

flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml
```diff
@@ -53,3 +53,8 @@ pipeline:
   name: source-database-sync-pipe
   parallelism: 4
   enable-schema-evolution: false
+  execution.checkpointing.interval: 10000
+  execution.checkpointing.mode: EXACTLY_ONCE
+  # yarn config
+  yarn.staging-directory: /tmp/flink-cdc
+  yarn.application.queue: flink-cdc
```
PatrickRen, line 60: Flink 1.19 has a significant change to the format of configuration (see [FLIP-366](https://cwiki.apache.org/confluence/display/FLINK/FLIP-366%3A+Support+standard+YAML+for+FLINK+configuration)), so actually I have some concerns of this functionality now. If we still use the legacy flattened format, there will be a gap between us and the Flink. 

Another thing in my mind is that how to resolve conflicts between pipeline configs and Flink configs. There are two level of conflict:

1. Naming conflict: what if a pipeline config has the same key as Flink but with totally different meaning
2. Semantic conflict: `parallelism` in pipeline config v.s. `parallelism.default` in Flink config. Which one has higher priority and which one to choose for users?

If we mix them together in the `pipeline` section, I'm afraid there might be a lot of explanations and conflict resolving works to do in the future.

PatrickRen (commented):
@joyCurry30 Thanks for the contribution! I commented with my concerns about introducing this one

flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml
joyCurry30: @PatrickRen Thanks for your comment. Actually, we can use standard YAML for Flink configuration by adding a YAML config translator. As for the conflict between pipeline configs and Flink configs, how about add a new ref, like 'env-config', for flink configs, which applying for flink execution env.

PatrickRen: I'm afraid this will make the YAML even more complex, with so many nested layers. And the name `env-config` is quite vague. Currently we read all Flink configurations from `FLINK_HOME/conf/flink-conf.yaml(config.yaml)`. What about letting users set Flink configurations directly there?

joyCurry30: I don't think this is a good idea. We might need to configure different Flink configurations for different tasks, and provide users with an entry point to adjust Flink configuration based on their tasks, such as checkpointing or task manager memory. If we use the same 'flink-conf.yaml' for every task, it will become very inflexible.

PR #3160, opened 2024-03-18
[FLINK-34680][cdc] Add overview page for CDC connectors and optimize connector docs

PatrickRen:
This pull request adds overview page for CDC connectors and optimizes connector docs.

PR #2681, opened 2023-11-10
[common] Introduce Configuration and related utilities

PatrickRen:
This pull request introduces `Configuration` and its related utilities such as `ConfigOption`, and migrate current usage of Flink `Configuration` to CDC's own one in `flink-cdc-composer` and `flink-cdc-cli`.

Please note that most code blocks are borrowed from Flink, and some deprecated and redundant methods are dropped.

PR #3194, opened 2024-03-25
[FLINK-34969][cdc-cli]Add support for both new and old Flink config files in Flink…

skymilong:
This commit introduces a simple logic check within the Flink CDC codebase to enable support for both the new (`config.yaml`) and old (`flink-conf.yaml`) Flink configuration file standards. By utilizing the `Files.exists` method, Flink CDC now first checks for the presence of `config.yaml` in the system. If `config.yaml` is found, it is used as the configuration file. Otherwise, Flink CDC falls back to using `flink-conf.yaml`. This enhancement ensures that Flink CDC can seamlessly transition between different versions of Flink, maintaining backward compatibility and facilitating easier upgrades for users.


PatrickRen:
@skymilong Thanks for the PR! According to the code contribution rule of Apache Flink, could you create an issue on [Jira](https://issues.apache.org/jira) for this and include the Jira ID in the PR title and commit message? You can take #3160 as an example. 

And the PR has code formatting issue. Please run `mvn spotless:apply` before pushing your commits. 

Thanks!

skymilong:
> @skymilong Thanks for the PR! According to the code contribution rule of Apache Flink, could you create an issue on [Jira](https://issues.apache.org/jira) for this and include the Jira ID in the PR title and commit message? You can take #3160 as an example.
> 
> And the PR has code formatting issue. Please run `mvn spotless:apply` before pushing your commits.
> 
> Thanks!

Thanks a bunch for your help. This is my first rodeo contributing code on GitHub, so I'm still trying to get the hang of things. I'll whip up a Jira issue like you suggested and pop the ID into the PR and commit messages. Also, I'll iron out those code formatting kinks. I'll get on it pronto and get the PR updated.
Appreciate your patience and all your help.

PatrickRen:
@skymilong Welcome to the community! Feel free to ask any questions

laglangyue (approved):
LGTM

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/utils/FlinkEnvironmentUtils.java
```diff
@@ -20,17 +20,24 @@
...
 
     public static Configuration loadFlinkConfiguration(Path flinkHome) throws Exception {
-        Path flinkConfPath = flinkHome.resolve(FLINK_CONF_DIR).resolve(FLINK_CONF_FILENAME);
+        Path flinkConfPath = flinkHome.resolve(FLINK_CONF_DIR).resolve(OLD_FLINK_CONF_FILENAME);
+        // If the old version of the configuration file does not exist, then attempt to use the new
+        // version of the file name.
+        if (!Files.exists(flinkConfPath)) {
+            flinkConfPath = flinkHome.resolve(FLINK_CONF_DIR).resolve(NEW_FLINK_CONF_FILENAME);
+        }
         return ConfigurationUtils.loadMapFormattedConfig(flinkConfPath);
```
PatrickRen, line 41: As the new `config.yaml` is not map-formatted anymore, I think we need to rewrite the parsing logic here. `ConfigurationUtils.loadMapFormattedConfig` assumes all configs are in flatten key:value format.

PatrickRen (changes requested):
@skymilong Thanks for the update! I left a comment.

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/utils/FlinkEnvironmentUtils.java
skymilong: Thank you for your feedback. I have rechecked the code and I agree with you. The new config.yaml is not in a map format anymore, so rewriting the parsing logic would be necessary. I am willing to take on this task and submit the necessary changes.

In addition, I think it might be beneficial to handle the old version of the configuration file through exception handling. Here is a possible way to do so:

```java
private static final String OLD_FLINK_CONF_FILENAME = "flink-conf.yaml";
private static final String FLINK_CONF_FILENAME = "config.yaml";
public static Configuration loadFlinkConfiguration(Path flinkHome) throws Exception {
    Path flinkConfPath = flinkHome.resolve(FLINK_CONF_DIR).resolve(FLINK_CONF_FILENAME);
    try {
        return ConfigurationUtils.loadMapFormattedConfig(flinkConfPath);
    } catch (Exception e) {
        LOG.warn(
                "Failed to load the new configuration file:{}. Trying to load the old configuration file:{}.",
                FLINK_CONF_FILENAME,
                OLD_FLINK_CONF_FILENAME);
        return ConfigurationUtils.loadMapFormattedConfig(
                flinkHome.resolve(FLINK_CONF_DIR).resolve(OLD_FLINK_CONF_FILENAME));
    }
}
```
This is my idea for exception handling. We first attempt to load the new configuration file. If an exception occurs (which could be due to the absence of the new configuration file), we then try to load the old configuration file. This approach not only ensures backward compatibility and allows for a smoother transition between different versions of Flink, but also results in two log notifications from the `loadMapFormattedConfig` method. I think this will make the notifications more user-friendly. Do you think this approach is feasible? Or do you have any better suggestions or ideas? I look forward to your feedback.

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/utils/FlinkEnvironmentUtils.java
```diff
@@ -20,18 +20,33 @@
...
     private static final String FLINK_CONF_DIR = "conf";
-    private static final String FLINK_CONF_FILENAME = "flink-conf.yaml";
+    private static final String OLD_FLINK_CONF_FILENAME = "flink-conf.yaml";
+    private static final String FLINK_CONF_FILENAME = "config.yaml";
 
     public static Configuration loadFlinkConfiguration(Path flinkHome) throws Exception {
         Path flinkConfPath = flinkHome.resolve(FLINK_CONF_DIR).resolve(FLINK_CONF_FILENAME);
-        return ConfigurationUtils.loadMapFormattedConfig(flinkConfPath);
+        try {
+            return ConfigurationUtils.loadMapFormattedConfig(flinkConfPath);
```
PatrickRen, line 41: What about renaming the method to `ConfigurationUtils#loadConfigFile`? This method doesn't process map-formatted config (`flink-conf.yaml`) only.

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/utils/ConfigurationUtils.java
```diff
@@ -37,15 +38,34 @@ public static Configuration loadMapFormattedConfig(Path configPath) throws Excep
...
                             "Failed to load config file \"%s\" to key-value pairs", configPath),
                     e);
         }
     }
+
+    private static Map<String, String> flattenConfigMap(Map<String, Object> configMap) {
+        Map<String, String> result = new HashMap<>();
+        flattenConfigMapHelper(configMap, "", result);
+        return result;
+    }
```
PatrickRen, line 57: This looks like just a very simple wrapper around `flattenConfigMapHelper`. We can just merge the logic into `flattenConfigMap` instead of having another helper.

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/utils/ConfigurationUtils.java
```diff
@@ -37,15 +38,34 @@ public static Configuration loadMapFormattedConfig(Path configPath) throws Excep
...
+
+    private static void flattenConfigMapHelper(
+            Map<String, Object> configMap, String currentPath, Map<String, String> result) {
+        for (Map.Entry<String, Object> entry : configMap.entrySet()) {
+            String updatedPath =
+                    currentPath.isEmpty() ? entry.getKey() : currentPath + "." + entry.getKey();
+            if (entry.getValue() instanceof Map) {
+                flattenConfigMapHelper((Map<String, Object>) entry.getValue(), updatedPath, result);
+            } else {
```
PatrickRen, line 66: There should be a case for handling `List` type, according to https://github.com/apache/flink/blob/5bbcf8de79ce1979412879b919299ffa5a9b62fe/flink-core/src/main/java/org/apache/flink/configuration/GlobalConfiguration.java#L301-L307

PatrickRen (commented):
@skymilong Thanks for the update! I left some comments.

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/utils/ConfigurationUtils.java
skymilong: > There should be a case for handling `List` type, according to https://github.com/apache/flink/blob/5bbcf8de79ce1979412879b919299ffa5a9b62fe/flink-core/src/main/java/org/apache/flink/configuration/GlobalConfiguration.java#L301-L307

After reviewing this section of code and the previous PR https://github.com/apache/flink-cdc/pull/2681, I find myself a bit uncertain. Should we directly reference the relevant code from Flink, or should we copy over the YamlParserUtils#toYAMLString part of the code as done in PR https://github.com/apache/flink-cdc/pull/2681? I would greatly appreciate your advice on this.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/MemorySize.java
```diff
@@ -0,0 +1,421 @@
...
+/**
+ * MemorySize is a representation of a number of bytes, viewable in different units.
+ *
+ * <h2>Parsing</h2>
+ *
+ * <p>The size can be parsed from a text expression. If the expression is a pure number, the value
+ * will be interpreted as bytes.
+ */
+@PublicEvolving
+public class MemorySize implements java.io.Serializable, Comparable<MemorySize> {
```
PatrickRen, line 43: I think for `MemorySize` we should use the one the Flink instead of creating our own. It is a public API (marked as `@PublicEvolving`), and Flink CDC doesn't use this one in our production code. IIUC it is just used for parsing memory size expressions in configuration.

PatrickRen (commented):
@skymilong Thanks for the update! I left a comment.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/MemorySize.java
skymilong: > I think for `MemorySize` we should use the one the Flink instead of creating our own. It is a public API (marked as `@PublicEvolving`), and Flink CDC doesn't use this one in our production code. IIUC it is just used for parsing memory size expressions in configuration.

I apologize for this oversight, and thank you for your patience.

PatrickRen:
@skymilong Thanks for the update! Could you rebase the latest master branch? I just merged a commit for fixing unstable CI tests

PatrickRen (approved):
@skymilong Thanks for the update! LGTM

PR #3210, opened 2024-04-08
[minor][cdc-connector][mongo] Fix warning during MongoDB tests

yuxiqian:
Currently, when running `MongoDBStreamSplitReaderTest` and `MongoDBSnapshotSplitReaderTest`, tons of warning messages were be printed:

```
Warning: NumberLong: specifying a number as argument is deprecated and may lead to loss of precision, pass a string instead
... (repeated lines)
```

Console output

The problem lies in `chunk_test.js:22`, where `NumberLong` was called with a plain number instead of string, which has been deprecated by MongoDB engine due to precision loss concerns, though here `i` would never be greater than `20480`.

Converting it to string before passing it to `NumberLong` should fix this problem.


morazow (approved):
Looks good!

yuxiqian:
@PatrickRen Could you please take a look when available? Applying this patch could make mongo-cdc logs much clearer.

PR #3215, opened 2024-04-10
[FLINK-35072][cdc][doris] Support applying compatible `AlterColumnTypeEvent` to Doris sink

yuxiqian:
According to [Doris documentation](https://doris.apache.org/docs/sql-manual/sql-reference/Data-Definition-Statements/Alter/ALTER-TABLE-COLUMN/), altering column types dynamically is supported (via `ALTER TABLE ... MODIFY COLUMN` statement) when lossless conversion is available. However, now Doris pipeline connector has no support for any `AlterColumnTypeEvent`s, and raises RuntimeException all the time.

This PR adds `alterColumn` method which is not provided by Doris' `SchemaChangeManager` to apply `AlterColumnTypeEvent` to Doris sink. Only lossless conversions that were accepted by Doris will be allowed, and a `RuntimeException` will be raised for any incompatible narrowing casts.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/sink/DorisSchemaChangeManager.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/sink/DorisSchemaChangeManager.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.connectors.doris.sink;
+
+import org.apache.doris.flink.catalog.doris.FieldSchema;
+import org.apache.doris.flink.cfg.DorisOptions;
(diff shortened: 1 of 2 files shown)


yuxiqian:
I'm not familiar with Doris, could @lvyanquan @JNSimba take a look?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/sink/DorisSchemaChangeManager.java
```diff
@@ -0,0 +1,62 @@
...
+                        MODIFY_COLUMN_DDL,
+                        tableIdentifier,
+                        columnName,
+                        alterFieldSchema.getTypeString());
+
+        try {
+            return this.schemaChange(
+                    database, table, buildRequestParam(true, columnName), alterColumnDDL);
+        } catch (RuntimeException ex) {
+            if (ex.getMessage().contains("Nothing is changed. please check your alter stmt.")) {
```
lvyanquan, line 52: Considering future maintenance, could you please add some description about the source code of this error message? and is there a method to determine through exception class.

yuxiqian: @lvyanquan Sure, but this error message comes from [Doris API server side](https://github.com/apache/doris/blob/debb83f35d37cac9089b2e0282e2d555ad9c1d62/fe/fe-core/src/main/java/org/apache/doris/alter/SchemaChangeHandler.java#L1508), which is wrapped as a generic `RuntimeException` on client-side, and could only be identified by checking error message. Any suggestions about this?

lvyanquan: Got it. 
And I have checked the history of this code and it has been a long time since it was modified. It looks quite stable, so adding a comment to describe the source is enough.

lvyanquan (approved):
LGTM.

JNSimba (approved):
LGTM

yuxiqian:
Since @qg-lin is implementing this change in upstream doris-flink-connector, I'll mark this as draft for now.

yuxiqian:
Since it has been implemented in https://github.com/apache/doris-flink-connector/pull/408, I'll close it for now and bump dependency version to address this after Doris Flink connector 1.6.2 released.

PR #3218, opened 2024-04-11
[FLINK-35077][cdc] Add license checking CI scripts

yuxiqian:
Currently, Flink project has CI scripts checking if dependencies with incompatible licenses are introduced.

Flink CDC module heavily relies on external libraries (especially connectors), so running similar checking scripts during every CI would be helpful preventing developers introducing questionable dependencies by accident.


yuxiqian:
Hi @PatrickRen, I just pushed a hotfix commit to lock dependency version and suppress potential  breaking changes as follows:

```
The public API of some Rubyzip classes has been modernized to use named
parameters for optional arguments. Please check your usage of the
following classes:
  * `Zip::File`
  * `Zip::Entry`
  * `Zip::InputStream`
  * `Zip::OutputStream`

Please ensure that your Gemfiles and .gemspecs are suitably restrictive
to avoid an unexpected breakage when 3.0 is released (e.g. ~> 2.3.0).
See https://github.com/rubyzip/rubyzip for details. The Changelog also
lists other enhancements and bugfixes that have been implemented since
version 2.3.0.
```

Could you please help re-run the CI? Thanks!

PatrickRen (approved):
@yuxiqian Thanks for the contribution! LGTM

PR #3219, opened 2024-04-11
[hotfix] run mvn spotless to fix code style violations.

loserwang1024:
Currently, there is no spotless violation in ` flink-cdc-pipeline-connectors `. When run `mvn spotless:apply ` locally , will throw Exception.


loserwang1024:
@PatrickRen , CC

PatrickRen (approved):
@loserwang1024 Thanks for the PR. It looks like CI doesn't catch the issue and FLINK-35078 will fix it.

Issue #1879, opened 2023-01-18
MongoDB CDC will appear "the resume token was not found" error


bigjar:
when use mongodb cdc (sql) transfer the mongodb table to flink table store table，if enable the scan.incremental.snapshot, the job will appear the error: " the resume token was not found" immediately. 

**Environment :**
 - Flink version :  1.16.0
 - Flink CDC version: 2.3.0
 - Database and version:  mongodb 4.4  flink table store 0.3

**the sql is below:**

```
SET table.exec.resource.default-parallelism=10;

DROP TABLE IF EXISTS dim_mongo_xxx;
CREATE TABLE IF NOT EXISTS dim_mongo_xxx (
    _id STRING,
    threaten_type  STRING,
    credit_level STRING,
    threaten_score STRING,
    threaten_source STRING,
    created_time TIMESTAMP(3),
    updated_time TIMESTAMP(3),
    attack_time STRING,
    source STRING,
    ip STRING,
    domain STRING,
    device STRING,
    iphone STRING,
    PRIMARY KEY(_id) NOT ENFORCED
) WITH (
    'connector' = 'mongodb-cdc',
    'hosts' = 'xxxxx',
    'username' = 'xxxxx',
    'password' = 'xxxx',
    'database' = 'xxxx',
    'collection' = 'xxxxxxx',
    'heartbeat.interval.ms' = '3000',
    'poll.max.batch.size' = '2048',
    'scan.incremental.snapshot.enabled' = 'true'
);
CREATE CATALOG fts_catalog WITH (
    'type'='table-store',
    'warehouse'='hdfs://xxx/table_store'
);
USE CATALOG fts_catalog;
DROP TABLE IF EXISTS dim_fts_xxx;
CREATE TABLE IF NOT EXISTS dim_fts_xxx(
    _id STRING,
    threaten_type  STRING,
    credit_level STRING,
    threaten_score STRING,
    threaten_source STRING,
    created_time TIMESTAMP(3),
    updated_time TIMESTAMP(3),
    attack_time STRING,
    source STRING,
    ip STRING,
    domain STRING,
    device STRING,
    iphone STRING,
    PRIMARY KEY(_id) NOT ENFORCED
);
insert into fts_catalog.`default`.dim_fts_xxx select * from default_catalog.default_database.dim_mongo_xxx ;
```

**the exception is below:**

java.lang.RuntimeException: One or more fetchers have encountered exception
[...]

PR #3221, opened 2024-04-11
[FLINK-35079] Fallback to timestamp startup mode when resume token has expired

yuxiqian:
Closes [FLINK-35079](https://issues.apache.org/jira/browse/FLINK-35079).

When MongoDB CDC connector tries to create cursor with an expired resuming token during stream task fetching stage, it will crash with a fatal exception: `error due to Command failed with error 280 (ChangeStreamFatalError): cannot resume stream; the resume token was not found.`

This PR added fallback logic to create cursor with timestamp, which only runs when:
* Mongo CDC is in StreamTaskFetch stage
* a `ChangeStreamFatalError` (280) is raised
* Current `ChangeStreamOffset` has a valid timestamp field

Also tweaked @loserwang1024 's test case to cover resume token expiration scenario better.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
@@ -115,13 +115,12 @@ public class MongoDBStreamFetchTask implements FetchTask<SourceSplitBase> {
                 } catch (MongoCommandException e) {
                     if (MongoUtils.checkIfChangeStreamCursorExpires(e)) {
                         LOG.warn("Change stream cursor has expired, trying to recreate cursor");
-                        if (MongoUtils.checkIfResumeTokenExpires(e)) {
+                        boolean resumeTokenExpires = MongoUtils.checkIfResumeTokenExpires(e);
+                        if (resumeTokenExpires) {
                             LOG.warn(
                                     "Resume token has expired, fallback to timestamp restart mode");
-                            ((ChangeStreamOffset) streamSplit.getStartingOffset())
-                                    .clearResumeToken();
                         }
-                        changeStreamCursor = openChangeStreamCursor(descriptor);
+                        changeStreamCursor = openChangeStreamCursor(descriptor, resumeTokenExpires);
                         next = Optional.ofNullable(changeStreamCursor.tryNext());
                     } else {
                         throw e;
@@ -246,6 +245,11 @@ public class MongoDBStreamFetchTask implements FetchTask<SourceSplitBase> {
 
     private MongoChangeStreamCursor<BsonDocument> openChangeStreamCursor(
             ChangeStreamDescriptor changeStreamDescriptor) {
+        return openChangeStreamCursor(changeStreamDescriptor, false);
+    }
+
+    private MongoChangeStreamCursor<BsonDocument> openChangeStreamCursor(
+            ChangeStreamDescriptor changeStreamDescriptor, boolean forceTimestampStartup) {
         ChangeStreamOffset offset =
                 new ChangeStreamOffset(streamSplit.getStartingOffset().getOffset());
         ChangeStreamIterable<Document> changeStreamIterable =
@@ -254,7 +258,7 @@ public class MongoDBStreamFetchTask implements FetchTask<SourceSplitBase> {
         BsonDocument resumeToken = offset.getResumeToken();
         BsonTimestamp timestamp = offset.getTimestamp();
 
-        if (resumeToken != null) {
+        if (resumeToken != null && !forceTimestampStartup) {
             if (supportsStartAfter) {
                 LOG.info("Open the change stream after the previous offset: {}", resumeToken);
                 changeStreamIterable.startAfter(resumeToken);
@@ -268,6 +272,10 @@ public class MongoDBStreamFetchTask implements FetchTask<SourceSplitBase> {
             if (supportsStartAtOperationTime) {
                 LOG.info("Open the change stream at the timestamp: {}", timestamp);
                 changeStreamIterable.startAtOperationTime(timestamp);
+            } else if (forceTimestampStartup) {
+                LOG.error("Open change stream failed. Unable to resume from timestamp");
+                throw new FlinkRuntimeException(
+                        "Open change stream failed. Unable to resume from timestamp");
             } else {
                 LOG.warn("Open the change stream of the latest offset");
             }
@@ -301,9 +309,9 @@ public class MongoDBStreamFetchTask implements FetchTask<SourceSplitBase> {
                         e.getErrorMessage(),
                         e.getErrorCode());
                 throw new FlinkRuntimeException("Unauthorized $changeStream operation", e);
-            } else if (timestamp != null && MongoUtils.checkIfResumeTokenExpires(e)) {
-                ((ChangeStreamOffset) streamSplit.getStartingOffset()).clearResumeToken();
-                return openChangeStreamCursor(changeStreamDescriptor);
+            } else if (!forceTimestampStartup && MongoUtils.checkIfResumeTokenExpires(e)) {
+                LOG.info("Failed to open cursor with resume token, fallback to timestamp startup");
+                return openChangeStreamCursor(changeStreamDescriptor, true);
             } else {
                 LOG.error("Open change stream failed ", e);
                 throw new FlinkRuntimeException("Open change stream failed", e);
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
@@ -86,7 +86,6 @@ public class MongoDBStreamFetchTask implements FetchTask<SourceSplitBase> {
     private final Time time = new SystemTime();
     private boolean supportsStartAtOperationTime = true;
     private boolean supportsStartAfter = true;
-    private boolean resumeTokenExpired = false;
 
     public MongoDBStreamFetchTask(StreamSplit streamSplit) {
         this.streamSplit = streamSplit;
@@ -119,7 +118,8 @@ public class MongoDBStreamFetchTask implements FetchTask<SourceSplitBase> {
                         if (MongoUtils.checkIfResumeTokenExpires(e)) {
                             LOG.warn(
                                     "Resume token has expired, fallback to timestamp restart mode");
-                            resumeTokenExpired = true;
+                            ((ChangeStreamOffset) streamSplit.getStartingOffset())
+                                    .clearResumeToken();
                         }
                         changeStreamCursor = openChangeStreamCursor(descriptor);
                         next = Optional.ofNullable(changeStreamCursor.tryNext());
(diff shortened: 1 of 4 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
```diff
@@ -108,7 +110,32 @@ public void execute(Context context) throws Exception {
...
+                Optional<BsonDocument> next = Optional.empty();
+                try {
+                    next = Optional.ofNullable(changeStreamCursor.tryNext());
+                } catch (MongoQueryException e) {
+                    if (e.getErrorCode() == CHANGE_STREAM_FATAL_ERROR) {
+                        ChangeStreamOffset offset =
+                                new ChangeStreamOffset(streamSplit.getStartingOffset().getOffset());
+                        if (offset.getTimestamp() != null) {
+                            LOG.warn(
+                                    "Resume token {} has expired, fallback to timestamp startup mode.",
```
loserwang1024, line 122: Still initial startup mode  or latest mode, maybe "Resume token {} has expired, fallback to restart based on timestamp"?

 I'm not a mongodb expert. I wonder whether some data will loss or duplicated when fallback from resume token to timestamp? @Jiabao-Sun , CC

loserwang1024 (commented):
@yuxiqian , thanks for your contribution, which solves a long-standing issue. I have left some comment.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
yuxiqian: Addressed that.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
```diff
@@ -108,7 +110,32 @@ public void execute(Context context) throws Exception {
...
         try {
             while (taskRunning) {
-                Optional<BsonDocument> next = Optional.ofNullable(changeStreamCursor.tryNext());
+                Optional<BsonDocument> next;
+                try {
+                    next = Optional.ofNullable(changeStreamCursor.tryNext());
+                } catch (MongoQueryException e) {
+                    if (e.getErrorCode() == CHANGE_STREAM_FATAL_ERROR) {
+                        ChangeStreamOffset offset =
+                                new ChangeStreamOffset(streamSplit.getStartingOffset().getOffset());
```
Jiabao-Sun, line 119: Thanks @yuxiqian for this fix.
Why don't we catch this exception when we are in `openChangeStreamCursor`? 
Is there any consideration?

Jiabao-Sun: I checked the following stack trace in #1879, and this exception occurs when opening the cursor.

```
at com.mongodb.internal.operation.OperationHelper.withReadConnectionSource(OperationHelper.java:583)
at com.mongodb.internal.operation.CommandOperationHelper.executeCommand(CommandOperationHelper.java:189)
at com.mongodb.internal.operation.AggregateOperationImpl.execute(AggregateOperationImpl.java:195)
at com.mongodb.internal.operation.ChangeStreamOperation$1.call(ChangeStreamOperation.java:323)
at com.mongodb.internal.operation.ChangeStreamOperation$1.call(ChangeStreamOperation.java:319)
at com.mongodb.internal.operation.OperationHelper.withReadConnectionSource(OperationHelper.java:583)
at com.mongodb.internal.operation.ChangeStreamOperation.execute(ChangeStreamOperation.java:319)
at com.mongodb.internal.operation.ChangeStreamOperation.execute(ChangeStreamOperation.java:58)
at com.mongodb.client.internal.MongoClientDelegate$DelegateOperationExecutor.execute(MongoClientDelegate.java:184)
at com.mongodb.client.internal.ChangeStreamIterableImpl.execute(ChangeStreamIterableImpl.java:204)
at com.mongodb.client.internal.ChangeStreamIterableImpl.access$000(ChangeStreamIterableImpl.java:53)
at com.mongodb.client.internal.ChangeStreamIterableImpl$1.cursor(ChangeStreamIterableImpl.java:129)
at com.mongodb.client.internal.ChangeStreamIterableImpl$1.cursor(ChangeStreamIterableImpl.java:121)
at com.ververica.cdc.connectors.mongodb.source.reader.fetch.MongoDBStreamFetchTask.openChangeStreamCursor(MongoDBStreamFetchTask.java:237)
... 8 more
when use mongodb cdc (sql) transfer the mongodb table to flink table store table，if enable the scan.incremental.snapshot, the job will appear the error: " the resume token was not found" immediately.
```

Jiabao-Sun: I think we can refer to the logic in mongo-kafka to reopen the changestream.

https://github.com/mongodb/mongo-kafka/blob/2ffd620441b68bec19227dbfa641375d37148e39/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L102

yuxiqian: Yes, it would be much cleaner to move that checking logic to `openChangeStreamCursor` and implement the fallback logic along with existing routes.

CMIIW but seems `MongoCommandException` will only be thrown [when getNext or tryGetNext is called](https://github.com/mongodb/mongo/blob/a71feef352a1f08e3f9dbc2b840691e75775a370/src/mongo/db/pipeline/document_source_change_stream_ensure_resume_token_present.cpp#L133C31-L133C83), and we can't intentionally trigger an `MongoCommandException` by calling other methods like `hasNext` earlier.

@Jiabao-Sun Any suggestions about this?

Jiabao-Sun: Hi @yuxiqian, the `MongoCommandException` seems can be thrown by openChangeStreamCursor.

```
at com.mongodb.client.internal.ChangeStreamIterableImpl$1.cursor(ChangeStreamIterableImpl.java:121)
at com.ververica.cdc.connectors.mongodb.source.reader.fetch.MongoDBStreamFetchTask.openChangeStreamCursor(MongoDBStreamFetchTask.java:237)
```

image

yuxiqian: @Jiabao-Sun You're right, I've added resume token expiration checks in creating and iterating process.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
```diff
@@ -236,7 +250,7 @@ private MongoChangeStreamCursor<BsonDocument> openChangeStreamCursor(
         BsonDocument resumeToken = offset.getResumeToken();
         BsonTimestamp timestamp = offset.getTimestamp();
 
-        if (resumeToken != null) {
+        if (resumeToken != null && !resumeTokenExpired) {
```
Jiabao-Sun, line 253: It should be noted that `startAtOperationTime` is supported starting from MongoDB version 4.0. 

The prerequisite for executing the following code block is that the `resumeToken` does not exist, which occurs in two scenarios: 
1. a specified timestamp is used as the starting offset.
2. prior to version 4.0.7, when a collection has not been updated for a long time and the `postResumeToken` cannot be obtained, the current timestamp is used as the starting offset.

In the `else` branch of the code below, if we cannot start the Change Stream using `startAtOperationTime ` , we fallback to starting the Change Stream from the latest offset. 
Before MongoDB 4.0, we may lose data during the snapshot changes, but the likelihood of losing data during the snapshot is relatively low: if a collection is highly likely to change, we can obtain the `resumeToken` at its starting position; if a collection has not changed for a long time and we cannot obtain the `resumeToken` and `postResumeToken`, the likelihood of data loss during the snapshot is very low.

However, when we encounter an invalid `resumeToken` during runtime or when recovering from a checkpoint, there is a high possibility of data loss. I have reconsidered this issue, and if we cannot recover the change stream from the `resumeToken`, it is highly likely that we cannot recover it through `startAtOperationTime` either, as they correspond to the same position in the `oplog`.

In `mongo-kafka`, there is an explicit configuration called "tolerant-errors" provided to handle interruptions in the change stream. But in some scenarios, consistency requirements may outweigh availability, we should throw an exception to let user re-run the task from beginning.

```java
       else {
            if (supportsStartAtOperationTime) {
                LOG.info("Open the change stream at the timestamp: {}", timestamp);
                changeStreamIterable.startAtOperationTime(timestamp);
            } else {
                LOG.warn("Open the change stream of the latest offset");
            }
        }
```

yuxiqian:
Resume token could be refreshed by processing any data record or heartbeat event. When a resume token carried by a `ChangeStreamOffset` is found to be expired, resume token field is removed from that offset, and will fallback to timestamp mode automatically only after another event updates that offset. Though it's clearer than setting flag bits but it might not be the ideal solution since it changes given data structures. Comments are welcomed.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
```diff
@@ -283,6 +301,9 @@ private MongoChangeStreamCursor<BsonDocument> openChangeStreamCursor(
                         e.getErrorMessage(),
                         e.getErrorCode());
                 throw new FlinkRuntimeException("Unauthorized $changeStream operation", e);
+            } else if (timestamp != null && MongoUtils.checkIfResumeTokenExpires(e)) {
+                ((ChangeStreamOffset) streamSplit.getStartingOffset()).clearResumeToken();
+                return openChangeStreamCursor(changeStreamDescriptor);
```
yuxiqian, line 306: Recursively calling `openChangeStreamCursor` is safe here, since resume token is cleared here, and `checkResumeTokenExpires` would never be true again since no resume token is provided.

Jiabao-Sun: I prefer to add a parameter to `openChangeStreamCursor` to indicate forcing the use of timestamp mode instead of clearing the resumeToken.
WDYT?

yuxiqian: Yes, that would be cleaner to add a switch parameter instead of keeping global variable / status elsewhere. Pushed another commit to solve that.

Jiabao-Sun (approved):
Thanks @yuxiqian for the quick update.
LGTM.

PR #3226, opened 2024-04-16
[FLINK-35119][cdc-runtime] Change DataChangeEvent serialization and eserialization

zhongqishang:
Resolve [FLINK-35119](https://issues.apache.org/jira/browse/FLINK-35119).

The `before` and `after` of DataChangeEvent participate in serialization in case of null value.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/test/java/org/apache/flink/cdc/connectors/values/source/ValuesDataSourceITCase.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/test/java/org/apache/flink/cdc/connectors/values/source/ValuesDataSourceITCase.java
@@ -68,6 +68,7 @@ public class ValuesDataSourceITCase {
     private void executeDataStreamJob(ValuesDataSourceHelper.EventSetId type) throws Exception {
         StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
         env.enableCheckpointing(3000);
+        env.setParallelism(2);
         env.setRestartStrategy(RestartStrategies.noRestart());
         FlinkSourceProvider sourceProvider =
                 (FlinkSourceProvider) new ValuesDataSource(type).getEventSourceProvider();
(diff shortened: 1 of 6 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/test/java/org/apache/flink/cdc/connectors/values/source/ValuesDataSourceITCase.java
```diff
@@ -68,6 +68,7 @@ public void after() {
     private void executeDataStreamJob(ValuesDataSourceHelper.EventSetId type) throws Exception {
         StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
         env.enableCheckpointing(3000);
+        env.setParallelism(2);
```
zhongqishang, line 71: Just to speed up the running of `ValuesDataSourceITCase`, if you mind I can delete it or submit a separate PR.

yuxiqian: As reported in [FLINK-35127](https://issues.apache.org/jira/browse/FLINK-35127), using parallelism > 1 will crash this IT case, and get fixed in #3237 by removing  `HybridSource` entirely. Maybe this change needs further investigation.

zhongqishang: Glad to see it's fixed, I rollbacked the currently irrelevant commit.

Jiabao-Sun:
Thanks @zhongqishang for this fix.
Hi @yuxiqian, do you have time to review this PR?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/event/DataChangeEventSerializer.java
```diff
@@ -76,28 +72,18 @@ public DataChangeEvent deserialize(DataInputView source) throws IOException {
         OperationType op = opSerializer.deserialize(source);
         TableId tableId = tableIdSerializer.deserialize(source);
 
+        RecordData before = recordDataSerializer.deserialize(source);
+        RecordData after = recordDataSerializer.deserialize(source);
+        Map<String, String> meta = metaSerializer.deserialize(source);
```
yuxiqian, line 77: Is it safe to deserialize here if `op` isn't valid? Or we can validate op in advance.

yuxiqian (commented):
Thanks for @zhongqishang's great work, I just left a little comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/event/DataChangeEventSerializer.java
zhongqishang: @yuxiqian Thanks for your review.

> Is it safe to deserialize here if `op` isn't any known following one? 

IMO, the currently deserialized fields and order are consistent with those during serialization. Even if an unknown op operation occurs, deserialization will not cause an exception.

PR #3228, opened 2024-04-16
[FLINK-34878][cdc] Flink CDC pipeline transform supports CASE WHEN

aiwenmo:
This closes FLINK-34878.

This PR adds conditional functions for flink cdc transform operator.


aiwenmo:
Due to optimization of JaninoCompiler, this PR also solves the problem of outputting empty values when there are only constants in the calculation column.

PatrickRen (approved):
@aiwenmo Thanks for the patch! LGTM

Could you back port the commit to release-3.1?

PR #3230, opened 2024-04-17
[FLINK-35128][cdc-connector][cdc-base] Re-calculate the starting changelog offset after the new table added

loserwang1024:
In mysql cdc, re-calculate the starting binlog offset after the new table added in MySqlBinlogSplit#appendFinishedSplitInfos, while there lack of same action in StreamSplit#appendFinishedSplitInfos. This will cause data loss if any newly added table snapshot split's high_watermark is smaller.

Some unstable test problem occurs because of it.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/meta/split/StreamSplit.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/meta/split/StreamSplit.java
@@ -174,7 +174,7 @@ public class StreamSplit extends SourceSplitBase {
 
         return new StreamSplit(
                 streamSplit.splitId,
-                streamSplit.getStartingOffset(),
+                startingOffset,
                 streamSplit.getEndingOffset(),
                 splitInfos,
                 streamSplit.getTableSchemas(),
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/meta/split/StreamSplit.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/meta/split/StreamSplit.java
@@ -163,7 +163,15 @@ public class StreamSplit extends SourceSplitBase {
     // -------------------------------------------------------------------
     public static StreamSplit appendFinishedSplitInfos(
             StreamSplit streamSplit, List<FinishedSnapshotSplitInfo> splitInfos) {
+        // re-calculate the starting changelog offset after the new table added
+        Offset startingOffset = streamSplit.getStartingOffset();
+        for (FinishedSnapshotSplitInfo splitInfo : splitInfos) {
+            if (splitInfo.getHighWatermark().isBefore(startingOffset)) {
+                startingOffset = splitInfo.getHighWatermark();
+            }
+        }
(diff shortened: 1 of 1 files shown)


loserwang1024:
@PatrickRen , @morazow , CC

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/meta/split/StreamSplit.java
```diff
@@ -159,7 +159,15 @@ public String toString() {
...
+        for (FinishedSnapshotSplitInfo splitInfo : splitInfos) {
+            if (splitInfo.getHighWatermark().isBefore(startingOffset)) {
+                startingOffset = splitInfo.getHighWatermark();
+            }
+        }
         splitInfos.addAll(streamSplit.getFinishedSnapshotSplitInfos());
+
         return new StreamSplit(
                 streamSplit.splitId,
                 streamSplit.getStartingOffset(),
```
yuxiqian, line 173: CMIIW, but seems newly added code just calculated the earliest starting offset into `startingOffset` but didn't really use it to generate new `StreamSplit`. Maybe missed a change here?

```suggestion
        return new StreamSplit(
                streamSplit.splitId,
                startingOffset,
```

loserwang1024: It seems true.

loserwang1024: done it

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/meta/split/StreamSplit.java
```diff
@@ -163,10 +163,18 @@ public String toString() {
     // -------------------------------------------------------------------
     public static StreamSplit appendFinishedSplitInfos(
             StreamSplit streamSplit, List<FinishedSnapshotSplitInfo> splitInfos) {
+        // re-calculate the starting changelog offset after the new table added
+        Offset startingOffset = streamSplit.getStartingOffset();
+        for (FinishedSnapshotSplitInfo splitInfo : splitInfos) {
+            if (splitInfo.getHighWatermark().isBefore(startingOffset)) {
+                startingOffset = splitInfo.getHighWatermark();
+            }
+        }
```
morazow, line 172: Do we have to distinguish the high watermarks before the startingOffset? For example, if there are multiple high watermarks before startingOffset, which one should we take? Should it be the latest of those?

Or is taking any highWatermark if it is before the startingOffset is allright?

morazow (changes requested):
Thanks @loserwang1024 for the PR!

I have minor question about choosing highWatermarks.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/meta/split/StreamSplit.java
morazow: Got it, it will be always the min value

morazow (approved):

PatrickRen (approved):
@loserwang1024 Thanks for the patch! LGTM

PR #3235, opened 2024-04-18
[hotfix][values] Temporary fix for ValuesDataSource stuck in infinite loop

yuxiqian:
Currently, multiple parallelism would cause ValuesDataSourceITCase fail occasionally. This is a temporary fix to avoid blocking CI workflow.


yuxiqian:
cc @lvyanquan

lvyanquan (approved):
+1, Thanks for this.

lvyanquan:
Looks like there are other places that use ValueSource with paralism > 1, and I'm trying to find out the reason of this failure.

yuxiqian:
Fixed in #3237.

PR #3237, opened 2024-04-18
[FLINK-35127] remove HybridSource to avoid CI failure.

lvyanquan:
At present, CI testing often causes OOM due to ValuesSource, which may be caused by the use of HybridSource. Considering that this will block other prs, remove it first to restore CI.


lvyanquan:
@PatrickRen @yuxiqian PTAL.

lvyanquan:
Looks like HybridSource has a requirement for parallelism, but some of our tests use the core number of the machine itself https://github.com/apache/flink/blob/a312a3bdd258e0ff7d6f94e979b32e2bc762b82f/flink-connectors/flink-connector-base/src/test/java/org/apache/flink/connector/base/source/hybrid/HybridSourceITCase.java#L55

lvyanquan:
The failure in current CI contains a error message of 
```
(4185ef2c5087300d32f25ea842d3ec59_cbc357ccb763df2852fee8c4fc7d55f2_0_0) switched from RUNNING to FAILED with failure cause:
java.lang.IllegalStateException: The "before" field of UPDATE/DELETE message is null, please check the Postgres table has been set REPLICA IDENTITY to FULL level. You can update the setting by running the command in Postgres 'ALTER TABLE inventory.products REPLICA IDENTITY FULL'. Please see more in Debezium documentation: https://debezium.io/documentation/reference/1.9/connectors/postgresql.html#postgresql-replica-identity
	at org.apache.flink.cdc.connectors.postgres.table.PostgresValueValidator.validate(PostgresValueValidator.java:44) ~[classes/:?]
```
Which is unrelated with this pr.

And CC @loserwang1024.

PatrickRen (approved):
@lvyanquan Thanks for the patch! LGTM as a temporary solution. Still we need to investigate the root cause of this one.

PR #3247, opened 2024-04-24
[hotfix][docs] Use the download link of the release version instead of a snapshot version

ruanhang1993:
This PR update docs by using the download link of the release version instead of a snapshot version.


leonardBang (approved):
Thanks @ruanhang1993 for the fix, LGTM

PR #3255, opened 2024-04-25
[FLINK-35234][hotfix][cdc-common] Fix NullPointerException of org.apache.flink.cdc.common.configuration.ConfigurationUtils#convertToString

loserwang1024:
Fix exception like this:
```java
Caused by: java.lang.NullPointerException    at org.apache.flink.cdc.common.configuration.ConfigurationUtils.convertToString(ConfigurationUtils.java:133) ~[?:?]    at org.apache.flink.cdc.common.configuration.Configuration.toMap(Configuration.java:138) ~[?:?] 

```

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/configuration/ConfigurationUtils.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/configuration/ConfigurationUtils.java
@@ -45,6 +45,9 @@ public class ConfigurationUtils {
      */
     @SuppressWarnings("unchecked")
     public static <T> T convertValue(Object rawValue, Class<?> clazz) {
+        if (rawValue == null) {
+            return null;
+        }
         if (Integer.class.equals(clazz)) {
             return (T) convertToInt(rawValue);
         } else if (Long.class.equals(clazz)) {
@@ -70,7 +73,9 @@ public class ConfigurationUtils {
 
     @SuppressWarnings("unchecked")
     static <T> T convertToList(Object rawValue, Class<?> atomicClass) {
-        if (rawValue instanceof List) {
+        if (rawValue == null) {
+            return null;
+        } else if (rawValue instanceof List) {
             return (T) rawValue;
         } else {
             return (T)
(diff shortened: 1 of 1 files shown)


loserwang1024:
@yuxiqian , @PatrickRen , a mirror fix, CC

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/configuration/ConfigurationUtils.java
```diff
@@ -130,7 +130,9 @@ static Duration convertToDuration(Object o) {
     }
 
     static String convertToString(Object o) {
-        if (o.getClass() == String.class) {
+        if (o == null) {
```
yuxiqian, line 133: Maybe we can bring the check to `convertValue` function? Seems other `convertToXXX` functions also share this problem.

loserwang1024: you are right, just done it.

yuxiqian (approved):
Thanks for @loserwang1024's patch, LGTM.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/configuration/ConfigurationUtils.java
Jiabao-Sun: The inner null check seems redundant now.

Jiabao-Sun: > The inner null check seems redundant now.

@loserwang1024 WDYT?

leonardBang (commented):
@loserwang1024 Thanks for the contribution, a unit test is welcome

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/configuration/ConfigurationUtils.java
loserwang1024: > > The inner null check seems redundant now.

Maybe a java.lang.IllegalArgumentException with description is better than NullPointerException?

leonardBang (approved):
+1

PR #3261, opened 2024-04-25
[release] Update version to 3.2-SNAPSHOT

PatrickRen:
This pull request bumps the version to 3.2-SNAPSHOT for master branch.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@ limitations under the License.
     </distributionManagement>
 
     <properties>
-        <revision>3.0-SNAPSHOT</revision>
+        <revision>3.2-SNAPSHOT</revision>
         <java.version>1.8</java.version>
         <scala.binary.version>2.12</scala.binary.version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -62,7 +62,7 @@ limitations under the License.
     </distributionManagement>
 
     <properties>
-        <revision>3.0-SNAPSHOT</revision>
+        <revision>3.2-SNAPSHOT</revision>
```
leonardBang, line 65: Uh? should we also fix this in release-3.1 branch?

leonardBang, line 65: Uh? should we also fix this in release-3.1 branch?

leonardBang (approved):
Thanks @PatrickRen for the work, LGTM

pom.xml
PatrickRen: I did and pushed directly on release-3.1 just now. We never noticed this before...

PR #3267, opened 2024-04-27
[FLINK-35251][cdc][runtime] Fix bug of serializing derivation mapping in SchemaDerivation

PatrickRen:
This PR fixes bug of serializing derivation mapping in SchemaDerivation that mistakenly uses `out.write` instead of `out.writeInt` when serializing the size of the mapping, and also add a test case for validating serde of SchemaDerivation.


leonardBang (approved):
Thanks @PatrickRen for the fix, LGTM

PR #3269, opened 2024-04-28
[minor][docs] Fix route definition example in core concept docs

yuxiqian:
This PR fixes `route` definition YAML example in Flink CDC core concept documentations.


PatrickRen (approved):
@yuxiqian Thanks for the patch! LGTM

PR #3270, opened 2024-04-28
[BP-3.1][minor][docs] Fix route definition example in core concept docs

yuxiqian:
This PR fixes `route` definition YAML example in Flink CDC core concept documentations.


yuxiqian:
Rebased, cc @PatrickRen

PR #3272, opened 2024-04-28
[FLINK-35256][runtime] Fix transform node does not respect type nullability

yuxiqian:
This closes [FLINK-35256](https://issues.apache.org/jira/browse/FLINK-35256).

Flink CDC 3.1.0 brought transform feature, allowing column type / value transformation prior to data routing process. However after the transformation, column type marked as `NOT NULL` lost their annotation, causing some downstream sinks to fail since they require primary key to be NOT NULL.

Here's the minimum reproducible example about this problem:

```yaml
source:
  type: mysql
    ...

sink:
  type: starrocks
  name: StarRocks Sink
    ...

pipeline:
  name: Sync MySQL Database to StarRocks
  parallelism: 4

transform:
  source-table: reicigo.\.*
  projection: ID, UPPER(ID) AS UPID
```

MySQL source table schema definition is as follows:
```
CREATE TABLE table1 (
  ID VARCHAR(17) NOT NULL,
  ...
  PRIMARY KEY (ID)
);
```

The underlying reason is Calcite `SqlToRelConverter` used in DDL parsing removes redundant NOT NULL annotation after conversion. Backfilling this data from pre-transform columns should fix this problem.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
@@ -19,6 +19,7 @@ package org.apache.flink.cdc.runtime.parser;
 
 import org.apache.flink.api.common.io.ParseException;
 import org.apache.flink.cdc.common.schema.Column;
+import org.apache.flink.cdc.common.types.DataType;
 import org.apache.flink.cdc.common.types.DataTypes;
 import org.apache.flink.cdc.common.utils.StringUtils;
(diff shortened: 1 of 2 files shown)


lvyanquan:
I've tested that if namespace of TableId is not set and `__namespace_name__` column will be `null` string, so it won't cause error if MetadataColumns are not null.

PatrickRen (approved):
@yuxiqian Thanks for the patch! LGTM

PatrickRen:
@yuxiqian Could you also back port the patch onto `release-3.1`? Thanks

yuxiqian:
@PatrickRen Sure! Done in #3277.

PR #3279, opened 2024-04-29
[BP-3.1][FLINK-35255] DataSinkWriterOperator should override snapshotState and processWatermark method

yanghuaiGit:
… for the snapshotState and processWatermark methods


yanghuaiGit:
@PatrickRen  Excuse me, do you have time to take a look at this?

PR #3285, opened 2024-04-30
[FLINK-35272][cdc][runtime] Transform projection & filter feature overhaul

yuxiqian:
This closes [FLINK-35272](https://issues.apache.org/jira/browse/FLINK-35272).

Currently, pipeline jobs with transform (including projection and filtering) are constructed with the following topology:

```
SchemaTransformOp --> DataTransformOp --> SchemaOp
```

where schema projections are applied in `SchemaTransformOp` and data projection & filtering are applied in `DataTransformOp`. The idea is `SchemaTransformOp` might be embedded in `Sources` in the future to reduce payload data size transferred in Flink Job.

However, current implementation has a known defect that omits unused columns too early, causing some downstream-relied columns got removed after they arrived in `DataTransformOp`. See a example as follows:

```
# Schema is (ID INT NOT NULL, NAME STRING, AGE INT)
transform:
  - source-table: employee
    projection: id, upper(name) as newname
    filter: age > 18
```

Such transformation rules will fail since `name` and `age` columns are removed in `SchemaTransformOp`, and those data rows could not be retrieved in `DataTransformOp`, where the actual expression evaluation and filtering comes into effect.

This PR introduces a new design, renaming the transform topology as follows:

```
PreTransformOp --> PostTransformOp --> SchemaOp
```

where the `PreTransformOp` filters out columns, but only if:
* The column is not present in projection rules
* The column is not indirectly referenced by calculation and filtering expressions

If a column is explicity written down, it will be passed to downstream as-is. But for referenced columns, a special prefix will be added to their names. In the example above, a schema like `[id, newname, __PREFIX__name, __PREFIX__age]` will be generated to downstream. Notice that the expression evaluation and filtering will not come into effect for now, so a `DataChangeEvent` would be like `[1, null, 'Alice', 19]`.

Adding prefix is meant to deal with such cases:

```
# Schema is (ID INT NOT NULL, NAME STRING, AGE INT)
transform:
  - source-table: employee
    projection: id, upper(name) as name
    filter: age > 18
```

Here we need to distinguish the calculated column `(new) name` and the referenced original column `(old) name`. So after the name mangling process the schema would be like: `[id, name, __PREFIX__name]`.

Also, the filtering process is still done in `PostTransformOp` since user could write down a  filter expression that references calculated column, but their value won't be available until `PostTransformOp`'s evaluation. It also means in the following somewhat ambigious case:

```
# Schema is (ID INT NOT NULL, NAME STRING, AGE INT)
transform:
  - source-table: employee
    projection: id, age * 2 as age
    filter: age > 18
```

The filtering expression is applied to the calculated `age` column (doubled!) instead of the original one.


yuxiqian:
This PR is still in very early progress, looking for @aiwenmo & @lvyanquan's comments.

yuxiqian:
Updated based on previous comments, cc @aiwenmo

yuxiqian:
Thanks for @aiwenmo's kindly review, addressed comments above.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
```diff
@@ -362,25 +376,51 @@ private Optional<DataChangeEvent> processFilter(
...
+            valueList.add(fieldGetter.getFieldOrNull(recordData));
+        }
+
+        return tableInfo
+                .getRecordDataGenerator()
+                .generate(valueList.toArray(new Object[valueList.size()]));
+    }
+
     private boolean containFilteredComputedColumn(String projection, String filter) {
         boolean contain = false;
```
aiwenmo, line 425: Due to the filter always executing before the projection, this method is no longer useful.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformers.java
```diff
@@ -0,0 +1,61 @@
...
+
+/** Transformation rules used by {@link PostTransformOperator}. */
+public class PostTransformers {
+    private final Selectors selectors;
+
+    private final Optional<TransformProjection> projection;
+    private final Optional<TransformFilter> filter;
+
+    private final boolean containFilteredComputedColumn;
+
```
aiwenmo, line 34: Due to the filter always executing before the projection, this private variable is no longer useful.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
```diff
@@ -384,10 +481,73 @@ public static SqlSelect parseFilterExpression(String filterExpression) {
...
+        if (!isNullOrWhitespaceOnly(filterExpression)) {
             statement.append(" WHERE ");
             statement.append(filterExpression);
         }
         return parseSelect(statement.toString());
     }
+
+    public static SqlNode rewriteExpression(SqlNode sqlNode, Map<String, SqlNode> replaceMap) {
+        if (sqlNode instanceof SqlBasicCall) {
+            SqlBasicCall sqlBasicCall = (SqlBasicCall) sqlNode;
```
aiwenmo, line 493: Need to handle SqlCase.

yuxiqian: Replace `SqlCall` instead of `SqlBaseCall` to cover `SqlCase`.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
```diff
@@ -221,73 +240,67 @@ private TableInfo getTableInfoFromSchemaEvolutionClient(TableId tableId) throws
...
+                .forEach(
+                        testSchema -> {
+                            if (!testSchema.equals(firstSchema)) {
+                                throw new IllegalArgumentException(
+                                        String.format(
+                                                "Incompatible transform rules result found. Inferred schema: %s and %s",
+                                                firstSchema, testSchema));
+                            }
+                        });
+        return firstSchema;
```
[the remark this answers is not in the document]
yuxiqian: Added `SchemaUtils.mergeCompatibleUpstreamSchema` to resolve this.

yuxiqian:
Thanks @aiwenmo for reviewing, addressed review comments above.

aiwenmo (approved):
LGTM

yuxiqian:
cc @PatrickRen @lvyanquan

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaUtils.java
```diff
@@ -56,6 +67,177 @@ public static List<RecordData.FieldGetter> createFieldGetters(List<Column> colum
...
+            for (Schema schema : schemas) {
+                outputSchema = mergeSchema(outputSchema, schema);
+            }
+            return outputSchema;
+        }
+    }
+
+    /** Try to combine two schemas with potential incompatible type. */
+    @VisibleForTesting
+    public static Schema mergeSchema(@Nullable Schema lhs, Schema rhs) {
```
lvyanquan, line 100: Optimize parameter names.

lvyanquan (commented):
Thanks @yuxiqian for this, I've left some comments.

yuxiqian:
Thanks for @lvyanquan's comments! Addressed your comments in latest commits.

(Also fixed `testTimestampDiffTransform` bug that fails only on the first day of a month / year, cc @aiwenmo)

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformProcessor.java
```diff
@@ -0,0 +1,184 @@
...
+    public BinaryRecordData processData(BinaryRecordData payload, long epochTime) {
+        List<Object> valueList = new ArrayList<>();
+        List<Column> columns = tableInfo.getSchema().getColumns();
+
+        for (int i = 0; i < columns.size(); i++) {
+            if (cachedProjectionColumns.get(i) != null) {
+                ProjectionColumn projectionColumn = cachedProjectionColumns.get(i);
+                projectionColumnProcessorMap.putIfAbsent(
+                        projectionColumn.getColumnName(),
+                        ProjectionColumnProcessor.of(tableInfo, projectionColumn, timezone));
```
lvyanquan, line 126: We still need to calculate hashcode of projectionColumn.getColumnName() here even projectionColumnProcessorMap already contains this key, could we use List to reduce this calculation?

yuxiqian: Done. Considering `ProjectionColumnProcessor` has carried `ProjectionColumn` inside, I've combined two caches into one single `List`. CMIIW, seems all column processors must be created to correctly handle a `DataChangeEvent`, so the lazy initialization is redundant and has been removed.

yuxiqian: @melin Sorry but I didn't see any relevance with this PR. Would you mind opening another ticket and provide a minimum reproducible example about your feature request?

yuxiqian:
Fixed another problem in transform when combined with route & schema evolution. Really needs feedback from @leonardBang & @aiwenmo since this PR has gone a bit out of control.

leonardBang:
@yuxiqian Thanks for the improvement, could you rebase to latest master?

yuxiqian:
@leonardBang Done, rebased to master. However, since FLINK-35242 and FLINK-35272 has some overlapping changes, more rebasing might be necessary later.

yuxiqian:
The community has just reported another FLINK-35852 that could be closed by this PR. I've added another E2e test case to cover this issue, where applying transform rules on complex data types may yield broken data.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformProcessor.java
yuxiqian: @melin's issue has been tracked as FLINK-35774.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PreTransformChangeInfo.java
```diff
@@ -37,26 +37,27 @@
 import java.util.List;
 
 /** The TableInfo applies to cache schema change and fieldGetters. */
-public class TableChangeInfo {
```
leonardBang, line 40: Could you update the description?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformChangeInfo.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.flink.cdc.common.event.TableId;
+import org.apache.flink.cdc.common.schema.Schema;
+import org.apache.flink.cdc.common.utils.SchemaUtils;
+import org.apache.flink.cdc.runtime.typeutils.BinaryRecordDataGenerator;
+import org.apache.flink.cdc.runtime.typeutils.DataTypeConverter;
+
+import java.util.List;
+
+/** The TableInfo applies to cache schema and fieldGetters. */
+public class PostTransformChangeInfo {
```
leonardBang, line 30: do it too

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
```diff
@@ -30,74 +29,81 @@
...
 import java.util.stream.Collectors;
 
-/** A data process function that applies user-defined transform logics. */
-public class TransformDataOperator extends AbstractStreamOperator<Event>
+/**
+ * A data process function that performs column filtering, calculated column evaluation & final
+ * projection.
+ */
+public class PostTransformOperator extends AbstractStreamOperator<Event>
         implements OneInputStreamOperator<Event, Event> {
```
leonardBang, line 53: Could you add a `serialVersionUID` for all  serializable `StreamOperator` ?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaUtils.java
```diff
@@ -56,6 +67,181 @@ public static List<RecordData.FieldGetter> createFieldGetters(List<Column> colum
...
+                    String.format(
+                            "Unable to merge column %s and %s with different comments.",
+                            lColumn, rColumn));
+        }
+        return lColumn.copy(mergeDataType(lColumn.getType(), rColumn.getType()));
+    }
+
+    /** Try to combine given data types to a compatible wider data type. */
+    @VisibleForTesting
+    public static DataType mergeDataType(DataType lType, DataType rType) {
```
leonardBang, line 166: How about rename to `inferWiderType`  , `inferWiderColumn` and `inferWiderSchema` ?

flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/resources/ddl/data_types_test.sql
```diff
@@ -0,0 +1,28 @@
...
+-- limitations under the License.
+
+DROP TABLE IF EXISTS DATA_TYPES_TABLE;
+
+CREATE TABLE DATA_TYPES_TABLE (
+    ID INT NOT NULL,
+    TS DATETIME(0),
+    NUM DECIMAL(10, 2),
+    PRIMARY KEY (ID)
+);
```
leonardBang, line 23: The type is not so complex from my point, could you refer `fullTypesTest` of MySQL CDC Source？

flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/ComplexDataTypesE2eITCase.java
```diff
@@ -0,0 +1,334 @@
...
+                                + "  type: doris\n"
+                                + "  fenodes: doris:8030\n"
+                                + "  benodes: doris:8040\n"
+                                + "  username: %s\n"
+                                + "  password: \"%s\"\n"
+                                + "  table.create.properties.replication_num: 1\n"
+                                + "\n"
+                                + "transform:\n"
+                                + "  - source-table: %s.DATA_TYPES_TABLE\n"
+                                + "    projection: \\*, 'fine' AS FINE\n"
```
leonardBang, line 185: Could you add a filter to cover more logic as e2e case is more expensive than others?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformers.java
```diff
@@ -0,0 +1,53 @@
...
+package org.apache.flink.cdc.runtime.operators.transform;
+
+import org.apache.flink.cdc.common.schema.Selectors;
+
+import javax.annotation.Nullable;
+
+import java.util.Optional;
+
+/** Transformation rules used by {@link PostTransformOperator}. */
+public class PostTransformers {
```
leonardBang, line 27: A `TransformRule` should map a `PostTransformer` instead of a `PostTransformers` , right? could you also update the java doc ?

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/UnifiedTransformOperatorTest.java
```diff
@@ -0,0 +1,919 @@
...
+                .insertPostTransformed()
+                .deleteSource(1000, "Alice", 17)
+                .deletePreTransformed(1000, "Alice", 17)
+                .deletePostTransformed(1000, "Alice", 17, 1017)
+                .runTests()
+                .destroyHarness();
+
+        UnifiedTransformTestCase.of(
+                        tableId,
+                        "id + age as computed, *",
```
leonardBang, line 545: Could you add  test to cover cast case like `id + age as computed, CAST(computed AS BIGINT）as larger_computed`?

leonardBang (commented):
Thanks @yuxiqian for the contribution, I left some comments.

yuxiqian:
Done, rebased due to some conflicts with https://github.com/apache/flink-cdc/commit/26ff6d2a081181f3df7aa49d65d804c57c634122. Will add `CAST ... AS` tests after #3357 got merged.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformFilterProcessor.java
yuxiqian, line 1: I modified expression evaluation logic here since previous version could not correctly handle expressions like `filter: a > 1 and a < 2` and generates duplicated argument names which would crash the operator.

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/UnifiedTransformOperatorTest.java
yuxiqian: This implementation does not support writing an expression that references another calculated column, which could be easily circumvented by expanding it to something like `CAST(id + age AS BIGINT)`.

Could we address it in another PR considering it revolves many corner cases? Like which column to use when overlapped, and how to behave if one writes a recursive expression like `A as B, B as A`.

yuxiqian:
Resolved conflicts with #3357 and added `CAST ... AS` tests in UT / E2e cases. Seems CI failure is a known issue and should be fixed by #3449.

yuxiqian:
Rebased to latest `master`.

leonardBang (approved):
Thanks @yuxiqian for the update, LGTM now

yuxiqian:
Seems CI is failing due to an expired link in Doris docs. Pushed another commit to fix this.

PR #3291, opened 2024-05-06
[FLINK-35294][mysql] Use source config to check if the filter should be applied in timestamp starting mode

Shawn-Hx:
Since MySQL does not support the ability to quickly locate an binlog offset through a timestamp, the current logic for starting from a timestamp is to begin from the earliest binlog offset and then filter out the data before the user-specified position.

If the user restarts the job during the filtering process, this filter will become ineffective.

Use source config to check if the filter should be applied in timestamp starting mode.


leonardBang (approved):
Thanks @Shawn-Hx for the improvement, LGTM

PR #3292, opened 2024-05-06
[minor][cdc] Suppress false alarm in flink config loader

yuxiqian:
Currently, Flink CDC supports both new Flink config file `config.yaml` and fallback legacy config file `flink-conf.yaml`.

However, a false-alarm "File Not Found" exception will be printed in legacy compatible mode. Using `if - else` control flow to replace `try - catch` clause should fix this problem.


leonardBang (approved):
+1

PR #3299, opened 2024-05-07
[FLINK-35300][cdc][mysql] Improve MySqlStreamingChangeEventSource to skip null events in event deserializer

Shawn-Hx:
Improve MySqlStreamingChangeEventSource to skip null events in event deserializer.


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
```diff
@@ -257,6 +260,9 @@ public Event nextEvent(ByteArrayInputStream inputStream) throws IOException {
                         try {
                             // Delegate to the superclass ...
                             Event event = super.nextEvent(inputStream);
+                            if (event == null) {
+                                return null;
```
lvyanquan, line 264: There is no logic to go to this place because currently we don't need to read the binlog file on the local path using `com.github.shyiko.mysql.binlog.BinaryLogFileReader`.

leonardBang: +1, I also did't see why we need to skip the null events

PR #3300, opened 2024-05-07
[FLINK-35301][cdc] fix deadlock when loading driver classes

Shawn-Hx:
In JDK 8 and earlier version, if multiple threads invoke `Class.forName("com.xx.Driver")` simultaneously, it may cause jdbc driver deadlock.

[FLINK-19435](https://issues.apache.org/jira/browse/FLINK-19435)  analyzes this problem.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/relational/connection/JdbcConnectionPoolFactory.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/relational/connection/JdbcConnectionPoolFactory.java
@@ -23,6 +23,8 @@ import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import io.debezium.jdbc.JdbcConfiguration;
 
+import java.sql.DriverManager;
+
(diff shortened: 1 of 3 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/relational/connection/JdbcConnectionPoolFactory.java
```diff
@@ -43,6 +45,7 @@ public HikariDataSource createPooledDataSource(JdbcSourceConfig sourceConfig) {
         config.setMinimumIdle(MINIMUM_POOL_SIZE);
         config.setMaximumPoolSize(sourceConfig.getConnectionPoolSize());
         config.setConnectionTimeout(sourceConfig.getConnectTimeout().toMillis());
+        DriverManager.getDrivers();
```
leonardBang, line 48: Could you add a note for this?

Shawn-Hx: OK, I have added a description in this PR.

leonardBang (approved):
LGTM

PR #3312, opened 2024-05-10
[BP-3.1][FLINK-35323][runtime] Fix transform failure when one rule matches multiple schemas

yuxiqian:
This closes FLINK-35323.

Currently, Transform module doesn’t support handling multiple tables with different schema within one projection / filtering rule. This is not intended and turns out to be related to incorrectly handled processor caching strategy. Fixing the cache key should resolve this problem.


yuxiqian:
cc @aiwenmo

aiwenmo (approved):
LGTM

yuxiqian:
Could @PatrickRen help take a look? Thanks!

leonardBang (approved):

PR #3313, opened 2024-05-11
[FLINK-35323][runtime] Fix transform failure when one rule matches multiple tables with incompatible schema

yuxiqian:
This closes [FLINK-35323](https://issues.apache.org/jira/browse/FLINK-35323).

Currently, Transform module doesn’t support handling multiple tables with different schema within one projection / filtering rule. A minimum reproducible example is:

```yaml
transform:
  - source-table: DB.TABLE\.*
    projection: \*
```

If there are `DB.TABLE1` and `DB.TABLE2` in upstream source but with different schema, transform operator will panic. This is not intended and turns out to be related to incorrectly handled processor caching strategy. Fixing the cache key should resolve this problem.


aiwenmo (approved):
LGTM

yuxiqian:
Could @PatrickRen please take a look? Thanks!

leonardBang (approved):
LGTM, wait the CI

PR #3321, opened 2024-05-14
Update postgres-cdc.md

911432:
I changed the link.


yuxiqian:
Hi @911432, Flink CDC currently uses Debezium `1.9.7.Final`, so latest documentations might not apply. WDYT?

PR #3330, opened 2024-05-17
[FLINK-35386][cdc][docs] Build release-3.1 documentation and mark it as stable

GOODBOY008:


PatrickRen (approved):
@GOODBOY008 Thanks for the patch! LGTM

PR #3336, opened 2024-05-21
[FLINK-35354] Support host mapping in Flink tikv cdc

Mrart:
In tidb production environment deployment, there are usually two kinds of network: internal network and public network. When we use pd mode in tikv, we need to do network mapping, such as `spark.tispark.host_mapping` in https://github.com/pingcap/tispark/blob/master/docs/userguide_3.0.md. So I think we need support `host_mapping` in our Flink tikv cdc connector.


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/table/TiDBTableSource.java
```diff
@@ -70,12 +72,14 @@ public TiDBTableSource(
             String database,
             String tableName,
             String pdAddresses,
+            String hostMapping,
```
leonardBang, line 75: @Nullable

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/TDBSourceOptions.java
```diff
@@ -82,10 +90,13 @@ private TDBSourceOptions() {}
                     .withDescription("TiKV GRPC batch scan concurrency");
 
     public static TiConfiguration getTiConfiguration(
-            final String pdAddrsStr, final Map<String, String> options) {
+            final String pdAddrsStr, final String hostMapping, final Map<String, String> options) {
         final Configuration configuration = Configuration.fromMap(options);
 
         final TiConfiguration tiConf = TiConfiguration.createDefault(pdAddrsStr);
+        if (StringUtils.isNotBlank(hostMapping)) {
```
leonardBang, line 97: Could we use Flink's `Preconditions` check utils instead of introducing external dependency ?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/TDBSourceOptions.java
```diff
@@ -82,10 +90,13 @@ private TDBSourceOptions() {}
                     .withDescription("TiKV GRPC batch scan concurrency");
 
     public static TiConfiguration getTiConfiguration(
-            final String pdAddrsStr, final Map<String, String> options) {
+            final String pdAddrsStr, final String hostMapping, final Map<String, String> options) {
```
leonardBang, line 93: Could we add a new method named `getTiConfiguration(
            final String pdAddrsStr, final Map<String, String> options) {
            final String pdAddrsStr, final String hostMapping, final Map<String, String> options)` instead of changing the exits one ?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/TDBSourceOptions.java
```diff
@@ -57,6 +59,12 @@ private TDBSourceOptions() {}
                     .noDefaultValue()
                     .withDescription("TiKV cluster's PD address");
 
+    public static final ConfigOption<String> HOST_MAPPING =
+            ConfigOptions.key("host-mapping")
```
leonardBang, line 63: Could we also update connector documentation as well ?

leonardBang (commented):
Thanks @Mrart for the contribution, I left some comments

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/TDBSourceOptions.java
Mrart: hostMapping is an optional variable and returns tiConfiguration.Adding a new method would increase the complexity of the code. Change to Optional.of(new UriHostMapping(hostMapping)).ifPresent(tiConf::setHostMapping);

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/TDBSourceOptions.java
Mrart: Optional.of(new UriHostMapping(hostMapping)).ifPresent(tiConf::setHostMapping);

Mrart:
> I think we can specific a config prefix pass config to tikv client like cdc source connector pass config to debezium by `debezium.*`. This way will be flexible. @leonardBang @Mrart cc

tikv.* is configured on tikv config. hostmapping and pd config are similar to the database host configuration, and the current configuration looks more appropriate

Mrart:
> > I think we can specific a config prefix pass config to tikv client like cdc source connector pass config to debezium by `debezium.*`. This way will be flexible. @leonardBang @Mrart cc
> 
> tikv.* is configured on tikv config. hostmapping and pd config are similar to the database host configuration, and the current configuration looks more appropriate

Thanks for the help review.

Mrart:
@leonardBang Can you help me review it again？

PR #3341, opened 2024-05-21
[FLINK-35408][cdc][mysql] Add 30 min tolerance value when validating the time-zone setting

Shawn-Hx:
Now, MySQL CDC connector will retrieve the offset seconds between the configured timezone and UTC by executing the SQL statement below, and then compare it with the configured timezone.

```SQL
SELECT TIME_TO_SEC(TIMEDIFF(NOW(), UTC_TIMESTAMP())) 
```

For some MySQL instances, the validating for time-zone is too strict. We can add 30min tolerance value.


leonardBang (approved):
+1 for the minor improvements

PR #3347, opened 2024-05-22
[FLINK-35143][cdc-connector][mysql] Expose newly added tables capture in mysql pipeline connector.

qg-lin:


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MysqlPipelineNewlyAddedTableTCase.java
```diff
@@ -0,0 +1,478 @@
...
+        properties.put("database.user", customDatabase.getUsername());
+        properties.put("database.password", customDatabase.getPassword());
+        properties.put("database.serverTimezone", ZoneId.of("UTC").toString());
+        io.debezium.config.Configuration configuration =
+                io.debezium.config.Configuration.from(properties);
+        return DebeziumUtils.createMySqlConnection(configuration, new Properties());
+    }
+
+    @Test
+    public void testAddNewTableOneByOne() throws Exception {
```
loserwang1024, line 149: Great catch! 
Could you retract a common method like what NewlyAddedTableITCase#testNewlyAddedTableOneByOne does, so we can test different parallelism, different numbers of tables and different failover later?
And it will be better to add testRemoveTablesOneByOne.

qg-lin: Well, I'll do it later.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MysqlPipelineNewlyAddedTableITCase.java
```diff
@@ -0,0 +1,579 @@
...
+        public static class Builder {
+            private List<String> firstRoundInitTables;
+            private final List<String> firstRoundListenTables;
+            private final Integer firstRoundFetchSize;
+
+            private List<String> secondRoundInitTables;
+            private final List<String> secondRoundListenTables;
+            private final Integer secondRoundFetchSize;
+
+            public Builder(
```
loserwang1024, line 529: Why firstRoundInitTables and firstRoundListenTables are passed by constructor , while firstRoundInitTables is set by setter?

loserwang1024 (commented):
Do we need to test remove table , FailoverType and FailoverPhase which have test in SQL? @leonardBang , @PatrickRen , CC. 
Or I can do it in another pr.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MysqlPipelineNewlyAddedTableITCase.java
qg-lin: If we listen full database like 'db.\.*', we don't have to init any tables, because some tables have been initialized in the bofore() method(but I didn't implement this test case yet).

PR #3349, opened 2024-05-23
[FLINK-35129][cdc][postgres] Add checkpoint cycle option for commit offsets

morazow:
https://issues.apache.org/jira/browse/FLINK-35129

- Adds option for checkpoint cycle parameter
- Commits PG offsets on every multiple of checkpoint cycle
- Updates docs with new parameter definition
- Adds unit & integration tests to cover new functionality

### DIFF
--- a/docs/content/docs/connectors/flink-sources/postgres-cdc.md
+++ b/docs/content/docs/connectors/flink-sources/postgres-cdc.md
@@ -28,8 +28,7 @@ under the License.
 
 The Postgres CDC connector allows for reading snapshot data and incremental data from PostgreSQL database. This document describes how to setup the Postgres CDC connector to run SQL queries against PostgreSQL databases.
 
-Dependencies
-------------
+## Dependencies
 
 In order to setup the Postgres CDC connector, the following table provides dependency information for both projects using a build automation tool (such as Maven or SBT) and SQL Client with SQL JAR bundles.
 
@@ -45,8 +44,7 @@ Download [flink-sql-connector-postgres-cdc-3.0.1.jar](https://repo1.maven.org/ma
 
 **Note:** Refer to [flink-sql-connector-postgres-cdc](https://mvnrepository.com/artifact/com.ververica/flink-sql-connector-postgres-cdc), more released versions will be available in the Maven central warehouse.
 
-How to create a Postgres CDC table
-----------------
+## How to create a Postgres CDC table
 
 The Postgres CDC table can be defined as following:
 
@@ -76,8 +74,7 @@ CREATE TABLE shipments (
 SELECT * FROM shipments;
 ```
 
-Connector Options
-----------------
+## Connector Options
 
 <div class="highlight">
 <table class="colwidths-auto docutils">
@@ -236,12 +233,29 @@ Connector Options
           so it does not need to be explicitly configured 'execution.checkpointing.checkpoints-after-tasks-finish.enabled' = 'true'
       </td>
     </tr>
+    <tr>
+      <td>scan.lsn-commit.checkpoints-num-delay</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">3</td>
+      <td>Integer</td>
+      <td>The number of checkpoint delays before starting to commit the LSN offsets. <br>
+          The checkpoint LSN offsets will be committed in rolling fashion, the earliest checkpoint identifier will be committed first from the delayed checkpoints.
+      </td>
+    </tr>
     </tbody>
     </table>
 </div>
 <div>
 
-Note: `slot.name` is recommended to set for different tables to avoid the potential `PSQLException: ERROR: replication slot "flink" is active for PID 974` error. See more [here](https://debezium.io/documentation/reference/2.0/connectors/postgresql.html#postgresql-property-slot-name).
+### Notes
+
+#### `slot.name` option
+
+The `slot.name` is recommended to set for different tables to avoid the potential `PSQLException: ERROR: replication slot "flink" is active for PID 974` error. See more [here](https://debezium.io/documentation/reference/2.0/connectors/postgresql.html#postgresql-property-slot-name).
+
+#### `scan.lsn-commit.checkpoints-num-delay` option
+
+When consuming PostgreSQL logs, the LSN offset must be committed to trigger the log data cleanup for the corresponding slot. However, once the LSN offset is committed, earlier offsets become invalid. To ensure access to earlier LSN offsets for job recovery, we delay the LSN commit by `scan.lsn-commit.checkpoints-num-delay` (default value is `3`) checkpoints. This feature is available when config option `scan.incremental.snapshot.enabled` is set to true.
 
 ### Incremental Snapshot Options
 
@@ -340,8 +354,7 @@ The following options is available only when `scan.incremental.snapshot.enabled=
 </table>
 </div>
 
-Available Metadata
-----------------
+## Available Metadata
 
 The following format metadata can be exposed as read-only (VIRTUAL) columns in a table definition.
 
@@ -377,8 +390,7 @@ The following format metadata can be exposed as read-only (VIRTUAL) columns in a
   </tbody>
 </table>
 
-Limitation
---------
+## Limitation
 
 ### Can't perform checkpoint during scanning snapshot of tables when incremental snapshot is disabled
 
@@ -417,8 +429,7 @@ CREATE TABLE products (
 );
 ```
 
-Features
---------
+## Features
 
 ### Incremental Snapshot Reading (Experimental)
 
@@ -522,8 +533,7 @@ public class PostgreSQLSourceExample {
 }
 ```
 
-Data Type Mapping
-----------------
+## Data Type Mapping
 
 <div class="wy-table-responsive">
 <table class="colwidths-auto docutils">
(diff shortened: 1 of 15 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/config/PostgresSourceConfigFactory.java
```diff
@@ -100,7 +102,9 @@ public PostgresSourceConfig create(int subtaskId) {
 
         // The PostgresSource will do snapshot according to its StartupMode.
         // Do not need debezium to do the snapshot work.
-        props.put("snapshot.mode", "never");
+        props.setProperty("snapshot.mode", "never");
+
+        props.setProperty("checkpoint.cycle", String.valueOf(checkpointCycle));
```
loserwang1024, line 107: I don't know what "checkpoint.cycle" does? Debezium's offet commit cycle? Flink cdc have already been responsible for submitting offset, please not let Debezium do it again(turn off)

loserwang1024 (commented):
Great catch! I have left some comment.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/config/PostgresSourceConfigFactory.java
morazow: Ahh yes, this was my poor attempt to provide the parameter with properties to the dialect.

I just realized that this is for Debezium, removing it. And I will consider it differently 👍

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresDialect.java
```diff
@@ -217,6 +220,11 @@ public JdbcSourceFetchTaskContext createFetchTaskContext(JdbcSourceConfig taskSo
 
     @Override
     public void notifyCheckpointComplete(long checkpointId, Offset offset) throws Exception {
```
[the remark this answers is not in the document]
morazow: Yes, maybe PostgresSourceReader is good option.
I didn't want to have it in IncrementalSourceReaderWithCommit since this is only PG related feature.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresSourceReader.java
```diff
@@ -104,6 +108,10 @@ public List<SourceSplitBase> snapshotState(long checkpointId) {
 
     @Override
     public void notifyCheckpointComplete(long checkpointId) throws Exception {
+        this.checkpointCount = (this.checkpointCount + 1) % this.checkpointCycle;
```
loserwang1024, line 111: @morazow, great job. I generally agree with your approach. However, I currently have a different perspective. Instead of committing at every third checkpoint cycle (rolling window), I prefer to commit the offsets three checkpoints in advance (sliding window).

For a detailed design, we can store successful checkpoint IDs in a min heap, whose size is three (as decided by the configuration). When a checkpoint is successfully performed, we can push its ID into the heap and take the minimum checkpoint ID value, then commit it. By doing this, we always have three checkpoints whose offsets have not been recycled.

(P.S.: Let's log the heap at each checkpoint, so users can know from which checkpoint IDs they can restore.)

@leonardBang , @ruanhang1993 , CC, WDYT?

leonardBang: +1 for @loserwang1024 ‘s comment

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/test/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresSourceReaderTest.java
```diff
@@ -0,0 +1,75 @@
+package org.apache.flink.cdc.connectors.postgres.source.reader;
```
loserwang1024, line 1: Please add license and javadoc comment of class, nor compile will fails

morazow: Yeap, added 👍

loserwang1024:
LGTM! @leonardBang @ruanhang1993 , CC, WDYT of the option name `checkpoint.cycle`

loserwang1024 (approved):

morazow:
I don’t like it :)) How about, `checkpoint.rolling-window.size`? This seems to express the config option clearer?

docs/content/docs/connectors/flink-sources/postgres-cdc.md
```diff
@@ -236,6 +236,13 @@ Connector Options
           so it does not need to be explicitly configured 'execution.checkpointing.checkpoints-after-tasks-finish.enabled' = 'true'
       </td>
     </tr>
+    <tr>
+      <td>checkpoint.cycle</td>
```
leonardBang, line 240: Naming is hard, how about `scan.lsn-commit.checkpoints-num-delay` ?

morazow: Thanks @leonardBang, it is clearer and describes the intent. I'll refactor it accordingly

morazow:
Thanks all, applied your suggestions. Please have a look 👍

loserwang1024:
The finally advice, you can do same thing in org.apache.flink.cdc.debezium.DebeziumSourceFunction#notifyCheckpointComplete , or you mention in configuration document that this option is only when scan.incremental.snapshot.enabled = true.

morazow:
Hey @loserwang1024,

Thanks for the suggestions! I was actually wondering about it. 

Since this is only feature for Postgres, initially I considered adding it in PostgresSQLSource, but this is `@Deprecated` now.

Should I still add it to the PostgresSQLSource. Or adding note on the docs would be fine?

loserwang1024:
> Should I still add it to the PostgresSQLSource. Or adding note on the docs would be fine?

Hi, @leonardBang , WDYT?

leonardBang:
adding note on the docs makes sense to me as the old API is marked as @Deprecated

docs/content/docs/connectors/flink-sources/postgres-cdc.md
```diff
@@ -236,6 +236,15 @@ Connector Options
           so it does not need to be explicitly configured 'execution.checkpointing.checkpoints-after-tasks-finish.enabled' = 'true'
       </td>
     </tr>
+    <tr>
+      <td>scan.lsn-commit.checkpoints-num-delay</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">3</td>
+      <td>Integer</td>
+      <td>The number of checkpoint delays before starting to commit the LSN offsets. <br>
+          The checkpoint LSN offsets will be committed in rolling fashion, the earliest checkpoint identifier will be committed first from the delayed checkpoints.
```
leonardBang, line 245: Could you explain to users why we need delay the offset commit?

morazow: Sure, I think the previous version had a reason, I will add this:

```
This will enable continuous recycling of log files, preventing disk space issues.
```

morazow: Done, please have a look @leonardBang, thanks

docs/content/docs/connectors/flink-sources/postgres-cdc.md
```diff
@@ -236,6 +236,17 @@ Connector Options
...
+          This will enable continuous recycling of log files, preventing disk space issues. <br>
+          This feature is not available in `PostgreSQLSource` since it is deprecated.
```
leonardBang, line 247: When consuming PostgreSQL logs, the LSN offset must be committed to trigger the log data cleanup for the corresponding slot. However, once the LSN offset is committed, earlier offsets become invalid. To ensure access to earlier LSN offsets for job recovery, we delay the LSN commit by 3 checkpoints by default. This feature is available when config option `scan.incremental.snapshot.enabled` is set to true.

morazow: Great, would it be fine to add it this to the another `Remarks` section below? 

I could revert to previous version and add this note to the below. Or should I add it to the options description?

morazow: Hey @leonardBang, I have committed what I meant, please have another look

leonardBang (approved):
Thanks @morazow for the update, LGTM

loserwang1024 (approved):
LGTM.

PR #3350, opened 2024-05-23
[FLINK-35421]fixed schema operator blocking when restart

hk-lrzy:
https://issues.apache.org/jira/browse/FLINK-35421
1. i move refresh action to the `setup` method and remove checker of whether it restored from checkpoint or not.
2. i add some testing for it to double check when `SchemaOperator` restart, the  new `CreateTableEvent` chould be responsed `SchemaOperator` as expect.


lvyanquan (approved):
Good catch!

lvyanquan (approved):

PR #3354, opened 2024-05-23
[BP-3.0][FLINK-35431][doc] Migrate references in Flink CDC documentation from Debezium 1.9 to 2.0

GOODBOY008:


leonardBang (approved):
Thanks @GOODBOY008 for the backport.

yuxiqian:
Seems Debezium team has brought 1.9 docs back in https://github.com/debezium/debezium.github.io/commit/1da0427ad77ccc9e858649bce3178b5bb6930387 and links like https://debezium.io/documentation/reference/2.0/connectors/db2.html#db2-temporal-types are available again. Not sure if we should revert this or when will CDC upgrade Debezium base package. Thoughts? @GOODBOY008 @leonardBang

leonardBang:
Nice catch，use 1.9 is better than 2.0

PR #3357, opened 2024-05-23
[FLINK-34877][cdc] Flink CDC pipeline transform supports type conversion

aiwenmo:
This closes [FLINK-34877](https://issues.apache.org/jira/browse/FLINK-34877).

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
@@ -511,39 +511,66 @@ public class SystemFunctionUtils {
         return null;
     }
 
-    public static String convertString(Object object) {
+    public static String castToString(Object object) {
+        if (object == null) {
+            return null;
+        }
         return object.toString();
     }
 
-    public static Byte convertByte(Object object) {
-        return Byte.valueOf(convertString(object));
+    public static Byte castToByte(Object object) {
+        if (object == null) {
+            return null;
+        }
+        return Byte.valueOf(castToString(object));
     }
 
-    public static Boolean convertBoolean(Object object) {
-        return Boolean.valueOf(convertString(object));
+    public static Boolean castToBoolean(Object object) {
+        if (object == null) {
+            return null;
+        }
+        return Boolean.valueOf(castToString(object));
     }
 
-    public static Short convertShort(Object object) {
-        return Short.valueOf(convertString(object));
+    public static Short castToShort(Object object) {
+        if (object == null) {
+            return null;
+        }
+        return Short.valueOf(castToString(object));
     }
 
-    public static Integer convertInteger(Object object) {
-        return Integer.valueOf(convertString(object));
+    public static Integer castToInteger(Object object) {
+        if (object == null) {
+            return null;
+        }
+        return Integer.valueOf(castToString(object));
     }
 
-    public static Long convertLong(Object object) {
-        return Long.valueOf(convertString(object));
+    public static Long castToLong(Object object) {
+        if (object == null) {
+            return null;
+        }
+        return Long.valueOf(castToString(object));
     }
 
-    public static Float convertFloat(Object object) {
-        return Float.valueOf(convertString(object));
+    public static Float castToFloat(Object object) {
+        if (object == null) {
+            return null;
+        }
+        return Float.valueOf(castToString(object));
     }
 
-    public static Double convertDouble(Object object) {
-        return Double.valueOf(convertString(object));
+    public static Double castToDouble(Object object) {
+        if (object == null) {
+            return null;
+        }
+        return Double.valueOf(castToString(object));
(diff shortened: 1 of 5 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -298,4 +312,33 @@ private static Java.Rvalue generateNoOperandTimestampFunctionOperation(String op
...
+            case "FLOAT":
+                return "convertFloat";
+            case "DOUBLE":
+                return "convertDouble";
+            case "DECIMAL":
+                return "convertBigDecimal";
+            case "CHAR":
+            case "VARCHAR":
+            case "STRING":
+            case "`STRING`":
```
yuxiqian, line 337: Seems only the STRING type is specially treated to handle the quoted version. Will other types be quoted too? If so, maybe we can trim typename first before `switch` cases.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -298,4 +312,33 @@ private static Java.Rvalue generateNoOperandTimestampFunctionOperation(String op
...
+                return "convertString";
+            default:
+                throw new ParseException(
+                        "Unsupported data type convert: " + sqlDataTypeSpec.toString());
```
yuxiqian, line 341:
```suggestion
                throw new ParseException(
                        "Unsupported data type cast: " + sqlDataTypeSpec.toString());
```

Using cast might be better since user writes `cast` instead of `convert` keyword in YAML

yuxiqian (commented):
Thanks for @aiwenmo's great work! Just left some minor comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -510,4 +510,40 @@ public static Object coalesce(Object... objects) {
         }
         return null;
     }
+
+    public static String convertString(Object object) {
+        return object.toString();
```
lvyanquan, line 515: What would happen if we pass a NULL Object to cast function? Does we allow this happen?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -510,4 +510,40 @@ public static Object coalesce(Object... objects) {
         }
         return null;
     }
+
+    public static String convertString(Object object) {
```
lvyanquan, line 514: By the way, it's better to name them with `castToxxxx`.

lvyanquan (approved):
LGTM.

aiwenmo:
@yuxiqian @leonardBang PTAL

yuxiqian:
Is it possible to add actual values conversion checks besides existing syntax parsing tests? And it should cover cases where `NULL` value is passed as conversion argument.

leonardBang:
> Is it possible to add actual values conversion checks besides existing syntax parsing tests? And it should cover cases where `NULL` value is passed as conversion argument.

@aiwenmo I think @yuxiqian's idea is reasonable, could you add more value tests ?

leonardBang (approved):
Thanks @aiwenmo for the update, LGTM

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/TransformDataOperatorTest.java
```diff
@@ -555,6 +574,64 @@ void testTimestampDiffTransform() throws Exception {
                 .isEqualTo(new StreamRecord<>(insertEventExpect));
     }
 
+    @Test
+    void testNullCastTransform() throws Exception {
```
yuxiqian, line 578: Great to see this! Could you please insert more non-null data row to test 1) non-null value casting and verify results and 2) unsupported type conversion failure?

leonardBang: @yuxiqian Looks like @aiwenmo is busy on other things, as this comment is a minor test-related improvement and this PR block some on-going PRs, I'll merge this PR soon.

leonardBang (approved):
+1 from my side

aiwenmo:
Hi. I will commit the test tonight. And the code has been completed.

leonardBang:
> Hi. I will commit the test tonight. And the code has been completed.

Thanks @aiwenmo ，I'll wait your test code improvement in this PR

yuxiqian (approved):
Thanks for @aiwenmo's great work! @leonardBang PTAL

PR #3362, opened 2024-05-24
[FLINK-35447][doc-ci] Flink CDC Document document file had removed but website can access

GOODBOY008:
Solution:
Deletes files from the destination directory that are not present in the source directory.


leonardBang (approved):
+1

PR #3129, opened 2024-03-12
[FLINK-34653] Support table merging with route

PatrickRen:
This pull request supports complex table merging with route, including:
- Keeping all columns from source tables
- Type coercion if type of columns have conflict
- Schema evolution under table merging

PR #3365, opened 2024-05-27
[FLINK-35441][tests] Add migration test cases

yuxiqian:
This closes FLINK-35441.

Adding backwards compatible test with `3.0.0` and `3.1.0` version by serializing objects on lower versions and deserializing them on snapshot version.

* `SchemaManager` test runs on `3.1.0` only since commit https://github.com/apache/flink-cdc/commit/bcad5d9d11b9c37182510a43292e3019356173d1 and #3129 changes `SchemaRegistry` serialization fields and breaks backwards compatibility.
* `SchemaRegistry` test runs on `3.1.0` only, since it stores `SchemaManager` in its state.
* `ChangeTableInfo` test runs on `3.1.0` only since Transform feature does not exist in `3.0.x`.

Also noticed that there's no much classes using versioned serialization protocol, which couldn't guarantee backwards compatibility. Needs more discussion on how to refactor them.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -37,6 +37,8 @@ runtime:
   - flink-cdc-runtime/**/*
 e2e-tests:
   - flink-cdc-e2e-tests/**/*
+migration-tests:
+  - flink-cdc-migration-tests/**/*
 base:
   - flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/**/*
 debezium:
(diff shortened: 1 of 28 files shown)


yuxiqian:
I'm not sure if it's OK to perform migration tests like so, could @leonardBang @PatrickRen please provide some advices on this? Thanks!

leonardBang:
Looks like we can close this PR as we've finish the tests in https://github.com/apache/flink-cdc/pull/3370 @yuxiqian

PR #3369, opened 2024-05-28
[FLINK-35464] Fixes operator state backwards compatibility from CDC 3.0.x

yuxiqian:
This closes FLINK-35441 and FLINK-35464.

Flink CDC 3.1 changes how SchemaRegistry [de]serializes state data, which causes any checkpoint states saved with earlier version could not be restored in version 3.1.0. This PR adds serialization versioning for state payloads and ensures 3.0.x state could be successfully restored.

Unfortunately 3.1.0 introduces breaking changes without bumping serialization version, so this release will be excluded from state compatibility guarantee.

---

Running some real savepoint restore tests with https://github.com/yuxiqian/migration-test reveals that:

Before this change: (Snapshots compiled from `release-3.1` and `master`)

|  From \ To   | 3.0.0 | 3.0.1 | 3.1.0 | 3.1-SNAPSHOT | 3.2-SNAPSHOT |
| ------------ | ----- | ----- | ----- | ------------ | -------------|
| 3.0.0        | ❓    | ❓    | ❌     | ❌            | ❌           |
| 3.0.1        |       | ❓    | ❌     | ❌            | ❌           |
| 3.1.0        |       |       | ✅     | ✅            | ✅           |
| 3.1-SNAPSHOT |       |       |        | ✅            | ✅           |    
| 3.2-SNAPSHOT |       |       |        |               | ✅           |

> ✅ - Compatible, ❌ - Not compatible, ❓ - Target version doesn't support `--from-savepoint`

After this change: (Snapshots compiled from `FLINK-35464-BP-3.1` and `FLINK-35464`)

|  From \ To   | 3.0.0 | 3.0.1 | 3.1.0 | 3.1-SNAPSHOT | 3.2-SNAPSHOT |
| ------------ | ----- | ----- | ----- | ------------ | -------------|
| 3.0.0        | ❓    | ❓    | ❌     | ✅            | ✅           |
| 3.0.1        |       | ❓    | ❌     | ✅            | ✅           |
| 3.1.0        |       |       | ✅     | ❌            | ❌           |
| 3.1-SNAPSHOT |       |       |        | ✅            | ✅           |    
| 3.2-SNAPSHOT |       |       |        |               | ✅           |

### DIFF
--- a/flink-cdc-migration-tests/pom.xml
+++ b/flink-cdc-migration-tests/pom.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.flink</groupId>
+        <artifactId>flink-cdc</artifactId>
+        <version>${revision}</version>
+    </parent>
+
+    <artifactId>flink-cdc-migration-tests</artifactId>
+    <name>flink-cdc-migration-tests</name>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>flink-cdc-release-3.0.0</module>
+        <module>flink-cdc-release-3.0.1</module>
+        <module>flink-cdc-release-3.1.0</module>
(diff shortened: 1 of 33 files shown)


PatrickRen (commented):
@yuxiqian Thanks for the PR! The patch and test themselves looks good to me, but I'm a bit concerning about introducing new modules for each version, even patch versions   (like 3.0.1). As the project evolves there must be a lot of versions in the future, so number of modules could grow very fast. Not sure if this might burden the Maven build process.

yuxiqian:
Thanks for @PatrickRen's comments! I too agree that it's not very elegant to package all history versions just for compatibility test, but seems to be inevitable since there's no such thing like C's macro or conditional compilation in Java.

What I'm thinking of is 1) moving those dummy projects to somewhere outside (like `yuxiqian/dummy-cdc-migration-test` repo), or 2) write a script to dynamically generate dummy projects during the CI. WDYT?

flink-cdc-migration-tests/pom.xml
```diff
@@ -19,13 +19,10 @@ limitations under the License.
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <parent>
```
PatrickRen: Actually you can keep the <parent> section here. Removing `flink-cdc-migration-tests` from the <modules> section in the parent pom should be enough.

pom.xml
```diff
@@ -40,7 +40,6 @@ limitations under the License.
         <module>flink-cdc-connect</module>
         <module>flink-cdc-runtime</module>
         <module>flink-cdc-e2e-tests</module>
-        <module>flink-cdc-migration-tests</module>
```
PatrickRen: Is it possible to use profiles to control whether to build `flink-cdc-migration-tests` module?

PatrickRen (commented):
@yuxiqian Thanks for the update! I left some comments.

yuxiqian:
Thanks for the tips! Addressed your comments to simplify pom files.

PatrickRen (approved):
@yuxiqian Thanks for the update! LGTM

PR #3396, opened 2024-06-06
[FLINK-35540][cdc-common][cdc-connector][mysql] fix lost table when database and table are with the same name

qg-lin:
https://issues.apache.org/jira/browse/FLINK-35540

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/schema/Selectors.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/schema/Selectors.java
@@ -98,15 +98,15 @@ public class Selectors {
                     Predicates.setOf(
                             tableInclusions, Predicates.RegExSplitterByComma::split, (str) -> str);
             for (String tableSplit : tableSplitSet) {
-                Set<String> tableIdSet =
-                        Predicates.setOf(
+                List<String> tableIdList =
+                        Predicates.listOf(
                                 tableSplit, Predicates.RegExSplitterByDot::split, (str) -> str);
-                Iterator<String> iterator = tableIdSet.iterator();
-                if (tableIdSet.size() == 1) {
+                Iterator<String> iterator = tableIdList.iterator();
+                if (tableIdList.size() == 1) {
                     selectors.add(new Selector(null, null, iterator.next()));
-                } else if (tableIdSet.size() == 2) {
+                } else if (tableIdList.size() == 2) {
                     selectors.add(new Selector(null, iterator.next(), iterator.next()));
-                } else if (tableIdSet.size() == 3) {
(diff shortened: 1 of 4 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/testutils/UniqueDatabase.java
```diff
@@ -58,10 +62,19 @@ public class UniqueDatabase {
...
+            MySqlContainer container,
+            String databaseName,
+            String username,
+            String password,
+            boolean withIdentifier) {
         this(
                 container,
                 databaseName,
-                Integer.toUnsignedString(new Random().nextInt(), 36),
+                withIdentifier ? getIdentifier() : StringUtils.EMPTY,
```
yuxiqian, line 77: The identifier suffix is meant to avoid test case conflicts since multiple cases might run simultaneously on a single MySQL database instance. It should be fine now since there's only one test case using this class, but might cause trouble in the future if more testcases are added.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/testutils/UniqueDatabase.java
yuxiqian, line 1: Generally I would prefer not to change `UniqueDatabase`, and some alternative approaches are 1) Keep using UniqueDatabase to test identical DB and table name scenario, create tables during tests or 2) Create another class like `CustomDatabase` which allows specifying database name directly without appending suffix. Looking forward to hearing @qg-lin's thoughts.

yuxiqian (commented):
Thanks @qg-lin, these changes look good to me! I just left some comments on test cases.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/testutils/UniqueDatabase.java
qg-lin: Good idea! I'll rewrite the case.

yuxiqian (approved):
Thanks for @qg-lin's nice work! Also please run `mvn spotless:apply` to fix format violations before it could be merged.

cc @leonardBang

leonardBang (approved):
LGTM

PR #3400, opened 2024-06-07
use lazy initialization to resolve circular dependency between the `DataFieldSerializer`, `DataTypeSerializer`, and `RowTypeSerializer` classes that causes NPE for certain RowTypes like nested row type

umeshdangat:
There is a circular dependency between the `DataFieldSerializer`, `DataTypeSerializer`, and `RowTypeSerializer` classes. This circular dependency can lead to incomplete initialization of the serializers, causing a `NullPointerException` when you attempt to use them.

Here's a breakdown of the problem:

1. `DataFieldSerializer` has a `DataTypeSerializer` instance.
2. `DataTypeSerializer` has a `RowTypeSerializer` instance.
3. `RowTypeSerializer` has a `ListSerializer<DataField>` instance, which in turn uses `DataFieldSerializer`.

This circular dependency can cause the serializers to be in an incomplete state when they are first accessed, leading to the `NullPointerException`. 

Code below is enough to re-create this NPE
```
        RowType innerMostRowType = RowType.of(DataTypes.BIGINT());
        RowType outerRowType = RowType.of(
                DataTypes.ROW(DataTypes.FIELD("outerRow", innerMostRowType)));

        DataTypeSerializer serializer = new DataTypeSerializer();


        assertNotNull(RowTypeSerializer.INSTANCE, "RowTypeSerializer.INSTANCE is null");
        RowType copiedRow = (RowType) serializer.copy(outerRowType);

```

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/DataFieldSerializer.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/DataFieldSerializer.java
@@ -39,7 +39,14 @@ public class DataFieldSerializer extends TypeSerializerSingleton<DataField> {
     public static final DataFieldSerializer INSTANCE = new DataFieldSerializer();
 
     private final StringSerializer stringSerializer = StringSerializer.INSTANCE;
-    private final DataTypeSerializer dataTypeSerializer = new DataTypeSerializer();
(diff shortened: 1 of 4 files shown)


yuxiqian:
Could @leonardBang please reopen this?

umeshdangat:
PR #3434 also has this change! Merging that should enable us to close this.

PR #3414, opened 2024-06-14
[FLINK-35237] Allow Sink to Choose HashFunction in PrePartitionOperator

dingxin-tech:
https://issues.apache.org/jira/browse/FLINK-35237

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.common.sink;
+
+import org.apache.flink.cdc.common.data.RecordData;
+import org.apache.flink.cdc.common.event.DataChangeEvent;
+import org.apache.flink.cdc.common.event.OperationType;
+import org.apache.flink.cdc.common.event.TableId;
+import org.apache.flink.cdc.common.schema.Schema;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+/** the default implementation of hash function. */
+public class DefaultHashFunctionProvider implements HashFunctionProvider {
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public HashFunction getHashFunction(TableId tableId, Schema schema) {
+        return new DefaultHashFunction(schema);
+    }
+
+    static class DefaultHashFunction implements HashFunction {
+        private final List<RecordData.FieldGetter> primaryKeyGetters;
+
+        public DefaultHashFunction(Schema schema) {
+            primaryKeyGetters = createFieldGetters(schema);
+        }
+
+        @Override
+        public Integer apply(DataChangeEvent event) {
+            List<Object> objectsToHash = new ArrayList<>();
+            // Table ID
+            TableId tableId = event.tableId();
+            Optional.ofNullable(tableId.getNamespace()).ifPresent(objectsToHash::add);
+            Optional.ofNullable(tableId.getSchemaName()).ifPresent(objectsToHash::add);
+            objectsToHash.add(tableId.getTableName());
+
+            // Primary key
+            RecordData data =
+                    event.op().equals(OperationType.DELETE) ? event.before() : event.after();
+            for (RecordData.FieldGetter primaryKeyGetter : primaryKeyGetters) {
+                objectsToHash.add(primaryKeyGetter.getFieldOrNull(data));
+            }
+
+            // Calculate hash
+            return (Objects.hash(objectsToHash.toArray()) * 31) & 0x7FFFFFFF;
+        }
+
+        private List<RecordData.FieldGetter> createFieldGetters(Schema schema) {
+            List<RecordData.FieldGetter> fieldGetters =
+                    new ArrayList<>(schema.primaryKeys().size());
+            schema.primaryKeys().stream()
+                    .mapToInt(
+                            pk -> {
+                                int index = schema.getColumnNames().indexOf(pk);
+                                if (index == -1) {
+                                    throw new IllegalStateException(
+                                            String.format(
+                                                    "Unable to find column \"%s\" which is defined as primary key",
+                                                    pk));
+                                }
+                                return index;
+                            })
+                    .forEach(
+                            primaryKeyPosition ->
+                                    fieldGetters.add(
+                                            RecordData.createFieldGetter(
+                                                    schema.getColumns()
+                                                            .get(primaryKeyPosition)
+                                                            .getType(),
+                                                    primaryKeyPosition)));
+            return fieldGetters;
+        }
+    }
+}
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunction.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunction.java
@@ -17,26 +17,12 @@
 
 package org.apache.flink.cdc.common.sink;
 
-import org.apache.flink.cdc.common.annotation.PublicEvolving;
+import org.apache.flink.cdc.common.event.DataChangeEvent;
 
-/**
- * {@code DataSink} is used to write change data to external system and apply metadata changes to
- * external systems as well.
- */
-@PublicEvolving
-public interface DataSink {
+import java.util.function.Function;
 
-    /** Get the {@link EventSinkProvider} for writing changed data to external systems. */
-    EventSinkProvider getEventSinkProvider();
-
-    /** Get the {@link MetadataApplier} for applying metadata changes to external systems. */
-    MetadataApplier getMetadataApplier();
-
-    /**
-     * Get the {@link HashFunctionProvider} for calculating hash value when partition by primary
-     * ley.
-     */
-    default HashFunctionProvider getHashFunctionProvider() {
-        return new DefaultHashFunctionProvider();
-    }
+/** use for {@code PrePartitionOperator} when calculating hash code of primary key. */
+public interface HashFunction extends Function<DataChangeEvent, Integer> {
+    @Override
+    Integer apply(DataChangeEvent event);
 }
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunctionProvider.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunctionProvider.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.common.sink;
+
+import org.apache.flink.cdc.common.event.DataChangeEvent;
+import org.apache.flink.cdc.common.event.TableId;
+import org.apache.flink.cdc.common.schema.Schema;
+
+import java.io.Serializable;
+
+/**
+ * Provide {@link HashFunction} to help {@code PrePartitionOperator} to shuffle {@link
+ * DataChangeEvent} to designated subtask. This is usually beneficial for load balancing, when
+ * writing to different partitions/buckets in {@link DataSink}, add custom implementation to further
+ * improve efficiency.
+ */
+public interface HashFunctionProvider extends Serializable {
+
+    /**
+     * Gets a hash function based on the given table ID and schema, to help {@code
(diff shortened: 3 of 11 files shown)


flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunction.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.common.sink;
+
+import org.apache.flink.cdc.common.event.DataChangeEvent;
+
+import java.util.function.Function;
+
+/** use for PrePartitionOperator when calculating hash code of primary key. */
```
lvyanquan, line 24: Used to calculate hash code of given DataChangeEvent by primary key, which will help {@link PrePartitionOperator} to shuffle DataChangeEvent to designated subtask.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunctionProvider.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.common.sink;
+
+import org.apache.flink.cdc.common.schema.Schema;
+
+import java.io.Serializable;
+
+/** use for PrePartitionOperator. */
```
lvyanquan, line 24: Provide {@link HashFunction} to help {@link PrePartitionOperator} to shuffle DataChangeEvent to designated subtask. This is usually beneficial for load balancing, when writing to different partitions/buckets in {@link DataSink}, add custom Implementation to further improve efficiency.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunction.java
lvyanquan, line 24: {@link PrePartitionOperator}

lvyanquan (commented):
Thanks for this great contribution, LGTM. left some minor comments about java doc.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunction.java
dingxin-tech: Actually, `Sink` in `cdc-common` cannot link `PrePartitionOperator` in `cdc-runtime`.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunctionProvider.java
dingxin-tech: Sure, thank you for perfect comments so clear.

lvyanquan (approved):
LGTM.
And CC @PatrickRen  @leonardBang

lvyanquan:
What's more, considering that the number of buckets and parallelism may not be consistent, should we remove the constraint on [EventPartitioner](https://github.com/apache/flink-cdc/blob/2bd2e4ce24ec0cc6a11129e3e3b32af6a09dd977/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/partitioning/EventPartitioner.java#L25)?

dingxin-tech:
> What's more, considering that the number of buckets and parallelism may not be consistent, should we remove the constraint on [EventPartitioner](https://github.com/apache/flink-cdc/blob/2bd2e4ce24ec0cc6a11129e3e3b32af6a09dd977/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/partitioning/EventPartitioner.java#L25)?

Although the number of buckets and parallelism will differ, we can only distribute based on parallelism rather than buckets, right? We have already distributed the hash values to the various parallelisms [here](https://github.com/apache/flink-cdc/blob/2bd2e4ce24ec0cc6a11129e3e3b32af6a09dd977/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/partitioning/PrePartitionOperator.java#L104C43-L104C64), so I think there's no need to change anything here.

lvyanquan:
Got it,  There is indeed no need for adjustment.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunctionProvider.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Provide {@link HashFunction} to help PrePartitionOperator to shuffle DataChangeEvent to
+ * designated subtask. This is usually beneficial for load balancing, when writing to different
+ * partitions/buckets in {@link DataSink}, add custom Implementation to further improve efficiency.
+ */
+public interface HashFunctionProvider extends Serializable {
+    HashFunction getHashFunction(Schema schema);
```
lvyanquan, line 30: Can we add `open`/`close` method and call in PrePartitionOperator?

lvyanquan, line 30: Can we pass `TableId` in `getHashFunction` method  too? 
When integrating third-party SDK, we may not want to rebuild the HashFunction using the Schema defination of FlinkCDC, but instead directly obtain thecalculation method in the SDK based on the table name.

dingxin-tech: Yes, I encountered this issue while implementing the MaxCompute Connector. I resolved it by converting the Flink Schema to the MaxCompute Schema.

You are right, passing the TableId would be a better approach.

dingxin-tech: Sure, if you feel it's necessary.

lvyanquan (approved):

lvyanquan (approved):

dingxin-tech:
Can this PR be merged before the other [PR](https://github.com/apache/flink-cdc/pull/3254)? Both PRs are marked for inclusion in version 3.2, but the other PR depends on this one. I will need some time to make the necessary adjustments.

lvyanquan:
@leonardBang  @PatrickRen can you help to review and merge this?

lvyanquan:
@yuxiqian CC.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
```diff
@@ -0,0 +1,96 @@
...
+        @Override
+        public Integer apply(DataChangeEvent event) {
+            List<Object> objectsToHash = new ArrayList<>();
+            // Table ID
+            TableId tableId = event.tableId();
+            Optional.ofNullable(tableId.getNamespace()).ifPresent(objectsToHash::add);
+            Optional.ofNullable(tableId.getSchemaName()).ifPresent(objectsToHash::add);
+            objectsToHash.add(tableId.getTableName());
+
+            // Primary key
```
yuxiqian, line 56: This comment seems misleading.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
```diff
@@ -0,0 +1,96 @@
...
+                                            ++i;
+                                        }
+                                        if (i >= schema.getColumnCount()) {
+                                            throw new IllegalStateException(
+                                                    String.format(
+                                                            "Unable to find column \"%s\" which is defined as primary key",
+                                                            pk));
+                                        }
+                                        return i;
+                                    })
```
yuxiqian, line 85: This seems could be replaced with `schema.getColumnNames().indexOf(pk)`

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
```diff
@@ -0,0 +1,96 @@
...
+                            .toArray();
+            for (int primaryKeyPosition : primaryKeyPositions) {
+                fieldGetters.add(
+                        RecordData.createFieldGetter(
+                                schema.getColumns().get(primaryKeyPosition).getType(),
+                                primaryKeyPosition));
+            }
```
yuxiqian, line 92: Chain this with previous `Stream<String>`?

yuxiqian (commented):
Thanks for @dingxin-tech's contribution. I wonder if  `DefaultHashFunctionProvider` implementation could be improved when migrating from existing `PrePartitionOperator#HashFunction`.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
dingxin-tech: Is "Primary Key" causing missleading?
why?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
dingxin-tech: Sure, in order to be consistent with the original logic, keep checking for illegal values.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
dingxin-tech: Sure.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/DefaultHashFunctionProvider.java
yuxiqian: Never mind. Was wondering if something like "Extract Table ID" / "Extract primary key columns" would be better, but I think current comments are clear enough.

yuxiqian (approved):

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunctionProvider.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public interface HashFunctionProvider extends Serializable {
+    HashFunction getHashFunction(TableId tableId, Schema schema);
+
+    // --------------------------------------------------------------------------------------------
+    //  Default life cycle methods
+    // --------------------------------------------------------------------------------------------
+    default void open() {}
+
+    default void close() {}
```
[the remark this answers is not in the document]
dingxin-tech: Since we provided the `getHashFunction` method based on `TableId`, connector implementers might use `TableId` to obtain the actual schema of the database and perform some caching operations. We can establish connections and initialize caches in the open method. These two lifecycle methods were added following @lvyanquan 's suggestion, and he might respond with further additions.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunctionProvider.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Provide {@link HashFunction} to help PrePartitionOperator to shuffle DataChangeEvent to
+ * designated subtask. This is usually beneficial for load balancing, when writing to different
+ * partitions/buckets in {@link DataSink}, add custom Implementation to further improve efficiency.
+ */
+public interface HashFunctionProvider extends Serializable {
+    HashFunction getHashFunction(TableId tableId, Schema schema);
```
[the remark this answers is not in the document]
dingxin-tech: Sure.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunction.java
dingxin-tech: done.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/HashFunctionProvider.java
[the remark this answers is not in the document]
lvyanquan: Usually, it will need to obtain partition or bucket information based on TableId.
I wonder if it is possible to cache catalog or connection here to reuse objects or connections.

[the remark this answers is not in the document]
leonardBang: A function with lifecycle management makes sense to me, why the factory need to care the resource management? could we push the logic that `use TableId to obtain partition or buckets info` to `HashFunction` internal? The HashFunction need to open/close  may be better, what do you think?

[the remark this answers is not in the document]
lvyanquan: Actually, HashFunctionalProvider is similar to the role of a `catalog`, and HashFunction is similar to the role of a `table`, so the open/lose method is called on the `catalog` role.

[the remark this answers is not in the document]
leonardBang: Well, Catalog is used as a metadata manager instance instead of a `table` factory,  `table` is not constructed by the instance. A metadata manager has its life cycle makes sense to me, a factory own lifecycle confuse me a little. Come back to the function itself, Flink also has many functions as well as the function factories, I didn't see the necessary why a function factory need to care the required resource in runtime, but function manages its resources is pretty common.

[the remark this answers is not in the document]
leonardBang: You can also visit flink code about:

- RichFunction & SourceFunctionProvider

- InputFormat & InputFormatProvider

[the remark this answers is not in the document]
lvyanquan: So the key point is that what we need is a metadata manager or a function factory. I agree that what we need is one function factory, though it may need the assistance of specific metadata manager.

And the resource utilization is indeed overthinking, because we can release database connection after extracting the parameters for calculation.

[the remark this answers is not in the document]
leonardBang: Yeah, that's what I want to propose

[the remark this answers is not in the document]
dingxin-tech: In my opinion, I believe that HashFunction should not have lifecycle functions. It is merely a hash function, not a Flink operator, and it will be cached and disappear automatically. Its lifecycle can be perceived as a kind of "constant" after creation, so we should not concern ourselves with its lifecycle.

Based on this viewpoint, whether HashFunctionProvider should have lifecycle functions depends solely on whether we need to reuse some resources created at runtime when creating a HashFunction.

In fact, I also think that currently, the connectors do not have any tasks that need to be done within lifecycle functions. Therefore, I have removed the lifecycle functions for now.

dingxin-tech:
hi, @leonardBang, can you help to review again and merge this?

leonardBang:
> hi, @leonardBang, can you help to review again and merge this?

I append one commit to polish the interface and package, Could you also take a look ?

dingxin-tech:
> > hi, @leonardBang, can you help to review again and merge this?
> 
> I append one commit to polish the interface and package, Could you also take a look ?

sure and looks good for me.

leonardBang (approved):

PR #3421, opened 2024-06-18
[minor][cdc-connector][sqlserver] Fix some words error

gong:
Fix some words error


gong:
@leonardBang @GOODBOY008 PTAL

PR #3429, opened 2024-06-20
[FLINK-35633][cli] Verify YAML pipeline definitions early

yuxiqian:
This closes FLINK-35633.

Currently, transform & route rules will be passed to Flink operators as strings, and any malformed configurations will not be exposed until runtime, which makes testing and debugging harder.

This PR adds early verification logic in Flink CDC CLI, and tries to provide more clear and descriptive error messages for input YAML files.


yuxiqian:
Not a priority for now.

PR #3434, opened 2024-06-24
[FLINK-37017][cdc-common] Supports map and array types binary record data

umeshdangat:


umeshdangat:
I was writing a custom flink cdc source connector. Input data has complex nested types. That is how I ran into the issue of BinaryRecordData not supporting complex types such as ARRAY and MAP, including AbstractBinaryWriter not supporting those.  Most of the code here is copied from flink source (including relevant tests).

I was also trying to get my data eventually into paimon, using paimon sink. Thus I have made relevant changes to PaimonWriterHelper as well to handle complex types.

Lastly, I came across a bug due to circular dependency between DataFieldSerializer, DataTypeSerializer, and RowTypeSerializer classes that causes NPE for certain RowTypes like nested row type. I have baked that change in this PR but also have a separate one [here with more explanation](https://github.com/apache/flink-cdc/pull/3400)

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -0,0 +1,574 @@
...
+ * <pre>
+ * [size(int)] + [null bits(4-byte word boundaries)] + [values or offset&length] + [variable length part].
+ * </pre>
+ */
+public final class BinaryArrayData extends BinarySection implements ArrayData {
+
+    /** Offset for Arrays. */
+    private static final int BYTE_ARRAY_BASE_OFFSET = UNSAFE.arrayBaseOffset(byte[].class);
+
+    private static final int BOOLEAN_ARRAY_OFFSET = UNSAFE.arrayBaseOffset(boolean[].class);
```
loserwang1024, line 54: Now that an array only contain one type, why should we get all kind of ${type}_array_offset? I think one offset is enough.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -0,0 +1,574 @@
...
+
+    // The number of elements in this array
+    private int size;
+
+    /** The position to start storing array elements. */
+    private int elementOffset;
+
+    public BinaryArrayData() {}
+
+    private void assertIndexIsValid(int ordinal) {
```
loserwang1024, line 113: What about rename ordinal to index? Same name style with BinaryRecordData would be easier to read.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -0,0 +1,574 @@
...
+        return size;
+    }
+
+    @Override
+    public void pointTo(MemorySegment[] segments, int offset, int sizeInBytes) {
+        // Read the number of elements from the first 4 bytes.
+        final int size = BinarySegmentUtils.getInt(segments, offset);
+        assert size >= 0 : "size (" + size + ") should >= 0";
+
+        this.size = size;
```
loserwang1024, line 133: set sizeInBytes, offset and segments by super.pointTo

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -0,0 +1,574 @@
...
+import org.apache.flink.core.memory.MemorySegment;
+import org.apache.flink.core.memory.MemorySegmentFactory;
+
+import java.lang.reflect.Array;
+
+import static org.apache.flink.core.memory.MemoryUtils.UNSAFE;
+
+/**
+ * A binary implementation of {@link ArrayData} which is backed by {@link MemorySegment}s.
+ *
```
loserwang1024, line 39: Please add more description link what BinaryRecordData does. For example, he first 4 bytes stores size(or length) of array....

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -0,0 +1,574 @@
...
+
+    private static final int BOOLEAN_ARRAY_OFFSET = UNSAFE.arrayBaseOffset(boolean[].class);
+    private static final int SHORT_ARRAY_OFFSET = UNSAFE.arrayBaseOffset(short[].class);
+    private static final int INT_ARRAY_OFFSET = UNSAFE.arrayBaseOffset(int[].class);
+    private static final int LONG_ARRAY_OFFSET = UNSAFE.arrayBaseOffset(long[].class);
+    private static final int FLOAT_ARRAY_OFFSET = UNSAFE.arrayBaseOffset(float[].class);
+    private static final int DOUBLE_ARRAY_OFFSET = UNSAFE.arrayBaseOffset(double[].class);
+
+    public static int calculateHeaderInBytes(int numFields) {
+        return 4 + ((numFields + 31) / 32) * 4;
```
loserwang1024, line 62: I don't understand what it means? Could you please give me more details?
1. what's the meaning of each data of this formula？
2. why not  ((4 +numFields + 31) / 32) * 4;
3. you mean each header is 4 bytes? But in calculateFixLengthPartSize, BIGINT or DOUBLE will return 8.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -0,0 +1,574 @@
...
+                segments, elementOffset, values, BYTE_ARRAY_BASE_OFFSET, size);
+        return values;
+    }
+
+    @Override
+    public short[] toShortArray() {
+        checkNoNull();
+        short[] values = new short[size];
+        BinarySegmentUtils.copyToUnsafe(
+                segments, elementOffset, values, SHORT_ARRAY_OFFSET, size * 2);
```
loserwang1024, line 451: Why not get SHORT_ARRAY_OFFSET when used rather than when initialization?

loserwang1024 (commented):
Hi, @umeshdangat , thanks for your hard job. The PR is so big and critical that I just review part of BinaryArrayData. Here is some comments.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
umeshdangat: As per my understanding:
The calculateHeaderInBytes method is used to determine the size of the header for a binary array in bytes. The header consists of the size of the array AND a bitmap indicating which elements are null.

- Fixed Size Path: The 4 at the beginning represents the number of bytes used to store the size of the array. This is a fixed overhead for any binary array, regardless of the number of elements.
- Null Bitmap: The second part of the calculation is ((numFields + 31) / 32) * 4. This part determines the number of bytes required to store the null bitmap.


**Why ((numFields + 31) / 32) * 4?:**

numFields: This is the number of elements in the array.
+ 31 and / 32: This is a mathematical trick to ensure that we have enough bits to represent all the elements in the array. Each element in the array needs 1 bit in the null bitmap.
" +31" ensures that we round up to the next multiple of .32 (to ensure word boundary) so we always allocate enough bits
"/32" converts the number of elements to number of 32 bit words needed. 
Example:
numFields = 64
1) fixed part will have 4 bytes with integer value 64
2) (64+31)/32 = 95/32 = 2.96875 = 3 (rounded up)
     then convert words to bytes: 3*4 = 12 bytes for bitmap

so for example 64 entries we have 
4 bytes fixed size plus
12 bytes for bitmap (that is 12*8 = 96 bits so space for upto 96 entries)

Overall the formula  at top of this file

`[size(int)] + [null bits(4-byte word boundaries)] + [values or offset&length] + [variable length part].`

represents how data is stored. 
Example 1:  array of 64 integers
[4 bytes (size)] + [12 bytes (null bits)] + [256 bytes (values)] = 272 bytes total

Example 2: array of 64 variable-length elements:
[4 bytes (size)] + [12 bytes (null bits)] + [64 * 8 bytes (offset & length)] + [variable length data]

loserwang1024:
@umeshdangat , thanks for you explain. More comment detail in java docs is always a good thing.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
umeshdangat: done.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
umeshdangat: done.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
umeshdangat: done

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
umeshdangat: added docString to this method.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
umeshdangat: are you suggesting removal of static variable SHORT_ARRAY_OFFSET and computing it directly instead?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
umeshdangat: The reason we maintain different offsets for each primitive type is because our implementation provides methods to convert the binary data into specific primitive type arrays, such as `toShortArray()` and `toIntArray()`. Each primitive type array has a different base offset in memory, which is why we initialize the respective offsets:

- SHORT_ARRAY_OFFSET for short[]
- INT_ARRAY_OFFSET for int[]
- And similarly for other types like boolean, long, float, and double.
These offsets are used to accurately copy the binary data into the appropriate primitive type arrays.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
loserwang1024: I just wonder why design like it.

umeshdangat:
@loserwang1024 @leonardBang can we re-open this? I would still like to get this looked at and merged as I feel it is valuable to handle MapData and ArrayData in BinaryRecordData (as supported in flink source)

herunkang2018:
@umeshdangat feel free to reopen it. And I would like to review this PR if there are any updates.

lvyanquan:
Hi, @umeshdangat gentle ping.
I think that this pr is an important base for many other following features, and we can try to merge it as soon as possible.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/schema/DataFieldSerializer.java
```diff
@@ -39,7 +39,14 @@ public class DataFieldSerializer extends TypeSerializerSingleton<DataField> {
     public static final DataFieldSerializer INSTANCE = new DataFieldSerializer();
 
     private final StringSerializer stringSerializer = StringSerializer.INSTANCE;
-    private final DataTypeSerializer dataTypeSerializer = new DataTypeSerializer();
+    private DataTypeSerializer dataTypeSerializer;
+
+    private DataTypeSerializer getDataTypeSerializer() {
```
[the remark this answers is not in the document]
umeshdangat: > I came across a bug due to circular dependency between DataFieldSerializer, DataTypeSerializer, and RowTypeSerializer classes that causes NPE for certain RowTypes like nested row type. I have baked that change in this PR but also have a separate one https://github.com/apache/flink-cdc/pull/3400

[explained earlier here](https://github.com/apache/flink-cdc/pull/3434#issuecomment-2192072774)

umeshdangat:
> I guess that some classes refer to Flink like [BinaryArrayData](https://github.com/apache/flink/blob/b48ccce1ce201ea76e11f089f43e8c115b96a992/flink-table/flink-table-common/src/main/java/org/apache/flink/table/data/binary/BinaryArrayData.java#L52) , so I think that we should avoid adding unnecessary change and add enough tests.

The reason to not reuse [BinaryArrayData](https://github.com/apache/flink/blob/b48ccce1ce201ea76e11f089f43e8c115b96a992/flink-table/flink-table-common/src/main/java/org/apache/flink/table/data/binary/BinaryArrayData.java#L52) from flink source code is that flink-cdc appears to have its own interface defined for [ArrayData](https://github.com/apache/flink-cdc/blob/master/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/ArrayData.java#L42) which is then used in [RecordData](https://github.com/apache/flink-cdc/blob/master/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/RecordData.java#L154). RecordData in flink-cdc essentially a port of [RowData](https://github.com/apache/flink/blob/b48ccce1ce201ea76e11f089f43e8c115b96a992/flink-table/flink-table-common/src/main/java/org/apache/flink/table/data/RowData.java#L192) interface from flink.

Also there are existing re-implementations of flink classes like GenericArrayData in flink-cdc. So there is prior art to this pattern as well.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/MapDataSerializer.java
```diff
@@ -0,0 +1,327 @@
...
+
+    @Override
+    public TypeSerializerSnapshot<MapData> snapshotConfiguration() {
+        return new MapDataSerializer.MapDataSerializerSnapshot(
+                keyType, valueType, keySerializer, valueSerializer);
+    }
+
+    /** {@link TypeSerializerSnapshot} for {@link MapDataSerializer}. */
+    public static final class MapDataSerializerSnapshot implements TypeSerializerSnapshot<MapData> {
+        private static final int CURRENT_VERSION = 3;
```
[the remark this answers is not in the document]
umeshdangat: done

flink-cdc-common/src/test/java/org/apache/flink/cdc/common/data/binary/BinarySegmentUtilsTest.java
```diff
@@ -0,0 +1,196 @@
...
+        bytes2[3] = 81;
+        bytes2[4] = 81;
+
+        assertThat(BinaryRecordDataDataUtil.byteArrayEquals(bytes1, bytes2, 4)).isTrue();
+        assertThat(BinaryRecordDataDataUtil.byteArrayEquals(bytes1, bytes2, 5)).isFalse();
+        assertThat(BinaryRecordDataDataUtil.byteArrayEquals(bytes1, bytes2, 0)).isTrue();
+    }
+
+    //    @Test
+    //    public void testBoundaryEquals() {
```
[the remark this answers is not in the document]
umeshdangat: uncommented test and added relevant utility helper code

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/serializer/data/writer/BinaryArrayWriterTest.java
```diff
@@ -0,0 +1,565 @@
...
+    //        RawValueDataSerializer<String> serializer =
+    //                new RawValueDataSerializer<>(StringSerializer.INSTANCE);
+    //        writer.writeRawValue(0, generic, serializer);
+    //        writer.setNullAt(1);
+    //        writer.complete();
+    //
+    //        RawValueData<String> newGeneric = array.getRawValue(0);
+    //        assertThat(newGeneric).satisfies(matching(equivalent(generic, serializer)));
+    //        assertThat(array.isNullAt(1)).isTrue();
+    //    }
```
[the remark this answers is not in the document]
umeshdangat: There is no `RawValueData` interface and dont think its even needed yet in flink-cdc codebase? It does exist flink. I removed this test case now.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -0,0 +1,574 @@
...
+            case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
+            case ARRAY:
+            case MAP:
+            case ROW:
+                // long and double are 8 bytes;
+                // otherwise it stores the length and offset of the variable-length part for types
+                // such as is string, map, etc.
+                return 8;
+            case TIMESTAMP_WITH_TIME_ZONE:
+                throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
umeshdangat: I was using the [BinaryArrayData](https://github.com/apache/flink/blob/master/flink-table/flink-table-common/src/main/java/org/apache/flink/table/data/binary/BinaryArrayData.java#L99) implementation from flink source where it appears to do the same.  

We cannot simply re-use the flink BinaryArrayData for reasons I mention in [this comment](https://github.com/apache/flink-cdc/pull/3434#issuecomment-2518863198)

umeshdangat:
@lvyanquan @loserwang1024 @herunkang2018 @leonardBang 
I believe I have addressed all outstanding comments on this PR. Could we continue moving the review forward?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
umeshdangat: Its a port of the BinaryArrayData from flink source as mentioned [here](https://github.com/apache/flink-cdc/pull/3434#issuecomment-2518863198)

leonardBang:
@umeshdangat Thanks for your updating, and we will have a final review for this PR ASAP.

loserwang1024 (approved):
LGTM!

umeshdangat:
@loserwang1024 thanks for your review. Is this PR now good to merge?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -0,0 +1,623 @@
...
+                return 4;
+            default:
+                throw new IllegalArgumentException();
+        }
+    }
+
+    // The number of elements in this array
+    private int size;
+
+    /** The position to start storing array elements. */
```
herunkang2018, line 159: Minor comment, would you like to unify code comment format? Seems here use two comment formats.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryMapData.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.flink.cdc.common.types.DataType;
+import org.apache.flink.core.memory.MemorySegment;
+import org.apache.flink.core.memory.MemorySegmentFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.cdc.common.utils.Preconditions.checkArgument;
+
+/**
```
herunkang2018, line 31: Could you add more description for this class?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryMapData.java
```diff
@@ -0,0 +1,118 @@
...
+ * [4 byte(keyArray size in bytes)] + [Key BinaryArray] + [Value BinaryArray].
+ *
+ * <p>{@code BinaryMap} are influenced by Apache Spark UnsafeMapData.
```
herunkang2018, line 34: Is this a typo, BinaryMap -> BinaryMapData?

herunkang2018 (commented):
A few minor comments, other looks good to me. By the way, could you rename the PR's title to include the Flink's Jira number and Flink CDC's module that you changed, like cdc-runtime?

leonardBang (approved):
Thanks @umeshdangat for the contribution and @loserwang1024 for the review, CI passed, merging...

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryMapData.java
leonardBang: @herunkang2018 thanks for you comment, I think they're nice supplement for this PR, but I plan to merge this PR firstly as it waited to long, and we can improve it after that

PR #3435, opened 2024-06-26
[FLINK-35700][cli] Loosen CDC pipeline options validation

yuxiqian:
This closes FLINK-35700.

FLINK-35121 adds pipeline configuration validation, rejecting any unknown options, which turns out to be too strict, and it's not possible to create customized configuration extensions. Also, Flink doesn't reject unknown entries in flink-conf / config.yaml, just silently ignores them. It might be better for CDC to follow such behavior.


yuxiqian:
@leonardBang @loserwang1024 PTAL

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/PipelineDef.java
```diff
@@ -150,19 +148,6 @@ public static void validatePipelineDefinition(Configuration configuration)
...
-                        key -> {
-                            if (!optionKeys.contains(key)) {
-                                throw new ValidationException(
-                                        String.format("Unknown configuration key `%s`", key));
-                            }
-                        });
-
         options.forEach(
                 option -> {
                     if (!configuration.getOptional(option).isPresent()
```
loserwang1024, line 153: To be honest, I also don't understand why unspecified configuration without default value is error?

yuxiqian: Since required / optional configuration in pipeline is vague (and undocumented), I've removed all pipeline-level validations.

leonardBang:
> @leonardBang @loserwang1024 PTAL

+1 to follow Flink Conf‘s validation style

loserwang1024:
LGTM!

leonardBang (approved):
+1

PR #3438, opened 2024-06-27
[FLINK-35713][cdc-compose] Add sink PARALLELISM for flink-cdc.

proletarians:


yuxiqian:
Thanks for @proletarians' contribution! Kindly remind that `mvn spotless:apply` should be run before committing to fix any code style violations.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/pipeline/PipelineOptions.java
yuxiqian, line 1: In the future, defining multiple sink in one single pipeline job might be supported, and one might want to define parallelisms for each sink individually. Will it be better if we put `sink.parallelism` as an option in `sink:` block instead of a global pipeline option?

proletarians: Got it! I will follow up on this.

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
```diff
@@ -93,6 +93,7 @@ private FlinkPipelineComposer(StreamExecutionEnvironment env, boolean isBlocking
     @Override
     public PipelineExecution compose(PipelineDef pipelineDef) {
         int parallelism = pipelineDef.getConfig().get(PipelineOptions.PIPELINE_PARALLELISM);
+        int sinkParallelism = pipelineDef.getConfig().get(PipelineOptions.SINK_PARALLELISM);
```
ruanhang1993, line 96: What if users do not set the SINK_PARALLELISM setting?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/pipeline/PipelineOptions.java
```diff
@@ -45,6 +45,15 @@ public class PipelineOptions {
                     .noDefaultValue()
                     .withDescription("Parallelism of the pipeline");
 
+
+
+    public static final ConfigOption<Integer> SINK_PARALLELISM =
+            ConfigOptions.key("sink.parallelism")
+                    .intType()
+                    .noDefaultValue()
+                    .withDescription("Parallelism of the sink in the pipeline");
```
ruanhang1993, line 54: Please add some tests for the setting.

yuxiqian:
Hi @proletarians, you may want to create individual branches for each PR to avoid mixing different changes together.

PR #3440, opened 2024-06-28
[FLINK-35718][mysql] cherry pick DBZ-5333 from debezium MySqlErrorHandler and add logs.

lvyanquan:
copy from https://github.com/debezium/debezium/blob/1.9/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlErrorHandler.java.


lvyanquan:
@ruanhang1993 CC.

PR #3454, opened 2024-07-05
[FLINK-35762] Cache hashCode for TableID & Transform descriptors

yuxiqian:
This closes FLINK-35762.

As previously discussed offline with @lvyanquan, HashMap lookup could be slow since we must calculate key hash code every time, though most key objects are immutable.

A common practice in Java world is caching object hash code in a transient field, just like what `java.lang.String` does. In CDC code, `TableId` & transform rule types are frequently used as HashMap key type, and their hash code could be safely cached since their instances are immutable.


lvyanquan (approved):
LGTM.

PR #3460, opened 2024-07-08
[minor][test] Remove duplicate test code

Karl-WangSK:
find some duplicate test code, need to remove to save test time


Karl-WangSK:
@yuxiqian @leonardBang PTAL

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/source/NewlyAddedTableITCase.java
```diff
@@ -214,29 +214,6 @@ public void testNewlyAddedTableForExistsPipelineThriceWithAheadBinlog() throws E
...
-            throws Exception {
-        testNewlyAddedTableOneByOne(
-                1,
-                FailoverType.NONE,
-                FailoverPhase.NEVER,
-                true,
-                "address_hangzhou",
-                "address_beijing");
-    }
-
```
leonardBang, line 239: @Karl-WangSK Thanks for your contribution, but they are different cases, why we need to remove the two cases?

PR #3468, opened 2024-07-10
[FLINK-35805][transform] Add `__data_event_type__` metadata column

yuxiqian:
This closes FLINK-35805.

Currently, event operation type (insert, update, or delete) was encoded in all `DataChangeEvents`. However it's not possible to access them in transform blocks since it is not exposed as metadata column. Adding this should allow such transform usages:

```yaml
transform:
  - projection: \*, __data_event_type__ == '-D' AS logical_delete_tag
```

### DIFF
--- a/flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposerITCase.java
+++ b/flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposerITCase.java
@@ -342,6 +342,66 @@ class FlinkPipelineComposerITCase {
 
     @ParameterizedTest
     @EnumSource
+    void testOpTypeMetadataColumn(ValuesDataSink.SinkApi sinkApi) throws Exception {
+        FlinkPipelineComposer composer = FlinkPipelineComposer.ofMiniCluster();
+
+        // Setup value source
+        Configuration sourceConfig = new Configuration();
+        sourceConfig.set(
+                ValuesDataSourceOptions.EVENT_SET_ID,
+                ValuesDataSourceHelper.EventSetId.TRANSFORM_TABLE);
+        SourceDef sourceDef =
+                new SourceDef(ValuesDataFactory.IDENTIFIER, "Value Source", sourceConfig);
+
+        // Setup value sink
+        Configuration sinkConfig = new Configuration();
+        sinkConfig.set(ValuesDataSinkOptions.MATERIALIZED_IN_MEMORY, true);
+        sinkConfig.set(ValuesDataSinkOptions.SINK_API, sinkApi);
+        SinkDef sinkDef = new SinkDef(ValuesDataFactory.IDENTIFIER, "Value Sink", sinkConfig);
+
+        // Setup transform
+        TransformDef transformDef =
+                new TransformDef(
+                        "default_namespace.default_schema.table1",
+                        "*,concat(col1,'0') as col12,__data_event_type__ as rk",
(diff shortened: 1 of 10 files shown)


aiwenmo (approved):
LGTM

flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposerITCase.java
```diff
@@ -335,6 +337,65 @@ void testTransform(ValuesDataSink.SinkApi sinkApi) throws Exception {
...
+        // Execute the pipeline
+        PipelineExecution execution = composer.compose(pipelineDef);
+        execution.execute();
+
+        // Check the order and content of all received events
+        String[] outputEvents = outCaptor.toString().trim().split("\n");
+        assertThat(outputEvents)
+                .containsExactly(
+                        "CreateTableEvent{tableId=default_namespace.default_schema.table1, schema=columns={`col1` STRING,`col2` STRING,`__op_type__` STRING NOT NULL,`col12` STRING,`op` STRING}, primaryKeys=col1, partitionKeys=col12, options=({key1=value1})}",
+                        "DataChangeEvent{tableId=default_namespace.default_schema.table1, before=[], after=[1, 1, INSERT, 10, INSERT], op=INSERT, meta=()}",
```
lvyanquan, line 390: Should we keep in consistence with https://nightlies.apache.org/flink/flink-cdc-docs-release-3.1/docs/connectors/flink-sources/mysql-cdc/#available-metadata to use `row_kind` as key and return `+I`, `-D`,`+U` ? But we don't actually have `+U` & `-U` kind, and This also has historical issues for `database_name` and `table_name` as we used another keys for them before.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformDataOperator.java
```diff
@@ -412,4 +417,8 @@ private void clearOperator() {
         this.transformFilterProcessorMap = null;
         TransformExpressionCompiler.cleanUp();
     }
+
+    private String opTypeToRowKind(OperationType opType, boolean isBefore) {
+        return String.format("%c%c", opType.name().charAt(0), (isBefore ? '-' : '+'));
```
lvyanquan, line 422: opType.name().charAt(0) should be put at second position.

lvyanquan (approved):

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
```diff
@@ -77,6 +77,7 @@ public class TransformParser {
     public static final String DEFAULT_NAMESPACE_NAME = "__namespace_name__";
     public static final String DEFAULT_SCHEMA_NAME = "__schema_name__";
     public static final String DEFAULT_TABLE_NAME = "__table_name__";
+    public static final String DEFAULT_ROW_KIND = "__row_kind__";
```
leonardBang, line 80:
```suggestion
    public static final String DEFAULT_DATA_EVENT_TYPE = "__data_event_type__";
```

yuxiqian:
Rebased with `master` and cleaned up metadata column definitions for a bit.

leonardBang (approved):
Thanks @yuxiqian for the update, LGTM

leonardBang:
CI passed, merging...

PR #3471, opened 2024-07-12
[Kafka]Kafka sink numRecordsOutByIncremental metrics

ChengJie1053:
numRecordsOutByIncremental metrics collects incremental statistics

numRecordsOutByIncremental has three metrics, respectively is numRecordsOutByIncrementalInsert, numRecordsOutByIncrementalUpdate, numRecordsOutByIncrementalDelete

image

2024.7.20
image

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/main/java/org/apache/flink/cdc/connectors/kafka/sink/KafkaSinkWriteMetrics.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/main/java/org/apache/flink/cdc/connectors/kafka/sink/KafkaSinkWriteMetrics.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.connectors.kafka.sink;
+
+import org.apache.flink.cdc.common.event.OperationType;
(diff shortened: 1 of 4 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/main/java/org/apache/flink/cdc/connectors/kafka/sink/KafkaSinkWriteMetrics.java
```diff
@@ -0,0 +1,80 @@
...
+    private final Map<String, Counter> numRecordsOutMap = new ConcurrentHashMap();
+
+    public KafkaSinkWriteMetrics(MetricGroup metricGroup) {
+        this.metricGroup = metricGroup;
+    }
+
+    public void numRecordsOutInsertIncrease(TableId tableId) {
+        Counter counter =
+                numRecordsOutMap.computeIfAbsent(
+                        tableId.identifier() + OperationType.INSERT,
```
yuxiqian, line 51: As previously pointed out by @lvyanquan, identifier string generating & hashing could be expensive considering it must be done for every DataChangeEvent. What about splitting `INSERT`, `UPDATE`, and `DELETE` counters into three `Map<TableId, Counter>` hash map, so the `TableId` key hash code could be safely cached after we merge #3454?

ChengJie1053: Ok, thank you for helping me review the code, I will modify this part

lvyanquan:
I have two points to be discussed.
1. Does 'numRecordsOutByIncremental' refer to the incremental phase here? We don't have a way to determine whether we have reached this phase in the sink now.
2. If we want to add these metrics, I suggest adding it uniformly though [DataSinkWriterOperator](https://github.com/apache/flink-cdc/blob/94a0415475e7040cc6a61e9bd8aca9c3ccea80ad/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataSinkWriterOperator.java#L63) for all pipeline sinks.

ChengJie1053:
> I have two points to be discussed.
> 
> 1. Does 'numRecordsOutByIncremental' refer to the incremental phase here? We don't have a way to determine whether we have reached this phase in the sink now.
> 2. If we want to add these metrics, I suggest adding it uniformly though [DataSinkWriterOperator](https://github.com/apache/flink-cdc/blob/94a0415475e7040cc6a61e9bd8aca9c3ccea80ad/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataSinkWriterOperator.java#L63) for all pipeline sinks.

Ok, thanks for reviewing the code for me
1.`numRecordsOutByIncremental` the name may be ambiguous, I will modify the name
2.I adding it uniformly though DataSinkWriterOperator for all pipeline sinks.

PR #3475, opened 2024-07-16
[docs] Add schema evolution type supporting status for pipeline sink connectors

yuxiqian:
Currently, not all pipeline sink connectors support every schema change events, and some have extra limitations when applying schema evolution events. Explicitly stating them in documentations should be helpful for users.


yuxiqian:
I'll leave it as draft since FLINK-35243 is expected to add more schema change events. More revision is needed on this.

PR #3476, opened 2024-07-17
add note **Please note that the lib here refers to the lib of flink-c…

Lordeath:
add note **Please note that the lib here refers to the lib of flink-cdc, not the lib of flink. Do not put pipeline-related packages under the lib of flink, otherwise it will cause class loading related problems**


Lordeath:
重复修改，取消这次PR

PR #3483, opened 2024-07-20
[minor][cdc-runtime]The processedTableIds add is executed twice

ChengJie1053:
image

The processedTableIds add is executed twice


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataSinkWriterOperator.java
```diff
@@ -172,7 +172,6 @@ public void processElement(StreamRecord<Event> element) throws Exception {
             emitLatestSchema(changeEvent.tableId());
             processedTableIds.add(changeEvent.tableId());
```
yuxiqian, line 173: Should this be removed or the outside one?

yuxiqian (approved):
Nice catch!

PR #3488, opened 2024-07-23
[FLINK-35874][cdc-connector][mysql] Check pureBinlogPhaseTables set before call getBinlogPosition method

qiaozongmi:
This closes [FLINK-35874]
Optimize and avoid frequent invocations of getBinlogPosition by moving the check pureBinlogPhaseTables.contains(tableId) in the hasEnterPureBinlogPhase method before calling getBinlogPosition. This way, if the SourceRecord belongs to a pure binlog phase table, we can directly return true without the need for the highly performance-consuming getBinlogPosition method.


qiaozongmi:
@leonardBang PTAL

lvyanquan (approved):
Nice catch, I've found and checked this optimization before in local code.

leonardBang (approved):
+1

PR #3496, opened 2024-07-25
[docs][minor] Add transform piece for pipeline example

leonardBang:
Add transform piece for pipeline example


ruanhang1993 (approved):
LGTM.

PR #3502, opened 2024-07-31
[FLINK-35936][cdc-connector][paimon] Fix paimon cdc schema evolution failure when restart job

MOBIN-F:
paimon cdc schema evolution failure when restart job

Minimal reproduce step：

1. stop flink-cdc-mysql-to-paimon pipeline job
2. alter mysql table schema, such as add column
2. start pipeline job
3. the newly added column was not synchronized to the paimon table


leonardBang:
Thanks @MOBIN-F for the contribution, could you add description for the PR ?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/test/java/org/apache/flink/cdc/connectors/paimon/sink/PaimonMetadataApplierTest.java
```diff
@@ -416,4 +416,87 @@ public void testAddColumnWithPosition(String metastore)
         Assertions.assertEquals(
                 tableSchema, catalog.getTable(Identifier.fromString("test.table1")).rowType());
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"filesystem", "hive"})
+    public void testOfflineAlterExistsTableSchema(String metastore)
```
lvyanquan, line 422: So the expected scenario to be solved is actually a scenario where the table structures of downstream that was previously created are inconsistent with upstream(mysql)?

MOBIN-F: Yes

lvyanquan: I think this change carries risks, I tend to add a sink option to explicitly support this matter:
When the upstream and downstream table structures are inconsistent, users can manually modify the table structure to adapt.
If this is done in the program, in the case where the user configures the wrong `route` rule (which is not a particularly rare situation), the situation where the table structure does not correspond does not throw an exception. We will forcibly modify downstream tables to adapt, and the user will not be aware of this error.

MOBIN-F: thanks, You are right, I only considered the \* situation before

MOBIN-F: @lvyanquan Could you please review it again? thank you

lvyanquan:
Hi @MOBIN-F.
I think that this can be closed as it was solved in https://github.com/apache/flink-cdc/pull/4081.
I believe that modifying the data to adapt to the downstream table structure is a safer solution compared to modifying the downstream table structure through the task to accommodate the data. What do you think?

MOBIN-F:
> Hi @MOBIN-F. I think that this can be closed as it was solved in #4081. I believe that modifying the data to adapt to the downstream table structure is a safer solution compared to modifying the downstream table structure through the task to accommodate the data. What do you think?

agree~

PR #3473, opened 2024-07-15
[FLINK-35072][doris] Support applying AlterColumnTypeEvent to Doris sink

yuxiqian:
This closes FLINK-35072.

Thanks to @qg-lin's great work, doris-flink-connector 1.6.2 has implemented column type modification API, so Flink CDC could implement this ticket more easily.

~~Notice that this PR isn't ready for merge yet, since doris-flink-connector 1.6.2 artifact has not been released. Local compilation & tests are required to verify this change for now.~~

PR #3339, opened 2024-05-21
[FLINK-35242] Supports per-SE type configuration & "lenient" evolution behavior

yuxiqian:
This closes FLINK-35242.

* Adds `TRY_EVOLVE` behavior that tolerates exception during metadata applying process
* Supports configuring various behavior for each schema evolution type with `include.schema.changes` and `exclude.schema.changes` sink option
* Adds schema evolution IT cases

---

With this change, Schema Operator now stores both "upstream" schema (which always keep up with the structure coming from pipeline source **after the transformation**) and "evolved" schema (the schema that actually applied to sink). They might differ when:

* Schema change events are partially or fully ignored (by fine-grained options / `IGNORE` behavior)
* Some schema change events failed to apply but pipeline keeps executing (in `TRY_EVOLVE` mode)

If these cases occur, `SchemaOperator` needs to cast upstream schema to match the actual evolved version (by adding / removing columns, converting types, etc.) with tolerance (fill in `null` if meaningful casts are not possible.)

PR #3503, opened 2024-07-31
[hotfix][doris] DorisMetadataApplier fails to reject unsupported schema change events

yuxiqian:
This fixed `DorisMetadataApplier` failed to throw exceptions for an unsupported SchemaChangeEvent, which was introduced in #3473.


yuxiqian:
Addressed by #3339.

PR #3504, opened 2024-08-01
[FLINK-35938][cdc-connect][paimon] use filterAndCommit API for retried Committables.

lvyanquan:
use filterAndCommit API for retried Committables.


leonardBang (approved):
+1， thanks @lvyanquan for the hotfix

PR #3506, opened 2024-08-02
[hotfix][cdc-connector][mongodb] Fix unstable test cases for snapshot back-filling

yuxiqian:
This fixes a rarely happened failure case, observed in https://github.com/apache/flink-cdc/actions/runs/10209766277/job/28253831363?pr=3388 by @loserwang1024.

`testSnapshotOnlyModeWithDMLPreHighWaterMark` case will fail, if there's no operation or heartbeat events between event `#a` (the ChangeStream opLog caused by the last event in the hook) and event `#b` (the calculated high watermark that limits the bounded back-filling stream fetch task), `#a` and `#b` will point to the same offset in change stream, thus the last event of hook will be missed since back-filling task only reads data between `[loW, hiW)` (high watermark not included).

This is a workaround by insert a dummy event in another collection to forcefully push opLog forward, so `#a` will not be the same as the high watermark `#b`.


yuxiqian:
Could @Jiabao-Sun @loserwang1024 please take a look?

Jiabao-Sun (approved):
Thanks @yuxiqian, LGTM.

PR #3511, opened 2024-08-06
[FLINK-35715][common] Ignore the compact optimize for mysql timestamp type in BinaryRecordData

ruanhang1993:
This PR fix FLINK-35715 and change to use same logic for timestamp types with different precision in the BinaryRecordData.


yuxiqian:
Seems Flink TIMESTAMP has a range limit from 0000-01-01 00:00:00.000000000 to 9999-12-31 23:59:59.999999999[1]. According to current binary encoding format, the millisecond part will be `253,402,300,799,999` at most, far less than the long limit (`9,223,372,036,854,775,807`). The higher 16 bits of `long` is actually wasted.

Is it possible to encode the "is compact or not" information on these unused higher bits?

Ignore my comment if it's too hard to implement.

[1] https://nightlies.apache.org/flink/flink-docs-release-1.19/docs/dev/table/types/

ruanhang1993:
> Seems Flink TIMESTAMP has a range limit from 0000-01-01 00:00:00.000000000 to 9999-12-31 23:59:59.999999999[1]. According to current binary encoding format, the millisecond part will be `253,402,300,799,999` at most, far smaller than the largest value that `Long` could carry (`9,223,372,036,854,775,807`). The higher 16 bits are actually wasted.
> 
> Is it possible to encode the "is compact or not" information on these unused higher bits?
> 
> Ignore my comment if it's too hard to implement.
> 
> [1] https://nightlies.apache.org/flink/flink-docs-release-1.19/docs/dev/table/types/

Thanks for this suggestion. I think this could be implemented in the next version as an optimization.

yuxiqian (approved):

leonardBang (approved):
Thanks @ruanhang1993 for the improve and @yuxiqian for the review, +1

PR #3513, opened 2024-08-07
[FLINK-35991][FLINK-36034][transform] Resolve conflicts in transform SQL operator tables

yuxiqian:
This closes FLINK-35991 and FLINK-36034.

Currently, transform operator chains self-defined SqlFunctions and standard function tables, which might be a problem since there are some conflicting function declarations.

Getting rid of `SqlStdOperatorTable` and declaring supported functions only should resolve this problem.

Also, current implementation unreasonably relies on Flink table planner, causing trouble in Kubernetes deployment mode. Getting rid of it should solve this.

### DIFF
--- a/flink-cdc-runtime/pom.xml
+++ b/flink-cdc-runtime/pom.xml
@@ -46,34 +46,6 @@ limitations under the License.
         </dependency>
         <dependency>
             <groupId>org.apache.flink</groupId>
-            <artifactId>flink-table-planner_2.12</artifactId>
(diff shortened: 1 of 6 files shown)


yuxiqian:
cc @aiwenmo

leonardBang (approved):
Thanks @yuxiqian for the fix， LGTM， wait the CI green

leonardBang:
CI passed, merging...

PR #3514, opened 2024-08-07
[tests] Run E2e tests on Flink 1.20 & Split E2e tests

yuxiqian:
This bumps Flink version used by E2e test cases.

* 1.19.0 -> 1.19.1
*  NEW   -> 1.20.0

Also noticed that after this change, the whole E2e test would take nearly 90 minutes to finish, which is close to default GitHub action timeout limit. Separating pipeline and source E2e cases might help.


yuxiqian:
cc @GOODBOY008

leonardBang (approved):
+1

PR #3516, opened 2024-08-07
[hotfix][e2e] Add missing default values field in E2e test

yuxiqian:
FLINK-34638 just implemented default value synchronization support.

However, `MySqlE2eITCase` wasn't updated correspondingly to validate check data records with default values, causing E2e cases' failure.


leonardBang (approved):
Thanks @yuxiqian for the quick fix, LGTM

PR #3520, opened 2024-08-08
[FLINK-36007][[cdc-composer] cdc load factory and add jar in once search

loserwang1024:
As shown https://issues.apache.org/jira/browse/FLINK-36007 , once search is enough for cdc load factory and add jar url.

### DIFF
--- a/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/utils/FactoryDiscoveryUtils.java
+++ b/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/utils/FactoryDiscoveryUtils.java
@@ -20,9 +20,10 @@ package org.apache.flink.cdc.composer.utils;
 import org.apache.flink.cdc.common.annotation.Internal;
 import org.apache.flink.cdc.common.factories.Factory;
 
-import org.codehaus.commons.nullanalysis.NotNull;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


loserwang1024:
@leonardBang @ruanhang1993 ,CC

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/utils/FactoryDiscoveryUtils.java
```diff
@@ -20,6 +20,7 @@
 import org.apache.flink.cdc.common.annotation.Internal;
 import org.apache.flink.cdc.common.factories.Factory;
 
+import org.codehaus.commons.nullanalysis.NotNull;
```
leonardBang, line 23: import javax.annotation.Nonnull;

leonardBang (commented):
Thanks @loserwang1024 for the optimization, left one minor comment

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/utils/FactoryDiscoveryUtils.java
loserwang1024: done

leonardBang (approved):

PR #3522, opened 2024-08-09
[FLINK-35442][cdc-connect][kafka] add key.format and partition.strategy option

lvyanquan:
To make sure the same record sending to the same partition after job restarted.


PatrickRen (approved):
@lvyanquan Thanks for the contribution! LGTM

PR #3523, opened 2024-08-09
[FLINK-36023][cdc-composer] Hotfix: Flink CDC K8S Native Application …

loserwang1024:
As shown in https://issues.apache.org/jira/browse/FLINK-36023, there is some problem:
* if user summit job not in Flink K8S Native Application Mode, and put jar in /mypackage/usrlib will be casted to &{FLINK_HOME}/usrlib.
* if user summit job in Flink K8S Native Application Mode , jar in local:///opt/flink/usrlib/ have already been added to classpath of JM and TM, thus no need to add jar anymore.


PatrickRen (approved):
@loserwang1024 Thanks for the patch! LGTM

PR #3528, opened 2024-08-09
[FLINK-35984][transform] Fix: Job crashes when metadata column names present in transform rules

MOBIN-F:
since metadata column existence check was done by searching identifier string in statement, without considering any syntax info

In the transform rule (projection: \*, '_namespace_nameschema_nametable_name_' AS string_literal) mentioned in FLINK-35984, Caused by: java.lang.IllegalArgumentException: wrong number of arguments is thrown because of the "\*" symbol. The PreTransformOperator introduced in FLINK-35272 fixes the "\*" problem, but there are still some scenarios where Caused by: java.lang.IllegalArgumentException: wrong number of arguments is thrown due to metaColumn name conflicts. This PR will fix these bugs.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
@@ -499,7 +499,7 @@ public class TransformParser {
     }
 
     private static List<Column> copyFillMetadataColumn(List<Column> columns) {
-        // Add meteColumn for SQLValidater.validate
+        // Add metaColumn for SQLValidater.validate
         List<Column> columnsWithMetadata = new ArrayList<>(columns);
         columnsWithMetadata.add(Column.physicalColumn(DEFAULT_NAMESPACE_NAME, DataTypes.STRING()));
         columnsWithMetadata.add(Column.physicalColumn(DEFAULT_SCHEMA_NAME, DataTypes.STRING()));
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
@@ -499,7 +499,7 @@ public class TransformParser {
     }
 
     private static List<Column> copyFillMetadataColumn(List<Column> columns) {
-        // Add metaColumn for SQLValidater.validate
+        // Add metaColumn for SQLValidator.validate
         List<Column> columnsWithMetadata = new ArrayList<>(columns);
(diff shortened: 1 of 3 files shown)


MOBIN-F:
cc @yuxiqian

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
```diff
@@ -498,29 +498,15 @@ private static SqlSelect parseProjectionExpression(String projection) {
         return parseSelect(statement.toString());
     }
 
-    private static List<Column> copyFillMetadataColumn(
-            String transformStatement, List<Column> columns) {
+    private static List<Column> copyFillMetadataColumn(List<Column> columns) {
+        // Add meteColumn for SQLValidater.validate
```
leonardBang, line 502: metaColumn

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/UnifiedTransformOperatorTest.java
```diff
@@ -917,6 +917,52 @@ public void testMetadataAndCalculatedTransform() throws Exception {
                 .destroyHarness();
     }
 
+    @Test
+    public void testMetadataTransformIncludeMetaColumnString() throws Exception {
+        TableId tableId = TableId.tableId("my_company", "my_branch", "schema_nullability");
+        UnifiedTransformTestCase.of(
+                        tableId,
+                        "id, name, age, id + age as computed, __table_name__ as metaCol1, '__table_name__' as metaCol2, '__namespace__name__schema__name__table__name__' as metaCol3, UPPER(__schema_name__) as metaCol4",
```
leonardBang, line 925: could you also add `__namespace_name__ as metaColNameSpaceName`, `__schema_name__ as metaColSchemaName` in this test?

leonardBang (commented):
Thanks @MOBIN-F for the contribution, I only left two minor comments

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/UnifiedTransformOperatorTest.java
MOBIN-F: done

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
MOBIN-F: done

yuxiqian (approved):
Looks great!

leonardBang (approved):
+1, wait the CI green

PR #3537, opened 2024-08-13
[FLINK-35985][transform] transform rules support substring function

MOBIN-F:
Currently, the startIndex of flink-cdc.substr starts from 0, while the startIndex of flink.substr starts from 1, and the startIndex of flink-cdc.substr cannot be less than 0. This is inconsistent with flink.substr and may cause trouble for users. This PR will solve these problems and provide a substring ([Refer to flink-sql substring function](https://github.com/apache/flink/blob/6fcf4555391487661ffab1e72f6d0e2875bca6a4/flink-table/flink-table-runtime/src/main/java/org/apache/flink/table/runtime/functions/SqlFunctionUtils.java#L653)) function to keep consistency with Flink SQL


flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperatorTest.java
```diff
@@ -1499,7 +1499,18 @@ void testBuildInFunctionTransform() throws Exception {
...
+        testExpressionConditionTransform("SUBSTR('ABC', 0) = 'ABC'");
+        testExpressionConditionTransform("SUBSTR('ABC', 1) = 'ABC'");
+        testExpressionConditionTransform("SUBSTR('ABC', 2, 2) = 'BC'");
+        testExpressionConditionTransform("SUBSTR('ABC', 2, 100) = 'BC'");
+        testExpressionConditionTransform("SUBSTRING('ABC', -1) = 'C'");
+        testExpressionConditionTransform("SUBSTRING('ABC', -2, 2) = 'BC'");
+        testExpressionConditionTransform("SUBSTRING('ABC', 0) = 'ABC'");
+        testExpressionConditionTransform("SUBSTRING('ABC', 1) = 'ABC'");
+        testExpressionConditionTransform("SUBSTRING('ABC', 2, 2) = 'BC'");
+        testExpressionConditionTransform("SUBSTRING('ABC', 2, 100) = 'BC'");
```
yuxiqian, line 1513: Seems Flink uses `SUBSTRING ... FROM ... FOR ... ` style instead of plain function call. Could this be supported?

MOBIN-F: yes,In flinksql, both substring('ABC',2,2) and substring('ABC' FROM 2 FOR 2) are supported.
done

yuxiqian (approved):
Thanks for @MOBIN-F's quick fix!

yuxiqian:
Could @ruanhang1993 please take a look and remove the Stale label?

leonardBang (approved):
Thanks @MOBIN-F  for the fix and @yuxiqian for the review, +1

PR #3538, opened 2024-08-14
Optimize log level

melin:


yuxiqian:
Hi @melin, seems the error log message you want to emit is `Handling schema request` in `SchemaRegistry`, which will be printed when an **schema lookup request** occurs. Changing it to DEBUG level is reasonable.

However, the log you've just modified will be printed only once for `CreateTableEvent`s, which is significant and should not be classified to DEBUG level.

Maybe the log code you want to change is
```java
LOG.info("Handling evolved schema request: {}", getEvolvedSchemaRequest);
```
and
```java
LOG.info("Handling original schema request: {}", getOriginalSchemaRequest);
```

in SchemaRegistry.java?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaManager.java
```diff
@@ -224,7 +224,7 @@ private void handleCreateTableEvent(
                 !schemaExists(schemaMap, event.tableId()),
                 "Unable to apply CreateTableEvent to an existing schema for table \"%s\"",
                 event.tableId());
-        LOG.info("Handling schema change event: {}", event);
+        LOG.debug("Handling schema change event: {}", event);
```
yuxiqian, line 227: Seems this is not the text that floods your output log.

yuxiqian (approved):
Thanks for @melin's work, changes look reasonable to me.

PR #3540, opened 2024-08-14
[FLINK-36054][cdc][build] Fix Flink CDC parent pom and release scripts

PatrickRen:
This pull request fixes several issues in Flink CDC parent pom and release script:

In parent pom:
- Parent pom should use artifact ID `flink-cdc-parent` instead of `flink-cdc`
- Distribution management pointing to Sonatype should be removed, as we should use Apache repository
- Parent pom should include correct SCM information
- Parent pom should use `apache-release` profile for releasing

In release script:
- LICENSE file should be included in the binary release
- Source release prefix should be `flink-cdc` instead of `apache-flink-cdc`


leonardBang (approved):
+1

PR #3548, opened 2024-08-18
[FLINK-36081][cdc-base][mysql] Fix that MySQL source connector missing some columns data of newly added tables

kevinwangcs:
**Problem Description:**

When adding a new table, the Flink CDC MySQL source connector experiences missing data for some columns of the newly added table.

**Reproduction Scenario:**

1. Remove a table from a cdc job that is running normally, then start the job with resume functionality.
2. Perform a column addition operation on the removed table.
3. Add the table back to the job. The job continues to run without interruption upon table addition, but data for the newly added columns is missing in the synchronized data.

**Cause Analysis:**

The issue arises because the MySQL CDC Source maintains the table schema in state. When adding a new table, it recovers the schema from the previous state. Since the prior schema exists and represents the structure before the column addition, the MySQL CDC Source provides the downstream with data based on the schema cached in the state. Consequently, records outputted to downstream systems are missing the fields corresponding to the newly added columns.

**Proposed Solution:**

Upon removing a table from the cdc job, it is necessary to also correspondingly remove the table from the MySQLBinlogSplit.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/split/MySqlBinlogSplit.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/split/MySqlBinlogSplit.java
@@ -183,30 +183,47 @@ public class MySqlBinlogSplit extends MySqlSplit {
      *
      * <p>When restore from a checkpoint, the finished split infos may contain some splits from the
      * deleted tables. We need to remove these splits from the total finished split infos and update
-     * the size.
+     * the size, while also removing the outdated tables from the table schemas of binlog split.
      */
     public static MySqlBinlogSplit filterOutdatedSplitInfos(
             MySqlBinlogSplit binlogSplit, Tables.TableFilter currentTableFilter) {
-        Set<TableId> tablesToRemove =
+        Map<TableId, TableChange> filteredTableSchemas =
+                binlogSplit.getTableSchemas().entrySet().stream()
+                        .filter(entry -> currentTableFilter.isIncluded(entry.getKey()))
+                        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+        Set<TableId> tablesToRemoveInFinishedSnapshotSplitInfos =
                 binlogSplit.getFinishedSnapshotSplitInfos().stream()
                         .filter(i -> !currentTableFilter.isIncluded(i.getTableId()))
                         .map(split -> split.getTableId())
                         .collect(Collectors.toSet());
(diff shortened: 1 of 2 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/split/MySqlBinlogSplit.java
```diff
@@ -183,17 +183,29 @@ public static MySqlBinlogSplit appendFinishedSplitInfos(
      *
      * <p>When restore from a checkpoint, the finished split infos may contain some splits from the
      * deleted tables. We need to remove these splits from the total finished split infos and update
-     * the size.
+     * the size, while also removing the outdated tables from the table schemas of binlog split.
      */
     public static MySqlBinlogSplit filterOutdatedSplitInfos(
             MySqlBinlogSplit binlogSplit, Tables.TableFilter currentTableFilter) {
+        Map<TableId, TableChange> newTableSchemas =
```
ruanhang1993, line 190:
```suggestion
        Map<TableId, TableChange> filteredTableSchemas =
```

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/split/MySqlBinlogSplit.java
```diff
@@ -183,17 +183,29 @@ public static MySqlBinlogSplit appendFinishedSplitInfos(
...
+                binlogSplit.getTableSchemas().entrySet().stream()
+                        .filter(entry -> currentTableFilter.isIncluded(entry.getKey()))
+                        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
         Set<TableId> tablesToRemove =
```
ruanhang1993, line 195:
```suggestion
        Set<TableId> tablesToRemoveInFinishedSnapshotSplitInfos =
```

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/split/MySqlBinlogSplit.java
kevinwangcs: Thanks for the review, [ruanhang1993](https://github.com/ruanhang1993), I've applied your suggestions.

ruanhang1993 (approved):
LGTM

PR #3549, opened 2024-08-18
[FLINK-36082][pipeline-connector][kafka] fix NotSerializableException for KafkaDataSink.

lvyanquan:
Avoid using lambda class.


lvyanquan:
@PatrickRen  @yuxiqian  CC as you may interested.

PatrickRen (approved):
@lvyanquan Thanks for the patch! LGTM

PR #3556, opened 2024-08-19
[FLINK-36076] Set isSchemaChangeApplying as isSchemaChangeApplying for thread safe.

loserwang1024:
As discussed in https://issues.apache.org/jira/browse/FLINK-36076 , Current, we apply schema change asynchronous.  
SchemaChangeThreadPool will set isSchemaChangeApplying as true when apply schema change event. And the main thread which handle operater event will check isSchemaChangeApplying.
 
For thread safe, this should be volatile.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistryRequestHandler.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistryRequestHandler.java
@@ -94,7 +94,7 @@ public class SchemaRegistryRequestHandler implements Closeable {
     private final Set<Integer> flushedSinkWriters;
 
     /** Status of the execution of current schema change request. */
-    private boolean isSchemaChangeApplying;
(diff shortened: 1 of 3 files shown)


loserwang1024:
@leonardBang @PatrickRen @ruanhang1993 , CC

leonardBang:
The PR title and commit content confuse me a little, @loserwang1024 could you check your PR ?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistryRequestHandler.java
```diff
@@ -94,7 +94,7 @@ public class SchemaRegistryRequestHandler implements Closeable {
     private final Set<Integer> flushedSinkWriters;
 
     /** Status of the execution of current schema change request. */
-    private boolean isSchemaChangeApplying;
+    private transient boolean isSchemaChangeApplying;
```
leonardBang, line 97: volatile?

leonardBang (approved):
Thanks for the update, +1

PR #3557, opened 2024-08-19
[FLINK-36092] Fix schema evolution with wildcarded transform rules

yuxiqian:
Currently, transform doesn't work very well with schema evolution.

The main issue is that TransformOperator (including Pre- and Post-) simply passes all upstream schema change events to downstream, which isn't correct under most cases.

Here's corresponding operations this PR tries to do in various transform rule definitions:

|                      | `*, ...`                                                         | `..., *, ...`                                   | `..., *`                                 | No asterisk                                 |
|----------------------|------------------------------------------------------------------|-------------------------------------------------|------------------------------------------|---------------------------------------------|
| AddColumnEvent       | Yes, but LAST position needs reordering (LAST -> AFTER) | Yes, but FIRST / LAST position needs reordering | Yes, but FIRST position needs reordering | No                                          |
| AlterColumnTypeEvent | Yes                                                              | Yes                                             | Yes                                      | Maybe, if that column was referenced           |
| RenameColumnEvent    | Yes                                                              | Yes                                             | Yes                                      | No, exception if that column was referenced |
| DropColumnEvent      | Yes                                                              | Yes                                             | Yes                                      | No, exception if that column was referenced |


yuxiqian:
Considering this is blocking users from using schema evolution with transform blocks, could @leonardBang @aiwenmo please take a look?

leonardBang (commented):
Thanks @yuxiqian for the fix, the change generally looks good to me except one comment

leonardBang:
@yuxiqian The CI failed

yuxiqian:
Seems Postgres and OceanBase CI is hanging, unlikely to be relevant to this PR.

leonardBang (approved):
CI passed after retry, merging...

PR #3561, opened 2024-08-21
[FLINK-35056] Fix conversion of SQL Server TIMESTAMP to Flink

morozov:
When introspecting a TIMESTAMP column from SQL Server, the connector should use the scale (in Debezium schema terms) as the value for the Flink TIMESTAMP precision, not length.

Otherwise, the connector attempts to convert `datetime` SQL Server type, which Debezium introspects as `datetime(23, 3)`, to Flink's `TIMESTAMP(23)` and fails:

> org.apache.flink.table.api.ValidationException: Timestamp precision must be between 0 and 9 (both inclusive).


leonardBang (approved):
Thanks @morozov for the fix, LGTM

leonardBang:
@morozov Would you like to open a PR to backport the fix to `release-3.2` branch?

PR #3562, opened 2024-08-21
Add insert-only change stream option

henneberger:
Adds the option to do an insert-only stream for connectors that support specifying a changelog mode. This allows debezium connectors to be imported into flink as stream tables to take advantage of flink's strong stream processing capabilities. There is currently no way to convert a all/upsert changelog stream to an insert-only stream with just the flink table api so this appears to be the only workaround.


henneberger:
Let me know if there is anything I should add or do to help close out this PR. Thanks!

PR #3563, opened 2024-08-22
[FLINK-36114] Make `SchemaRegistryRequestHandler` thread safe by blocking more schema change events

yuxiqian:
This closes FLINK-36114.

Currently, SchemaRegistry asynchronously receives schema change requests from SchemaOperator, and results of multiple requests might got mixed up together, causing incorrect logic flow in multiple parallelism cases.

Changing SchemaRegistry's behavior to accept requests in serial should resolve this problem.

---

This PR also changes MySQL pipeline source behavior.

Previously, some subTask might not emit CreateTableEvents if there are no corresponding SnapshotSplit assigned to them. This might cause problems if they are assigned some StreamSplits later, since their downstream Transform nodes never received CreateTableEvent and doesn't know about the schema.

By forcing MySQL source emitting CreateTableEvents when transiting from Snapshot to BinLog stage should resolve this problem.

---

This PR also adjusts Pipeline E2e test case parallelism to 1 to 4 to verify this change.


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
```diff
@@ -435,13 +421,26 @@ private void handleSchemaChangeEvent(TableId tableId, SchemaChangeEvent schemaCh
 
     private SchemaChangeResponse requestSchemaChange(
             TableId tableId, SchemaChangeEvent schemaChangeEvent) {
-        return sendRequestToCoordinator(new SchemaChangeRequest(tableId, schemaChangeEvent));
+        while (true) {
+            SchemaChangeResponse response =
+                    sendRequestToCoordinator(new SchemaChangeRequest(tableId, schemaChangeEvent));
+            if (response.isRejected()) {
+                try {
+                    LOG.info("Schema Registry is busy now, waiting for next request...");
```
leonardBang, line 429: Do we have a timeout seetting?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistryRequestHandler.java
```diff
@@ -107,16 +101,69 @@ public SchemaRegistryRequestHandler(
...
+                                            .ifPresent(pe::fillPreSchema);
+                                }
+                            }
+                        });
+                currentDerivedSchemaChangeEvents = new ArrayList<>(derivedSchemaChangeEvents);
+                return CompletableFuture.completedFuture(
+                        wrap(new SchemaChangeResponse(derivedSchemaChangeEvents)));
+            }
+        } else {
+            LOG.info("There is already a schema change request in progress. Rejected {}.", request);
```
leonardBang, line 164: how about just tell SchemaOperator a BUSY signal instead of a REJECT signal？

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistryRequestHandler.java
```diff
@@ -107,16 +101,69 @@ public SchemaRegistryRequestHandler(
...
+            LOG.info(
+                    "Received schema change event request {} from table {}. Start to buffer requests for others.",
+                    request.getSchemaChangeEvent(),
+                    request.getTableId().toString());
+            SchemaChangeEvent event = request.getSchemaChangeEvent();
+            if (schemaManager.isOriginalSchemaChangeEventRedundant(event)) {
+                LOG.info("Event {} has been addressed before, ignoring it.", event);
+                finishCurrentSchemaChangeRequest();
+                return CompletableFuture.completedFuture(
+                        wrap(new SchemaChangeResponse(Collections.emptyList())));
```
leonardBang, line 137: Could we add some message in SchemaChangeResponse body to let clients know more details? it will be helpful for troubleshooting

leonardBang:
Thanks @yuxiqian for the contribution, I left some comments @loserwang1024 Would you like to take a look at this PR whrn you have time?

loserwang1024 (commented):
I have left some comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
```diff
@@ -412,36 +404,62 @@ private void handleSchemaChangeEvent(TableId tableId, SchemaChangeEvent schemaCh
...
+            throw new IllegalStateException("Unexpected response status " + response);
         }
     }
 
     private SchemaChangeResponse requestSchemaChange(
-            TableId tableId, SchemaChangeEvent schemaChangeEvent) {
-        return sendRequestToCoordinator(new SchemaChangeRequest(tableId, schemaChangeEvent));
+            TableId tableId, SchemaChangeEvent schemaChangeEvent)
+            throws InterruptedException, TimeoutException {
+        long schemaEvolveTimeOutMillis = System.currentTimeMillis() + rpcTimeOutInMillis;
```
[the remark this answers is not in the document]
yuxiqian: This has been exposed as a pipeline configuration as `schema-operator.rpc-timeout`.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
```diff
@@ -412,36 +404,62 @@ private void handleSchemaChangeEvent(TableId tableId, SchemaChangeEvent schemaCh
...
-        // The request will need to send a FlushEvent or block until flushing finished
+        // The request will block if another schema change event is being handled
         SchemaChangeResponse response = requestSchemaChange(tableId, schemaChangeEvent);
-        if (!response.getSchemaChangeEvents().isEmpty()) {
-            LOG.info(
-                    "Sending the FlushEvent for table {} in subtask {}.",
-                    tableId,
-                    getRuntimeContext().getIndexOfThisSubtask());
+        if (response.isAccepted()) {
+            LOG.info(">{} Sending the FlushEvent for table {}.", subTaskId, tableId);
```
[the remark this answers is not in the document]
yuxiqian: This is intended to imitate how Values data sink prints subTaskId to distinguish them:

```java
if (print) {
    String prefix = numSubtasks > 1 ? subtaskIndex + "> " : "";
    // print the detail message to console for verification.
    System.out.println(
            prefix
                    + ValuesDataSinkHelper.convertEventToStr(
                            event,
                            fieldGetterMaps.get(((ChangeEvent) event).tableId())));
}
```

leonardBang (approved):
Thanks @yuxiqian for the cooperation fix, LGTM, I believe your effort will make the schema evolution more stable.

PR #3573, opened 2024-08-25
[FLINK-36150][pipeline-connector/mysql] tables.exclude is still valid if scan.binlog.newly-added-table.enabled is true.

loserwang1024:
As described in https://issues.apache.org/jira/browse/FLINK-36150 , this PR aims to pass `tables.exclude` to MySqlSourceConfig and will use it when find tables from mysql database.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/schema/Selectors.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/schema/Selectors.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.connectors.mysql.schema;
+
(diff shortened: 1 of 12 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/schema/Selectors.java
```diff
@@ -0,0 +1,119 @@
...
+import io.debezium.relational.TableId;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Predicate;
+
+/** Selectors for filtering tables. */
+public class Selectors {
```
ruanhang1993, line 31: Why do we need to add this class ?

ruanhang1993 (commented):
Could you please add some IT tests for this in the mysql source cdc connector?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/schema/Selectors.java
loserwang1024: Because we need to translate tables.exclude to a filter. We can not get it by dbzMySqlConfig. 
```suggestion
public class Selectors {z
```

loserwang1024: Because we need to translate tables.exclude to a filter. We can not get it by dbzMySqlConfig. getTableFilters which need the whole debezium config.
I alse want to use org.apache.flink.cdc.common.schema.Selectors, but  it is not so good:
1. the param is org.apache.flink.cdc.common.event.TableId rather than io.debezium.relational.TableId, so I can use it directly in mysql cdc source.
2. org.apache.flink.cdc.common.schema.Selectors.SelectorsBuilder#includeTables will parse "databaseName.tableName" as <null, databaseName, tableName>, the org.apache.flink.cdc.connectors.mysql.source.utils.TableDiscoveryUtils#listTables will return <databaseName, null, tableName>. Thus, the selector cannot filter table from mysql cdc.

loserwang1024:
> Could you please add some IT tests for this in the mysql source cdc connector?

I have done it in org.apache.flink.cdc.connectors.mysql.source.MysqlPipelineNewlyAddedTableITCase#testScanBinlogNewlyAddedTableEnabledAndExcludeTables

ruanhang1993 (approved):

PR #3574, opened 2024-08-25
[FLINK-36128] Promote LENIENT as the default schema change behavior

yuxiqian:
This closes FLINK-36128.

Currently, default schema evolution mode "EVOLVE" could not handle exceptions, and might not be able to restore from existing state correctly after failover. Before we can "trigger checkpoints on demand" which is not possible prior to in Flink 1.19, making "LENIENT" a default option might be more suitable.

This PR also contains a hotfix patch for PrePartitionOperator broadcasting failure.

### DIFF
--- a/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
+++ b/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
@@ -19,6 +19,8 @@ package org.apache.flink.cdc.cli.parser;
 
 import org.apache.flink.cdc.common.configuration.Configuration;
 import org.apache.flink.cdc.common.event.SchemaChangeEventType;
+import org.apache.flink.cdc.common.event.SchemaChangeEventTypeFamily;
+import org.apache.flink.cdc.common.pipeline.SchemaChangeBehavior;
 import org.apache.flink.cdc.common.utils.StringUtils;
 import org.apache.flink.cdc.composer.definition.PipelineDef;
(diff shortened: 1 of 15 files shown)


flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
```diff
@@ -172,6 +183,23 @@ private SinkDef toSinkDef(JsonNode sinkNode) {
...
+            // default.
+            Arrays.stream(SchemaChangeEventTypeFamily.ALL)
+                    .map(SchemaChangeEventType::getTag)
+                    .forEach(includedSETypes::add);
+        }
+
+        if (excludedSETypes.isEmpty()
+                && SchemaChangeBehavior.LENIENT.equals(schemaChangeBehavior)) {
+            // In lenient mode, we exclude DROP_TABLE and TRUNCATE_TABLE by default. This could be
+            // overridden by manually specifying excluded types.
```
leonardBang, line 197: minor: we should add schema behavior  document ASAP

leonardBang (approved):
Thanks @yuxiqian for the fix, + 1 for the point we have to enable lenient mode as default before flink 1.19, LGTM

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
yuxiqian: Tracked with FLINK-36151.

leonardBang (approved):
CI passed, merging

PR #3576, opened 2024-08-26
[hotfix] Keep upstream pending requests in order to avoid checkpoint hanging & state inconsistency in timestamp startup mode

yuxiqian:
This PR fixes:

* Schema registry hanging due to out-of-order requests
* Fix job failure when visiting DropTableEvent without a prior CreateTableEvent when starting up with timestamp mode


leonardBang (approved):

PR #3577, opened 2024-08-27
[minor][cdc-runtime] Run schema coordinator logic asynchronously to avoid blocking the main thread

yuxiqian:
This tweaks `SchemaRegistry` coordinator to let it handle requests / trigger checkpoints in background threads to avoid blocking the main thread.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistry.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistry.java
@@ -128,7 +128,7 @@ public class SchemaRegistry implements OperatorCoordinator, CoordinationRequestH
                 executorService,
                 metadataApplier,
                 routes,
-                SchemaChangeBehavior.EVOLVE);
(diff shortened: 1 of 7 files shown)


leonardBang:
@loserwang1024 Would you like to review this PR when you have time ?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistry.java
```diff
@@ -113,18 +119,27 @@ public class SchemaRegistry implements OperatorCoordinator, CoordinationRequestH
...
             MetadataApplier metadataApplier,
             List<RouteRule> routes) {
-        this(operatorName, context, metadataApplier, routes, SchemaChangeBehavior.EVOLVE);
+        this(
+                operatorName,
+                context,
+                executorService,
+                metadataApplier,
+                routes,
+                SchemaChangeBehavior.EVOLVE);
```
leonardBang, line 131: LENIENT ？

leonardBang (approved):
Thanks @yuxiqian for the fix, +1. The failed case reason is clear and not related to this PR, will merge this PR first.

PR #3579, opened 2024-08-28
[hotfix][mysql] Fix primary key restraints missing when using inline `PRIMARY KEY` declaration syntax

yuxiqian:
Currently, it is possible to parse correct CreateTableEvents from Binlog with the following syntax:

```sql
CREATE TABLE table1 (
   id INT,
   ... -- other columns
   PRIMARY KEY (id)
);
```

However, inlined primary key declaration syntax is not possible:

```sql
CREATE TABLE table1 (
   id INT PRIMARY KEY,
   ... -- other columns
);
```

This should be resolved by adding missing `enterPrimaryKeyColumnConstraint` handlers in `CustomColumnDefinitionParserListener`.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
@@ -81,8 +81,8 @@ public class CustomAlterTableParserListener extends MySqlParserBaseListener {
         TableId tableId = parser.parseQualifiedTableId(ctx.tableName().fullId());
         if (parser.databaseTables().forTable(tableId) == null) {
             tableEditor = parser.databaseTables().editOrCreateTable(tableId);
-            super.enterColumnCreateTable(ctx);
         }
+        super.enterColumnCreateTable(ctx);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
```diff
@@ -113,14 +113,17 @@ public void exitColumnCreateTable(MySqlParser.ColumnCreateTableContext ctx) {
...
+                            new CreateTableEvent(
+                                    toCdcTableId(tableEditor.tableId()), builder.build()));
                 },
                 tableEditor);
-        Schema.Builder builder = Schema.newBuilder();
-        tableEditor.columns().forEach(column -> builder.column(toCdcColumn(column)));
-        if (tableEditor.hasPrimaryKey()) {
-            builder.primaryKey(tableEditor.primaryKeyColumnNames());
-        }
-        changes.add(new CreateTableEvent(toCdcTableId(tableEditor.tableId()), builder.build()));
```
lvyanquan, line 123: Is this change necessary or for code style standardization？

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlPipelineITCase.java
```diff
@@ -648,6 +648,20 @@ public void testSchemaChangeEvents() throws Exception {
                                     .physicalColumn("timestamp6_c", DataTypes.TIMESTAMP_LTZ(6))
                                     .primaryKey("id")
                                     .build()));
+
+            // Test create table DDL with inline primary key
+            statement.execute(
+                    String.format(
+                            "CREATE TABLE `%s`.`newlyAddedTable2`(id SERIAL PRIMARY KEY);",
```
lvyanquan, line 655: Further, we can add one more test case like:
```
CREATE TABLE my_table (
    id INT,
    name VARCHAR(255),
    PRIMARY KEY (id, name)
);
```

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
yuxiqian: I thought it makes more sense to put it inside `runIfNotNull` block, since referencing `tableEditor` from the outside implicitly assumes `tableEditor` is not null.

lvyanquan (approved):

leonardBang (approved):

PR #3580, opened 2024-08-28
[minor][tests] Fix `testDanglingDroppingTableDuringBinlogMode` tests case due to imprecise timestamp startup

yuxiqian:
In order to test if pipeline job could correctly handle dangling `DropTableEvent`s (those ones without preceding `CreateTableEvent`s) during binlog reading phase, this PR uses `specific-offset` startup mode to cover such case instead of `timestamp` mode which is not so precise and prone to cause CI failure.

### DIFF
--- a/flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/MysqlE2eITCase.java
+++ b/flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/MysqlE2eITCase.java
@@ -354,6 +354,8 @@ public class MysqlE2eITCase extends PipelineTestEnvironment {
             }
         }
 
+        // We start reading binlog from the tail of current position and file to avoid reading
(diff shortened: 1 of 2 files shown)


flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/MysqlE2eITCase.java
```diff
@@ -333,16 +335,32 @@ public void testSchemaChangeEvents() throws Exception {
...
+                Statement statement = connection.createStatement()) {
+            statement.execute("CREATE TABLE live_fast(ID INT PRIMARY KEY);");
+        }
+
+        String logFileName = null;
+        Long logPosition = null;
+
+        try (Connection connection = mysqlInventoryDatabase.getJdbcConnection();
+                Statement statement = connection.createStatement()) {
+            ResultSet rs = statement.executeQuery("SHOW BINARY LOGS;");
```
leonardBang, line 350: IIUC we can add a note: one more transaction to push the binlog file latest position move forward ?

leonardBang (approved):
Thanks for the update, LGTM

PR #3590, opened 2024-08-29
[FLINK-36178][pipeline-connector/mysql] Parse `CREATE TABLE ... LIKE …` syntax.

lvyanquan:
Support parsing SQL like following and emit new SchemaChangeEvent to downstream.
```
CREATE TABLE `user_log_copy` LIKE `user_log`
```
and
```
CREATE TABLE `user_log_copy` AS SELECT * FROM `user_log`
```

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
@@ -34,6 +34,7 @@ import io.debezium.ddl.parser.mysql.generated.MySqlParser;
 import io.debezium.ddl.parser.mysql.generated.MySqlParserBaseListener;
 import io.debezium.relational.Column;
 import io.debezium.relational.ColumnEditor;
+import io.debezium.relational.Table;
 import io.debezium.relational.TableEditor;
 import io.debezium.relational.TableId;
 import org.antlr.v4.runtime.tree.ParseTreeListener;
@@ -74,6 +75,29 @@ public class CustomAlterTableParserListener extends MySqlParserBaseListener {
         this.parser = parser;
         this.listeners = listeners;
         this.changes = changes;
+    }
+
+    @Override
+    public void exitCopyCreateTable(MySqlParser.CopyCreateTableContext ctx) {
+        TableId tableId = parser.parseQualifiedTableId(ctx.tableName(0).fullId());
+        TableId originalTableId = parser.parseQualifiedTableId(ctx.tableName(1).fullId());
+        Table original = parser.databaseTables().forTable(originalTableId);
+        if (original != null) {
+            parser.databaseTables()
+                    .overwriteTable(
+                            tableId,
+                            original.columns(),
+                            original.primaryKeyColumnNames(),
+                            original.defaultCharsetName());
+            parser.signalCreateTable(tableId, ctx);
+            Schema.Builder builder = Schema.newBuilder();
+            original.columns().forEach(column -> builder.column(toCdcColumn(column)));
+            if (!original.primaryKeyColumnNames().isEmpty()) {
+                builder.primaryKey(original.primaryKeyColumnNames());
+            }
+            changes.add(new CreateTableEvent(toCdcTableId(tableId), builder.build()));
+        }
+        super.exitCopyCreateTable(ctx);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


lvyanquan:
The test case failure in OceanBase is unrelated with this pr.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
```diff
@@ -76,6 +77,29 @@ public CustomAlterTableParserListener(
         this.changes = changes;
     }
 
+    @Override
+    public void exitCopyCreateTable(MySqlParser.CopyCreateTableContext ctx) {
+        TableId tableId = parser.parseQualifiedTableId(ctx.tableName(0).fullId());
+        TableId originalTableId = parser.parseQualifiedTableId(ctx.tableName(1).fullId());
+        Table original = parser.databaseTables().forTable(originalTableId);
```
leonardBang, line 84: Do we need to consider some config options that makes the parser only parse captured table list?

ruanhang1993: I think we need to skip parsing the filtered tables.

leonardBang:
@ruanhang1993 Would you like to help review this PR?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
lvyanquan: The original sql was already filtered by table name in https://github.com/apache/flink-cdc/blob/32ec9f57e6be7f8152d5e27cbf2423dc23055b6b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java#L660

lvyanquan:
Hi @yuxiqian @leonardBang, PTAL.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
```diff
@@ -82,6 +83,29 @@ public CustomAlterTableParserListener(
...
+            parser.signalCreateTable(tableId, ctx);
+            Schema.Builder builder = Schema.newBuilder();
+            original.columns().forEach(column -> builder.column(toCdcColumn(column)));
+            if (!original.primaryKeyColumnNames().isEmpty()) {
+                builder.primaryKey(original.primaryKeyColumnNames());
+            }
+            changes.add(new CreateTableEvent(toCdcTableId(tableId), builder.build()));
+        }
+        super.exitCopyCreateTable(ctx);
+    }
```
yuxiqian, line 113: Print a warning log if `CREATE TABLE LIKE` original table isn't presented in the capture list?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlPipelineITCase.java
yuxiqian, line 1: Can we add similar cases for MySQL CDC pipeline connectors with `scan.binlog.newly-added-table.enabled`?

IIUC if we can't obtain new schema from `CREATE TABLE ... LIKE ...` DDL, triggering failover and re-scanning table schemas should recover the job.

yuxiqian: Also, `MySqlPipelineRecordEmitter` may detect this case by checking if there are "DataChangeEvents" without corresponding "CreateTableEvent" and throw exceptions in advance.

yuxiqian (commented):
Thanks for @lvyanquan's nice work, I left some minor comments.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
lvyanquan: Done.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlPipelineITCase.java
lvyanquan: The MySqlDataSource here already used regular expressions for matching newly added tables.

PR #3592, opened 2024-09-02
[FLINK-35982] Fix transform metadata config doesn't work if no projection block was provided

yuxiqian:
This closes FLINK-35982.

One may readjust source table primary keys, partition keys, table options by specifying them in a Transform block like this:

```yaml
transform:
  - source-table: db.tbl
    projection: '*'
    primary-keys: order_id, product_name
    partition-keys: order_id
    table-options: bucket=1
```

However, if the projection field is omitted (which fallbacks to default behavior that does not change any source table columns), such configuration will not take effect:

```yaml
transform:
  - source-table: db.tbl
    primary-keys: order_id, product_name
    partition-keys: order_id
    table-options: bucket=1 # These options will not apply
```

The underlying reason is transform blocks without `projection` was simply ignored by `PreTransformOperator`, which isn't accurate since other options like primary keys, partition keys should work without projecting original columns.


yuxiqian:
cc @ruanhang1993 @lvyanquan

lvyanquan (approved):

leonardBang (approved):
+1

PR #3595, opened 2024-09-05
[FLINK-36211][pipeline-connector/kafka] shade org.apache.flink.streaming.connectors.kafka

lvyanquan:
To avoid conflict with flink-connector-kafka jar.


ruanhang1993 (approved):
LGTM

PR #3616, opened 2024-09-22
[FLINK-36397][cdc-connector][mysql] Obtain the high watermark offset and table data in the same transaction during the Snapshot phase

qiaozongmi:
This closes [[FLINK-36397](https://issues.apache.org/jira/projects/FLINK/issues/FLINK-36397)]
Obtain the high watermark offset and table data in the same transaction during the Snapshot phase,This ensures that the data retrieved is consistent and avoids any potential data loss.


qiaozongmi:
@leonardBang  PTAL

PR #3621, opened 2024-09-25
[hotfix] Fix Stale label automatically removed with no update

yuxiqian:
This fixes `Stale` bot not working as expected due to workflow interactions.

Currently, `approve_label` workflow will trigger update for all PRs' update time, causing Stale label automatically removed though this PR has no updates at all.

Removing auto-remove feature should fix this.


leonardBang (approved):
+1

PR #3623, opened 2024-09-29
[FLINK-3640] Close MetadataApplier when the job stops

morozov:
This pull request implements closing `MetadataApplier` when the job stops. This allows `MetadataApplier` implementation to release the resources that they may use to apply schema changes (e.g. a JDBC connection).

## Choice of the interface

1. `SchemaRegistryRequestHandler` (which will close `MetadataApplier`) implements `Closeable`, so it would make sense to declare `MetadataApplier` as `Closeable` as well.
2. `PaimonMetadataApplier` internally instantiates a Paimon `Catalog`, which it ideally should close, and which is `Autocloseable`.

To me, it makes more sense to declare `MetadataApplier` as `AutoCloseable`. The reason is that `Closeable` is an IO-specific interface, which is declared in `java.io` and throws `IOException` in `close()`. `AutoCloseable` is more generic and seems to be more suitable for `MetadataApplier`, whose implementations doesn't necessarily perform IO.

## Testing
I didn't find existing tests focusing on `SchemaRegistryRequestHandler` or `SchemaRegistry`, so I dind't implement a unit test. Testing of the changes in `SchemaRegistryRequestHandler` can be done by implementing some logging in `ValuesMetadataApplier#close` and running any test that uses it (e.g. `FlinkPipelineComposerITCase`)

## Additional scope
Just as a demonstration of where this new functionality can be applied, I implemented `close()` in `PaimonMetadataApplier`. I didn't test it and am not sure that this is the right thing to do. I can revert this part, if necessary.

## Further considerations
The problem of closing resources seems also relevant for the `MetadataAccessor` interface. Specifically, `MySqlMetadataAccessor` disconnects from the database after each call but some other implementations may require maintaining a single connection.


flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/MetadataApplier.java
```diff
@@ -50,4 +50,7 @@ default boolean acceptsSchemaEvolutionType(SchemaChangeEventType schemaChangeEve
     default Set<SchemaChangeEventType> getSupportedSchemaEvolutionTypes() {
         return Arrays.stream(SchemaChangeEventTypeFamily.ALL).collect(Collectors.toSet());
     }
+
+    @Override
+    default void close() throws Exception {}
```
yuxiqian, line 55: Adding a JavaDocs for this method would be nice.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistryRequestHandler.java
```diff
@@ -348,6 +348,12 @@ public void close() throws IOException {
         if (schemaChangeThreadPool != null) {
             schemaChangeThreadPool.shutdown();
         }
+
+        try {
+            metadataApplier.close();
+        } catch (Exception e) {
+            throw new IOException("Failed to close metadata applier.", e);
+        }
```
yuxiqian, line 356: I've tested locally with a dummy MetadataApplier that always throws exception but gets no output. Seems OperatorCoordinators will be closed by calling `OperatorCoordinatorHandler#disposeAllOperatorCoordinators`, where exceptions thrown here are ignored without being logged:

```java
// DefaultOperatorCoordinatorHandler.java
@Override
public void disposeAllOperatorCoordinators() {
    coordinatorMap.values().forEach(IOUtils::closeQuietly);
}

// IOUtils.java
public static void closeQuietly(AutoCloseable closeable) {
    try {
        if (closeable != null) {
            closeable.close();
        }
    } catch (Throwable ignored) {
    }
}
```

Maybe we can add some logging message if metadata appliers are not closed as expected?

yuxiqian (commented):
Thanks for @morozov's great work! Just left some minor comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/coordinator/SchemaRegistryRequestHandler.java
morozov: Good catch 👍 Thanks for the detailed explanation.

yuxiqian (approved):
Looks good! cc @ruanhang1993 @lvyanquan

lvyanquan (approved):

morozov:
@ruanhang1993 could you take a look, please?

ruanhang1993 (approved):
LGTM.

morozov:
@ruanhang1993, thanks for the review. What are the next steps to get this merged?

leonardBang:
As we forget merge the PR timely, I update this to latest master branch, will merge once the CI passed.

leonardBang:
CI passed, merging...

PR #3629, opened 2024-10-08
[FLINK-36442][postgres-cdc] Add insert-only change stream option

nandorsoma:
[FLINK-36442](https://issues.apache.org/jira/browse/FLINK-36442)

Adds the option to do an insert-only stream for connectors that support specifying a changelog mode. This allows debezium connectors to be imported into flink as stream tables to take advantage of flink's strong stream processing capabilities. There is currently no way to convert a all/upsert changelog stream to an insert-only stream with just the flink table api so this appears to be the only workaround.

This PR duplicates #3562 but follows community rules.
cc @henneberger

### DIFF
--- a/docs/content/docs/connectors/flink-sources/postgres-cdc.md
+++ b/docs/content/docs/connectors/flink-sources/postgres-cdc.md
@@ -166,7 +166,7 @@ SELECT * FROM shipments;
       <td>optional</td>
       <td style="word-wrap: break-word;">all</td>
       <td>String</td>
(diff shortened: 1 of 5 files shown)


herunkang2018:
@nandorsoma Thanks for your contribution. Currently we can use 'debezium.skipped.operations'='u,d' option to skip the update/delete operation, to filter out the insert-only stream.

nandorsoma:
> @nandorsoma Thanks for your contribution. Currently we can use 'debezium.skipped.operations'='u,d' option to skip the update/delete operation, to filter out the insert-only stream.

Thanks for the reply. Let me try it, I didn't know about this property.

PR #3631, opened 2024-10-10
[FLINK-36463][cdc][mysql] Forbid to override some debezium options

Shawn-Hx:
FlinkCDC uses debezium for capturing and parsing incremental data, but during the snapshot phase, FlinkCDC uses its own designed read implementation, so for FlinkCDC, some debezium options do not take effect, such as 'debezium.snapshot.query.mode'.


ruanhang1993 (approved):
LGTM.

PR #3633, opened 2024-10-10
[hotfix] Fix Java 11 target compatibility & add tests

yuxiqian:
This PR:

* Fixes Java 11 incompatibility by unifying `maven-shade-plugin` versions.
* Automatically selects higher version Java target when using JDK 11+.
* Adds Java 11 into CI test matrices

### DIFF
--- a/.github/workflows/flink_cdc.yml
+++ b/.github/workflows/flink_cdc.yml
@@ -124,7 +124,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        jdk: [ 8 ]
+        jdk: [ '8', '11' ]
         module: [ "core",
                   "pipeline_connectors",
                   "mysql",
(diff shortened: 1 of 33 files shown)


.github/workflows/flink_cdc_migration_test.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           submodules: true
       - uses: actions/setup-java@v4
         with:
-          java-version: 8
+          java-version: 11
```
ruanhang1993, line 31: Do we need to add jdk 8 here?

yuxiqian: I'm a little unsure about how to run CI tests across Java versions. According to @GOODBOY008's opinion, it is not suggested to have too much GitHub action jobs at once, since there's a hard parallelism limit on GitHub runners (20 for free plan, 40 / 60 for paid users), and it would be expensive to have too many jobs in CI matrix.

yuxiqian: Added Java version matrix into migration test CI. Now there are 57 jobs to run.

yuxiqian: Sounds good, I'll have a try.

yuxiqian:
Done, extracted common logic into `flink_cdc_base` and `flink_cdc_migration_test_base`, and kept minimal configurations in `flink_cdc_java_8` and `flink_cdc_java_11` workflow. @GOODBOY008 PTAL

yuxiqian:
Thanks for @GOODBOY008's review, fixed and rebased.

yuxiqian:
Apologies to @GOODBOY008, but seems modifying workflow names cause some side effects that breaks README building status badge. See #3807 for a hotfix.

PR #3636, opened 2024-10-11
[FLINK-36474] Support merging timestamp columns when routing

yuxiqian:
This closes FLINK-36474.

For now, it's not possible to merge timestamp-family data (including `Timestamp`, `ZonedTimestamp`, and `LocalZonedTimestamp`) without some awkward transform rules and explicit castings. This PR adds the following type coercions:

* ~~`TIME`s,~~ `TIMESTAMP`s, `TIMESTAMP_TZ`s, and `TIMESTAMP_LTZ`s with promoted precision

For example, `TIMESTAMP(3)` and `TIMESTAMP(6)` will be considered compatible and data with lower precision will be automatically escalated to the higher one.

* `TIMESTAMP_TZ`s and `TIMESTAMP_LTZ`s will be converted to `TIMESTAMP` if necessary.

I reused some code from `SystemFunctionUtils#castToTimestamp`, so this implicit behavior should be equivalent to writing explicit castings manually.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
@@ -20,8 +20,11 @@ package org.apache.flink.cdc.runtime.operators.schema;
 import org.apache.flink.api.java.tuple.Tuple3;
 import org.apache.flink.cdc.common.annotation.Internal;
 import org.apache.flink.cdc.common.annotation.VisibleForTesting;
+import org.apache.flink.cdc.common.data.LocalZonedTimestampData;
 import org.apache.flink.cdc.common.data.RecordData;
 import org.apache.flink.cdc.common.data.StringData;
+import org.apache.flink.cdc.common.data.TimestampData;
+import org.apache.flink.cdc.common.data.ZonedTimestampData;
 import org.apache.flink.cdc.common.event.DataChangeEvent;
 import org.apache.flink.cdc.common.event.DropTableEvent;
 import org.apache.flink.cdc.common.event.Event;
@@ -72,6 +75,8 @@ import javax.annotation.Nullable;
 
 import java.io.Serializable;
 import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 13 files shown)


yuxiqian:
Could @lvyanquan @aiwenmo please take a look?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
```diff
@@ -609,6 +643,34 @@ public Object getFieldOrNull(RecordData recordData) {
...
+                int precisionDifference =
+                        ((TimeType) originalType).getPrecision()
+                                - ((TimeType) destinationType).getPrecision();
+                if (precisionDifference > 0) {
+                    return value / (int) Math.pow(10, precisionDifference);
+                } else if (precisionDifference < 0) {
+                    return value * (int) Math.pow(10, -precisionDifference);
+                } else {
+                    return value;
+                }
```
aiwenmo, line 658: Currently, time supports only a precision of 0. We need to open another PR to implement a new feature that   custom precision for time, and then implement a new feature that support merging time columns when routing.

yuxiqian: Agree that `TIME` precision is a different issue and could be dealt with in another PR.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
```diff
@@ -624,4 +672,21 @@ public void snapshotState(StateSnapshotContext context) throws Exception {
...
+            return TimestampData.fromLocalDateTime(
+                    LocalDateTime.ofInstant(
+                            ((LocalZonedTimestampData) object).toInstant(), ZoneId.of(timezone)));
+        } else if (object instanceof ZonedTimestampData) {
+            return TimestampData.fromLocalDateTime(
+                    LocalDateTime.ofInstant(
+                            ((ZonedTimestampData) object).toInstant(), ZoneId.of(timezone)));
+        } else {
+            return TimestampData.fromLocalDateTime(LocalDateTime.parse(String.valueOf(object)));
+        }
```
aiwenmo, line 690: This code may never be executed.

aiwenmo (approved):
LGTM

lvyanquan (approved):
LGTM.

yuxiqian:
Squashed & rebased with `master`. Could @ruanhang1993 please take a look?

leonardBang (approved):
+1， merging...

PR #3575, opened 2024-08-26
[FLINK-36151] Add schema evolution related docs

yuxiqian:
This closes FLINK-36151 by adding missing schema evolution related concepts into Flink CDC docs.

PR #3640, opened 2024-10-12
[FLINK-36151][BP-3.2] Add schema evolution related docs

yuxiqian:
Backported from #3575, and removed Truncate / Drop Table events specification that does not exist in release 3.2.


leonardBang (approved):
+1, wait the CI green

PR #3642, opened 2024-10-14
[FLINK-36525] Support for AI Model Integration for Data Processing

proletarians:
The goal is to extend flink-cdc with the capability to invoke AI models during the data stream processing workflow, with a particular focus on supporting array data structures. This feature will allow users to easily integrate embedding models to handle array-based data, such as lists of text, numeric data, or other multi-dimensional arrays, within their Flink jobs.


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -22,6 +22,11 @@
 import org.apache.flink.cdc.common.data.ZonedTimestampData;
 import org.apache.flink.cdc.common.utils.DateTimeUtils;
 
+import dev.langchain4j.data.document.Metadata;
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.data.segment.TextSegment;
+import dev.langchain4j.model.openai.OpenAiEmbeddingModel;
+import okhttp3.*;
```
yuxiqian, line 29: Don't use `*` import.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -30,16 +35,14 @@
...
+import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.TimeZone;
-import java.util.UUID;
+import java.util.*;
```
yuxiqian, line 42: Ditto.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -77,6 +80,53 @@ public static int currentDate(long epochTime, String timezone) {
...
+    }
+
+    public static String getEmbedding(String input, String apiKey, String model) {
+        if (input == null || input.trim().isEmpty()) {
+            LOG.debug("Empty or null input provided for embedding.");
+            return "";
+        }
+
+        try {
+            // 确保 OpenAiEmbeddingModel 已初始化
```
yuxiqian, line 104: Remove Chinese comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -77,6 +80,53 @@ public static int currentDate(long epochTime, String timezone) {
...
+            if (embeddingModel == null) {
+                initializeOpenAiEmbeddingModel(apiKey, "https://api.openai.com/v1/");
```
yuxiqian, line 106: Is the endpoint hard-encoded here? Why we still need this function and passing apiKeys manually? Should these be configured in `models:` rule block?

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/ModelDef.java
```diff
@@ -0,0 +1,92 @@
...
+        @Override
+        public void open(FunctionContext context) throws Exception {
+            // 初始化逻辑，如建立API连接等
+        }
+
+        @Override
+        public void close() throws Exception {
+            // 清理逻辑，如关闭API连接等
+        }
+    }
```
yuxiqian, line 61: What is this class for? Seems it's completely useless.

pom.xml
```diff
@@ -41,6 +41,7 @@ limitations under the License.
         <module>flink-cdc-runtime</module>
         <module>flink-cdc-e2e-tests</module>
         <module>flink-cdc-pipeline-udf-examples</module>
+        <module>flink-cdc-openai-udf</module>
```
yuxiqian, line 44: Maybe putting it as a part of `SystemFunctionUtils` is enough? Creating a new top-level package seems too much.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/metadata/TransformSqlOperatorTable.java
```diff
@@ -178,6 +178,40 @@ public void lookupOperatorOverloads(
...
+                            SqlTypeFamily.STRING, SqlTypeFamily.STRING, SqlTypeFamily.STRING),
+                    SqlFunctionCategory.USER_DEFINED_FUNCTION // Category of the function
+                    );
+
+    // Define the AI_EMBEDDING function
+    public static final SqlFunction GET_EMBEDDING =
+            new SqlFunction(
+                    "GET_EMBEDDING",
+                    SqlKind.OTHER_FUNCTION,
+                    ReturnTypes.explicit(SqlTypeName.VARCHAR), // 返回类型是数组
```
yuxiqian, line 197: `SqlTypeName.VARCHAR), // 返回类型是数组`

This comment seems a little untrue. Also, after we have user-defined-models, are these definitions still necessary?

yuxiqian (changes requested):
Hi @proletarians, thanks for working on this huge PR! However I'm a little confused by some of these codes.

Seems you have implemented a new interface (with `ModelDef`) allowing users to configure model parameters without creating a new model. But I also see some functions in `SystemFunctionUtils` that have API keys and endpoints hard-encoded here. Apart some obviously unwanted test classes, I also see some weird things in `flink-cdc-openai-udf` package, where UDF functions in there seems not configurable, and temporary API endpoints and keys are hard-encoded inside. How are users supposed to use them?


Could you please kindly share your design concepts about this feature?

Also, I noticed that there's no any test cases (except those written by umeshdangat). Though it's not easy to test real AI models in CI, I think some kind of mocking test is required.

As for code-style, please run `mvn spotless:check` to ensure all rules have been enforced before committing. Some changes are required including:

* Comments should not be written in Chinese.
* License headers and JavaDocs have standard formats. Do not change or alter them. For newly created classes, make sure they have been properly added.
* Wildcard import `import java.*` and omitting if-block braces are not allowed.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/model/ModelUdf.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.stream.Collectors;
+
+public class ModelUdf implements UserDefinedFunction {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ModelUdf.class);
+    private static final String DEFAULT_API_KEY =
+            "sk-WegHEuogRpIyRSwaF5Ce6fE3E62e459dA61eFaF6CcF8C79b";
+    private static final String DEFAULT_MODEL_NAME = "text-embedding-ada-002";
+    private static final int DEFAULT_TIMEOUT_SECONDS = 30;
+    private static final String DEFAULT_BASE_URL = "https://api.gpt.ge/v1/";
```
yuxiqian, line 31: Passing these "default" API keys and endpoints aren't meaningful. Request tokens might ran out, endpoints might expire.

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/translator/TransformTranslator.java
```diff
@@ -85,11 +80,45 @@ public DataStream<Event> translatePostTransform(
...
-                        .map(udf -> Tuple2.of(udf.getName(), udf.getClasspath()))
-                        .collect(Collectors.toList()));
+                allFunctions.stream().map(this::udfDefToTuple2).collect(Collectors.toList()));
         return input.transform(
                 "Transform:Data", new EventTypeInfo(), postTransformFunctionBuilder.build());
     }
+
+    private List<UdfDef> convertModelsToUdfs(List<ModelDef> models) {
+        return models.stream().map(this::modelToUdf).collect(Collectors.toList());
+    }
```
yuxiqian, line 95: This conversion is quite worrying, since obviously `ModelDef` carries more information (the params), and the conversion could not be done without some hacky approaches like pasting params into the name field.

Modifying the `addUdfFunctions` function, making it accepting a more universal data structure (instead of `Name - Classpath` `Tuple2`) makes more sense to me.

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/ModelDef.java
yuxiqian: Put this in documentation if this was meant to be an implementation reference.

ruanhang1993:
Provided by #3753 . Close this PR.

PR #3647, opened 2024-10-17
[FLINK-36558] Fix column metadata parsing compatibility with MySQL 8.0.17 / 8.0.18

yuxiqian:
This closes FLINK-36558.

It is known that MySQL 8.0.17 and 8.0.18 writes Array-typed key index TableMap data into Binlog differently.

* MySQL 8.0.17 used a different internal enum value for TYPED_ARRAY column type
* MySQL 8.0.17 and 8.0.18 writes extraneous bytes after expected column metadata

The "extra byte" issue has been fixed since MySQL 8.0.19. Since MySQL takes Binlog format as an internal implementation detail[1] and doesn't guarantee its backwards compatibility, MySQL CDC has to deal with the compatibility problem itself.

The actual code change is limited to `TableMapEventDataDeserializer.java` where the encoded "column metadata bytes size" is used to determine how many bytes should be sent to `readMetadata` function for parsing. Any unused bytes will be ignored and will not affect parsing the following fields.

[1]   https://bugs.mysql.com/bug.php?id=105545


lvyanquan (approved):
Thanks @yuxiqian for this detailed description, and this change looks good to me.

leonardBang (approved):
Thanks @yuxiqian for the fix and @lvyanquan for the review, LGTM

PR #3648, opened 2024-10-17
[FLINK-36560][pipeline-connector][paimon] fix the issue that timestamp_ltz field is not correctly converted.

lvyanquan:
This test case has already indicated this issue, but it was initially overlooked.
1. For Paimon & Kafka, they already support timestamp with local time zone type, so we don't need add extra conversion using zoneId.
2. For Doris & Starrocks, they only have datetime type and don't support timestamp with local time zone, refer to https://doris.apache.org/zh-CN/docs/1.2/sql-manual/sql-reference/Data-Types/DATETIMEV2/ and https://docs.starrocks.io/docs/sql-reference/data-types/date-types/DATE/, so we need to convert timestamp_ltz to timestamp using zoneid of database, Therefore, we do not need to modify the implementation.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonWriterHelper.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonWriterHelper.java
@@ -32,7 +32,6 @@ import org.apache.paimon.data.Timestamp;
 import org.apache.paimon.types.RowKind;
 
 import java.time.ZoneId;
-import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -111,15 +110,11 @@ public class PaimonWriterHelper {
             case TIMESTAMP_WITH_TIME_ZONE:
                 fieldGetter =
                         row ->
-                                Timestamp.fromLocalDateTime(
-                                        ZonedDateTime.ofInstant(
-                                                        row.getLocalZonedTimestampData(
-                                                                        fieldPos,
-                                                                        DataTypeChecks.getPrecision(
-                                                                                fieldType))
-                                                                .toInstant(),
-                                                        zoneId)
-                                                .toLocalDateTime());
+                                Timestamp.fromInstant(
+                                        row.getLocalZonedTimestampData(
+                                                        fieldPos,
+                                                        DataTypeChecks.getPrecision(fieldType))
+                                                .toInstant());
                 break;
             case ROW:
                 final int rowFieldCount = getFieldCount(fieldType);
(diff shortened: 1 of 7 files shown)


lvyanquan:
@beryllw @yuxiqian @ruanhang1993 could you please help to review this?

yuxiqian (approved):
Thanks for @lvyanquan's quick fix! Change looks reasonable.

Irrelevant to this PR, just noticed that Paimon project runs CI in different timezones randomly chosen every time [(code here)](https://github.com/apache/paimon/blob/aeffc89ff9e83b5c19664372b9e4644e7a988f10/.github/workflows/flink-cdc-tests.yml#L50-L54). Maybe we could also adapt this method to expose more time-zone related bugs.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonWriterHelper.java
```diff
@@ -111,15 +110,11 @@ private static RecordData.FieldGetter createFieldGetter(
...
-                                Timestamp.fromLocalDateTime(
-                                        ZonedDateTime.ofInstant(
-                                                        row.getLocalZonedTimestampData(
-                                                                        fieldPos,
-                                                                        DataTypeChecks.getPrecision(
-                                                                                fieldType))
-                                                                .toInstant(),
-                                                        zoneId)
-                                                .toLocalDateTime());
+                                Timestamp.fromInstant(
```
beryllw, line 113: Should we need to fix Kafka/StarRocks/Doris?

https://github.com/apache/flink-cdc/blob/dd69756d3fc1e3f5261b286added5397acbd88ad/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/sink/DorisRowConverter.java#L107-L115

https://github.com/apache/flink-cdc/blob/dd69756d3fc1e3f5261b286added5397acbd88ad/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java#L186-L195

https://github.com/apache/flink-cdc/blob/dd69756d3fc1e3f5261b286added5397acbd88ad/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/main/java/org/apache/flink/cdc/connectors/kafka/json/TableSchemaInfo.java#L163-L174

herunkang2018: Good catch, I think we should fix this issue in other connectors.

lvyanquan: Done fix in Paimon & Kafka connector.
For Doris & Starrocks, they  only have datetime type and don't support timestamp with local time zone, refer to https://doris.apache.org/zh-CN/docs/1.2/sql-manual/sql-reference/Data-Types/DATETIMEV2/ and https://docs.starrocks.io/docs/sql-reference/data-types/date-types/DATE/, so we need to convert timestamp_ltz to timestamp using zoneid of database, Therefore, we do not need to modify the implementation. I added tests here.
And I've picked one fix from https://github.com/apache/flink-cdc/pull/3655.

leonardBang (approved):
+1

herunkang2018 (approved):
LGTM.

PR #3651, opened 2024-10-18
[FLINK-36565] Route module allows merging Decimals with various precisions

yuxiqian:
This closes FLINK-36565.

Currently, table-merging router does not allow coercing `Decimal` fields with different precisions from upstream, which means some lossless conversions are actually not possible, including:

* `TINYINT`, `SMALLINT`, `INT`, `BIGINT` to `DECIMAL(p > 17, s >= 0)`
* low precision `DECIMAL` to high precision ones

---

I've baked the changes in #3636 (FLINK-36474) into this PR since there are some common changes between them.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
@@ -20,8 +20,11 @@ package org.apache.flink.cdc.runtime.operators.schema;
 import org.apache.flink.api.java.tuple.Tuple3;
 import org.apache.flink.cdc.common.annotation.Internal;
 import org.apache.flink.cdc.common.annotation.VisibleForTesting;
+import org.apache.flink.cdc.common.data.LocalZonedTimestampData;
 import org.apache.flink.cdc.common.data.RecordData;
 import org.apache.flink.cdc.common.data.StringData;
+import org.apache.flink.cdc.common.data.TimestampData;
+import org.apache.flink.cdc.common.data.ZonedTimestampData;
 import org.apache.flink.cdc.common.event.DataChangeEvent;
 import org.apache.flink.cdc.common.event.DropTableEvent;
 import org.apache.flink.cdc.common.event.Event;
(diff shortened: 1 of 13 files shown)


yuxiqian:
Could @melin please take a look?

lvyanquan:
@yuxiqian could you rebase to master as https://github.com/apache/flink-cdc/pull/3636 merged?

yuxiqian:
Done, rebased with `master`.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/SchemaOperator.java
```diff
@@ -611,14 +614,42 @@ public Object getFieldOrNull(RecordData recordData) {
...
+                } else if (originalField instanceof Long) {
+                    decimalValue = BigDecimal.valueOf((Long) originalField, 0);
+                } else if (originalField instanceof DecimalData) {
+                    decimalValue = ((DecimalData) originalField).toBigDecimal();
+                } else {
+                    return fail(
+                            new IllegalArgumentException(
+                                    String.format(
+                                            "Cannot fit type \"%s\" into a DECIMAL column. "
+                                                    + "Currently only BYTE / SHORT / INT / LONG / DECIMAL can be accepted by a DECIMAL column",
```
lvyanquan, line 646: Can we support merge FLOAT/DOUBLE to DECIMAL like Flink does？ 
https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/types/#casting

yuxiqian: It's reasonable to have this merging path, I will try to implement it.

yuxiqian: Unfortunately, it's not possible to determine a "wide enough" Decimal type to store FLOAT / DOUBLE losslessly (since floating point numbers' expression range doesn't have an uplimit -- it could be up to `Infinity`), so maybe it's a not good idea to let this type conversion happen implicitly. Explicit `CAST`ing might be better, I'll open another ticket for this.

lvyanquan: Makes sense to me.

lvyanquan (approved):
LGTM.

leonardBang (approved):
+1, thanks @yuxiqian and @lvyanquan for the contribution

PR #3632, opened 2024-10-10
[FLINK-36461] Fix schema evolution failure with un-transformed tables

yuxiqian:
This closes FLINK-36461.

Currently, such transform configuration will fail for any schema change events from tables except `foo.bar.baz`:

```yaml
transform:
  - source-table: foo.bar.baz
    projection: \*
```

The exception message is as follows:

```
java.lang.ArrayIndexOutOfBoundsException: -1
at java.util.ArrayList.elementData(ArrayList.java:424)
at java.util.ArrayList.get(ArrayList.java:437)
at org.apache.flink.cdc.common.utils.SchemaUtils.lambda$transformSchemaChangeEvent$11(SchemaUtils.java:371)
at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1384)
at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
at org.apache.flink.cdc.common.utils.SchemaUtils.transformSchemaChangeEvent(SchemaUtils.java:386)
at org.apache.flink.cdc.runtime.operators.transform.PreTransformOperator.cacheChangeSchema(PreTransformOperator.java:272)
at org.apache.flink.cdc.runtime.operators.transform.PreTransformOperator.processElement(PreTransformOperator.java:248)
at org.apache.flink.streaming.runtime.tasks.CopyingChainingOutput.pushToOperator(CopyingChainingOutput.java:75)
at org.apache.flink.streaming.runtime.tasks.CopyingChainingOutput.collect(CopyingChainingOutput.java:50)
at org.apache.flink.streaming.runtime.tasks.CopyingChainingOutput.collect(CopyingChainingOutput.java:29)
at org.apache.flink.streaming.runtime.tasks.SourceOperatorStreamTask$AsyncDataOutputToOutput.emitRecord(SourceOperatorStreamTask.java:310)
at org.apache.flink.streaming.api.operators.source.SourceOutputWithWatermarks.collect(SourceOutputWithWatermarks.java:110)
[...]

PR #3665, opened 2024-10-24
[FLINK-36596] Fix unable to schema evolve with project-less transform rules

yuxiqian:
This closes FLINK-36596. ~~Requires #3632 being merged first.~~

As described in docs, omitting `transform[*].projection` field in YAML definition should be equivalent to `projection: \*` expression. However, now schema change events will be dropped when there are no projection fields explicitly written down. For example, the following definition:

```yaml
transform:
  - source-table: db.tbl
    filter: x > 0
```

could not apply upstream schema change events to downstream due to the lacking of `projection` block, which would be surprising.

Unifying the syntax of `projection: \*` and omitting `projection` field should fix this.


lvyanquan (approved):
LGTM.

leonardBang (approved):
+1

PR #3666, opened 2024-10-25
[FLINK-36520] CDC Supports MySQL 8.x and above

yuxiqian:
This closes FLINK-36520, FLINK-36590.

Currently, trying to use MySQL CDC with MySQL 8.4 would cause the following exception:

```
Caused by: org.apache.flink.util.FlinkRuntimeException: Read split MySqlSnapshotSplit{tableId=customer.customers, splitId='customer.customers:0', splitKeyType=[`id` BIGINT NOT NULL], splitStart=null, splitEnd=null, highWatermark=null} error due to org.apache.flink.util.FlinkRuntimeException: Cannot read the binlog filename and position via 'SHOW MASTER STATUS'. Make sure your server is correctly configured.
```

The reason is MySQL has changed some questionable words like `MASTER` and `SLAVE` to more appropriate ones, so some commands are no longer compatible. Legacy MySqlSource (based on Debezium) wasn't fixed since it's not possible to upgrade Debezium dependency for now, and it has been deprecated for a long time anyway.

Marked this PR as a draft, needs more discussion and testing over this.


leonardBang:
could you investigate binlog compress would been supported via this PR?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
```diff
@@ -90,6 +95,41 @@ public MySqlConnection(
         this.connectionConfig = connectionConfig;
         this.mysqlFieldReader = fieldReader;
         this.urlPattern = connectionConfig.getUrlPattern();
+
+        probeShowBinaryLogStatement();
```
morozov: This method both returns the result of the probe and stores it in a property, which isn't nice from the CQRS standpoint (a method should either return a value or produce side effects, but not both).

Would it make sense to split it into `private String probeShowBinaryLogStatement()` and `public String getShowBinaryLogStatement()`? The constructor would call `probe()` and store the result in a property, and `get()` would return the property. In this case, the field can be made final, and the `if (showBinaryLogStatement != null)` condition could be removed.

Alternatively, if we want `showBinaryLogStatement` to be lazily initialized, the call to `probeShowBinaryLogStatement()` should be removed from the constructor. The first call will initialize it.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
```diff
@@ -90,6 +94,40 @@ public MySqlConnection(
...
+                // Attempt to query
+                query(MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT, rs -> {});
+                showBinaryLogStatement = MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT;
+            } catch (SQLException e) {
+                LOGGER.info(
+                        "Probing with {} failed, try {}. Caused by: {}",
+                        MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT,
+                        MYSQL_CLASSIC_SHOW_BINARY_LOG_STATEMENT,
+                        e);
+                query(MYSQL_CLASSIC_SHOW_BINARY_LOG_STATEMENT, rs -> {});
```
morozov: If we're going to fall back to this statement anyways, why probe it? The corresponding [code](https://github.com/debezium/debezium/blob/8ca8f451f85c6674f37fc1c1d899a4b556b69553/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java#L36-L47) in Debezium looks simpler.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
```diff
@@ -90,6 +94,40 @@ public MySqlConnection(
...
+            try {
+                // Attempt to query
+                query(MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT, rs -> {});
+                showBinaryLogStatement = MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT;
+            } catch (SQLException e) {
+                LOGGER.info(
+                        "Probing with {} failed, try {}. Caused by: {}",
+                        MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT,
+                        MYSQL_CLASSIC_SHOW_BINARY_LOG_STATEMENT,
+                        e);
```
morozov: Would it be sufficient to just log the exception message? It isn't an error, so someone looking at the logs may see a stack trace, think that there is some problem, but in fact there is none.

morozov:
Correct me if I'm wrong, but it looks like using the proper SQL command for showing binary logs is the only major change here. The rest is about refactoring the test suite.

@leonardBang is the support for binlog compression a blocker for these changes to get merged?

> [...] needs more discussion and testing over this.

@yuxiqian do you have any specific discussion topics in mind? As for testing, I believe having the test suite pass would be a great start, although it's not clear why the CI didn't run the tests on this PR.

yuxiqian:
Thanks @morozov's kindly review!

> @leonardBang is the support for binlog compression a blocker for these changes to get merged?

I did some experiment earlier, and noticed that binlog compression would break current binlog backfilling code. However, considering this feature was introduced in MySQL 8.0.20, perhaps it could be handled independently.

> @yuxiqian do you have any specific discussion topics in mind? As for testing, I believe having the test suite pass would be a great start, although it's not clear why the CI didn't run the tests on this PR.

Seems there are a lot of merging conflicts blocking the CI. Will resolve them soon.

.github/workflows/flink_cdc.yml
```diff
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         java-version: [ '8' ]
-        module: [ 'core', 'pipeline_connectors', 'mysql', 'postgres', 'oracle', 'mongodb6', 'mongodb7', 'sqlserver', 'tidb', 'oceanbase', 'db2', 'vitess' ]
+        module: [ 'core', 'pipeline_connectors', 'mysql57', 'mysql80', 'mysql84', 'mysql90', 'postgres', 'oracle', 'mongodb6', 'mongodb7', 'sqlserver', 'tidb', 'oceanbase', 'db2', 'vitess' ]
```
sjmudd: Do not use 'mysql90" as a label. Oracle has horrible versioning and basically you probably need to do one of the following: label as "mysql9x" as 9.X.0 changes every 3 months until it's expected to be 9.7 LTS in April 2026 and then numbers become normal again. e.g. expected to be 9.7.0, 9.7.1, ...

Basically 9.0 should not be used 9.3.0 is current innovation release and it'll be 9.4.0 on ~15th July. So until 9.X is confirmed to be LTS either don't mention it or label it 'mysql9x'.

Reference: https://dev.mysql.com/blog-archive/introducing-mysql-innovation-and-long-term-support-lts-versions/

Clearly do what makes sense but 'mysql90' will never be a useful label.

JeffHu88:
Hi, could you please let me know when this issue will go live? Thanks!

j4k4:
A release of this would be very helpful as MySQL 8.0 is nearing it's end of life.  
E.g. AWS stops supporting version 8.0 soon.

shengjk:
Hi, I would like to ask why we do not use Debezium directly, but implement it ourselves? Debezium already supports MySQL 8 and above, and Debezium updates much faster than we do.

yuxiqian:
> Hi, I would like to ask why we do not use Debezium directly, but implement it ourselves? Debezium already supports MySQL 8 and above, and Debezium updates much faster than we do.

Thanks for @shengjk's nice suggestion! For a long time we've been stuck at Debezium 1.9.x and monkey-patching Debezium, mainly for keeping compatibility with Java 8.

FLINK-36605 also suggests a similar idea, but it is more likely to happen after bumping to Flink 2.x and dropping Java 8 support.

zhangshenghang:
@yuxiqian Hi, Is there any problem? Why is this PR in draft status? Does it have an expected merge date?

yuxiqian:
> @yuxiqian Hi, Is there any problem? Why is this PR in draft status? Does it have an expected merge date?

Thanks @zhangshenghang for tracing this issue!

IIRC, one concern from maintainers is we can't provide decent support for some newly introduced MySQL features, like Binlog compression, VECTOR data types, etc., without upgrading Debezium dependencies, which requires dropping Java 8 and Flink 1.x support. So this PR is, at best, a transitional solution.

The next feature release (3.6 or 4.0) has not been kicked-off, but there's some discussions on upgrading Debezium versions here (https://issues.apache.org/jira/browse/FLINK-36605), so I'm not sure if it's appropriate to introduce this temporary change at this time.

Anyway, I'll try to resolve all conflicts later this week so it could be marked ready for review.

zhangshenghang:
> > @yuxiqian Hi, Is there any problem? Why is this PR in draft status? Does it have an expected merge date?
> 
> Thanks @zhangshenghang for tracing this issue!
> 
> IIRC, one concern from maintainers is we can't provide descent support for some newly introduced MySQL features, like Binlog compression, VECTOR data types, etc., without upgrading Debezium dependencies, which requires dropping Java 8 and Flink 1.x support. So this PR is, at best, a transitional solution.
> 
> The next feature release (3.6 or 4.0) has not been kicked-off, but there's some discussions on upgrading Debezium versions here (https://issues.apache.org/jira/browse/FLINK-36605), so I'm not sure if it's appropriate to introduce this temporary change at this time.
> 
> Anyway, I'll try to resolve all conflicts later this week so it could be marked ready for review.

@yuxiqian Thank you for your reply. Yes, upgrading to JDK 11 would be a significant change. If there could be a temporary transitional solution for 1.8, that would also be a good option.

zhangshenghang:
> > > @yuxiqian Hi, Is there any problem? Why is this PR in draft status? Does it have an expected merge date?
> > 
> > 
> > Thanks @zhangshenghang for tracing this issue!
> > IIRC, one concern from maintainers is we can't provide descent support for some newly introduced MySQL features, like Binlog compression, VECTOR data types, etc., without upgrading Debezium dependencies, which requires dropping Java 8 and Flink 1.x support. So this PR is, at best, a transitional solution.
> > The next feature release (3.6 or 4.0) has not been kicked-off, but there's some discussions on upgrading Debezium versions here (https://issues.apache.org/jira/browse/FLINK-36605), so I'm not sure if it's appropriate to introduce this temporary change at this time.
> > Anyway, I'll try to resolve all conflicts later this week so it could be marked ready for review.
> 
> @yuxiqian Thank you for your reply. Yes, upgrading to JDK 11 would be a significant change. If there could be a temporary transitional solution for 1.8, that would also be a good option.

@yuxiqian Hi，I saw a discussion here about upgrading Debezium to version 2.7.x: https://issues.apache.org/jira/browse/FLINK-36605.
However, it seems that version 2.7.x also does not support MySQL 8.4. 
MySQL 8.4 supported starting from version 3.2
image

yuxiqian:
Thanks @zhangshenghang for confirming this! Debezium 3.2 requiring Java 17 seems not very likely, since Flink 2.x still supports Java 11 (deprecated though). I'll fix CI test cases first.

Romeoy:
Hi @yuxiqian , much appreciate! Is there any update for the plan to support mysql 8.4? As mysql 8.0 will not been matained soon for many cloud service providers, I believe many companies will met this issue as well. By the way, I saw another PR to resolve same issue https://github.com/apache/flink-cdc/pull/4196 .

sridhard:
Hi,
When this will be released?

nevfo:
Hello team, when will this feature be supported? Since MySQL 8.0 is approaching EOL soon, a large number of users will be facing this issue.

fmorillo7694:
Any news on this?

cgnie:
`Hope to see support for MySQL 8.x+ versions sooner.`

leonardBang:
Looks like many users/devs are waiting this feature, would you like to contribute this ?@yuxiqian

yuxiqian:
Thanks everyone for the interest in this PR! I've addressed morozov and sjmudd's comments, added basic compatibility tests, and marked this ready for review.

I hope we can parameterize the entire test case with multiple MySQL versions for better coverage, but that might be suitable for another PR.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
```diff
@@ -769,4 +784,27 @@ public <T extends DatabaseSchema<TableId>> Object getColumnValue(
...
+                    "Successfully found show binary log statement with `{}`.",
+                    MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT);
+            return MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT;
+        } catch (SQLException e) {
+            LOGGER.info(
+                    "Probing with {} failed, fallback to classic {}. Caused by: {}",
+                    MYSQL_NEW_SHOW_BINARY_LOG_STATEMENT,
+                    MYSQL_CLASSIC_SHOW_BINARY_LOG_STATEMENT,
+                    e.getMessage());
+            return MYSQL_CLASSIC_SHOW_BINARY_LOG_STATEMENT;
```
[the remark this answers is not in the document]
yuxiqian: Such strict checking seems unnecessary, it may break on some MySQL-compatible servers. Falling back to legacy syntax is safer.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -0,0 +1,717 @@
...
+            super.lastSnapshotRecord(snapshotContext);
+        }
+    }
+
+    @Override
+    protected void postSnapshot() throws InterruptedException {
+        // We cannot be sure that the last event as the last one
+        // - last table could be empty
+        // - data snapshot was not executed
+        // - the last table schema snaphsotted is not monitored and storing of monitored is disabled
```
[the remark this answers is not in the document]
yuxiqian: This file is copied from Debezium project, I would prefer keeping it as-is.

leonardBang:
Thanks @yuxiqian for the update, current PR is in a good status, LGTM. 
We can improve the multiple MySQL version tests later.

PR #3668, opened 2024-10-28
[FLINK-36610] MySQL CDC supports parsing gh-ost generated schema changes

yuxiqian:
This closes FLINK-36610.

GitHub's [gh-ost](https://github.com/github/gh-ost/) project allows us to alter MySQL table schema without triggers. It is implemented based on a "replicate and swap" technique, which could not be correctly recognized by MySQL CDC for now, since `RENAME TABLE` DDL isn't supported yet.

This PR adds basic support for table altering schema change events emitted by gh-ost.


MOBIN-F:
hi, @yuxiqian [pt-osc](https://docs.percona.com/percona-toolkit/pt-online-schema-change.html) is a product with similar functions to gh-ost. I think we should also support it. I am willing to do this part of the work. Can I submit code on this PR?

yuxiqian:
@MOBIN-F Sure! Feel free to work on this PR if helpful.

yuxiqian:
Added testcases for DataStream / Table API, too.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/RecordUtils.java
```diff
@@ -489,4 +529,75 @@ private static Optional<WatermarkKind> getWatermarkKind(SourceRecord record) {
...
+                return OSC_TABLE_ID_PATTERN.matcher(tableName).matches();
+            }
+
+            return false;
+        } catch (JsonProcessingException e) {
+            return false;
+        }
+    }
+
+    private static final Pattern OSC_TABLE_ID_PATTERN = Pattern.compile("^_(.*)_(gho|new)$");
```
czy006, line 593: As I know, gh-ost also has **_del** and **_ghc** tables. Do we need to join them **^_(.*)_(gho|ghc|del|new)$** ?

yuxiqian: Seems it's not necessary since actual schema evolution DDLs are always performed on `_XXX_gho` tables with gh-ost and `_XXX_new` with os-ptc.

czy006 (approved):
LGTM

mielientiev:
Hello @czy006 @MOBIN-F 
Any idea when this PR could be merged?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
```diff
@@ -281,4 +281,12 @@ public class MySqlDataSourceOptions {
                     .withDescription(
                             "List of readable metadata from SourceRecord to be passed to downstream, split by `,`. "
                                     + "Available readable metadata are: op_ts.");
+
+    @Experimental
+    public static final ConfigOption<Boolean> PARSE_ONLINE_SCHEMA_CHANGES =
+            ConfigOptions.key("scan.parse.online.schema.changes")
```
lvyanquan, line 287: We can add `enabled` to key name to indicate that the configOption is boolean type.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/RecordUtils.java
```diff
@@ -384,6 +390,40 @@ public static TableId getTableId(SourceRecord dataRecord) {
         return new TableId(dbName, null, tableName);
     }
 
+    public static SourceRecord setTableId(
+            SourceRecord dataRecord, TableId originalTableId, TableId tableId) {
```
lvyanquan, line 394: We can move the logic of isOnLineSchemaChangeEvent and setTableId to [MySqlStreamingChangeEventSource#handleQueryEvent](https://github.com/apache/flink-cdc/blob/ddb5f00df5bab21e8c91e4a31574d446d8b7a3f7/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java#L660) to simplify the code.

yuxiqian: Thanks for the suggestion, but I noticed that it's not very convenient to peel it out there since `SchemaChangeEvent` is still in raw state (DDL expression only) and no tableId information available yet.

More specifically, there's no TableId information available here and must be parsed manually from DDL string.

```
// MySqlStreamingChangeEventSource.java

eventDispatcher.dispatchSchemaChangeEvent(
                        partition,
                        tableId,
                        (receiver) -> {
                            try {
                                receiver.schemaChangeEvent(schemaChangeEvent);
                            } catch (Exception e) {
                                throw new DebeziumException(e);
                            }
                        });
```

Any suggestions on this?

lvyanquan: Can we reuse and modify the tableId here?
https://github.com/apache/flink-cdc/blob/ddb5f00df5bab21e8c91e4a31574d446d8b7a3f7/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java#L664-L667.

And we can use 
```
schemaChangeEvents.get(i).getType().equals(SchemaChangeEvent.SchemaChangeEventType.ALTER)
```
to judge the type of raw DDL.

yuxiqian: Seems we can't retrieve TableIds here because `schemaChangeEvent.getTables()` is always empty for schema changes on shadow tables.

lvyanquan: Got it, that's because the temporary table is not included in the table filter condition.
We can still get the table name from source struct of schemaChangeEvent, but this loses the meaning of doing so.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/RecordUtils.java
```diff
@@ -489,4 +529,75 @@ private static Optional<WatermarkKind> getWatermarkKind(SourceRecord record) {
...
+            // DROP TRIGGER IF EXISTS `db`.`pt_osc_db_test_tb1_ins`
+            //
+            // Among all these, we only need the "ALTER" one that happens on the `_gho`/`_new`
+            // table.
+            String ddl =
+                    mapper.readTree(value.getString(HISTORY_RECORD_FIELD))
+                            .get("ddl")
+                            .asText()
+                            .toLowerCase();
+            if (ddl.toLowerCase().startsWith("alter")) {
```
lvyanquan, line 581: This toLowerCase() is unnecessary?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlOnLineSchemaMigrationITCase.java
```diff
@@ -0,0 +1,591 @@
...
+        env.setParallelism(1);
+        TableId tableId = TableId.tableId(customerDatabase.getDatabaseName(), "customers");
+        MySqlSourceConfigFactory configFactory =
+                new MySqlSourceConfigFactory()
+                        .hostname(MYSQL8_CONTAINER.getHost())
+                        .port(MYSQL8_CONTAINER.getDatabasePort())
+                        .username(TEST_USER)
+                        .password(TEST_PASSWORD)
+                        .databaseList(customerDatabase.getDatabaseName())
+                        .tableList(customerDatabase.getDatabaseName() + "\\.customers")
```
lvyanquan, line 174: I think it would be more reliable to validate capturing the changes of entire database and making twice online table change changes.

lvyanquan:
Hi @leonardBang, @ruanhang1993 could you please help to trigger CI check?

yuxiqian:
Seems there's some [compatibility issue](https://forums.percona.com/t/percona-toolkit-execution-failure/34610) between Percona Toolkit 3.7 and MySQL. Locked to a lower version should solve this.

PR #3673, opened 2024-10-30
[FLINK-36193] Supports applying TRUNCATE / DROP table ddl...

yuxiqian:
…to Paimon, StarRocks, and Doris.

This closes FLINK-36193.

FLINK-35243 adds Truncate / Drop table events support at framework level, but no sink connector supports them actually (except `values`). This PR brings schema evolution ability to several compatible sink connectors where such DDLs are supported.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/sink/DorisSchemaChangeManager.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/sink/DorisSchemaChangeManager.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.connectors.doris.sink;
+
+import org.apache.doris.flink.cfg.DorisOptions;
+import org.apache.doris.flink.exception.IllegalArgumentException;
+import org.apache.doris.flink.sink.schema.SchemaChangeManager;
+
+import java.io.IOException;
+
+import static org.apache.doris.flink.catalog.doris.DorisSystem.identifier;
+
+/** An enriched version of Doris' {@link SchemaChangeManager}. */
+public class DorisSchemaChangeManager extends SchemaChangeManager {
+    public DorisSchemaChangeManager(DorisOptions dorisOptions) {
+        super(dorisOptions);
+    }
+
+    public boolean truncateTable(String databaseName, String tableName)
+            throws IOException, IllegalArgumentException {
+        String createTableDDL =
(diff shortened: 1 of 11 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/sink/DorisSchemaChangeManager.java
```diff
@@ -0,0 +1,47 @@
...
+
+/** An enriched version of Doris' {@link SchemaChangeManager}. */
+public class DorisSchemaChangeManager extends SchemaChangeManager {
+    public DorisSchemaChangeManager(DorisOptions dorisOptions) {
+        super(dorisOptions);
+    }
+
+    public boolean truncateTable(String databaseName, String tableName)
+            throws IOException, IllegalArgumentException {
+        String createTableDDL =
```
lvyanquan, line 36: truncateTableDDL

yuxiqian:
Rebased & resolved conflicts, would @lvyanquan like to take a look?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/PaimonMetadataApplier.java
```diff
@@ -325,14 +317,32 @@ private void applyAlterColumnType(AlterColumnTypeEvent event) throws SchemaEvolv
...
         } catch (Catalog.TableNotExistException
                 | Catalog.ColumnAlreadyExistException
                 | Catalog.ColumnNotExistException e) {
             throw new SchemaEvolveException(event, e.getMessage(), e);
         }
     }
+
+    private void applyTruncateTable(TruncateTableEvent event) throws SchemaEvolveException {
+        try (BatchTableCommit batchTableCommit =
+                catalog.getTable(tableIdToIdentifier(event)).newBatchWriteBuilder().newCommit()) {
```
lvyanquan, line 330: As We are currently experiencing some unexpected situations while truncating table with option of `deletion-vectors.enabled: true`. How about throwing exceptions when truncate these tables? WDYT.

yuxiqian: Yeah it should be safer throw an exception in such configuration.

lvyanquan (approved):
LGTM.

lvyanquan:
Hi, @leonardBang could you help to trigger the CI?

leonardBang (approved):

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/PaimonMetadataApplier.java
gongzexin: @yuxiqian Hello, I encountered the following error while using flink-cdc-pipeline-connector-paimon:

`Unable to truncate a table with deletion vectors enabled.`

I then found this PR. Could you explain why `TRUNCATE` cannot be executed on a Paimon table with `deletion-vectors.enabled=true`? From what I understand, Paimon itself allows this operation, so I would like to understand the reason why Flink CDC prohibits it.

gongzexin: > @yuxiqian Hello, I encountered the following error while using [flink-cdc-pipeline-connector-paimon](https://issues.apache.org/jira/browse/FLINK-cdc-pipeline-connector-paimon):
> 
> `Unable to truncate a table with deletion vectors enabled.`
> 
> I then found this PR. Could you explain why `TRUNCATE` cannot be executed on a Paimon table with `deletion-vectors.enabled=true`? From what I understand, Paimon itself allows this operation, so I would like to understand the reason why Flink CDC prohibits it.

@lvyanquan or do you know the reason?

PR #3676, opened 2024-10-31
[hotfix][tests] Tolerate schema change timeout failure and retry, suppress false alarms

yuxiqian:
This doesn't fix the root cause of this failure at all. It just adjusts the testing configuration and allows job failover and restart the job when the "flush event stuck" problem occurs, instead of reporting CI failure immediately.


yuxiqian:
Low priority but should improve CI passing rate, cc @ruanhang1993

yuxiqian:
After some pressure tests, it turns out that job just keeps stuck even after restarts. Seems some special pattern of event mixture will 100% cause FlushEvent to stuck. This won't make anything better, closing it.

PR #3677, opened 2024-10-31
[FLINK-36636][cdc] Timestamp data supports comparison in cdc pipeline transform

aiwenmo:
Timestamp data can not be compared in cdc pipeline transform.

When projecting or filtering code containing timestamp comparisons, the program will encounter errors.

```
TO_TIMESTAMP('2024-01-01 00:00:00') < TO_TIMESTAMP('2024-08-01 00:00:00')
```

Error:

```
org.apache.flink.api.common.InvalidProgramException: Expression cannot be compiled. This is a bug. Please file an issue.
Expression: import static org.apache.flink.cdc.runtime.functions.SystemFunctionUtils.*;toTimestamp("2024-01-01 00:00:00", __time_zone__) < toTimestamp("2024-08-01 00:00:00", __time_zone__) ? 1 : 0
...
...
Caused by: org.codehaus.commons.compiler.CompileException: Line 1, Column 176: Operator "<" not allowed on reference operands 
```

This PR has already solved the problem, and its modifications mainly include the following:

1. Support comparing two identical types which implements Comparable.
2. Support comparing two different numerical types which extends Number.
3. Fix incorrect descriptions in the document: TIMESTAMP_DIFF.
4. Fix existing unit tests and add type comparison unit tests.

### DIFF
--- a/docs/content.zh/docs/core-concept/transform.md
+++ b/docs/content.zh/docs/core-concept/transform.md
@@ -92,10 +92,10 @@ Flink CDC uses [Calcite](https://calcite.apache.org/) to parse expressions and [
 |----------------------|-----------------------------|-----------------------------------------------------------------|
 | value1 = value2      | valueEquals(value1, value2) | Returns TRUE if value1 is equal to value2; returns FALSE if value1 or value2 is NULL. |
 | value1 <> value2     | !valueEquals(value1, value2) | Returns TRUE if value1 is not equal to value2; returns FALSE if value1 or value2 is NULL. |
-| value1 > value2      | value1 > value2             | Returns TRUE if value1 is greater than value2; returns FALSE if value1 or value2 is NULL. |
-| value1 >= value2     | value1 >= value2            | Returns TRUE if value1 is greater than or equal to value2; returns FALSE if value1 or value2 is NULL. |
-| value1 < value2      | value1 < value2             | Returns TRUE if value1 is less than value2; returns FALSE if value1 or value2 is NULL. |
-| value1 <= value2     | value1 <= value2            | Returns TRUE if value1 is less than or equal to value2; returns FALSE if value1 or value2 is NULL. |
+| value1 > value2      | greater(value1, value2)     | Returns TRUE if value1 is greater than value2; returns FALSE if value1 or value2 is NULL. |
+| value1 >= value2     | greaterOrEqual(value1, value2) | Returns TRUE if value1 is greater than or equal to value2; returns FALSE if value1 or value2 is NULL. |
+| value1 < value2      | less(value1, value2)        | Returns TRUE if value1 is less than value2; returns FALSE if value1 or value2 is NULL. |
+| value1 <= value2     | lessOrEqual(value1, value2) | Returns TRUE if value1 is less than or equal to value2; returns FALSE if value1 or value2 is NULL. |
 | value IS NULL        | null == value               | Returns TRUE if value is NULL.                                  |
 | value IS NOT NULL    | null != value               | Returns TRUE if value is not NULL.                              |
 | value1 BETWEEN value2 AND value3 | betweenAsymmetric(value1, value2, value3) | Returns TRUE if value1 is greater than or equal to value2 and less than or equal to value3. |
--- a/docs/content.zh/docs/core-concept/transform.md
+++ b/docs/content.zh/docs/core-concept/transform.md
@@ -92,10 +92,10 @@ Flink CDC uses [Calcite](https://calcite.apache.org/) to parse expressions and [
 |----------------------|-----------------------------|-----------------------------------------------------------------|
 | value1 = value2      | valueEquals(value1, value2) | Returns TRUE if value1 is equal to value2; returns FALSE if value1 or value2 is NULL. |
 | value1 <> value2     | !valueEquals(value1, value2) | Returns TRUE if value1 is not equal to value2; returns FALSE if value1 or value2 is NULL. |
-| value1 > value2      | greater(value1, value2)     | Returns TRUE if value1 is greater than value2; returns FALSE if value1 or value2 is NULL. |
-| value1 >= value2     | greaterOrEqual(value1, value2) | Returns TRUE if value1 is greater than or equal to value2; returns FALSE if value1 or value2 is NULL. |
-| value1 < value2      | less(value1, value2)        | Returns TRUE if value1 is less than value2; returns FALSE if value1 or value2 is NULL. |
-| value1 <= value2     | lessOrEqual(value1, value2) | Returns TRUE if value1 is less than or equal to value2; returns FALSE if value1 or value2 is NULL. |
+| value1 > value2      | greaterThan(value1, value2)     | Returns TRUE if value1 is greater than value2; returns FALSE if value1 or value2 is NULL. |
+| value1 >= value2     | greaterThanOrEqual(value1, value2) | Returns TRUE if value1 is greater than or equal to value2; returns FALSE if value1 or value2 is NULL. |
+| value1 < value2      | lessThan(value1, value2)        | Returns TRUE if value1 is less than value2; returns FALSE if value1 or value2 is NULL. |
+| value1 <= value2     | lessThanOrEqual(value1, value2) | Returns TRUE if value1 is less than or equal to value2; returns FALSE if value1 or value2 is NULL. |
 | value IS NULL        | null == value               | Returns TRUE if value is NULL.                                  |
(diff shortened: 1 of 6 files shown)


aiwenmo:
PTAL @leonardBang @yuxiqian

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperatorTest.java
```diff
@@ -1483,6 +1496,107 @@ void testCastErrorTransform() throws Exception {
                 .hasRootCauseMessage("For input string: \"1.0\"");
     }
 
+    @Test
+    void testCompareTransform() throws Exception {
+        PostTransformOperator transform =
+                PostTransformOperator.newBuilder()
+                        .addTransform(
+                                COMPARE_TABLEID.identifier(),
+                                "col1, 2.1 > CAST(1 AS DOUBLE) as numerical_equal,"
```
yuxiqian, line 1505: It would be nice if we could construct the test case as follows:

* Create upstream schema containing comparable objects, like `(id INT, c1 FLOAT, c2 DOUBLE, c3 TIMESTAMP, ...)`
* Write projection expressions like `id, c1 < 5, c2 > 2.5, c3 <= TO_TIMESTAMP('2024-01-01 00:00:00') ...`
* Post multiple rows into test harness, and checking if results are as expected. For example, row `(1, 4, 3.5, 2023-01-01)` should yield `(1, true, true, true)` and row `(2, 10, 0, '2024-11-01)` should yield `(2, false, false, false)`.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -645,4 +645,32 @@ private static String castObjectIntoString(Object object) {
         }
         return String.valueOf(object);
     }
+
+    public static boolean greater(Comparable comparable1, Comparable comparable2) {
```
yuxiqian, line 649: To keep consistency with [Flink](https://github.com/apache/flink/blob/ed50ae21fc78fa98b015566303f9a80c88278df1/flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/BuiltInFunctionDefinitions.java#L579) and function names, maybe we can use `greaterThan`, `greaterThanOrEqual`, `lessThan`, `lessThanOrEqual`?

docs/content.zh/docs/core-concept/transform.md
```diff
@@ -131,18 +131,18 @@ Flink CDC uses [Calcite](https://calcite.apache.org/) to parse expressions and [
...
+| CURRENT_DATE                                          | currentDate() | Returns the current SQL date in the local time zone. |
+| CURRENT_TIMESTAMP                                     | currentTimestamp() | Returns the current SQL timestamp in the local time zone, the return type is TIMESTAMP_LTZ(3). |
+| NOW()                                                 | now() | Returns the current SQL timestamp in the local time zone, this is a synonym of CURRENT_TIMESTAMP. |
+| DATE_FORMAT(timestamp, string)                        | dateFormat(timestamp, string) | Converts timestamp to a value of string in the format specified by the date format string. The format string is compatible with Java's SimpleDateFormat. |
+| TIMESTAMP_DIFF(timepointunit, timepoint1, timepoint2) | timestampDiff(timepointunit, timepoint1, timepoint2) | Returns the (signed) number of timepointunit between timepoint1 and timepoint2. The unit for the interval is given by the first argument, which should be one of the following values: SECOND, MINUTE, HOUR, DAY, MONTH, or YEAR. |
```
yuxiqian, line 143: Maybe we can leave changes of `TIMESTAMP_DIFF` in FLINK-36647

docs/content/docs/core-concept/transform.md
```diff
@@ -131,18 +131,18 @@ Flink CDC uses [Calcite](https://calcite.apache.org/) to parse expressions and [
...
+| CURRENT_DATE                                          | currentDate() | Returns the current SQL date in the local time zone. |
+| CURRENT_TIMESTAMP                                     | currentTimestamp() | Returns the current SQL timestamp in the local time zone, the return type is TIMESTAMP_LTZ(3). |
+| NOW()                                                 | now() | Returns the current SQL timestamp in the local time zone, this is a synonym of CURRENT_TIMESTAMP. |
+| DATE_FORMAT(timestamp, string)                        | dateFormat(timestamp, string) | Converts timestamp to a value of string in the format specified by the date format string. The format string is compatible with Java's SimpleDateFormat. |
+| TIMESTAMP_DIFF(timepointunit, timepoint1, timepoint2) | timestampDiff(timepointunit, timepoint1, timepoint2) | Returns the (signed) number of timepointunit between timepoint1 and timepoint2. The unit for the interval is given by the first argument, which should be one of the following values: SECOND, MINUTE, HOUR, DAY, MONTH, or YEAR. |
```
yuxiqian, line 143: Ditto

yuxiqian (commented):
Thanks for @aiwenmo's great work! Just left some trivial comments.

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperatorTest.java
aiwenmo: thx. I will  construct the test case in the way you described.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
aiwenmo: I agree with you.

docs/content.zh/docs/core-concept/transform.md
aiwenmo: The current description in the document is incorrect. This modification is to correct the document.

docs/content/docs/core-concept/transform.md
aiwenmo: Ditto

docs/content.zh/docs/core-concept/transform.md
yuxiqian: I see, this change makes sense.

aiwenmo:
PTAL@leonardBang

leonardBang:
> _ No description provided. _

Thanks @aiwenmo for the contribution, minor: a description for PR is welcome

aiwenmo:
A link in a document unrelated to this PR is invalid, I will fix it.
```
ERROR: 1 dead links found!
  [✖] https://vitess.io/docs/get-started/local-docker/ → Status: 404
```
The new link information is as follows:
```
https://vitess.io/docs/20.0/get-started/local-docker/
```

leonardBang:
@aiwenmo Would you  like to append a hotfix commit to fix the doc dead link ?

aiwenmo:
> Would you like to append a hotfix commit to fix the doc dead link ?

Listen to your advice. What do you think? @leonardBang

yuxiqian:
@aiwenmo Could you please rebase this to `master` to re-trigger the CI?

aiwenmo:
Rebased. @yuxiqian

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -846,4 +846,49 @@ private static String castObjectIntoString(Object object) {
...
+        } else if (Number.class.isAssignableFrom(leftClass)
+                && Number.class.isAssignableFrom(rightClass)) {
+            return Double.compare(((Number) lhs).doubleValue(), ((Number) rhs).doubleValue());
+        } else {
+            throw new RuntimeException(
+                    "Comparison of unsupported data types: "
+                            + leftClass.getName()
+                            + " and "
+                            + rightClass.getName());
+        }
```
yuxiqian, line 864: Will it be faster if we use `obj instanceof` instead of reflections?

aiwenmo:
PTAL @yuxiqian

yuxiqian (approved):
Thanks for @aiwenmo's great work, this will simplify conditional expressions quite a lot.

leonardBang (commented):
Thanks @aiwenmo for the contribution and @yuxiqian for the review work, +1

leonardBang (approved):
CI passed, merging....

PR #3680, opened 2024-11-01
[FLINK-36690][runtime] Fix schema operator hanging under extreme parallelized pressure

yuxiqian:
This closes FLINK-36690 by fixing schema operator hanging glitch under extreme parallelized pressure.

---

After FLINK-36114, `SchemaOperator`s will ask for schema evolve permission first, before sending `FlushEvent`s to downstream sinks.

Normal case

However, Flink regards `FlushEvent`s as normal data records and might block it to align checkpoint barriers. It might cause the following deadlock situation:

* `SchemaOperator` A has obtained schema evolution permission
* `SchemaOperator` B does not get the permission and hangs
* `SchemaOperator` A sends a `FlushEvent`, but after a checkpoint barrier
* `SchemaOperator` B received a checkpoint barrier after the schema change event (which is blocked)

Bad case

Now, neither A nor B can post any event records to downstream, and the entire job blocks with the following iconic error message (in TM):

```
0> Schema Registry is busy now, waiting for next request...
1> Schema Registry is busy now, waiting for next request...
2> Schema Registry is busy now, waiting for next request...
[Repeated lines]
```

---

This PR changes the schema evolution permission requesting workflow by:

* `SchemaOperator`s emit `FlushEvent` immediately when they received a `SchemaChangeEvent`.
* A schema change request could only be permitted when a) SchemaRegistry is IDLE and b) the requesting `SchemaOperator` has finished data flushing already.

Since `FlushEvent` might be emitted from multiple `SchemaOperator`s simultaneously, a nonce value that is uniquely bound to a schema change event is added into `FlushEvent` payload. `WAITING_FOR_FLUSH` stage is no longer necessary since this state will not block the `SchemaRegistry` but one single `SchemaOperator` now.

Proposed

It should be noted that current schema evolution design implicitly assumes that for each table, it won't be schema evolving in subTask A and emits normal data change events without blocking in subTask B at the same time.

So, after a `SchemaOperator` successfully triggered a flush event, there can't be any more uncommitted dirty data got written down since 1) any following data from this subTask is still being blocked and 2) other subTask can't carry any data belonging to this tableId (according to our previous guarantee).

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/event/FlushEvent.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/event/FlushEvent.java
@@ -28,12 +28,23 @@ public class FlushEvent implements Event {
     /** The schema changes from which table. */
     private final TableId tableId;
 
-    public FlushEvent(TableId tableId) {
+    /**
+     * Nonce code to distinguish flush events corresponding to each schema change event from
+     * different subTasks.
+     */
+    private final long nonce;
+
+    public FlushEvent(TableId tableId, long nonce) {
         this.tableId = tableId;
+        this.nonce = nonce;
     }
 
     public TableId getTableId() {
         return tableId;
+    }
+
+    public long getNonce() {
+        return nonce;
     }
 
     @Override
@@ -45,11 +56,16 @@ public class FlushEvent implements Event {
             return false;
         }
         FlushEvent that = (FlushEvent) o;
-        return Objects.equals(tableId, that.tableId);
+        return Objects.equals(tableId, that.tableId) && Objects.equals(nonce, that.nonce);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(tableId);
+        return Objects.hash(tableId, nonce);
+    }
+
+    @Override
+    public String toString() {
+        return "FlushEvent{" + "tableId=" + tableId + ", nonce=" + nonce + '}';
     }
 }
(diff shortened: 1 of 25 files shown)


yuxiqian:
Need more eyes on this PR since this PR tweaks schema evolution communication process. cc @leonardBang @ruanhang1993

yuxiqian:
Based on previous discussions, I've made the following changes:

* Adjusted flush event nonce generating algorithm. Now the higher 32 bits are timestamp and lower 32 bits are Java hash code (like snowflake ID).
* added parallelized schema change cases in MySQL integrated test & route e2e test.

As we're not going to have it in 3.2.1, I would propose downgrading parallelized E2e cases to single-parallelism ones (#3718) to avoid breaking our CI pipeline.

leonardBang:
Thanks @yuxiqian for the contribution, it's great that you've added the flow picture to make the PR easy to catch.

yuxiqian:
@leonardBang Will this PR be reviewed soon? I'm planning to implement FLINK-36763 based on this.

leonardBang:
> @leonardBang Will this PR be reviewed soon? I'm planning to implement [FLINK-36763](https://issues.apache.org/jira/browse/FLINK-36763) based on this.

Added to my todo list

Shawn-Hx (commented):
Thanks for the great work! Left some minor comments.

yuxiqian:
Thanks for @Shawn-Hx's review! Addressed your comments.

Sorry to make your review tougher, but considering FLINK-36763 is expected to modify current codebase anyway, I've baked changes in this PR into #3801. Looking forward to hearing from you on that, too.

yuxiqian:
Baked into #3801.

PR #3695, opened 2024-11-07
[BP-3.2][FLINK-35982] Fix unable to transform metadata without projections

yuxiqian:
Cherry-picked from #3592.


yuxiqian:
This requires #3702 being merged first.

PR #3698, opened 2024-11-07
[FLINK-36647][transform] Support Timestampdiff and Timestampadd function in cdc pipeline transform

aiwenmo:
Support TimestampDiff function in cdc pipeline transform.

Now. The TimestampDiff function implemented in the current code of CDC is as follows:

```
TIMESTAMP_DIFF('DAY', TO_TIMESTAMP('1970-01-01 00:00:00'), TO_TIMESTAMP('1970-01-02 00:00:00')) = 1 
```

The Timestampdiff and Timestampadd function implemented in Flink SQL is as follows:
| SQL Function | Table Function |  |
| :---- | :----: | ----: |
| TIMESTAMPDIFF(timepointunit, timepoint1, timepoint2) | timestampDiff(TIMEPOINTUNIT, TIMEPOINT1, TIMEPOINT2) | Returns the (signed) number of timepointunit between timepoint1 and timepoint2. The unit for the interval is given by the first argument, which should be one of the following values: SECOND, MINUTE, HOUR, DAY, MONTH, or YEAR. |
| TIMESTAMPADD(timeintervalunit, interval, timepoint) | N/A |  |

I have implemented the Timestampdiff and Timestampadd function with the same definition as Flink.

```
TIMESTAMPDIFF(DAY, TO_TIMESTAMP('1970-01-01 00:00:00'), TO_TIMESTAMP('1970-01-02 00:00:00')) = 1  
TIMESTAMPADD(DAY, 1, TO_TIMESTAMP('1970-01-02 00:00:00'))
```

Now. Not support TIMESTAMP_TZ.
Todo: Upgrade Apache Calcite into 1.37.0 to support TIMESTAMP_TZ.
https://issues.apache.org/jira/browse/FLINK-37123


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -184,6 +231,53 @@ public static int timestampDiff(String symbol, long fromDate, long toDate) {
...
+                field = Calendar.DAY_OF_YEAR;
+                break;
+            case "MONTH":
+                field = Calendar.MONTH;
+                break;
+            case "YEAR":
+                field = Calendar.YEAR;
+                break;
+            default:
+                LOG.error("Unsupported timestamp add: {}", timeintervalunit);
```
yuxiqian, line 274: I think it's not very meaningful to log here since it was evaluated in Janino engine.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -184,6 +231,53 @@ public static int timestampDiff(String symbol, long fromDate, long toDate) {
...
+                throw new RuntimeException("Unsupported timestamp add: " + timeintervalunit);
```
yuxiqian, line 275: "Unsupported timestamp interval unit. Supported types are: SECOND, MINUTE, HOUR, ..."

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -184,6 +231,53 @@ public static int timestampDiff(String symbol, long fromDate, long toDate) {
...
+        return timestampadd(timeintervalunit, interval, timepoint.getMillisecond());
+    }
+
+    public static TimestampData timestampadd(
+            String timeintervalunit, int interval, TimestampData timepoint) {
+        return timestampadd(timeintervalunit, interval, timepoint.getMillisecond());
+    }
+
+    public static TimestampData timestampadd(
+            String timeintervalunit, int interval, long timepoint) {
```
yuxiqian, line 250: Use camel-case names for parameters, like `timeIntervalUnit`. Please also check other newly-added functions' parameters.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -184,6 +231,53 @@ public static int timestampDiff(String symbol, long fromDate, long toDate) {
...
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(new Date(timepoint));
+        int field;
+        switch (timeintervalunit) {
```
yuxiqian, line 254: Maybe we can downcase the unit first?

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperatorTest.java
```diff
@@ -732,12 +831,107 @@ void testTimestampDiffTransform() throws Exception {
...
+                        transformFunctionEventEventOperatorTestHarness.getOutputRecords().poll())
+                .isEqualTo(
+                        new StreamRecord<>(
+                                new CreateTableEvent(TIMESTAMPADD_TABLEID, TIMESTAMPADD_SCHEMA)));
+        transform.processElement(new StreamRecord<>(insertEvent));
+        Assertions.assertThat(
+                        transformFunctionEventEventOperatorTestHarness.getOutputRecords().poll())
+                .isEqualTo(new StreamRecord<>(insertEventExpect));
+
+        DataChangeEvent insertEvent2 =
```
yuxiqian, line 915: Currently, all test data are hard-encoded inside, and it's not easy to extend it to cover more cases. I wonder if organizing test cases in this way would be more effective?

* Input schema is columns in TIMESTAMP, TIMESTAMP_TZ, and TIMESTAMP_LTZ
* Append multiple computed rows to utilize newly-added functions in transform rules
* Insert multiple rows (with some extreme cases, like possibly out-of-range cases)

For example, our input schema might be `TIMESTAMP ts`, and transform rule could be `TIMESTAMPADD(DAY, 1, ts) AS one_day_later, ...`.

Now, after inserting original data `2020-01-01` and `2023-12-31`, we may get `2020-01-02` and `2024-01-01` as expected results.

Anyway, it's just an example. Please consider adding more corner use cases here.

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/parser/TransformParserTest.java
```diff
@@ -249,6 +249,8 @@ public void testTranslateFilterToJaninoExpression() {
                 "TO_DATE(dt, 'yyyy-MM-dd')", "toDate(dt, \"yyyy-MM-dd\", __time_zone__)");
         testFilterExpression("TO_TIMESTAMP(dt)", "toTimestamp(dt, __time_zone__)");
         testFilterExpression("TIMESTAMP_DIFF('DAY', dt1, dt2)", "timestampDiff(\"DAY\", dt1, dt2)");
+        testFilterExpression("TIMESTAMPDIFF(DAY, dt1, dt2)", "timestampdiff(\"DAY\", dt1, dt2)");
+        testFilterExpression("TIMESTAMPADD(DAY, 1, dt)", "timestampadd(\"DAY\", 1, dt)");
```
yuxiqian, line 253: Consider add cases for all valid timestamp units (and invalid ones).

yuxiqian: Also, why some functions require quoting timeUnits (like `'DAY'`) while some doesn't? Any reasons behind this?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
aiwenmo: I will remove it.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
aiwenmo: Get!

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
aiwenmo: okay

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/parser/TransformParserTest.java
aiwenmo: The single quotes added in TIMESTAMP_DIFF are for the old implementation scheme. Now a scheme that can be achieved without single quotes has been developed.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -108,83 +108,305 @@ public static TimestampData toTimestamp(String str, String format, String timezo
...
+        if (fromTimestamp == null || toTimestamp == null) {
+            return null;
+        }
         return timestampDiff(
-                symbol, fromTimestamp.getEpochMillisecond(), toTimestamp.getEpochMillisecond());
+                timeIntervalUnit,
+                fromTimestamp.getEpochMillisecond(),
+                ZoneId.systemDefault().getId(),
+                toTimestamp.getEpochMillisecond(),
+                ZoneId.systemDefault().getId());
```
yuxiqian, line 123: I think it's potentially risky to use `ZoneId.systemDefault` by default since crossing Daylight Saving Time boundaries might cause unexpected results. Why not just stick to UTC and remove the "TimeZone" parameter from all `timestampdiff` function (they're not exposed to user anyway)?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -108,83 +108,305 @@ public static TimestampData toTimestamp(String str, String format, String timezo
...
+            long fromDate,
+            String fromDateZoneId,
+            long toDate,
+            String toDateZoneId) {
+        if (fromDateZoneId == null || fromDateZoneId.isEmpty()) {
+            fromDateZoneId = "UTC";
+        }
+        if (toDateZoneId == null || toDateZoneId.isEmpty()) {
+            toDateZoneId = "UTC";
+        }
```
yuxiqian, line 308: Ditto, why introducing timezone-aware timestampDiff?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -108,83 +108,305 @@ public static TimestampData toTimestamp(String str, String format, String timezo
...
+            return null;
+        }
+        return TimestampData.fromMillis(
+                timestampadd(timeIntervalUnit, interval, timePoint.getMillisecond()));
+    }
+
+    private static long timestampadd(String timeIntervalUnit, int interval, long timePoint) {
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTimeZone(TimeZone.getTimeZone("UTC"));
+        calendar.setTime(new Date(timePoint));
```
yuxiqian, line 381: Can we cache a static `Calendar` and avoid creating one every time when this method got invoked?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/metadata/TransformSqlReturnTypes.java
```diff
@@ -188,4 +188,12 @@ public RelDataType inferReturnType(SqlOperatorBinding opBinding) {
             return opBinding.getTypeFactory().leastRestrictive(types);
         }
     }
+
+    public static final SqlReturnTypeInference ARG2_TIMESTAMP_FORCE_NULLABLE =
+            new OrdinalReturnTypeInference(2) {
+                @Override
+                public RelDataType inferReturnType(SqlOperatorBinding opBinding) {
+                    return super.inferReturnType(opBinding);
+                }
```
yuxiqian, line 197: Is this override necessary?

yuxiqian (commented):
Thanks for @aiwenmo's quick response, left some general comments.

Not sure if there's any potential incompatibilities after bumping Calcite version, needs hearing from @leonardBang and @ruanhang1993.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -108,83 +108,203 @@ public static TimestampData toTimestamp(String str, String format, String timezo
...
-    public static int timestampDiff(
-            String symbol, ZonedTimestampData fromTimestamp, ZonedTimestampData toTimestamp) {
-        return timestampDiff(symbol, fromTimestamp.getMillisecond(), toTimestamp.getMillisecond());
+    public static Integer timestampdiff(
+            String timeIntervalUnit,
+            TimestampData fromTimestamp,
+            LocalZonedTimestampData toTimestamp,
+            String timezone) {
+        return timestampDiff(timeIntervalUnit, fromTimestamp, toTimestamp, timezone);
     }
```
yuxiqian, line 181: suggestion: since `timestampdiff` is merely a wrapper of `timestampDiff`, we can put proxy methods in a single place and add comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -108,83 +108,203 @@ public static TimestampData toTimestamp(String str, String format, String timezo
...
-                symbol, fromTimestamp.getMillisecond(), toTimestamp.getEpochMillisecond());
+    public static Integer timestampdiff(
+            String timeIntervalUnit,
+            LocalZonedTimestampData fromTimestamp,
+            TimestampData toTimestamp,
+            String timezone) {
+        return timestampDiff(timeIntervalUnit, fromTimestamp, toTimestamp, timezone);
+    }
+
+    public static Integer timestampDiff(
```
yuxiqian, line 207: nit: Will this method be directly invoked by generated expressions? If not, maybe we can mark it `private` and change its name (or move it to `DateTimeUtils`) to avoid adding another overloading versions into it.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/typeutils/DataTypeConverter.java
```diff
@@ -129,6 +132,8 @@ public static RelDataType convertCalciteType(
                 return typeFactory.createSqlType(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE);
             case TIMESTAMP_WITHOUT_TIME_ZONE:
                 return typeFactory.createSqlType(SqlTypeName.TIMESTAMP);
+            case TIMESTAMP_WITH_TIME_ZONE:
```
yuxiqian, line 135: nit: maybe add a TODO comment with JIRA ticket here

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -108,83 +108,203 @@ public static TimestampData toTimestamp(String str, String format, String timezo
...
+                field = Calendar.MINUTE;
+                break;
             case "HOUR":
-                return second.intValue() / 3600;
+                field = Calendar.HOUR;
+                break;
             case "DAY":
-                return second.intValue() / (24 * 3600);
+                field = Calendar.DAY_OF_YEAR;
+                break;
```
yuxiqian, line 293: Seems `DAY_OF_MONTH`, `DAY_OF_YEAR`, `DAY_OF_WEEK` are all synonyms of `Calendar.DATE` when calling `Calendar#add`. Maybe just use `DATE` for clarity?

yuxiqian (approved):
Thanks @aiwenmo for your quick response! Just left some minor comments.

Would @leonardBang like to review it?

aiwenmo:
PTAL @yuxiqian

leonardBang (approved):
LGTM, wait the CI green

PR #3564, opened 2024-08-22
[FLINK-36105] Fix unable to restore from checkpoint with Flink 1.20

yuxiqian:
This closes FLINK-36105.

Some manual testing reveals that Flink CDC job could not restore from savepoint when using Flink 1.20.0. Turns out FLINK-34455 (https://github.com/apache/flink/pull/24320) has moved `RestoreMode` enum class from `flink-runtime` to `flink-core`, causing incompatiblity since CDC is still developed upon Flink 1.18.

Using some reflection hackaround to circumvent this for now.

PR #3701, opened 2024-11-08
[BP-3.2][FLINK-36105] Fix unable to restore from checkpoint with Flink 1.20

yuxiqian:
Cherry-picked from #3564.


leonardBang (approved):
+1

PR #3713, opened 2024-11-11
[FLINK-36699][common] Fix nullability when converting cdc data type to flink data type

Shawn-Hx:
Correct the nullability for data type ARRAY and MAP in method `DataTypeUtils#toFlinkDataType`.


ruanhang1993 (approved):
LGTM.

PR #3691, opened 2024-11-07
[FLINK-36285][doris] Fix unable to alter column type without default value specified

yuxiqian:
This closes FLINK-36285.

Thanks to @qg-lin's great work in https://github.com/apache/doris-flink-connector/pull/490, doris-flink-connector now supports altering column type without passing default value explicitly. Bumping dependency version should close this issue.

PR #3725, opened 2024-11-13
[FLINK-36285][doris] Fix unable to alter column type without default value specified

yuxiqian:
Cherry-picked from #3691.


yuxiqian:
Upgrading dependency version might break compatibility. Just exclude this from 3.2.1.

PR #3729, opened 2024-11-13
[FLINK-36669][pipeline-connector][ES] The flink version of ES should use parent's pom

Gianzie:
[FLINK-36669](https://issues.apache.org/jira/browse/FLINK-36669)

When I was looking at the interaction between cdc-pipeline and flink, I found that there were multiple versions of flink implementations in the source code. The pom file of ES pipeline did not use the parent flink version like other connector pipelines. I think this is a small mistake because there is no special statement here.


leonardBang:
Thank @Gianzie for the contribution,  +1 for the change, but we need to confirm that es connector works well with Flink 1.19 first , because recently the master branch bumped the flink version to 1.19

PR #3744, opened 2024-11-20
[FLINK-36750][paimon] Restore SinkWrite from state to keep the information of previous SinkWrite when schema evolution happen.

lvyanquan:
Use replace API instead of removing and recreating StoreSinkWrite to keep the write state consistent with previous  StoreSinkWrite.


lvyanquan:
@yuxiqian @leonardBang  Could you please help to review this?

leonardBang (approved):
+1

leonardBang:
CI passed, merging...

PR #3749, opened 2024-11-21
[FLINK-36754][transform] when projection is empty or null, it is treated as an asterisk

MOBIN-F:
When projection is empty or NULL, the projection should be treated as an asterisk. The following transform rule should throw an exception [different column counts], but the actual schema columns are id and name, which is not expected.
```
transform:
  - source-table: testdb.test_table1
    filter: age> 8 and age < 18
  - source-table: testdb.test_table1
    projection: id,UPPER(name) AS name
    filter: age >= 18 
```
 
The following transform rule will throw exception [different column counts]
```
transform:
  - source-table: testdb.test_table1
    projection: *
    filter: age> 8 and age < 18
  - source-table: testdb.test_table1
    projection: id,UPPER(name) AS name
    filter: age >= 18 
```
We should treat projection as asterisk when it is empty or NULL, to be consistent

### DIFF
--- a/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/TransformDef.java
+++ b/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/TransformDef.java
@@ -75,16 +75,16 @@ public class TransformDef {
         return sourceTable;
     }
 
-    public Optional<String> getProjection() {
-        return Optional.ofNullable(projection);
+    public String getProjection() {
+        return projection;
     }
 
     public boolean isValidProjection() {
         return !StringUtils.isNullOrWhitespaceOnly(projection);
     }
 
-    public Optional<String> getFilter() {
-        return Optional.ofNullable(filter);
+    public String getFilter() {
+        return filter;
     }
 
     public boolean isValidFilter() {
--- a/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/TransformDef.java
+++ b/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/TransformDef.java
@@ -20,7 +20,6 @@ package org.apache.flink.cdc.composer.definition;
 import org.apache.flink.cdc.common.utils.StringUtils;
 
 import java.util.Objects;
-import java.util.Optional;
 
 /**
  * Definition of a transformation.
@@ -91,8 +90,8 @@ public class TransformDef {
         return !StringUtils.isNullOrWhitespaceOnly(filter);
     }
 
-    public Optional<String> getDescription() {
-        return Optional.ofNullable(description);
+    public String getDescription() {
+        return description;
     }
 
     public String getPrimaryKeys() {
--- a/flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineTransformITCase.java
+++ b/flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineTransformITCase.java
@@ -382,6 +382,7 @@ class FlinkPipelineTransformITCase {
                         null,
                         null,
                         null,
+                        null,
                         null);
 
         TransformDef emptyProjection =
@@ -392,6 +393,7 @@ class FlinkPipelineTransformITCase {
                         null,
                         null,
                         null,
+                        null,
                         null);
 
         TransformDef asteriskProjection =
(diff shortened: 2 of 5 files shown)


flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineTransformITCase.java
```diff
@@ -354,6 +355,93 @@ void testMultiTransformMissingProjection(ValuesDataSink.SinkApi sinkApi) throws
...
+                                            Collections.emptyList()))
+                    .rootCause()
+                    .hasMessage(
+                            "Unable to merge schema columns={`id` BIGINT,`name` VARCHAR(255),`age` TINYINT,`description` STRING}, primaryKeys=id, options=() "
+                                    + "and columns={`id` BIGINT,`name` STRING}, primaryKeys=id, options=() with different column counts.");
+        }
+
+        for (TransformDef transformDef : transformDefList) {
+            assertThatThrownBy(
+                    () ->
```
yuxiqian, line 419: Also add some chaining assertions to make sure we're catching correct Exception and messages?

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/TransformDef.java
```diff
@@ -72,16 +72,16 @@ public String getSourceTable() {
...
     }
 
     public boolean isValidProjection() {
         return !StringUtils.isNullOrWhitespaceOnly(projection);
     }
 
-    public Optional<String> getFilter() {
-        return Optional.ofNullable(filter);
+    public String getFilter() {
+        return filter;
```
yuxiqian, line 84: CMIIW, This class has no accessibility annotation... so it's an `@Internal` class and breaking changes should be fine.

Maybe we can also tweak `getDescription` method to return nullable `String` for consistency?

yuxiqian (commented):
Thanks for @MOBIN-F's quick fix, the new implementation is much cleaner. Just left some minor concerns.

Also, could you please rebase this to `master` branch? Seems there are some conflicting changes in #3758.

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/definition/TransformDef.java
MOBIN-F: done~
ditto

flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineTransformITCase.java
```diff
@@ -370,6 +371,99 @@ void testMultiTransformMissingProjection(ValuesDataSink.SinkApi sinkApi) throws
...
+                                            Collections.emptyList()))
+                    .rootCause()
+                    .isExactlyInstanceOf(IllegalStateException.class)
+                    .hasMessage(
+                            "Unable to merge schema columns={`id` BIGINT,`name` VARCHAR(255),`age` TINYINT,`description` STRING}, primaryKeys=id, options=() "
+                                    + "and columns={`id` BIGINT,`name` STRING}, primaryKeys=id, options=() with different column counts.");
+        }
+
+        for (TransformDef transformDef : transformDefList) {
+            assertThatThrownBy(
```
yuxiqian, line 439: Seems this `assertThatTHrownBy` block still has no following assertions 🤕

MOBIN-F: Thanks for the review!!
I modified the test case to be more rigorous, please help review again~
Deleted redundant assertThatThrownBy and I shouldn't use runGenericTransformTest in for loop, which will cause duplicate data events.

yuxiqian (approved):
Looks nice except a minor comment. Would @leonardBang like to review this? 👀

leonardBang (approved):
+1

MOBIN-F:
found some failed tests, I will fix them later

leonardBang:
@MOBIN-F Could you check the failed tests via search `<<< FAILURE!`?

lvyanquan:
Hi, @MOBIN-F, rebase to master will help to fix ci failure here.

MOBIN-F:
For the issue of not throwing the [different column count] exception, I will temporarily disable the case. I will resolve it as soon as possible in FLINK-37132. The current PR is only to unify the behavior of projection to null, '', or .

MOBIN-F:
test error seems to be unrelated to this PR, can you trigger it again? @leonardBang

PR #3756, opened 2024-11-25
[legal] Add third-party OSS NOTICE

yuxiqian:
Now, Flink CDC's NOTICE file contains no third-party software notice, which violates corresponding open-source license that requires explicit mentioning during redistribution.

I added corresponding third-party open-source software licenses by referencing a properly written [NOTICE file example](https://github.com/apache/paimon/blob/master/NOTICE) from Paimon project.


yuxiqian:
@leonardBang @ruanhang1993 Please take a look as this might block releasing in the future.

aiwenmo:
Hi. What I can think of is that Apache Calcite and Janino still needs to be added.

yuxiqian:
Thanks @aiwenmo for pointing it out! Added missing software packages.

leonardBang (approved):
+1

PR #3758, opened 2024-11-25
[FLINK-36784][common] Support to add metadata columns for data in the meta fields of DataChangeEvent at transform

ruanhang1993:
This PR supports to add metadata in the meta field of DataChangeEvent at transform.

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/SupportedMetadataColumn.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/SupportedMetadataColumn.java
@@ -18,12 +18,13 @@
 package org.apache.flink.cdc.common.source;
 
 import org.apache.flink.cdc.common.annotation.Experimental;
+import org.apache.flink.cdc.common.event.DataChangeEvent;
 import org.apache.flink.cdc.common.types.DataType;
 
 import java.io.Serializable;
 import java.util.Map;
 
-/** A metadata column that the source supports. */
+/** A metadata column that the source supports to read from the meta field. */
 @Experimental
 public interface SupportedMetadataColumn extends Serializable {
     /** Column name. */
@@ -35,6 +36,11 @@ public interface SupportedMetadataColumn extends Serializable {
     /** The returned java class of the reader. */
     Class<?> getJavaClass();
 
-    /** Read the metadata from the dataChangeEvent. */
+    /**
+     * Read the metadata from the {@link DataChangeEvent#meta()}.
+     *
+     * @param metadata the metadata returned from {@link DataChangeEvent#meta()}
+     * @return the value of this metadata found by metadata name
+     */
     Object read(Map<String, String> metadata);
 }
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/SupportedMetadataColumn.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/SupportedMetadataColumn.java
@@ -17,23 +17,24 @@
 
 package org.apache.flink.cdc.common.source;
 
-import org.apache.flink.cdc.common.annotation.PublicEvolving;
+import org.apache.flink.cdc.common.annotation.Experimental;
+import org.apache.flink.cdc.common.types.DataType;
 
-/**
- * {@code DataSource} is used to access metadata and read change data from external systems. It can
- * read data from multiple tables simultaneously.
- */
-@PublicEvolving
-public interface DataSource {
+import java.io.Serializable;
+import java.util.Map;
 
-    /** Get the {@link EventSourceProvider} for reading events from external systems. */
-    EventSourceProvider getEventSourceProvider();
+/** A metadata column that the source supports. */
+@Experimental
+public interface SupportedMetadataColumn extends Serializable {
+    /** Column name. */
+    String getName();
 
-    /** Get the {@link MetadataAccessor} for accessing metadata from external systems. */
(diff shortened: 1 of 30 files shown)


flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/SupportedMetadataColumn.java
```diff
@@ -0,0 +1,40 @@
...
+    /** Column name. */
+    String getName();
+
+    /** The data type of this column in Flink CDC. */
+    DataType getType();
+
+    /** The returned java class of the reader. */
+    Class<?> getJavaClass();
+
+    /** Read the metadata from the dataChangeEvent. */
```
lvyanquan, line 38: Link to the actual class or variable.

lvyanquan (approved):
LGTM!

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformFilterProcessor.java
```diff
@@ -67,15 +73,22 @@ public static TransformFilterProcessor of(
...
+            List<Object> udfFunctionInstances,
+            SupportedMetadataColumn[] supportedMetadataColumns) {
         return new TransformFilterProcessor(
-                tableInfo, transformFilter, timezone, udfDescriptors, udfFunctionInstances);
+                tableInfo,
+                transformFilter,
+                timezone,
+                udfDescriptors,
+                udfFunctionInstances,
+                supportedMetadataColumns);
```
yuxiqian, line 84: Feels like we need something like `TransformContext` to gather all these things together instead of extending arguments lists forever, but it's not a must in this PR.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformFilterProcessor.java
```diff
@@ -135,6 +159,19 @@ private Object[] generateParams(BinaryRecordData record, long epochTime, String
...
+            for (SupportedMetadataColumn supportedMetadataColumn : supportedMetadataColumns) {
+                if (supportedMetadataColumn.getName().equals(columnName)) {
+                    params.add(supportedMetadataColumn.read(meta));
+                    foundInMeta = true;
+                    break;
+                }
+            }
+            if (foundInMeta) {
+                continue;
+            }
```
yuxiqian, line 173: Is it possible to let `supportedMetadataColumns()` return a `Map<String, SupportedMetadataColumn>`? This code would be executed for each stream data record and this array traversing seems a little inefficient.

yuxiqian (approved):
Thanks for @ruanhang1993's nice work, looks nice overall. Just left some optional comments that should not block this PR.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformFilterProcessor.java
ruanhang1993: Thanks for this suggestion. Let open a new issue for this optimization.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformFilterProcessor.java
ruanhang1993: `Map` is not `Serializable` for `TransformRule`. I could build a map for this class.

yuxiqian: `UnmodifiableMap` could be an option. Of course building one dynamically is fine, too.

lvyanquan (approved):
LGTM.

Updating  docs for the newly added option is also welcomed.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
```diff
@@ -272,4 +272,13 @@ public class MySqlDataSourceOptions {
...
                                     + "scan.binlog.newly-added-table.enabled: only do binlog-reading for newly added table during binlog reading phase.");
+
+    @Experimental
+    public static final ConfigOption<String> METADATA_LIST =
+            ConfigOptions.key("metadata.list")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "List of readable metadata from SourceRecord to be passed to downstream, split by `,`. "
+                                    + "Refer to MySqlReadableMetadata, available readable metadata are: table_name,database_name,op_ts,row_kind.");
```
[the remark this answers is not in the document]
ruanhang1993: There is only one metadata `op_ts` for mysql pipeline connector. We cannot use it to generate the description.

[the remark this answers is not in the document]
yuxiqian: I see, that makes sense.

leonardBang (approved):
LGTM

PR #3760, opened 2024-11-26
[FLINK-36790][cdc-connector][paimon] set waitCompaction with true dur…

stayrascal:
set waitCompaction with true during prepare common of PaimonWriter to avoid CME problem


lvyanquan (approved):
+1, good catch!

stayrascal:
> +1, good catch!

Thanks for reviewing, do you know how to rerun the CI since one pipeline failed?

lvyanquan:
The failure of CI is case of postgres connector, which is unrelated to the changes, I think that it didn't block this pr, we can trace it in another issue.

stayrascal:
> The failure of CI is case of postgres connector, which is unrelated to the changes, I think that it didn't block this pr, we can trace it in another issue.

ok, so what can I do and then the PR can be merged?

lvyanquan:
@ruanhang1993 @leonardBang  Could you please take a look at this?

leonardBang (approved):

stayrascal:
Hi @lvyanquan @leonardBang Thanks for review this PR, may I check what else I can do to merge this PR?

leonardBang:
@stayrascal Could you check the failed CI tests？

stayrascal:
> @stayrascal Could you check the failed CI tests？

Sorry for reply late, it looks that all CI cases are passed now. Previously the postgress connector test case failed but it's not related to this PR as lvyanquan mentioned.

PR #3767, opened 2024-11-28
[FLINK-36813][cdc-connectors][mysql] support mysql sync part columns

JNSimba:
**Background**
In some scenarios, MySQL synchronization only expects to synchronize specified fields instead of all fields in the table.
1. The user only has the permission for some fields in MySQL
2. The user has too many fields in a single table and only wants to synchronize some fields, for example, here https://github.com/apache/flink-cdc/discussions/3058

**Current situation**
For the incremental stage, you only need to configure the column.include.list property of debezium to support the synchronization of some fields in the incremental stage, refer to: https://debezium.io/documentation/reference/1.9/connectors/mysql.html#mysql-property-column-include-list

For the full snapshot stage, * is currently used in MySqlSnapshotSplitReadTask, refer to
```
if (isScanningData) {
return buildSelectWithRowLimits(
tableId, limitSize, "*", Optional.ofNullable(condition), Optional.empty()); 
```

**Solution**
We can refer to debezium [RelationalSnapshotChangeEventSource](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java#L752-L776), The user configures column.include.list, and then captures the specific columns in MySqlSnapshotSplitReadTask, and splices them when constructing Scan SQL.


JNSimba:
@leonardBang @ruanhang1993 PTAL, Thanks

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/StatementUtils.java
```diff
@@ -131,8 +131,12 @@ public static Object queryNextChunkMax(
     }
 
     public static String buildSplitScanQuery(
```
lvyanquan, line 133: Maybe we can add a test for the method as there is not test here before.

ruanhang1993:
@JNSimba Thanks for your PR.

I have some questions about the usage of [column.include.list](https://debezium.io/documentation/reference/1.9/connectors/mysql.html#mysql-property-column-include-list).

Consider that there are 3 tables db1.table1, db1.table2 and db1.table3 to read, whose schema are the same.
| name      | type | note |
| ----------- | ----------- | ----------- |
| id      | varchar       | primary key|
| name   | varchar        | |
| age   | int        | |
| address   | varchar        | |

For a table source as follows, will the change events contain all columns for `db1.table2` and `db1.table3` in incremental phase?
```sql
CREATE TABLE test_tbl (
    `id` STRING NOT NULL,
    name STRING,
    age INT,
    address STRING,
    primary key (`id`) not enforced
) WITH (
  'connector' = 'mysql-cdc',
  'database-name' = 'db1',
  'table-name' = 'table[0-9]+',
  'debezium.column.include.list' = 'db1.table1.id,db1.table1.name',
   ......
);
```

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
```diff
@@ -109,6 +117,21 @@ public MySqlSnapshotSplitReadTask(
         this.snapshotChangeEventSourceMetrics = snapshotChangeEventSourceMetrics;
         this.hooks = hooks;
         this.isBackfillSkipped = isBackfillSkipped;
+
+        // setting column filter
+        String columnIncludeList =
+                sourceConfig.getDbzConfiguration().getString(COLUMN_INCLUDE_LIST);
+        String columnExcludeList =
+                sourceConfig.getDbzConfiguration().getString(COLUMN_EXCLUDE_LIST);
```
ruanhang1993, line 125: It seems that only one of `COLUMN_INCLUDE_LIST` and `COLUMN_EXCLUDE_LIST` will take effect.
And `COLUMN_INCLUDE_LIST` has a higher priority.

Please add an warning for the useless setting.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
```diff
@@ -109,6 +117,21 @@ public MySqlSnapshotSplitReadTask(
...
+        if (columnIncludeList != null) {
+            this.columnFilter =
+                    Tables.ColumnNameFilterFactory.createIncludeListFilter(
+                            columnIncludeList, ColumnFilterMode.CATALOG);
+        } else {
+            this.columnFilter =
+                    Tables.ColumnNameFilterFactory.createExcludeListFilter(
+                            columnExcludeList, ColumnFilterMode.CATALOG);
+        }
```
ruanhang1993, line 134: Only one of `COLUMN_INCLUDE_LIST` and `COLUMN_EXCLUDE_LIST` will take effect in snapshot phase. 
But both of them will pass as the debezium setting in the incremental phase. 
Will this setting cause the different behavior between snapshot phase and incremental phase?

yuxiqian:
Hi @JNSimba, could you please rebase this PR with latest `master` when available?

* JUnit 4 style test annotations should be changed to JUnit 5 equivalents

  * `org.junit.Assert`
  * `org.hamcrest.*`

Running `mvn verify -DskipTests` locally could check if all these requirements have been satisfied.

ruanhang1993:
@JNSimba Could we provide this feature by `SupportsProjectionPushDown`? 
`debezium.column.include.list` and `debezium.column.exclude.list` is hard for users to understand and use. We could use `SupportsProjectionPushDown` interface to automatically generate the debezium setting. In this way we don't need to validate whether there is a conflict between the provided schema and the debezium setting.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
```diff
@@ -305,6 +329,24 @@ private void createDataEventsForTable(
...
+                                                table.id().catalog(),
+                                                table.id().schema(),
+                                                table.id().table(),
+                                                columnName))
+                        .map(columnName -> jdbcConnection.quotedColumnIdString(columnName))
+                        .collect(Collectors.toList());
+        if (columnNames.isEmpty()) {
+            columnNames.add("*");
+        }
+        return String.join(", ", columnNames);
```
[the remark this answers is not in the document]
JNSimba: done

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/StatementUtils.java
JNSimba: done

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
JNSimba: done

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
JNSimba: This will only take effect in the snapshot phase, because MySQL's binlog cannot be configured to capture partial columns. In the incremental phase, projection can be used to query partial columns.

JNSimba:
> @JNSimba Could we provide this feature by `SupportsProjectionPushDown`? `debezium.column.include.list` and `debezium.column.exclude.list` is hard for users to understand and use. We could use `SupportsProjectionPushDown` interface to automatically generate the debezium setting. In this way we don't need to validate whether there is a conflict between the provided schema and the debezium setting.

@ruanhang1993 `SupportsProjectionPushDown` is indeed an elegant approach, but some of the cdc logic seems to have problems in the push-down phase, for example, here

If you run the `MySqlConnectorITCase.testConsumingAllEvents` case, the chunkkey will be adjusted here. For example, if my project column does not have a chunkkey, an error will be reported

https://github.com/apache/flink-cdc/blob/b437a49e67a2ef84c59cde09f2bfc372f45b31fa/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java#L267C1-L274C37
```java
                Object[] chunkKey =
                        RecordUtils.getSplitKey(
                                splitKeyType, statefulTaskContext.getSchemaNameAdjuster(), target);
                for (FinishedSnapshotSplitInfo splitInfo : finishedSplitsInfo.get(tableId)) {
                    if (RecordUtils.splitKeyRangeContains(
                                    chunkKey, splitInfo.getSplitStart(), splitInfo.getSplitEnd())
                            && position.isAfter(splitInfo.getHighWatermark())) {
                        return true;
```

ruanhang1993:
> > @JNSimba Could we provide this feature by `SupportsProjectionPushDown`? `debezium.column.include.list` and `debezium.column.exclude.list` is hard for users to understand and use. We could use `SupportsProjectionPushDown` interface to automatically generate the debezium setting. In this way we don't need to validate whether there is a conflict between the provided schema and the debezium setting.
> 
> @ruanhang1993 `SupportsProjectionPushDown` is indeed an elegant approach, but some of the cdc logic seems to have problems in the push-down phase, for example, here
> 
> If you run the `MySqlConnectorITCase.testConsumingAllEvents` case, the chunkkey will be adjusted here. For example, if my project column does not have a chunkkey, an error will be reported
> 
> https://github.com/apache/flink-cdc/blob/b437a49e67a2ef84c59cde09f2bfc372f45b31fa/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java#L267C1-L274C37
> 
> ```java
>                 Object[] chunkKey =
>                         RecordUtils.getSplitKey(
>                                 splitKeyType, statefulTaskContext.getSchemaNameAdjuster(), target);
>                 for (FinishedSnapshotSplitInfo splitInfo : finishedSplitsInfo.get(tableId)) {
>                     if (RecordUtils.splitKeyRangeContains(
>                                     chunkKey, splitInfo.getSplitStart(), splitInfo.getSplitEnd())
>                             && position.isAfter(splitInfo.getHighWatermark())) {
>                         return true;
> ```

@JNSimba If the chunk key column is not contained in the projection, the snapshot phase will fail because of the lost of the chunk column.
In this case, I think we need to contain these columns in snapshot phase and incremental phase. And apply the projection in the emitter.

ruanhang1993:
@JNSimba In the backfill task, there may be one problem for projection. See `RecordUtils.upsertBinlog` in `SnapshotSplitReader#pollSplitRecords`.

For tables with PK, we use all PK columns to deduplicate the records.
For tables with no PK, we use all columns to deduplicate the records.

So the limit for the projection is as follows.
For tables with PK, we need to keep all PK columns.
For tables with no PK, we need to keep all columns.

JNSimba:
> @JNSimba In the backfill task, there may be one problem for projection. See `RecordUtils.upsertBinlog` in `SnapshotSplitReader#pollSplitRecords`.
> 
> For tables with PK, we use all PK columns to deduplicate the records. For tables with no PK, we use all columns to deduplicate the records.
> 
> So the limit for the projection is as follows. For tables with PK, we need to keep all PK columns. For tables with no PK, we need to keep all columns.

@ruanhang1993 Thanks for your suggestions. But I modified a version, but found some new problems: 

Like this case: **MySqlTimezoneITCase.testMySqlServerInBerlin**

The id is the primary key, the inconsistency between **the actual projected fields (date_c, time_c, datetime3_c, datetime6_c, timestamp_c , id)** and the fields to be projected in the **select (date_c, time_c, datetime3_c, datetime6_c, timestamp_c )** may cause some problems.

In non-incremental snapshot, **RowDataSerializer.copy** will force the judgment whether the type matches the data, so an error will be reported
https://github.com/apache/flink/blob/release-1.20.1/flink-table/flink-table-runtime/src/main/java/org/apache/flink/table/runtime/typeutils/RowDataSerializer.java#L122-L125
```java
// from.getArity()=6 and   types.length =5
if (from.getArity() != types.length) {
throw new IllegalArgumentException(
"Row arity: " + from.getArity() + ", but serializer arity: " + types.length);
}
```
For this case , it is normal when incrementalSnapshot=true, but it is wrong when false.

PR #3772, opened 2024-12-02
[docs] Cleanup pipeline example / remove 'will soon' from reference to transform support

rmoff:
- **Remove non-mandatory sections from the bare-basics example YAML**
- **At least one pipeline parameter must be specified**
- **s/will soon//**


leonardBang (approved):
+1, thanks @rmoff for the improvement, will merge once doc CI passed

PR #3773, opened 2024-12-02
[docs] Fix ToC to include H1

rmoff:
The existing Table of Contents on the right-hand side of the page does not display H1 elements. 
H1 are used throughout the docs as part of the logical presentation of content.
Therefore it makes sense to include them, as is already done in the main Flink docs.

CleanShot 2024-12-02 at 15 38 23


Thorne-coder (approved):
LGTM！

leonardBang (approved):

PR #3775, opened 2024-12-03
[FLINK-36842] FlinkPipelineComposer allows injecting StreamExecutionEnvironment

maxf-decodable:
As a user of the FlinkPipelineComposer API, I would like to run jobs with a pre-configured StreamExecutionEnvironment of my choosing, for commonality with executing non-Flink-CDC jobs. However, the API only allows me to pass a Configuration object, which FlinkPipelineComposer uses to build its own StreamExecutionEnvironment.

In addition, I would like to retrieve a JobClient from submitted jobs, again for commonality with non-Flink-CDC jobs.

### DIFF
--- a/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/PipelineExecution.java
+++ b/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/PipelineExecution.java
@@ -17,6 +17,8 @@
 
 package org.apache.flink.cdc.composer;
 
+import org.apache.flink.core.execution.JobClient;
+
 /** A pipeline execution that can be executed by a computing engine. */
 public interface PipelineExecution {
 
@@ -27,10 +29,18 @@ public interface PipelineExecution {
     class ExecutionInfo {
         private final String id;
         private final String description;
+        private final JobClient jobClient;
 
         public ExecutionInfo(String id, String description) {
             this.id = id;
             this.description = description;
+            this.jobClient = null;
+        }
+
+        public ExecutionInfo(String id, String description, JobClient jobClient) {
+            this.id = id;
+            this.description = description;
+            this.jobClient = jobClient;
         }
 
         public String getId() {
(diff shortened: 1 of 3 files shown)


flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
```diff
@@ -54,6 +54,11 @@ public class FlinkPipelineComposer implements PipelineComposer {
     private final StreamExecutionEnvironment env;
     private final boolean isBlocking;
 
+    public static FlinkPipelineComposer ofStreamExecutionEnvironment(
+            StreamExecutionEnvironment env) {
+        return new FlinkPipelineComposer(env, false);
+    }
```
sharonx, line 60: hmm to me this essentially expose the constructor and it's a bit opaque to set `isBlocking=false` statically here as you can have blocking mode for stream execution env too. My proposal is to make the constructor below public. 
```
  private FlinkPipelineComposer(StreamExecutionEnvironment env, boolean isBlocking) {
        this.env = env;
        this.isBlocking = isBlocking;
    }
```

@PatrickRen I see that you are tagged in the history of the class. Any opinions?

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/PipelineExecution.java
```diff
@@ -27,10 +29,18 @@ public interface PipelineExecution {
     class ExecutionInfo {
         private final String id;
         private final String description;
+        private final JobClient jobClient;
 
         public ExecutionInfo(String id, String description) {
             this.id = id;
             this.description = description;
+            this.jobClient = null;
```
sharonx, line 37: I'd try to avoid `null` usage as much as possible to avoid potential NPEs.  It does look like the `jobClient` is available in all the places that construct `ExecutionInfo`. So can we just change the constructor? If not, please add `@Nullable` wherever is applicable.

yuxiqian:
Recently, I also found this ability useful when writing an IT case that requires injecting a `LocalEnvironmentWithWebUI` to the composer.

@maxf-decodable Would you like to reopen this PR and rebase with the latest master branch?

PR #3776, opened 2024-12-04
[FLINK-36165][source-connector/mysql] Support capturing snapshot data with conditions

uicosp:
Hey, this is an implementation designed to capture snapshot data with filtering conditions.

For example, by specifying `scan.snapshot.filters: db.user_table:id > 200;`, we can synchronize only the user data where the id is greater than 200.

issue link: https://issues.apache.org/jira/browse/FLINK-36165


yuxiqian:
Thanks for @uicosp's great work! It's indeed a long awaited feature.

Seems Debezium has a similar option called [`snapshot.select.statement.overrides`](https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-property-snapshot-select-statement-overrides), which allows users to project out unwanted columns and filter rows based on custom predicates.

As these options aren't available in incremental framework, it would be nice if we could support both row and column level pushdown with similar syntax, since they're both related to tweaking snapshot querying SQL. WDYT?

Thorne-coder:
> Thanks for @uicosp's great work! It's indeed a long awaited feature.
> 
> Seems Debezium has a similar option called [`snapshot.select.statement.overrides`](https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-property-snapshot-select-statement-overrides), which allows users to project out unwanted columns and filter rows based on custom predicates.
> 
> As these options aren't available in incremental framework, it would be nice if we could support both row and column level pushdown with similar syntax, since they're both related to tweaking snapshot querying SQL. WDYT?

it is also support table api and datastream api  with  incremental framework

uicosp:
Column pruning is certainly a feature worth considering. In theory, it would only require minor adjustments to the current design, such as implementing a syntax like `scan.snapshot.filters: db.user:id>200:id,name,age`.

Taking this a step further, since the pipeline config already includes a transform option, we could leverage it to automatically extract filter conditions and projection fields, dynamically generating scan.snapshot.filters and thereby simplifying the config.

That said, I’m currently focused on other tasks and don't have time to work on this. Once I have time, I’d be happy to revisit and enhance this feature.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
```diff
@@ -242,12 +243,24 @@ private void createDataEventsForTable(
...
         final String selectSql =
                 StatementUtils.buildSplitScanQuery(
                         snapshotSplit.getTableId(),
                         snapshotSplit.getSplitKeyType(),
                         snapshotSplit.getSplitStart() == null,
-                        snapshotSplit.getSplitEnd() == null);
+                        snapshotSplit.getSplitEnd() == null,
+                        filter);
         LOG.info(
                 "For split '{}' of table {} using select statement: '{}'",
```
MOBIN-F, line 265: Since selectSql is applied to each split, assuming the job's filter condition is only **scan.snapshot.filters: test_table:id=803326**, it seems that the performance is not as good as jdbc data pull

Through testing, I found that there are many logs like the following：
2025-03-07 17:08:14,966 INFO  org.apache.flink.cdc.connectors.mysql.debezium.task.MySqlSnapshotSplitReadTask [] - Finished exporting **0 records** for split 'test_db.test_table:0'
2025-03-07 17:08:14,966 INFO  org.apache.flink.cdc.connectors.mysql.debezium.task.MySqlSnapshotSplitReadTask [] - Finished exporting **0 records** for split 'test_db.test_table:1'
2025-03-07 17:08:14,966 INFO  org.apache.flink.cdc.connectors.mysql.debezium.task.MySqlSnapshotSplitReadTask [] - Finished exporting **0 records** for split 'test_db.test_table:2'

mageniao2:
Does this function also apply to oracle.cdc ?

uicosp:
> Does this function also apply to oracle.cdc ?

no, this function only works for mysql connector.

mageniao2:
> > Does this function also apply to oracle.cdc ?
> 
> no, this function only works for mysql connector.

Thanks for the reply, does oracle connector have similar capabilities? I tried to configure `debezium.snapshot.select.statement.overrides` but it doesn't work.

uicosp:
> > > Does this function also apply to oracle.cdc ?
> > 
> > 
> > no, this function only works for mysql connector.
> 
> Thanks for the reply, does oracle connector have similar capabilities? I tried to configure `debezium.snapshot.select.statement.overrides` but it doesn't work.

I haven't used oracle, so I'm not sure. You can try to find the answer through the offical flink cdc docs or the oracle connector source code.

ThorneANN:
> debezium.snapshot.select.statement.overrides
`debezium.snapshot.select.statement.overrides` option only applicable to read data by lock algorithms，this is incremental snapshot for capturing data.
You can completely implement other incremental snapshot connectors methods by following the snapshot filtering of MySQL connector

mageniao2:
> > debezium.snapshot.select.statement.overrides
> > `debezium.snapshot.select.statement.overrides` option only applicable to read data by lock algorithms，this is incremental snapshot for capturing data.
> > You can completely implement other incremental snapshot connectors methods by following the snapshot filtering of MySQL connector

Thank you. I will follow your suggestion to attempt implementing a filtering feature based on an incremental snapshot connector

ThorneANN:
@yuxiqian @loserwang1024 PTAL

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/factory/MySqlDataSourceFactory.java
```diff
@@ -235,6 +236,27 @@ public DataSource createDataSource(Context context) {
             configFactory.chunkKeyColumn(chunkKeyColumnMap);
         }
 
+        String snapshotFilters = config.get(SCAN_SNAPSHOT_FILTERS);
+        if (snapshotFilters != null) {
+            Map<String, String> snapshotFilterMap = new HashMap<>();
+            for (String snapshotFilter : snapshotFilters.split(";")) {
+                String[] splits = snapshotFilter.trim().split(":");
```
lvyanquan, line 243: What if the filter in sql contains `;` or `:`, like `city != 'China:beijing'`?

uicosp: Sorry I’m late, I was busy with something else. I have modified the snapshotFilters parsing logic in MySqlDataSourceFactory to correctly handle semicolons and colons within filter conditions.

Key improvements:

1. parseSnapshotFilters method:
  - Intelligently parses semicolon delimiters while recognizing quoted content
  - Supports both single and double quotes
  - Semicolons within quotes are not treated as delimiters

2. parseTableAndCondition method:
  - Splits only on the first colon to avoid misinterpreting colons in condition values
  - Ensures proper separation of table names and conditions

yuxiqian: I think a better solution to this problem should allow passing nested YAML structures (LIST and MAP should be enough) as config option values instead of inventing more customized syntaxes, as they're hard to write and parse.

ThorneANN: > I think a better solution to this problem should allow passing nested YAML structures (LIST and MAP should be enough) as config option values instead of inventing more customized syntaxes, as they're hard to write and parse.

I agree with u , but i think that we should consider the difference with snapshot filters or cdc filters

uicosp: You are right. Actually, at the beginning I tried using nested YAML structures (LIST and MAP). However, I found that the YamlPipelineDefinitionParser.toSourceDef() method doesn’t support that; it only supports STRING values (because it uses Map<String, String> in Java).

I’m not sure whether this was intentional or simply a limitation of the current implementation. If I could use Map<String, Object>, I could switch to YAML structures, which is clearly more reasonable than parsing everything from strings.

uicosp:
This version adds support for configuring scan.snapshot.filters in the pipeline using a YAML list format.
Example:

```yaml
scan.snapshot.filters:
  - table: db1.user_table_[0-9]+
    filter: id > 100
  - table: db[1-2].[app|web]_order_\.* 
    filter: city != 'China:beijing'
```

Note:
The parameter type of `Configuration.fromMap` has been changed from `Map<String, String>` map to `Map<String, ?> map`.
@yuxiqian Please confirm that this change does not break existing design conventions.

yuxiqian (commented):
It appears that modifying config methods to support nested types should be safe, since all existing connectors will only declare `ConfigOption`s with primitive types.

Would @lvyanquan like to take a look?

ThorneANN:
@lvyanquan  Hi,Can you push this PR

nickdelnano:
This functionality would be really helpful. What is the status of this PR? Can it be merged?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/table/MySqlTableSource.java
```diff
@@ -241,6 +244,9 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
                             .parseOnLineSchemaChanges(parseOnlineSchemaChanges)
                             .useLegacyJsonFormat(useLegacyJsonFormat)
                             .assignUnboundedChunkFirst(assignUnboundedChunkFirst)
+                            .snapshotFilters(
+                                    escapeDot(database) + "." + escapeDot(tableName),
+                                    snapshotFilter)
                             .build();
```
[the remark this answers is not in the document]
uicosp: fixed https://github.com/apache/flink-cdc/pull/3776/changes/3d360c9053c734a2392d4c0f6f8e5e2304c28fe3

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/factory/MySqlDataSourceFactory.java
```diff
@@ -285,6 +288,19 @@ public DataSource createDataSource(Context context) {
...
+        if (snapshotFilters != null && !snapshotFilters.isEmpty()) {
+            Map<String, String> snapshotFiltersMap =
+                    snapshotFilters.stream()
+                            .collect(
+                                    Collectors.toMap(
+                                            it -> it.get(SNAPSHOT_FILTER_TABLE_KEY),
+                                            it -> it.get(SNAPSHOT_FILTER_FILTER_KEY)));
+            LOG.info("Add snapshotFilters {}.", snapshotFiltersMap);
+            configFactory.snapshotFilters(snapshotFiltersMap);
+        }
```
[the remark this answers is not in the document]
uicosp: fixed https://github.com/apache/flink-cdc/pull/3776/changes/67fae6df57871d47da78ffe9bdf555d6b4bcdb34

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/SnapshotFilterUtils.java
```diff
@@ -0,0 +1,71 @@
...
+        Map<Selectors, String> snapshotFilters = toSelector(filters);
+
+        String filter = null;
+        for (Selectors selector : snapshotFilters.keySet()) {
+            if (selector.isMatch(
+                    org.apache.flink.cdc.common.event.TableId.tableId(
+                            tableId.catalog(), tableId.table()))) {
+                filter = snapshotFilters.get(selector);
+                break;
+            }
```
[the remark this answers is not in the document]
uicosp: fixed https://github.com/apache/flink-cdc/pull/3776/changes/b7f13fdb42c3497b66f21480d4f0d03c076fbe90

uicosp:
@lvyanquan I've addressed the review comments and pushed the updates.
Could you please take another look when you have time? Thanks!

ThorneANN:
@leonardBang  PTAL,TKS.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceFactoryTest.java
```diff
@@ -330,6 +332,142 @@ void testAddChunkKeyColumns() {
                         });
     }
 
+    @Test
+    void testAddSnapshotFilters() {
```
ruanhang1993, line 337: We need more tests for this feature(IT, E2E test).

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/table/MySqlTableSourceFactory.java
```diff
@@ -206,6 +210,7 @@ public Set<ConfigOption<?>> optionalOptions() {
         options.add(MySqlSourceOptions.USE_LEGACY_JSON_FORMAT);
         options.add(MySqlSourceOptions.SCAN_INCREMENTAL_SNAPSHOT_UNBOUNDED_CHUNK_FIRST);
         options.add(MySqlSourceOptions.SCAN_READ_CHANGELOG_AS_APPEND_ONLY_ENABLED);
+        options.add(MySqlSourceOptions.SCAN_SNAPSHOT_FILTER);
```
ruanhang1993, line 213: I am considering whether we should implement `SupportsFilterPushDown` for the sql connector instead of a new config.

docs/content/docs/connectors/flink-sources/mysql-cdc.md
```diff
@@ -460,6 +460,13 @@ Only valid for cdc 1.x version. During a snapshot operation, the connector will
         For example updating an already updated value in snapshot, or deleting an already deleted entry in snapshot. These replayed change log events should be handled specially.
       </td>
     </tr>
+    <tr>
+      <td>scan.snapshot.filter</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">(none)</td>
+      <td>String</td>
+      <td>When reading a table snapshot, the rows of captured tables will be filtered using the specified filter expression (AKA a SQL WHERE clause). By default, no filter is applied, meaning the entire table will be synchronized. e.g. `id > 100`.</td> 
+    </tr>
```
ruanhang1993, line 477: It is better to add a new use example for this feature.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
```diff
@@ -242,12 +243,24 @@ private void createDataEventsForTable(
         long exportStart = clock.currentTimeInMillis();
         LOG.info("Exporting data from split '{}' of table {}", snapshotSplit.splitId(), table.id());
 
+        String filter =
+                SnapshotFilterUtils.getSnapshotFilter(
+                        sourceConfig.getSnapshotFilters(), table.id());
```
ruanhang1993: Could we store the filters as a class member variable instead of constructing it on every invocation?

ruanhang1993 (commented):
@uicosp Hi, the MySQL changes are significant, and it looks like there are still some issues that need further discussion. It may not be feasible to complete this in version 3.6. We can schedule this feature for the next release.

sd4324530:
Really looking forward to this functionality.

ThorneANN:
> @uicosp Hi, the MySQL changes are significant, and it looks like there are still some issues that need further discussion. It may not be feasible to complete this in version 3.6. We can schedule this feature for the next release.

Hi, I think that we do not to supple the `SupportsFilterPushDown` because of the Incremental snapshot framework.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceFactoryTest.java
uicosp: Added two tests in commit 85647b2:                               
- MySqlSourceITCase#testSnapshotFiltersMultipleTables — IT test 
  verifying per-table filters work independently across two tables   
  via the DataStream API.                      
- MysqlE2eITCase#testSnapshotFilters — E2E test verifying the    
  full pipeline path using the YAML-based scan.snapshot.filters      
  configuration.

docs/content/docs/connectors/flink-sources/mysql-cdc.md
uicosp: Added usage examples in commit cc477d5:
- docs/connectors/pipeline-connectors/mysql.md — Added "Snapshot
  Filters" section with a YAML pipeline example demonstrating the
  scan.snapshot.filters list syntax.
- docs/connectors/flink-sources/mysql-cdc.md — Added "Snapshot
  Filter" section with both SQL DDL and DataStream API examples
  showing the scan.snapshot.filter option usage.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlSnapshotSplitReadTask.java
uicosp: Optimized filter lookup in commit 9e8b5f5:
- Moved snapshot filter resolution from                          
  createDataEventsForTable() to the constructor of                   
  MySqlSnapshotSplitReadTask.
- The filter is now computed once and stored as a class member   
  variable (snapshotFilter), avoiding repeated lookups on every      
  method invocation.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/table/MySqlTableSourceFactory.java
uicosp: Thanks for the suggestion!
We actually discussed this with @ThorneANN earlier — the conclusion was that SupportsFilterPushDown is not suitable here because of the Incremental Snapshot framework's design.
The current config-based approach (scan.snapshot.filter) is intentional.

yuxiqian:
Due to inactivity, I'll removing this ticket from the 3.7 release plan. Feel free to keep working on this PR later.

PR #3780, opened 2024-12-05
[FLINK-36854] Fix comments and default expressions missing after being transformed

yuxiqian:
This closes FLINK-36854.

Consider the following simple schema:

`[id INT NOT NULL COMMENT 'id_field' DEFAULT 17]`

This would be sync to downstream sink databases that supports them intactly.

However, things would go wrong after being transformed like this:

```yaml
transform:
  - projection: id, id AS id_alias, id + 1 AS id_new
```

Here, `id_new` column should carry no comments / default expressions obviously, since it has nothing to do with original `id` technically.

However, for `id` and `id_alias` column, it's reasonable to expect that comments and default expressions to be kept. Worse still, the nullability of `id_alias` will be lost since the previous half is regarded as a complex expression, and nullability of an expression's return type is uncertain.

So, currently the transform will yields the following schema:

```
[
  id INT NOT NULL,
  id_alias INT,
  id_new INT
]
```

With this PR, deduced schema will be something like:

```
[
  id INT NOT NULL COMMENT 'id_field' DEFAULT 17,
  id_alias INT NOT NULL COMMENT 'id_field' DEFAULT 17,
  id_new INT
]
```

> Some extra parts have been changed in `TransformParser#generateProjectionColumns` since the previous version is a little cryptic.

### DIFF
--- a/flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposerITCase.java
+++ b/flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposerITCase.java
@@ -411,7 +411,7 @@ class FlinkPipelineComposerITCase {
         String[] outputEvents = outCaptor.toString().trim().split("\n");
         assertThat(outputEvents)
                 .containsExactly(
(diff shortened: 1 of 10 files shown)


yuxiqian:
Would @ruanhang1993 and @aiwenmo like to take a look?

aiwenmo:
> Would @ruanhang1993 and @aiwenmo like to take a look?

I will review it.

aiwenmo:
The supplement for renaming the data type of the original column is great.
For the moment, I can't see where the code needs to be optimized.

leonardBang (approved):
Thanks @yuxiqian @aiwenmo for the contribution, +1

PR #3785, opened 2024-12-09
[FLINK-36864] Fix unable to use numeric literals that goes beyond Int32 range

yuxiqian:
This closes FLINK-36864.

Flink SQL (and its underlying Calcite framework) allows integer literals within `Int64` range (`-9223372036854775808` ~ `9223372036854775807`), which is enough for most cases.

However, CDC YAML parser supports Int32 range (`-2147483648` ~ `2147483647`) only, because Java code only allows `Int32` literals without any suffix. This could be overridden by appending an extra "L" at the end of the literal.


aiwenmo:
This was overlooked before. Your fix is great.

leonardBang (approved):
Nice Catch, +1

PR #3787, opened 2024-12-10
[FLINK-36879][runtime] Support to convert delete as insert in transform

ruanhang1993:
Support to convert delete as insert in transform

### DIFF
--- a/flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml
+++ b/flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml
@@ -47,6 +47,7 @@ transform:
     partition-keys: product_name
     table-options: comment=app order
     description: project fields from source table
+    convert-delete-as-insert: true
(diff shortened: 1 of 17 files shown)


flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml
```diff
@@ -47,6 +47,7 @@ transform:
     partition-keys: product_name
     table-options: comment=app order
     description: project fields from source table
+    convert-delete-as-insert: true
```
yuxiqian, line 50: Docs in Core Concepts / Transform needs to be updated, too

yuxiqian (commented):
Thanks for @ruanhang1993's nice work, just left some trivial comments.

ruanhang1993:
@yuxiqian Thanks for the review. I will raise a new version for this PR and I have resolved these unrelated comments.

PR #3788, opened 2024-12-10
[FLINK-36877][pipeline-connector-kafka] Fix the output of canal-json for Kafka Sink when the record is deleted record.

lvyanquan:
An example of current output:
```
{
    "old": [
        {
            "id": 8,
            "name": "zhangsan3",
            "age": 30
        }
    ],
    "data": null,
    "type": "DELETE",
    "database": null,
    "table": "flink_penghai_test_yaml_student_copy2",
    "pkNames": [
        "id"
    ],
    "ts": 0
}
```

However, for canal-json, fields in delete record should be written to `data` field instead of `old` field.
Otherwise, we will get java.io.IOException: Corrupt Canal JSON message when reading with Kafka connector.


leonardBang (approved):
LGTM

lvyanquan:
The failure of test case is in postgres connector, and is not related with current pr.

PR #3794, opened 2024-12-11
FLINK-36891: MySQL CDC connector produces corrupted state in case of serialization failure

morozov:
Given that the output buffer is reused between calls to `serialize()`, it should be cleared in a `finally` block to account for potential exceptions.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializer.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializer.java
@@ -72,28 +72,32 @@ public class PendingSplitsStateSerializer implements SimpleVersionedSerializer<P
         }
         final DataOutputSerializer out = SERIALIZER_CACHE.get();
 
-        out.writeInt(splitSerializer.getVersion());
-        if (state instanceof SnapshotPendingSplitsState) {
-            out.writeInt(SNAPSHOT_PENDING_SPLITS_STATE_FLAG);
-            serializeSnapshotPendingSplitsState((SnapshotPendingSplitsState) state, out);
-        } else if (state instanceof BinlogPendingSplitsState) {
-            out.writeInt(BINLOG_PENDING_SPLITS_STATE_FLAG);
-            serializeBinlogPendingSplitsState((BinlogPendingSplitsState) state, out);
-        } else if (state instanceof HybridPendingSplitsState) {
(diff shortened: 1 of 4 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializer.java
```diff
@@ -72,28 +72,32 @@ public byte[] serialize(PendingSplitsState state) throws IOException {
...
+            } else {
+                throw new IOException(
+                        "Unsupported to serialize PendingSplitsState class: "
+                                + state.getClass().getName());
+            }
+
+            final byte[] result = out.getCopyOfBuffer();
+            // optimization: cache the serialized from, so we avoid the byte work during repeated
+            // serialization
+            state.serializedFormCache = result;
```
sharonx, line 95: I see the same chunk of code in the base folder https://github.com/apache/flink-cdc/blob/master/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/state/PendingSplitsStateSerializer.java#L83-L104. Should it be fixed too?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializerTest.java
```diff
@@ -102,6 +104,22 @@ public void testRepeatedSerializationCache() throws Exception {
...
+
+        final byte[] ser1 = serializer.serialize(state);
+        state.serializedFormCache = null;
+
+        PendingSplitsState unsupportedState = new UnsupportedPendingSplitsState();
+
+        assertThrows(IOException.class, () -> serializer.serialize(unsupportedState));
+
+        final byte[] ser2 = serializer.serialize(state);
+        assertEquals(ser1.length, ser2.length);
```
sharonx, line 120: Is it better to assert that `ser1` and `ser2` are the same, since there shouldn't be any changes to the state if an error happens based on my understanding.

morozov: For some reason, they aren't.

sharonx (approved):
LGTM

leonardBang (approved):
Thanks @morozov for the contribution and @sharonx for the review work, LGTM

PR #3795, opened 2024-12-12
[FLINK-35257][cdc-runtime]when schema.change.behavior is `exception`,…

yanghuaiGit:
When schema.change.behavior is exception, the task should throw an exception failure after all downstream data is flushed.

After ddl data is received, the task fails quickly. The data between checkPoint and ddl is lost


yanghuaiGit:
@yuxiqian PTAL, Thanks

yuxiqian:
Hi @yanghuaiGit, thanks for the contribution! Postponing exception throwing is reasonable, as it will avoid job going into unexpected state.

Would you like to check if this problem still persists in latest `master` branch?

PR #3796, opened 2024-12-12
[FLINK-36895] The JdbcSourceChunkSplitter#queryMin method passed a parameter with tableName/coiumnName reversed.

Mrart:
JdbcChunkUtils#queryMin()
final String minQuery =
String.format(
"SELECT MIN(%s) FROM %s WHERE %s > ?",
quotedColumnName, quotedTableName, quotedColumnName);
 
but JdbcSourceChunkSplitter #queryMin
 
protected Object queryMin(
JdbcConnection jdbc, TableId tableId, Column splitColumn, Object excludedLowerBound)
throws SQLException

{ return JdbcChunkUtils.queryMin( jdbc, jdbc.quotedColumnIdString(splitColumn.name()), jdbc.quotedTableIdString(tableId), excludedLowerBound); }
 
Error passing column/tableId parameter. Resulting in the generated sql "select tableId from columnID".

we need sql "select columnid from tableId"

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,25 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="IssueNavigationConfiguration">
-    <option name="links">
-      <list>
-        <IssueNavigationLink>
-          <option name="issueRegexp" value="FLINK\-\d+" />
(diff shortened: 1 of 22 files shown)


Mrart:
Please help me close. It error pr.

PR #3712, opened 2024-11-11
[tests][oceanbase] Fix oceanbase test failure, possibly caused by some interactions among cases

yuxiqian:
This fixes a mysterious test failure in OceanBaseMySQLModeITCase like [this](https://github.com/apache/flink-cdc/actions/runs/11720914709/job/32647267237). Recently it is happening quite frequently, possibly brought in by #3439.

This could be stably reproduced locally, by running `OceanBaseMySqlModeITCase` locally, and make sure `testMetadataColumns` is executed right after `testSnapshotOnly`.

This test case will fail, because instead of getting expected

```
+I(test,inventory,products,101,scooter,Small 2-wheel scooter,3.1400000000), 
+I(test,inventory,products,102,car battery,12V car battery,8.1000000000), 
+I(test,inventory,products,103,12-pack drill bits,12-pack of drill bits with sizes ranging from #40 to #3,0.8000000000), 
+I(test,inventory,products,104,hammer,12oz carpenter's hammer,0.7500000000), 
+I(test,inventory,products,105,hammer,14oz carpenter's hammer,0.8750000000), 
+I(test,inventory,products,106,hammer,16oz carpenter's hammer,1.0000000000), 
+I(test,inventory,products,107,rocks,box of assorted rocks,5.3000000000), 
+I(test,inventory,products,108,jacket,water resistent black wind breaker,0.1000000000), 
+I(test,inventory,products,109,spare tire,24 inch spare tire,22.2000000000), 
+U(test,inventory,products,106,hammer,18oz carpenter hammer,1.0000000000)
```

values, it is actually receiving:

```
+I(test,inventory,products,101,scooter,Small 2-wheel scooter,3.1400000000),
+I(test,inventory,products,102,car battery,12V car battery,8.1000000000), 
+I(test,inventory,products,103,12-pack drill bits,12-pack of drill bits with sizes ranging from #40 to #3,0.8000000000), 
+I(test,inventory,products,104,hammer,12oz carpenter's hammer,0.7500000000), 
+I(test,inventory,products,105,hammer,14oz carpenter's hammer,0.8750000000), 
+I(test,inventory,products,106,hammer,16oz carpenter's hammer,1.0000000000), 
+I(test,inventory,products,107,rocks,box of assorted rocks,5.3000000000), 
+I(test,inventory,products,108,jacket,water resistent black wind breaker,0.1000000000), 
+I(test,inventory,products,109,spare tire,24 inch spare tire,22.2000000000), 
[...]

PR #3802, opened 2024-12-13
[FLINK-36701][cdc-runtime] Add steps to get and emit schemaManager's latest evolvedSchema when SinkDataWriterOperator handles FlushEvent

Jzjsnow:
Currently, directly after a failover, when the pipeline first handles a schema change event (e.g. AddColumnEvent) and then a DataChangeEvent, it may cause the job to fail again as sink has repeatedly applied that schema change.
I add steps to get and emit schemaManager's latest evolvedSchema when SinkDataWriterOperator handles FlushEvent. If the operator doesn't have a local cache of the schema when handling the flushevent, it will request Schema Manager's latest evolved schema. At this point the evolved schema is the same with the sink table's, as the schema change event hasn't been applied to the evolved schema by the master yet.


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataSinkWriterOperator.java
yuxiqian, line 1: CDC Pipeline provides wrapper for both modern `sink2` and legacy `SinkFunction`s. Is `DataSinkFunctionOperator` prone to this issue, too?

Jzjsnow: Yes it is, thanks for the heads up, I have fixed it in the new commit for DataSinkFunctionOperator.

yuxiqian:
Thanks for @Jzjsnow's clear and detailed diagram!

Just noticed that we still lack failover recovering tests for pipeline jobs. Would you like to add some based on `MySqlSourceITCase#testMySqlParallelSource` and trigger TM/JM Failures among different phases?

Jzjsnow:
> Thanks for @Jzjsnow's clear and detailed diagram!
> 
> Just noticed that we still lack failover tests for pipeline jobs. Could we add some recovering tests like `MySqlSourceITCase#testMySqlParallelSource` and trigger TM/JM Failures in different phases? `flink-cdc-composer` unit tests might be a place for it.

@yuxiqian Thanks for the suggestion, I have added some tests to the `flink-cdc-runtime` unit test for testing DataSinkOperator's handling of schema change events, which was lacking. Among them, `DataSinkOperatorWithSchemaEvolveTest#testSchemaChangeEventAfterFailover` is used to test the underlying schema change process after failure recovery, which is applicable to the scenario of restarting after jm/tm failure.

Jzjsnow:
Here it looks like we have once again encountered a `OceanBaseMySQLModeITCase` test failure similar to the one in #3712. Any ideas on how to fix this?

yuxiqian:
It's worrying to notice similar OceanBase test case failures here, but it seems irrelevant to this PR. Will investigate this.

yuxiqian:
@Jzjsnow Should be fixed now... please rebase to master branch.

Jzjsnow:
@yuxiqian Thanks for the quick fix, now we've rebased to the master branch.

Jzjsnow:
The `Source E2E Tests#OceanBaseE2eITCase` test did not pass in this round of CIs, but it is not relevant to this PR. It seems that the 1 minute timeout set in the method `checkResultWithTimeout` (timeout=60000L)  is a bit too short, so that the row `id=111` in the sink table that should have been deleted has not been synchronized yet.

Here is the error log:
> 2025-01-03T07:38:53.8553280Z 1254259 [main] ERROR org.apache.flink.cdc.connectors.tests.OceanBaseE2eITCase - 
2025-01-03T07:38:53.8553917Z --------------------------------------------------------------------------------
2025-01-03T07:38:53.8554680Z Test testOceanBaseCDC[flinkVersion: 1.19.1](org.apache.flink.cdc.connectors.tests.OceanBaseE2eITCase) failed with:
2025-01-03T07:38:53.8555684Z array lengths differed, expected.length=10 actual.length=11; arrays first differed at element [10]; expected:<end of array> but was:<111,scooter,Big 2-wheel scooter ,5.18,null,null>
2025-01-03T07:38:53.8556858Z 	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:89)
2025-01-03T07:38:53.8557558Z 	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:28)
2025-01-03T07:38:53.8558017Z 	at org.junit.Assert.internalArrayEquals(Assert.java:534)
2025-01-03T07:38:53.8558492Z 	at org.junit.Assert.assertArrayEquals(Assert.java:285)
2025-01-03T07:38:53.8558895Z 	at org.junit.Assert.assertArrayEquals(Assert.java:300)
2025-01-03T07:38:53.8559341Z 	at org.apache.flink.cdc.common.test.utils.JdbcProxy.checkResult(JdbcProxy.java:70)
2025-01-03T07:38:53.8559996Z 	at org.apache.flink.cdc.common.test.utils.JdbcProxy.checkResultWithTimeout(JdbcProxy.java:93)
2025-01-03T07:38:53.8560819Z 	at org.apache.flink.cdc.connectors.tests.OceanBaseE2eITCase.testOceanBaseCDC(OceanBaseE2eITCase.java:179)

yuxiqian (commented):
Thanks for @Jzjsnow's great work, left some minor comments.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/event/FlushEvent.java
```diff
@@ -17,25 +17,49 @@
...
  */
 public class FlushEvent implements Event {
+    /** The schema changes from which table. */
+    private final List<TableId> tableIds;
 
     /** Which subTask ID this FlushEvent was initiated from. */
     private final int sourceSubTaskId;
 
-    public FlushEvent(int sourceSubTaskId) {
+    /** Flag indicating whether the FlushEvent is sent before a create table event. */
```
yuxiqian: Please modify JavaDocs, too

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/event/FlushEvent.java
```diff
@@ -17,25 +17,49 @@
...
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 
 /**
  * An {@link Event} from {@code SchemaOperator} to notify {@code DataSinkWriterOperator} that it
  * start flushing.
  */
 public class FlushEvent implements Event {
+    /** The schema changes from which table. */
```
yuxiqian: If I understand correctly, this `TableId` set is the "affected" tables that need to be flushed, not the "original table that initiates the schema change".

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/event/FlushEvent.java
```diff
@@ -17,25 +17,49 @@
...
+    private final List<TableId> tableIds;
 
     /** Which subTask ID this FlushEvent was initiated from. */
     private final int sourceSubTaskId;
 
-    public FlushEvent(int sourceSubTaskId) {
+    /** Flag indicating whether the FlushEvent is sent before a create table event. */
+    private final SchemaChangeEventType schemaChangeEventType;
+
+    public FlushEvent(int sourceSubTaskId, SchemaChangeEventType schemaChangeEventType) {
```
yuxiqian, line 38: Seems this constructor should be `@VisibleForTesting`. But I'm not quite sure why `BucketWrapperFlushEvent` is also calling this, since it might lose track of the `tableIds` field if sink declares an extra bucket assigning topology.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/bucket/BucketWrapperEventSerializer.java
```diff
@@ -82,14 +84,19 @@ public void serialize(Event event, DataOutputView dataOutputView) throws IOExcep
             BucketWrapperFlushEvent bucketWrapperFlushEvent = (BucketWrapperFlushEvent) event;
             dataOutputView.writeInt(bucketWrapperFlushEvent.getBucket());
             dataOutputView.writeInt(bucketWrapperFlushEvent.getSourceSubTaskId());
+            schemaChangeEventTypeEnumSerializer.serialize(
+                    bucketWrapperFlushEvent.getSchemaChangeEventType(), dataOutputView);
```
yuxiqian, line 88: Similar concern here, should we keep original `tableIds` field when wrapping FlushEvents?

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/serializer/event/EventSerializerTest.java
```diff
@@ -44,7 +45,12 @@ protected Class<Event> getTypeClass() {
 
     @Override
     protected Event[] getTestData() {
-        Event[] flushEvents = new Event[] {new FlushEvent(1), new FlushEvent(2), new FlushEvent(3)};
+        Event[] flushEvents =
+                new Event[] {
+                    new FlushEvent(1, SchemaChangeEventType.CREATE_TABLE),
+                    new FlushEvent(2, SchemaChangeEventType.CREATE_TABLE),
+                    new FlushEvent(3, SchemaChangeEventType.CREATE_TABLE)
+                };
```
yuxiqian, line 53: Please also test if FlushEvents with extra TableIds could be correctly [de]serialized.

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/serializer/event/PartitioningEventSerializerTest.java
```diff
@@ -48,7 +49,12 @@ protected Class<PartitioningEvent> getTypeClass() {
 
     @Override
     protected PartitioningEvent[] getTestData() {
-        Event[] flushEvents = new Event[] {new FlushEvent(1), new FlushEvent(2), new FlushEvent(3)};
+        Event[] flushEvents =
+                new Event[] {
+                    new FlushEvent(1, SchemaChangeEventType.CREATE_TABLE),
+                    new FlushEvent(2, SchemaChangeEventType.CREATE_TABLE),
+                    new FlushEvent(3, SchemaChangeEventType.CREATE_TABLE)
+                };
```
yuxiqian, line 57: Ditto

yuxiqian (commented):
Thanks for @Jzjsnow's quick response! Just left some trivial comments, please take a look when available.

yuxiqian (approved):
Looks good. Would @leonardBang like to take a look?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/event/FlushEvent.java
Jzjsnow: Thanks for the careful review. In order for the `BucketWrapperFlushEvent` to retain the full `FlushEvent`'s information, the sink table IDs and schema change event type are passed in its constructor. So this constructor for `FlushEvent` is no longer necessary and is removed.

leonardBang (approved):
Thanks @Jzjsnow for the improvement and @yuxiqian for the review work, the change looks good to me.

PR #3810, opened 2024-12-20
[FLINK-36580][pipeline-connector-mysql] Fix column.[in|ex]clude.list parameter causing schema inconsistency

MOBIN-F:
When users need to ignore some privacy fields or large fields from the collection end, column.include.list/column.exclude.list may be used. At this time, the pipeline may fail due to schema inconsistency.

```
source:
  type: mysql
  ....
  debezium.column.exclude.list: test_db.test_table.big_col

transform:
  - source-table: test_db.test_table
  projection: \*
  description: project fields and filter 
```

error will be reported:
1. ArrayIndexOutOfBoundsException
```
Caused by: java.lang.ArrayIndexOutOfBoundsException: 33370909
    at org.apache.flink.cdc.common.data.binary.BinarySegmentUtils.getLongMultiSegments(BinarySegmentUtils.java:731)
    at org.apache.flink.cdc.common.data.binary.BinarySegmentUtils.getLong(BinarySegmentUtils.java:721) 
```
2. columns are not aligned properly

PR main changes
1. Filter columns when getting table schema
2. Filter columns in snapshot phase
3. Filter columns event in binlog phase

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceFactoryTest.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceFactoryTest.java
@@ -36,6 +36,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static io.debezium.relational.RelationalDatabaseConnectorConfig.COLUMN_EXCLUDE_LIST;
+import static io.debezium.relational.RelationalDatabaseConnectorConfig.COLUMN_INCLUDE_LIST;
+import static io.debezium.relational.RelationalDatabaseConnectorConfig.TABLE_INCLUDE_LIST_ALREADY_SPECIFIED_ERROR_MSG;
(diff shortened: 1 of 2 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-debezium/src/main/java/org/apache/flink/cdc/debezium/table/DebeziumOptions.java
```diff
@@ -43,6 +47,10 @@ public static Properties getDebeziumProperties(Map<String, String> properties) {
             throw new IllegalArgumentException(
                     "table.include.list and table.exclude.list are not supported to set manually, please remove these options.");
         }
+        if (debeziumProperties.containsKey(COLUMN_INCLUDE_LIST.name())
+                && debeziumProperties.containsKey(COLUMN_EXCLUDE_LIST.name())) {
+            throw new IllegalArgumentException(TABLE_INCLUDE_LIST_ALREADY_SPECIFIED_ERROR_MSG);
+        }
```
MOBIN-F, line 53: COLUMN_INCLUDE_LIST and COLUMN_EXCLUDE_LIST cannot coexist https://github.com/debezium/debezium/blob/ce9ddb119eaebef4a78882146f20a3d3ebd7c30f/debezium-core/src/main/java/io/debezium/relational/RelationalDatabaseConnectorConfig.java#L743

PR #3812, opened 2024-12-23
[FLINK-36931][cdc] FlinkCDC YAML supports batch mode

aiwenmo:
# Premise

## MysqlCDC supports snapshot mode

MysqlCDC in Flink CDC (MySqlSource) supports **StartupMode.SNAPSHOT** and is of **Boundedness.BOUNDED**, and can run in **RuntimeExecutionMode.BATCH**.

## Streaming VS Batch

Stream mode is suitable for job types including: jobs with high real-time requirements; in non-real-time scenarios, **stateless jobs** with many Shuffle steps; jobs that require continuous and stable data processing capabilities; jobs with **small states**, **simple topologies** and low fault tolerance costs.

Batch mode is suitable for job types including: in non-real-time scenarios, jobs with **a large number of stateful operators**; jobs that require **high resource utilization**; jobs with **large states**, **complex topologies** and low fault tolerance costs.

# Expectation

## Full snapshot synchronization

The FlinkCDC YAML job **only reads the full snapshot data** of the database and then writes it to the target database in Streaming or Batch mode. It is mainly used for full catch-up.

Currently, the SNAPSHOT startup strategy of the FlinkCDC YAML job can run correctly in the Streaming mode; it cannot run correctly in the Batch mode.

## Full-incremental offline

The FlinkCDC YAML job collects full snapshot data + incremental log data from the final Offset of the full-incremental snapshot algorithm to the current EndingOffset for the first run; for subsequent runs, it collects from the last EndingOffset to the current EndingOffset.

The job runs in Batch mode. Users can **schedule the job periodically**, tolerate data delays for a certain period of time (such as hourly or daily), and ensure eventual consistency. Since the periodically scheduled incremental job only collects logs between the last EndingOffset and the current EndingOffset, duplicate full collection of data is avoided.

# Test

## Full snapshot synchronization in Batch mode

1. After removing the PartitionOperator, all operators will be chained into one PipelinedRegion and can run correctly;
2. When there are multiple PipelinedRegions, only the first PipelinedRegion is in the jobgraph and it cannot run correctly;
3. After removing the SchemaOperator, when there are multiple PipelinedRegions, a correct jobgraph can also be generated, but the sink requires the registration of the coordinator operator.

# Solution

## Use StartupMode.SNAPSHOT + Streaming for full snapshot synchronization

There is no need to modify the source code. For MysqlCDC, after specifying StartupMode.SNAPSHOT, the full snapshot synchronization job of the entire database can be run in the streaming mode. Although it is not the optimal solution, this capability can be achieved currently.

## Expand the FlinkPipelineComposer applicable to the Batch mode to support full Batch synchronization

Topology graph: Source -> PreTransform -> PostTransform -> Router -> PartitionBy -> Sink

There are no change events in the Batch mode, and Schema Evolution does not need to be considered. In addition, the automatic table creation is completed before the job starts.
The field derivation of transform can be placed before the job starts instead of during runtime. Other operations such as the derivation of Router can also be placed before the job starts.
Workload: Implement the Batch construction strategy of FlinkPipelineComposer. Router needs to be independent, and Sink needs to be extended or transformed to support the implementation that does not require a coordinator (it would be better if Batch writing can be achieved).

## Expand StartupMode to support users specifying the Offset range to support incremental offline synchronization

Allow users to specify the collection Offset range of the binlog, and then the user's own platform records the EndingOffset of each execution, as well as the periodic scheduling by the platform.

# Discussion

1.Is it necessary to implement support for Batch mode because the benefits brought by Batch are small or the performance is not as good as Streaming. Specifically, which Batch optimizations can be used?

2.Whether the full-incremental offline method should be implemented (users can periodically schedule incremental log synchronization)?

# Code implementation
Topology graph: Source -> BatchPreTransform -> PostTransform -> SchemaBatchOperator -> PartitionBy(Batch) -> BatchSink
ps: The data flow only contains CreateTableEvent and DataChangeEvent (insert).
# Implementation ideas
image

1.Source first sends all CreateTableEvents, then sends snapshot data.
2.BatchPreTransform doesn't need to cache the state and resume, and PostTransform is no changes in other cases.
3.When SchemaBatchOperator receives the CreateTableEvent, it is only stored in the cache and no events are sent.
4.When SchemaBatchOperator receives the first DataChangeEvent, the widest downstream table structure is deduced based on the router rule, and then the table creation statement is executed in the external data source. Subsequently, the wide table structure is sent to BatchPrePartition.
5.BatchPrePartition broadcasts the CreateTableEvent to BatchPostPartition. BatchPrePartition partitions and distributes the DataChangeEvent to PostPartition based on the table ID and primary key information.
6.BatchPostPartition issues the CreateTableEvent and DataChangeEvent to BatchSink, and BatchSink performs batch writing.
# Implementation effect
Computing node 1: Source -> BatchPreTransform -> PostTransform -> SchemaBatchOperator -> BatchPrePartition
Computing node 2: BatchPostPartition -> BatchSink
Batch mode: Computing node 2 starts computing only after computing node 1 is completely finished.


aiwenmo:
# Code implementation

Topology graph: Source -> PreTransform -> PostTransform -> SchemaBatchOperator-> PartitionBy(Batch) -> BatchSink

1. add SchemaBatchOperator which removed the processing of  schema change event and removed Coordinator.
2. add RegularPrePartitionBatchOperator which removed SchemaEvolutionClient.
3. add DataBatchSinkFunctionOperator and DataBatchSinkWriterOperator which removed SchemaEvolutionClient.
4. remove SchemaRegistry in batch mode.

aiwenmo:
5. DataSource will send CreateTableCompletedEvent after sending all CreateTableEvent.
6. add CreateTableCompletedEvent to notify SchemaBatchOperator to merge all CreateTableEvent.

aiwenmo:
During the test, a new bug was discovered and has been fixed. This PR relies on this fix. #3826

lvyanquan (commented):
Thanks @aiwenmo for this contribution, left some comments.

And a end-to-end test is also welcomed.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PreBatchTransformOperator.java
```diff
@@ -0,0 +1,389 @@
...
+        } else if (event instanceof DropTableEvent) {
+            preTransformProcessorMap.remove(((DropTableEvent) event).tableId());
+            output.collect(new StreamRecord<>(event));
+        } else if (event instanceof TruncateTableEvent) {
+            output.collect(new StreamRecord<>(event));
+        } else if (event instanceof SchemaChangeEvent) {
+            SchemaChangeEvent schemaChangeEvent = (SchemaChangeEvent) event;
+            preTransformProcessorMap.remove(schemaChangeEvent.tableId());
+            cacheChangeSchema(schemaChangeEvent)
+                    .ifPresent(e -> output.collect(new StreamRecord<>(e)));
```
lvyanquan, line 211: No DropTableEvent, TruncateTableEvent and other SchemaChangeEvents will be sent, so we can only deal with CreateTableEvent and DataChangeEvent.

lvyanquan:
I think an e2e test that run in batch mode with transform module is necessary to verify the whole pipeline is runnable.

aiwenmo:
Hi. I'm in the process of coding and testing.

flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/MySqlToDorisE2eITCase.java
```diff
@@ -336,6 +336,92 @@ public void testSyncWholeDatabase() throws Exception {
...
+                String.format(
+                        "source:\n"
+                                + "  type: mysql\n"
+                                + "  hostname: mysql\n"
+                                + "  port: 3306\n"
+                                + "  username: %s\n"
+                                + "  password: %s\n"
+                                + "  tables: %s.\\.*\n"
+                                + "  server-id: 5400-5404\n"
+                                + "  server-time-zone: UTC\n"
```
lvyanquan, line 352: Do we need to add `scan.startup.mode: snapshot` in MySQL source config?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
```diff
@@ -89,7 +90,16 @@ public MySqlPipelineRecordEmitter(
...
+        if (StartupOptions.snapshot().equals(sourceConfig.getStartupOptions())) {
+            // In snapshot mode, we simply emit all schemas at once.
+            createTableEventCache.forEach(
+                    createTableEvent -> {
+                        output.collect(createTableEvent);
+                        TableId tableId = TableId.parse(createTableEvent.tableId().identifier());
+                        alreadySendCreateTableTables.add(tableId);
+                    });
+            alreadySendCreateTableForBinlogSplit = true;
+        } else if (isLowWatermarkEvent(element) && splitState.isSnapshotSplitState()) {
```
lvyanquan, line 102: You may need to rebase to solve the conflict as there is another change in MySqlPipelineRecordEmitter.java that introduced in https://github.com/apache/flink-cdc/commit/eee0cc06f7bf842034d0e22bc45f8ffb811f7f78#diff-7da556d0b9e83dcd71a4a2319c7588c813a5c14ac0dfc180fd096ae7b2a4a5b0.

flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/MySqlToDorisE2eITCase.java
aiwenmo: It is currently needed. Since batch mode is only applicable to snapshots, should it be optimized so that snapshot is the default option if not specified?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
aiwenmo: Hi. I'm rebasing and testing.

flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/MySqlToDorisE2eITCase.java
lvyanquan: It will be great if we reduce this config, and we can extract pipeline config from Context when using `createDataSource(Context context)` method.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlSource.java
```diff
@@ -200,7 +201,22 @@ public SplitEnumerator<MySqlSplit, PendingSplitsState> createEnumerator(
         validator.validate();
 
         final MySqlSplitAssigner splitAssigner;
-        if (!sourceConfig.getStartupOptions().isStreamOnly()) {
+        // In batch mode, only split snapshots.
```
lvyanquan, line 204: This change will also modify the behavior of Stream mode, should we update the comment to `In snapshot-moly mode`?

lvyanquan, line 204: This change will also modify the behavior of Stream mode, should we update the comment to `In snapshot-only startup option`?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
```diff
@@ -90,7 +91,16 @@ public MySqlPipelineRecordEmitter(
...
             SourceRecord element, SourceOutput<Event> output, MySqlSplitState splitState)
             throws Exception {
-        if (isLowWatermarkEvent(element) && splitState.isSnapshotSplitState()) {
+        if (StartupOptions.snapshot().equals(sourceConfig.getStartupOptions())) {
+            // In snapshot mode, we simply emit all schemas at once.
+            if (alreadySendCreateTableTables.isEmpty()) {
+                createTableEventCache.forEach(
+                        (tableId, createTableEvent) -> {
+                            output.collect(createTableEvent);
+                            alreadySendCreateTableTables.add(tableId);
```
lvyanquan, line 100: alreadySendCreateTableTables is useless in snapshot-only mode.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaDerivator.java
```diff
@@ -341,4 +355,57 @@ public Optional<DataChangeEvent> coerceDataRecord(
...
+            }
+            Schema mergedSchema = null;
+            for (Schema toBeMergedSchema : toBeMergedSchemas) {
+                if (mergedSchema == null) {
+                    mergedSchema = toBeMergedSchema;
+                    continue;
+                }
+                mergedSchema =
+                        SchemaMergingUtils.getLeastCommonSchema(mergedSchema, toBeMergedSchema);
+            }
```
yuxiqian, line 393: call `SchemaUtils#getCommonSchema`

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/TableIdRouter.java
```diff
@@ -76,7 +77,7 @@ public List<TableId> route(TableId sourceTableId) {
         return routingCache.getUnchecked(sourceTableId);
     }
 
-    private List<TableId> calculateRoute(TableId sourceTableId) {
+    public List<TableId> calculateRoute(TableId sourceTableId) {
```
yuxiqian, line 80: Don't modify this, it's an internal method. Call `TableIdRouter#route` externally so the results could be cached.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/TableIdRouter.java
```diff
@@ -95,4 +96,23 @@ private TableId resolveReplacement(
         }
         return TableId.parse(route.f1);
     }
+
+    public List<Set<TableId>> groupSourceTablesByRouteRule(Set<TableId> tableIdSet) {
```
yuxiqian, line 100: I doubt if it's really a generic and reusable method, without corresponding JavaDocs and test cases. Maybe just write it as a `for` loop in `SchemaDerivator`?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataSinkWriterOperatorFactory.java
```diff
@@ -36,17 +36,37 @@ public class DataSinkWriterOperatorFactory<CommT>
...
 
     private final Sink<Event> sink;
+    private final boolean isBatchMode;
     private final OperatorID schemaOperatorID;
 
     public DataSinkWriterOperatorFactory(Sink<Event> sink, OperatorID schemaOperatorID) {
         this.sink = sink;
+        this.isBatchMode = false;
+        this.schemaOperatorID = schemaOperatorID;
+    }
```
yuxiqian, line 46: Better remove this constructor and specify streaming / batch mode explicitly

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataSinkWriterOperatorFactory.java
```diff
@@ -59,6 +79,9 @@ public <T extends StreamOperator<CommittableMessage<CommT>>> T createStreamOpera
 
     @Override
     public Class<? extends StreamOperator> getStreamOperatorClass(ClassLoader classLoader) {
+        if (isBatchMode) {
+            return DataBatchSinkWriterOperator.class;
+        }
         return DataSinkWriterOperator.class;
```
yuxiqian, line 85: nit: wrap streaming case in `else { ... }`

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/translator/PartitioningTranslator.java
```diff
@@ -46,6 +47,34 @@ public DataStream<Event> translateRegular(
...
+                            "BatchPrePartition",
+                            new PartitioningEventTypeInfo(),
+                            new RegularPrePartitionBatchOperator(
+                                    downstreamParallelism, hashFunctionProvider))
+                    .setParallelism(upstreamParallelism)
+                    .partitionCustom(new EventPartitioner(), new PartitioningEventKeySelector())
+                    .map(new PostPartitionProcessor(), new EventTypeInfo())
+                    .name("BatchPostPartition")
+                    .setParallelism(downstreamParallelism);
+        }
```
yuxiqian, line 77: ditto: prefer

```java
if (isBatchMode) {
    // batch code
} else {
    // streaming code
}
```

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PreBatchTransformOperator.java
```diff
@@ -0,0 +1,337 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+/**
+ * A data process function that filters out columns which aren't (directly & indirectly) referenced
+ * in batch mode. There is no need to consider the task failover scenario in batch mode, as it
+ * should be re - emitted from the source.
+ */
+public class PreBatchTransformOperator extends AbstractStreamOperator<Event>
```
yuxiqian, line 57: "PreTransform" is a phrase. Call it `BatchPreTransformOperator`?

yuxiqian (changes requested):
Thanks for @aiwenmo's work, just left some comments.

My major concern is I've seen a lot of copy-and-paste codes from streaming mode, which makes maintaining code base much harder in the future. I would suggest extracting common parts into an abstract parent class (for example, put common partitioning logic in `AbstractPrePartitionOperator`) and extend it in `StreamingPrePartitionOperator` and `BatchPrePartitionOperator`.

Ignore that if we don't have enough time to finish it before 3.4.0.

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/translator/DataSinkTranslator.java
```diff
@@ -158,8 +176,7 @@ private <CommT> void addCommittingTopology(
                     ((WithPreCommitTopology<Event, CommT>) sink).addPreCommitTopology(written);
         }
 
-        // TODO: Hard coding stream mode and checkpoint
-        boolean isBatchMode = false;
+        // TODO: Hard coding checkpoint
         boolean isCheckpointingEnabled = true;
```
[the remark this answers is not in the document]
aiwenmo: Some sinks need to rely on checkpointing to complete the writing.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaDerivator.java
```diff
@@ -134,6 +135,19 @@ public static Set<Schema> reverseLookupDependingUpstreamSchemas(
...
+            final TableId evolvedTableId,
+            final Set<TableId> allOriginalTables,
+            final Schema originalSchema) {
+        return reverseLookupDependingUpstreamTables(
+                        tableIdRouter, evolvedTableId, allOriginalTables)
+                .stream()
+                .map(utid -> originalSchema)
+                .collect(Collectors.toSet());
+    }
+
```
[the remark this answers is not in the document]
aiwenmo: This code is useless and should be deleted.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaDerivator.java
```diff
@@ -341,4 +355,57 @@ public Optional<DataChangeEvent> coerceDataRecord(
 
         return Optional.of(dataChangeEvent);
     }
+
+    /** Deduce merged CreateTableEvent in batch mode. */
+    public static List<CreateTableEvent> deduceMergedCreateTableEventInBatchMode(
```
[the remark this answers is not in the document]
aiwenmo: At present, it only merges CreateTableEvent in batch mode.

aiwenmo:
PTAL @leonardBang @lvyanquan @yuxiqian

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/TableIdRouter.java
lvyanquan: @aiwenmo  could you add some feedback about this comment？

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/TableIdRouter.java
lvyanquan: Ditto.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/TableIdRouter.java
aiwenmo: I've already replaced `TableIdRouter#calculateRoute` with `TableIdRouter#route`.I forgot to remove the change here.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/TableIdRouter.java
aiwenmo: It makes use of "routes". In the latest code, I've added documentation for it.
Now that I think about it, it can also be placed in SchemaDerivator. I'll give it a try tonight.

aiwenmo: Sorry. The attempt couldn't be achieved. The current way of writing is more optimal.

yuxiqian:
Hi @aiwenmo, could you please rebase this PR with latest `master` when available?

* JUnit 4 style test annotations should be changed to JUnit 5 equivalents

  * `org.junit.Assert`
  * `org.hamcrest.*`

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/factory/MySqlDataSourceFactory.java
```diff
@@ -129,6 +130,13 @@ public DataSource createDataSource(Context context) {
         ZoneId serverTimeZone = getServerTimeZone(config);
         StartupOptions startupOptions = getStartupOptions(config);
 
+        // Batch mode only supports StartupMode.SNAPSHOT.
+        Configuration pipelineConfiguration = context.getPipelineConfiguration();
+        if (pipelineConfiguration != null
+                && pipelineConfiguration.contains(PipelineOptions.PIPELINE_BATCH_MODE_ENABLED)) {
+            startupOptions = StartupOptions.snapshot();
+        }
```
yuxiqian, line 138: Just a suggestion, what about throwing an exception explicitly if one enables batch mode with a non-snapshotting source? That would prevent some silent behavior change of `startupOptions` config.

Alternatively, we may add a interface in `DataSourceFactory` like this to make things clearer:

```java
@PublicEvolving
public interface DataSourceFactory extends Factory {

    /** Creates a {@link DataSource} instance. */
    DataSource createDataSource(Context context);

    /** Checking if this {@link DataSource} could be created in batch mode. */
    boolean supportsBatchPipeline(Context context);
}
```

and verifies it during translating pipeline job graph. WDYT?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
```diff
@@ -90,7 +94,16 @@ public MySqlPipelineRecordEmitter(
     protected void processElement(
             SourceRecord element, SourceOutput<Event> output, MySqlSplitState splitState)
             throws Exception {
-        if (isLowWatermarkEvent(element) && splitState.isSnapshotSplitState()) {
+        if (StartupOptions.snapshot().equals(sourceConfig.getStartupOptions())) {
+            // In snapshot mode, we simply emit all schemas at once.
+            if (!alreadySendAllCreateTableTables) {
```
yuxiqian, line 99: In snapshot mode, we will:
  * obtain and check startup mode, and check the flag
In any other modes, we will:
  * obtain and check startup mode

I would suggest naming `alreadySendAllCreateTableTables` => `shouldEmitAllCtesInSnapshotMode`, and set it to `true` in snapshot mode, `false` (in other modes), and the checking could be simplified to:

```java
if (shouldEmitAllCtesInSnapshotMode) {
    createTableEventCache.forEach(
        (tableId, createTableEvent) -> output.collect(createTableEvent)
    );
    shouldEmitAllCtesInSnapshotMode = false;
}
```

so we don't have to check the startup mode every time when we receive a SourceRecord.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
```diff
@@ -90,7 +94,16 @@ public MySqlPipelineRecordEmitter(
...
+                createTableEventCache.forEach(
+                        (tableId, createTableEvent) -> {
+                            output.collect(createTableEvent);
+                            alreadySendAllCreateTableTables = true;
```
yuxiqian, line 103: nit: move `alreadySendAllCreateTableTables = true` out of the loop

flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/TransformE2eITCase.java
yuxiqian, line 1: I think TransformE2e and UdfE2e has nothing to do with batch mode. Shall we refactor original cases with `@ParameterizedTest`, or just remove these cases to avoid code inflation and longer CI execution time?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaDerivator.java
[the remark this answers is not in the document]
yuxiqian: Yes, but `SchemaDerivator` itself shouldn't be aware of streaming / batch execution mode. Similar initial deducing logic could be ported to streaming mode later.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaBatchOperator.java
yuxiqian, line 1: minor: use consistent names for new operator classes: either `BatchXXXOperator` (like `BatchPreTransformOp`) or `XXXBatchOperator` (`SchemaBatchOperator`). 

Personally I prefer the former one since it is easy to distinguish it from normal Streaming operators.

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaDerivatorTest.java
```diff
@@ -599,4 +600,113 @@ void testNormalizeSchemaChangeEventsInIgnoreMode() {
                                 new DropTableEvent(NORMALIZE_TEST_TABLE_ID)))
                 .isEmpty();
     }
+
+    @Test
+    void testDeduceMergedCreateTableEventInBatchMode() {
```
yuxiqian, line 605: ditto, remove `inBatchMode`

yuxiqian (changes requested):
Thanks for @aiwenmo's quick response.

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/translator/PartitioningTranslator.java
```diff
@@ -46,15 +47,44 @@ public DataStream<Event> translateRegular(
...
+        } else {
+            return input.transform(
+                            "PrePartition",
+                            new PartitioningEventTypeInfo(),
+                            new RegularPrePartitionOperator(
+                                    schemaOperatorID, downstreamParallelism, hashFunctionProvider))
+                    .setParallelism(upstreamParallelism)
+                    .partitionCustom(new EventPartitioner(), new PartitioningEventKeySelector())
+                    .map(new PostPartitionProcessor(), new EventTypeInfo())
+                    .name("PostPartition");
```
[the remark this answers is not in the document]
aiwenmo: No modification is needed after discussion.

yuxiqian (approved):
Thanks for @aiwenmo's great work!

lvyanquan:
Hi, @aiwenmo. Overall, it looks good to me and we can verify its reliability through the rich testing you have added.

However, perhaps we still need some log information to assist in analysis in case of abnormal situations, so you can refer to https://github.com/apache/flink-cdc/pull/3960/files and add some logs in BatchPreTransformOperator//DataBatchSinkWriterOperator/DataBatchSinkFunctionOperator to help better troubleshooting.

lvyanquan:
Hi @leonardBang, PTAL.

docs/content.zh/docs/core-concept/data-pipeline.md
```diff
@@ -115,4 +115,5 @@ under the License.
 |-----------------|---------------------------------------|-------------------|
 | name            | 这个 pipeline 的名称，会用在 Flink 集群中作为作业的名称。 | optional          |
 | parallelism     | pipeline的全局并发度，默认值是1。                 | optional          |
-| local-time-zone | 作业级别的本地时区。                            | optional          |
\ No newline at end of file
+| local-time-zone | 作业级别的本地时区。                            | optional          |
+| batch-mode.enabled | 仅使用批处理模式来同步当前快照数据。                    | optional          |
```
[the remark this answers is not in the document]
aiwenmo: Hi. @ruanhang1993 .The previous writing was consistent with what you described, but @lvyanquan  suggested changing it to this form. I think it's necessary to communicate and confirm.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/pipeline/PipelineOptions.java
```diff
@@ -45,6 +45,12 @@ public class PipelineOptions {
                     .defaultValue(1)
                     .withDescription("Parallelism of the pipeline");
 
+    public static final ConfigOption<Boolean> PIPELINE_BATCH_MODE_ENABLED =
+            ConfigOptions.key("batch-mode.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription("Run pipeline job in batch mode instead of streaming mode");
```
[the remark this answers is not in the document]
aiwenmo: ditto.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaManager.java
```diff
@@ -73,6 +73,12 @@ public SchemaManager() {
         behavior = SchemaChangeBehavior.EVOLVE;
```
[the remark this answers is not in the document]
aiwenmo: I will update the code tonight.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaManager.java
```diff
@@ -73,6 +73,12 @@ public SchemaManager() {
...
     }
 
+    public SchemaManager(SchemaChangeBehavior schemaChangeBehavior) {
+        evolvedSchemas = new ConcurrentHashMap<>();
+        originalSchemas = new ConcurrentHashMap<>();
+        behavior = schemaChangeBehavior;
```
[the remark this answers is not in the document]
aiwenmo: I will update the code tonight.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/TableIdRouter.java
```diff
@@ -95,4 +96,35 @@ private TableId resolveReplacement(
...
+     *
+     * @param: tableIdSet
+     * @return: java.util.List<java.util.Set<org.apache.flink.cdc.common.event.TableId>>
+     *
+     * Group the source tables that conform to the same routing rule together.
+     * The total number of groups is less than or equal to the number of routing rules.
+     * For the source tables within each group, their table structures will be merged to obtain
+     * the widest table structure in that group. The structures of all tables within the group
+     * will be expanded to this widest table structure.
+     */
```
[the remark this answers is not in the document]
aiwenmo: ditto.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataBatchSinkFunctionOperator.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * An operator that processes records to be written into a {@link SinkFunction} in batch mode.
+ *
+ * <p>The operator is a proxy of {@link StreamSink} in Flink.
+ *
+ * <p>The operator is always part of a sink pipeline and is the first operator.
+ */
+@Internal
+public class DataBatchSinkFunctionOperator extends StreamSink<Event> {
```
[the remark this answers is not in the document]
aiwenmo: ditto.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataBatchSinkWriterOperator.java
```diff
@@ -0,0 +1,207 @@
...
+ * An operator that processes records to be written into a {@link Sink} in batch mode.
+ *
+ * <p>The operator is a proxy of SinkWriterOperator in Flink.
+ *
+ * <p>The operator is always part of a sink pipeline and is the first operator.
+ *
+ * @param <CommT> the type of the committable (to send to downstream operators)
+ */
+@Internal
+public class DataBatchSinkWriterOperator<CommT>
```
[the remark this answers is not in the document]
aiwenmo: ditto.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/partitioning/RegularPrePartitionBatchOperator.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Operator for processing events from {@link SchemaOperator} before {@link EventPartitioner} with
+ * regular topology in batch mode.
+ */
+@Internal
+public class RegularPrePartitionBatchOperator extends AbstractStreamOperator<PartitioningEvent>
```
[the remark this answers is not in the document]
aiwenmo: ditto.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/factories/DataSourceFactory.java
```diff
@@ -28,4 +28,10 @@ public interface DataSourceFactory extends Factory {
 
     /** Creates a {@link DataSource} instance. */
     DataSource createDataSource(Context context);
+
+    /** Checking if this {@link DataSource} could be created in batch mode. */
+    default void verifyBatchMode(Context context) {
```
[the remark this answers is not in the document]
yuxiqian: +1 for @ruanhang1993's suggestion.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataBatchSinkFunctionOperator.java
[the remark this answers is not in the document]
yuxiqian: Are you referring to this comment I left earlier?

> minor: use consistent names for new operator classes: either BatchXXXOperator (like BatchPreTransformOp) or XXXBatchOperator (SchemaBatchOperator).

> Personally I prefer the former one since it is easy to distinguish it from normal Streaming operators.

I think it's essentially the same as what Hang just suggested: favor `BatchDataSinkFunctionOperator` instead of `DataBatchSinkFunctionOperator`.

aiwenmo:
Hi. @ruanhang1993 All checks have passed in my repository. Please run the failed work again. thx

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/pipeline/PipelineOptions.java
```diff
@@ -45,6 +45,12 @@ public class PipelineOptions {
                     .defaultValue(1)
                     .withDescription("Parallelism of the pipeline");
 
+    public static final ConfigOption<RuntimeMode> PIPELINE_RUNTIME_MODE =
```
leonardBang, line 48: Flink has `execution.runtime-mode` configuration, and I assume the two configuration has same semantics, could we follow Flink's design[1] ?
[1] https://github.com/apache/flink/blob/7c03bc834b642ff40828739961a9a4100b9fb073/flink-core-api/src/main/java/org/apache/flink/api/common/RuntimeExecutionMode.java#L39C5-L39C14

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/factories/DataSourceFactory.java
```diff
@@ -28,4 +28,7 @@ public interface DataSourceFactory extends Factory {
 
     /** Creates a {@link DataSource} instance. */
     DataSource createDataSource(Context context);
+
+    /** Checking if this {@link DataSource} could be created in batch mode. */
+    default void verifyRuntimeMode(Context context) {}
```
leonardBang, line 33: This is not intuitive to me, why do we need to introduce a method to verify runtime mode(especially the java doc said batch mode) in Factory?

flink-cdc-cli/src/test/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParserTest.java
```diff
@@ -366,6 +366,7 @@ private void testSchemaEvolutionTypesParsing(
                                     .put("parallelism", "4")
                                     .put("schema.change.behavior", "evolve")
                                     .put("schema-operator.rpc-timeout", "1 h")
+                                    .put("runtime-mode", "STREAMING")
```
leonardBang, line 369: The change is unnecessary if the configuration has a default value IIUC?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
```diff
@@ -71,6 +72,9 @@ public class MySqlPipelineRecordEmitter extends MySqlRecordEmitter<Event> {
     // Used when startup mode is initial
     private Set<TableId> alreadySendCreateTableTables;
 
+    // Used when startup mode is snapshot
+    private boolean shouldEmitAllCtesInSnapshotMode = true;
```
leonardBang, line 76: `Ctes` means?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
```diff
@@ -90,7 +94,15 @@ public MySqlPipelineRecordEmitter(
     protected void processElement(
             SourceRecord element, SourceOutput<Event> output, MySqlSplitState splitState)
             throws Exception {
-        if (isLowWatermarkEvent(element) && splitState.isSnapshotSplitState()) {
+        if (shouldEmitAllCtesInSnapshotMode
+                && StartupOptions.snapshot().equals(sourceConfig.getStartupOptions())) {
```
leonardBang, line 98: We can calculate `Boolean isBounded = StartupOptions.snapshot().equals(sourceConfig.getStartupOptions()` in constructor instead of calculate it per element.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/factory/MySqlDataSourceFactory.java
```diff
@@ -274,6 +276,22 @@ public DataSource createDataSource(Context context) {
...
+        StartupOptions startupOptions = getStartupOptions(config);
+        // Batch mode only supports StartupMode.SNAPSHOT.
+        Configuration pipelineConfiguration = context.getPipelineConfiguration();
+        if (pipelineConfiguration != null
+                && pipelineConfiguration.contains(PipelineOptions.PIPELINE_RUNTIME_MODE)
+                && RuntimeMode.BATCH.equals(
+                        pipelineConfiguration.get(PipelineOptions.PIPELINE_RUNTIME_MODE))
+                && !StartupOptions.snapshot().equals(startupOptions)) {
+            throw new IllegalArgumentException(
+                    "Batch mode is only supported for MySQL source in snapshot mode.");
```
leonardBang, line 291:
```suggestion
                    "Only "snapshot" of MySQLDataSource StartupOption is supported in BATCH pipeline, but actual MySQLDataSource StartupOption is {}.");
```

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/sink/DataSinkWriterOperatorFactory.java
```diff
@@ -36,17 +36,31 @@ public class DataSinkWriterOperatorFactory<CommT>
...
+        this.isBatchMode = isBatchMode;
         this.schemaOperatorID = schemaOperatorID;
     }
 
     @SuppressWarnings("unchecked")
     @Override
     public <T extends StreamOperator<CommittableMessage<CommT>>> T createStreamOperator(
             StreamOperatorParameters<CommittableMessage<CommT>> parameters) {
+
+        if (isBatchMode) {
```
leonardBang, line 54: just a personal flavor,I think use `isBounded` in code is more aligned with flink's stream concept(unbounded stream and bounded stream )

leonardBang (commented):
Thanks @aiwenmo for the great work, the PR keeps a good status, I left some comments

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/factories/DataSourceFactory.java
aiwenmo: Hi. @leonardBang The purpose of adding this method is to remind developers to handle streaming and batch processing when developing new data sources. It's also feasible to remove the definition of this method and let developers implement it themselves in the createDataSource method. So, should we remove it now? I'm glad to hear your opinion.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/pipeline/PipelineOptions.java
aiwenmo: Regarding this issue, there have been numerous exchanges of views and modifications with other reviewers before. Now, I want to know the following information more clearly:
1.Should we change to the same parameter definition as that of Flink, with our parameter having a lower priority than Flink's own parameters?
2.Is it feasible to use "snapshot" along with Flink's own parameters for control? In this way, when users enable the batch mode, they need to explicitly specify Flink's `execution.runtime-mode` parameter.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
aiwenmo: `Ctes` means CreateTableEvents. How do you think it should be modified to make it clearer?

leonardBang: Clear is better than long in my opinion

leonardBang (approved):
Thanks @aiwenmo for the great work and all for the review, LGTM

PR #3821, opened 2024-12-27
[FLINK-36970][runtime] Merge result of data type BIGINT and DOUBLE should be DOUBLE instead of STRING

Shawn-Hx:
In SchemaMergingUtils#getLeastCommonType, the merge result of BIGINT and DOUBLE is STRING now.

However, considering JSON format, JSON numbers can be integers or floating point, which can be inferred as BIGINT and  DOUBLE respectively. In this case, a JSON number field will be inferred as STRING, which is confusing.

The updated data type merge tree is:

```mermaid
graph TD;
    BOOLEAN --> STRING;
    
    CHAR --> STRING;
    VARCHAR --> STRING;
    BINARY --> STRING;
    VARBINARY --> STRING;

    TINYINT --> FLOAT;
    SMALLINT --> FLOAT;
    FLOAT --> DOUBLE;
    INT --> DOUBLE;
    
    TINYINT --> SMALLINT --> INT --> BIGINT --> DOUBLE --> STRING;
    BIGINT --> DECIMAL;
    DECIMAL --> STRING;

    TIMESTAMP --> TIMESTAMP_LTZ --> TIMESTAMP_TZ --> STRING;

    TIME --> STRING;

    ROW -->  STRING;
    ARRAY --> STRING;
    MAP --> STRING;
```

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
@@ -810,7 +810,9 @@ public class SchemaMergingUtils {
         mergingTree.put(DoubleType.class, ImmutableList.of(doubleType, stringType));
         mergingTree.put(FloatType.class, ImmutableList.of(floatType, doubleType, stringType));
         mergingTree.put(DecimalType.class, ImmutableList.of(stringType));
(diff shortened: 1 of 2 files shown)


yuxiqian (approved):
This change looks reasonable. Could @Shawn-Hx please rebase this and @leonardBang trigger the CI?

leonardBang (approved):
LGTM

PR #3826, opened 2024-12-30
[FLINK-36981][cdc] Fix the bug of transform merging with route in YAML task

aiwenmo:
# Situation
In FlinkPipelineComposerITCase.testTransformMergingWithRoute:
```
// Create test dataset:
// Create table 1 [id, name, age]
// Table 1: +I[1, Alice, 18]
// Table 1: +I[2, Bob, 20]
// Table 1: -U[2, Bob, 20] +U[2, Bob, 30]
// Create table 2 [id, name, age, description]
// Table 2: +I[3, Charlie, 15, student]
// Table 2: +I[4, Donald, 25, student]
// Table 2: -D[4, Donald, 25, student]
// Rename column for table 1: name -> last_name
// Add column for table 2: gender
// Table 1: +I[5, Eliza, 24]
// Table 2: +I[6, Frank, 30, student, male] 
```
This is executed correctly.

```
// Create test dataset:
// Create table 1 [id, name, age]
// Create table 2 [id, name, age, description]
// Table 1: +I[1, Alice, 18]
// Table 1: +I[2, Bob, 20]
// Table 1: -U[2, Bob, 20] +U[2, Bob, 30]
// Table 2: +I[3, Charlie, 15, student]
// Table 2: +I[4, Donald, 25, student]
// Table 2: -D[4, Donald, 25, student]
// Rename column for table 1: name -> last_name
// Add column for table 2: gender
// Table 1: +I[5, Eliza, 24]
// Table 2: +I[6, Frank, 30, student, male] 
```
This will cause an exception.
```
Caused by: org.codehaus.commons.compiler.CompileException: Line 1, Column 76: Assignment conversion not possible from type "java.lang.Integer" to type "java.lang.Long" 
```

# Cause
When a conversion rule of transform matches multiple tables, only tables with the same structure are supported. However, in the scenario of sharding databases and tables, it is possible to match multiple tables with different structures.  
In the `Map<Tuple2<TableId, TransformProjection>, TransformProjectionProcessor>
            transformProjectionProcessorMap` ,`TransformProjection` is wrongly shared, resulting in `Schema` always having the schema of the latest matched table.

# Fix
1.Change `Map<Tuple2<TableId, TransformProjection>, TransformProjectionProcessor>
            transformProjectionProcessorMap` into `Map<Tuple2<TableId, String>, TransformProjectionProcessor>
            transformProjectionProcessorMap`.
2.When transformProjectionProcessor has none TableInfo, set TransformProjection into new transformProjectionProcessor from existed transformProjectionProcessor.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
@@ -81,7 +81,7 @@ public class PostTransformOperator extends AbstractStreamOperator<Event>
     private List<UserDefinedFunctionDescriptor> udfDescriptors;
     private transient Map<String, Object> udfFunctionInstances;
 
-    private transient Map<Tuple2<TableId, TransformProjection>, TransformProjectionProcessor>
+    private transient Map<Tuple2<TableId, String>, TransformProjectionProcessor>
             transformProjectionProcessorMap;
     private transient Map<Tuple2<TableId, TransformFilter>, TransformFilterProcessor>
             transformFilterProcessorMap;
@@ -355,12 +355,14 @@ public class PostTransformOperator extends AbstractStreamOperator<Event>
         for (PostTransformer transform : transforms) {
             Selectors selectors = transform.getSelectors();
             if (selectors.isMatch(tableId) && transform.getProjection().isPresent()) {
-                TransformProjection transformProjection = transform.getProjection().get();
+                TransformProjection transformProjection =
+                        TransformProjection.of(transform.getProjection().get().getProjection())
+                                .get();
                 if (transformProjection.isValid()) {
                     if (!transformProjectionProcessorMap.containsKey(
-                            Tuple2.of(tableId, transformProjection))) {
+                            Tuple2.of(tableId, transformProjection.getProjection()))) {
                         transformProjectionProcessorMap.put(
-                                Tuple2.of(tableId, transformProjection),
+                                Tuple2.of(tableId, transformProjection.getProjection()),
                                 TransformProjectionProcessor.of(
                                         transformProjection,
                                         timezone,
@@ -370,7 +372,7 @@ public class PostTransformOperator extends AbstractStreamOperator<Event>
                     }
                     TransformProjectionProcessor postTransformProcessor =
                             transformProjectionProcessorMap.get(
-                                    Tuple2.of(tableId, transformProjection));
+                                    Tuple2.of(tableId, transformProjection.getProjection()));
                     // update the columns of projection and add the column of projection into Schema
                     newSchemas.add(
                             postTransformProcessor.processSchemaChangeEvent(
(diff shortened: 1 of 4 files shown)


aiwenmo:
PTAL @yuxiqian @lvyanquan

yuxiqian:
`1.Change Map<Tuple2<TableId, TransformProjection>, TransformProjectionProcessor> transformProjectionProcessorMap into Map<Tuple2<TableId, String>, TransformProjectionProcessor> transformProjectionProcessorMap.`

Seems you're changing the key of this map from "[TableId, TransformProjection]` to `[Transform, TransformProjection#projection]`. However the problem is Projection string is a field of TransformProjection. If `TransformProjection` has collision, how can the `ProjectionString` be OK?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
```diff
@@ -434,23 +436,36 @@ private Optional<DataChangeEvent> processDataChangeEvent(DataChangeEvent dataCha
...
                                 TransformProjectionProcessor.of(
                                         tableInfo,
                                         transformProjection,
                                         timezone,
                                         udfDescriptors,
                                         getUdfFunctionInstances(),
                                         transform.getSupportedMetadataColumns()));
+                    } else if (!transformProjectionProcessorMap
+                            .get(Tuple2.of(tableId, transformProjection.getProjection()))
+                            .hasTableInfo()) {
```
yuxiqian, line 451: `hasTableInfo` could be renamed to something like `hasChangeInfo` or `hasPostTransformChangeInfo`

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
```diff
@@ -434,23 +436,36 @@ private Optional<DataChangeEvent> processDataChangeEvent(DataChangeEvent dataCha
...
+                        TransformProjectionProcessor transformProjectionProcessorWithoutTableInfo =
+                                transformProjectionProcessorMap.get(
+                                        Tuple2.of(tableId, transformProjection.getProjection()));
+                        transformProjectionProcessorMap.put(
+                                Tuple2.of(tableId, transformProjection.getProjection()),
+                                TransformProjectionProcessor.of(
```
yuxiqian, line 457: We may have a `copy` method in `TransformProjectionProcessor` to allow us overwriting specific field but keep others intact.

Otherwise, a lot of codes must be modified if we add one more field into it.

yuxiqian (changes requested):

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
ChaomingZhangCN: +1 Use TransformProjectionProcessor#copy() to create a non-shared object seems better.

yuxiqian (approved):
As discussed with @aiwenmo, we'll keep this PR minimal, focusing on bug-fix and leave transform module cleaning-up work later.

Could @leonardBang please trigger the CI?

leonardBang (approved):
Thanks @aiwenmo for the contribution and @yuxiqian for the review, LGTM

PR #3828, opened 2025-01-01
[FLINK-36985]remove Catalog.ColumnAlreadyExistException when apply applyAddColumnEventWithPosition in paimon

hiliuxg:
remove Catalog.ColumnAlreadyExistException when apply applyAddColumnEventWithPosition in paimon. because use may create table already before with the target column.


lvyanquan (approved):
LGTM.

lvyanquan:
@leonardBang  @ruanhang1993  could you please help to trigger the CI?

leonardBang (approved):
+1

PR #3829, opened 2025-01-02
[FLINK-36805][cdc-common] Add interface to support encryption of sensitive configuration items and provide base64 encoding implementation

Jzjsnow:
When Flink CDC uses pipeline to submit jobs, we need to set configuration items in plaintext in the definition file, including sensitive configurations such as passwords for connecting to source and sink components (e.g., mysql, doris, etc.), which may be a security risk. To avoid the use of plaintext passwords, we provide an interface (ConfigShade) by implementing which developers can customize the decryption method themselves. 

We also provide an implementation for base64 encoding first, not only as an example implementation of the interface, but also to solve the current problem of plaintext passwords.


flink-cdc-common/src/main/java/org/apache/flink/cdc/common/shade/ConfigShade.java
```diff
@@ -0,0 +1,54 @@
...
+     * ConfigShade}.
+     */
+    String getIdentifier();
+
+    /**
+     * Decrypt the content.
+     *
+     * @param content The content to decrypt
+     */
+    String decrypt(String content);
```
ChaomingZhangCN, line 53: Maybe we should use `String decrypt(String key, String content)`.

joyCurry30 (commented):
Nice contribution. Could you please add a test case where the configuration value contains sensitive words?

Jzjsnow:
> Nice contribution. Could you please add a test case where the configuration value contains sensitive words?

@joyCurry30 Ok, I've added a test to verify that yaml can be parsed and decrypted correctly when the configuration value contains sensitive keywords. PTAL

PR #3830, opened 2025-01-02
[FLINK-36958][pipeline-connector][kafka] Support csv/json/debezium-json/canal-json format options

MOBIN-F:
cdc-pipeline-connector-kafka supports key.format=csv/json, value.format=canal-json/debezium-json formats. There are format options in flink-formats for users to choose to use. cdc-pipeline should also support these format options

format options will also be used in the pipeline-connector-kafka-source


MOBIN-F:
@lvyanquan  I refer to the source code of the flink-json module and introduce format options. Can you review the code when you have time? thanks~

yuxiqian:
Thanks for Mobin's great work! Rebasing & migrating JUnit assertions might be necessary, too :)

MOBIN-F:
This PR overlaps with https://github.com/apache/flink-cdc/pull/3926. The difference is this PR verifies value.format and format options. When the verification fails, the error and prompt are more like flnk-sql-connector-kafka.

maybe https://github.com/apache/flink-cdc/pull/3926 is enough

lvyanquan:
After offline discuss with @MOBIN-F, we can skip this PR and trace in https://github.com/apache/flink-cdc/pull/3926

PR #3840, opened 2025-01-08
[hotfix] Remove duplicated code to improve performance

ChaomingZhangCN:
Already have LoadingCache.


ChaomingZhangCN:
@yuxiqian Please take a look~

yuxiqian (approved):
Not sure why I wrote that 😅

Also, it would be nice if we can add `@CheckReturnValue` annotation to these pure functions in `SchemaUtils`. Thus, such useless code will surely cause a `Result of 'SchemaUtils.createFieldGetters()' is ignored` warning.

yuxiqian (approved):
Thanks for @ChaomingZhangCN's quick fix! 

ping @leonardBang for more review

leonardBang (approved):
Thanks @ChaomingZhangCN for the improvement, LGTM

PR #3843, opened 2025-01-08
[Docs] Update mysql-to-doris.md, add scan.startup.mode config to make demo run more smoothly

maixiaohai:
When not set scan.startup.mode to earliest-offset, will use intial mode，the next part about incrementing data will not work. That's not very friendly to new starters.


leonardBang:
Thanks for your contribution @maixiaohai ! But the default initial mode will read snapshot data and then read incremental data(i.e. binlog data) automatically,  Could you explain more why incremental reading will not work?

PR #3844, opened 2025-01-08
[FLINK-36794] [cdc-composer/cli] pipeline cdc connector support multiple data sources

linjianchang:
pipeline cdc connector support multiple data sources
The relevant design and verification documents are as follows:
https://www.yuque.com/oniucium/lfk1te/ocp6m13kpgh1x9pg?singleDoc#

### DIFF
--- a/docs/content.zh/docs/connectors/pipeline-connectors/mysql.md
+++ b/docs/content.zh/docs/connectors/pipeline-connectors/mysql.md
@@ -50,9 +50,9 @@ MySQL CDC Pipeline 连接器允许从 MySQL 数据库读取快照数据和增量
 </table>
 </div>
 
-## 示例
+## 单数据源示例
 
-从 MySQL 读取数据同步到 Doris 的 Pipeline 可以定义如下：
+单数据源，从单个 MySQL 读取数据同步到 Doris 的 Pipeline 可以定义如下：
 
 ```yaml
 source:
@@ -74,6 +74,47 @@ sink:
 
 pipeline:
    name: MySQL to Doris Pipeline
+   parallelism: 4
+```
+
+## 多数据源示例
+
+多数据源，从多个mysql数据源读取数据同步到 Doris 的 Pipeline 可以定义如下：
+
+```yaml
+sources:
+   - type: mysql
+     name: MySQL multiple Source1
+     hostname: 127.0.0.1
+     port: 3306
+     username: admin
+     password: pass
+     tables: adb.\.*, bdb.user_table_[0-9]+, [app|web].order_\.*
+     server-id: 5400-5404
+     server-time-zone: Asia/Shanghai
+     source.unique.id: 1
+
+   - type: mysql
+       name: MySQL multiple Source2
+     hostname: 127.0.0.2
+     port: 3307
+     username: admin
+     password: pass
+     tables: adb.\.*, bdb.user_table_[0-9]+, [app|web].order_\.*
+     server-id: 5405-5409
+     server-time-zone: Asia/Shanghai
+     source.unique.id: 2
+
+sink:
+  type: doris
+  name: Doris Sink
+  fenodes: 127.0.0.1:8030
+  username: root
+  password: pass
+
+pipeline:
+   name: MySQL to Doris Pipeline
+   operator.uid.prefix: multiple-source
    parallelism: 4
 ```
 
--- a/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
+++ b/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
@@ -62,6 +62,7 @@ public class YamlPipelineDefinitionParser implements PipelineDefinitionParser {
 
     // Parent node keys
     private static final String SOURCE_KEY = "source";
+    private static final String MULTIPLE_SOURCE_KEY = "sources";
     private static final String SINK_KEY = "sink";
     private static final String ROUTE_KEY = "route";
     private static final String TRANSFORM_KEY = "transform";
@@ -70,6 +71,7 @@ public class YamlPipelineDefinitionParser implements PipelineDefinitionParser {
 
     // Source / sink keys
     private static final String TYPE_KEY = "type";
+    private static final String SOURCES = "sources";
     private static final String NAME_KEY = "name";
     private static final String INCLUDE_SCHEMA_EVOLUTION_TYPES = "include.schema.changes";
     private static final String EXCLUDE_SCHEMA_EVOLUTION_TYPES = "exclude.schema.changes";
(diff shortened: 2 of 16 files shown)


flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
```diff
@@ -126,16 +127,28 @@ private void translate(StreamExecutionEnvironment env, PipelineDef pipelineDef)
...
+            dataSource = sourceTranslator.createDataSource(sourceDef, pipelineDefConfig, env);
+            DataStream<Event> streamBranch =
+                    sourceTranslator.translate(sourceDef, dataSource, env, parallelism);
+            if (stream == null) {
+                stream = streamBranch;
+            } else {
+                stream = stream.union(streamBranch);
+            }
+        }
+        boolean isParallelMetadataSource = dataSource.isParallelMetadataSource();
```
ChaomingZhangCN, line 148: I think multi data sources should be regarded as parallelized.

docs/content.zh/docs/connectors/pipeline-connectors/mysql.md
```diff
@@ -77,6 +77,32 @@ pipeline:
    parallelism: 4
 ```
 
+## 多数据源示例
+
+单数据源，从多个 MySQL 读取数据同步到 Doris 的 Pipeline 可以定义如下：
+
+```yaml
+source:
+   type: mysql_mutiple
```
ChaomingZhangCN, line 86: Should we use a new key like 'sources' to describe multiple sources? The '_multiple' suffix in value seems a bit odd. Because the YAML content does not correspond one-to-one with the PipelineDef.

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
```diff
@@ -97,6 +97,13 @@ public class YamlPipelineDefinitionParser implements PipelineDefinitionParser {
 
     public static final String TRANSFORM_TABLE_OPTION_KEY = "table-options";
 
+    private static final String HOST_LIST = "host_list";
+    private static final String COMMA = ",";
+    private static final String HOST_NAME = "hostname";
+    private static final String PORT = "port";
+    private static final String COLON = ":";
+    private static final String MUTIPLE = "_mutiple";
```
ChaomingZhangCN, line 105: Should be `_multiple`.

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
linjianchang: Already modified

docs/content.zh/docs/connectors/pipeline-connectors/mysql.md
linjianchang: Already modified

ChaomingZhangCN:
```yaml
sources:
 - type: mysql
    name: mysql-instance-00
    hostname: localhost
    port: 3306
    ....
 - type: mysql
    name: mysql-instance-01
    hostname: localhost
    port: 3307
    ....
```
And the corresponding PipelineDef looks like this:
```java
public class PipelineDef {
    @Nullable private List<SourceDef> sources;
    private final SourceDef source;
    ...
}
```
If the `sources` is not null then we use these data sources, otherwise we use `source` to build up DataStream. In this way, the previous usage will not be affected.
I want to hear your opinion. @yuxiqian

yuxiqian: I do like @ChaomingZhangCN's proposed syntax for a fully multiple data source, they're intuitive and expressive, but might be a chore if users just want to connect to a MySQL cluster with multiple servers, as they have to copy all identical configurations to both source definition.

@linjianchang's solution for now seems like MySQL specific, especially for multi-host clusters. It could not be extended for hetero-sources (like concatenating data from different DBMS), or when one wants to use different configs for each node. These cases don't exist for now since all we have is MySQL source connector, but as we're modifying composer and YAML API (instead of MySQL connector itself), such possibility should be discussed more carefully.

As for multiple sources in pipeline itself, I remembered the idea has been informally discussed with @leonardBang and @PatrickRen long time ago, and the conclusion was running multiple sources in one single job actually makes the pipeline more fragile, since any single-point failure would easily escalate and cause a global failover. Things might have changed since then, still needs hearing from senior developers on this.

linjianchang: > ```yaml
> sources:
>  - type: mysql
>     name: mysql-instance-00
>     hostname: localhost
>     port: 3306
>     ....
>  - type: mysql
>     name: mysql-instance-01
>     hostname: localhost
>     port: 3307
>     ....
> ```
> 
> And the corresponding PipelineDef looks like this:
> 
> ```java
> public class PipelineDef {
>     @Nullable private List<SourceDef> sources;
>     private final SourceDef source;
>     ...
> }
> ```
> 
> If the `sources` is not null then we use these data sources, otherwise we use `source` to build up DataStream. In this way, the previous usage will not be affected. I want to hear your opinion. @yuxiqian @ChaomingZhangCN 

It has been modified according to comments, please  review it again, thanks!

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
linjianchang: @ChaomingZhangCN Hi,please help review when you have time,thanks!

docs/content.zh/docs/connectors/pipeline-connectors/mysql.md
```diff
@@ -77,6 +77,44 @@ pipeline:
...
+     port: 3306
+     username: admin
+     password: pass
+     tables: adb.\.*, bdb.user_table_[0-9]+, [app|web].order_\.*
+     server-id: 5400-5404
+     server-time-zone: Asia/Shanghai
+
+   - type: mysql
+       name: MySQL multiple Source2
+     hostname: 127.0.0.2
```
lvyanquan, line 99: It is essential to verify state compatibility.
Would adding or removing a Source from an existing job lead to state compatibility issues, or could reordering the Sources result in state inconsistency? Such incompatible operations must be exposed as errors to avoid silent handling.

linjianchang: > It is essential to verify state compatibility. Would adding or removing a Source from an existing job lead to state compatibility issues, or could reordering the Sources result in state inconsistency? Such incompatible operations must be exposed as errors to avoid silent handling.
@lvyanquan Thanks for the review！I have already made the modifications and tested them. Please refer to the following document
https://www.yuque.com/oniucium/lfk1te/yruqcmf9vbz4rxzv?singleDoc# 《multiple source从状态恢复验证》

PR #3849, opened 2025-01-10
[FLINK-35325][transform] Skip insufficient_quota error when running test case using ad model.

lvyanquan:
SKip  ci that were unstable.


flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineUdfITCase.java
```diff
@@ -897,7 +899,6 @@ void testTransformWithModel(ValuesDataSink.SinkApi sinkApi) throws Exception {
         // Execute the pipeline
         PipelineExecution execution = composer.compose(pipelineDef);
         execution.execute();
-
```
yuxiqian, line 900: nit: irrelevant change

ruanhang1993 (approved):
LGTM

PR #3855, opened 2025-01-14
[hotfix][postgres] PostgresSourceFetchTaskContext log do not output P…

hql0312:
log output format and text


hql0312:
@xiaom review please?

PR #3704, opened 2024-11-09
[FLINK-36682][mysql][mongodb] Add scan.chunk.assign.strategy to avoid OOM exception.

lvyanquan:
Introduce a new config option scan.chunk.assign.strategy to control the order of split assignment.

PR #3856, opened 2025-01-14
[FLINK-37120][cdc-connector] Add ending split chunk first to avoid TaskManager oom

beryllw:
Add ending split chunk first to avoid TaskManager oom

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/factory/MySqlDataSourceFactory.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/factory/MySqlDataSourceFactory.java
@@ -76,6 +76,7 @@ import static org.apache.flink.cdc.connectors.mysql.source.MySqlDataSourceOption
 import static org.apache.flink.cdc.connectors.mysql.source.MySqlDataSourceOptions.PORT;
 import static org.apache.flink.cdc.connectors.mysql.source.MySqlDataSourceOptions.SCAN_BINLOG_NEWLY_ADDED_TABLE_ENABLED;
 import static org.apache.flink.cdc.connectors.mysql.source.MySqlDataSourceOptions.SCAN_INCREMENTAL_CLOSE_IDLE_READER_ENABLED;
+import static org.apache.flink.cdc.connectors.mysql.source.MySqlDataSourceOptions.SCAN_INCREMENTAL_SNAPSHOT_ASSIGN_ENDING_CHUNK_FIRST;
 import static org.apache.flink.cdc.connectors.mysql.source.MySqlDataSourceOptions.SCAN_INCREMENTAL_SNAPSHOT_CHUNK_KEY_COLUMN;
 import static org.apache.flink.cdc.connectors.mysql.source.MySqlDataSourceOptions.SCAN_INCREMENTAL_SNAPSHOT_CHUNK_SIZE;
 import static org.apache.flink.cdc.connectors.mysql.source.MySqlDataSourceOptions.SCAN_NEWLY_ADDED_TABLE_ENABLED;
@@ -150,6 +151,8 @@ public class MySqlDataSourceFactory implements DataSourceFactory {
                 config.get(SCAN_BINLOG_NEWLY_ADDED_TABLE_ENABLED);
         boolean isParsingOnLineSchemaChanges = config.get(PARSE_ONLINE_SCHEMA_CHANGES);
         boolean useLegacyJsonFormat = config.get(USE_LEGACY_JSON_FORMAT);
+        boolean isAssignEndingChunkFirst =
+                config.get(SCAN_INCREMENTAL_SNAPSHOT_ASSIGN_ENDING_CHUNK_FIRST);
 
         validateIntegerOption(SCAN_INCREMENTAL_SNAPSHOT_CHUNK_SIZE, splitSize, 1);
         validateIntegerOption(CHUNK_META_GROUP_SIZE, splitMetaGroupSize, 1);
@@ -201,7 +204,8 @@ public class MySqlDataSourceFactory implements DataSourceFactory {
                         .scanNewlyAddedTableEnabled(scanNewlyAddedTableEnabled)
                         .parseOnLineSchemaChanges(isParsingOnLineSchemaChanges)
                         .treatTinyInt1AsBoolean(treatTinyInt1AsBoolean)
-                        .useLegacyJsonFormat(useLegacyJsonFormat);
+                        .useLegacyJsonFormat(useLegacyJsonFormat)
+                        .assignEndingChunkFirst(isAssignEndingChunkFirst);
 
         List<TableId> tableIds = MySqlSchemaUtils.listTables(configFactory.createConfig(0), null);
 
@@ -337,6 +341,7 @@ public class MySqlDataSourceFactory implements DataSourceFactory {
         options.add(USE_LEGACY_JSON_FORMAT);
         options.add(TREAT_TINYINT1_AS_BOOLEAN_ENABLED);
         options.add(PARSE_ONLINE_SCHEMA_CHANGES);
+        options.add(SCAN_INCREMENTAL_SNAPSHOT_ASSIGN_ENDING_CHUNK_FIRST);
         return options;
     }
 
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
@@ -313,4 +313,12 @@ public class MySqlDataSourceOptions {
                     .defaultValue(true)
                     .withDescription(
                             "Whether to use legacy json format. The default value is true, which means there is no whitespace before value and after comma in json format.");
+
+    @Experimental
+    public static final ConfigOption<Boolean> SCAN_INCREMENTAL_SNAPSHOT_ASSIGN_ENDING_CHUNK_FIRST =
+            ConfigOptions.key("scan.incremental.snapshot.unbounded-chunk-first.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Whether to assign the ending chunk first during snapshot reading phase. This might help reduce the risk of the TaskManager experiencing an out-of-memory (OOM) error when taking a snapshot of the largest unbounded chunk.  Defaults to false.");
 }
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/splitter/JdbcSourceChunkSplitter.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/splitter/JdbcSourceChunkSplitter.java
@@ -470,7 +470,13 @@ public abstract class JdbcSourceChunkSplitter implements ChunkSplitter {
             }
         }
         // add the ending split
-        splits.add(ChunkRange.of(chunkStart, null));
+        // assign ending split first, both the largest and smallest unbounded chunks are completed
+        // in the first two splits
+        if (sourceConfig.isAssignEndingChunkFirst()) {
+            splits.add(0, ChunkRange.of(chunkStart, null));
+        } else {
+            splits.add(ChunkRange.of(chunkStart, null));
+        }
         return splits;
     }
 
@@ -498,7 +504,7 @@ public abstract class JdbcSourceChunkSplitter implements ChunkSplitter {
             chunkEnd = nextChunkEnd(jdbc, chunkEnd, tableId, splitColumn, max, chunkSize);
         }
         // add the ending split
-        splits.add(ChunkRange.of(chunkStart, null));
+        splits.add(0, ChunkRange.of(chunkStart, null));
         return splits;
     }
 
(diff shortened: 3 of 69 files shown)


beryllw:
CI：https://github.com/beryllw/flink-cdc/actions/runs/12779452404

beryllw:
Adding the ending split chunk first changes the snapshot read result order. I will fix the unit test.

leonardBang:
THanks @beryllw for the contribution, the idea makes sense to me, but we need to consider the compatibility, a configuration is recommended

beryllw:
> THanks @beryllw for the contribution, the idea makes sense to me, but we need to consider the compatibility, a configuration is recommended

 If data consistency is not an issue, are there any compatibility concerns we need to address?

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/splitter/JdbcSourceChunkSplitter.java
```diff
@@ -498,7 +498,7 @@ private List<ChunkRange> splitUnevenlySizedChunks(
             chunkEnd = nextChunkEnd(jdbc, chunkEnd, tableId, splitColumn, max, chunkSize);
         }
         // add the ending split
-        splits.add(ChunkRange.of(chunkStart, null));
+        splits.add(0, ChunkRange.of(chunkStart, null));
```
lvyanquan, line 501: Good optimization, so both the largest and smallest unbounded chunks are completed in the first two splits, and subsequent chunks are bounded size.

lvyanquan:
What‘s more, could you add this optimization to `MongoDBChunkSplitter`?

beryllw:
> What‘s more, could you add this optimization to `MongoDBChunkSplitter`?

https://github.com/apache/flink-cdc/pull/3704#discussion_r1864686969

I missed this issue https://github.com/apache/flink-cdc/pull/3704, good idea. Maybe support `AssignStrategy` in base module is better? Do we implement this in a future PR or in this PR?

lvyanquan:
> I missed this issue #3704, good idea. Maybe support `AssignStrategy` in base module is better? Do we implement this in a future PR or in this PR?

My previous PR provided both `ascending_order` and `descending_order` order configurations, but I think it would be better to place both unbounded chunks at the beginning, so you can go ahead.

I think if there is no impact on restarting from the previous state, there is no compatibility issue, but adding a parameter to control it would be safer.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/factory/MySqlDataSourceFactory.java
```diff
@@ -336,6 +340,8 @@ public Set<ConfigOption<?>> optionalOptions() {
         options.add(INCLUDE_COMMENTS_ENABLED);
         options.add(USE_LEGACY_JSON_FORMAT);
         options.add(TREAT_TINYINT1_AS_BOOLEAN_ENABLED);
+        options.add(PARSE_ONLINE_SCHEMA_CHANGES);
+        options.add(SCAN_INCREMENTAL_SNAPSHOT_ASSIGN_ENDING_FIRST);
```
lvyanquan, line 344: Please update the [test](https://github.com/apache/flink-cdc/blob/master/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceFactoryTest.java#L250) for this optionalOptions.

lvyanquan:
I think adding a test in [MySqlSourceITCase](https://github.com/apache/flink-cdc/blob/8815f2b879fd48ad6211141b7f627eb1d143a02c/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlSourceITCase.java#L223) to test the case of restoring from failure will be better.

beryllw:
> I think adding a test in [MySqlSourceITCase](https://github.com/apache/flink-cdc/blob/8815f2b879fd48ad6211141b7f627eb1d143a02c/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlSourceITCase.java#L223) to test the case of restoring from failure will be better.

agree.

lvyanquan:
LGTM.
I'm worried if this parameter is a bit complicated, what about naming it `scan.incremental.assign-max-chunk-first.enabled`?
WDYT @beryllw @leonardBang?

beryllw:
> LGTM. I'm worried if this parameter is a bit complicated, what about naming it `scan.incremental.assign-max-chunk-first.enabled` to avoid using `ending`? As `chunk` is a concept in snapshot phase. WDYT @beryllw @leonardBang?

What about 'scan.incremental.assign-ending-chunk-first.enabled', when the write throughput is low, the ending chunk is not equal to the maximum chunk.

beryllw:
@leonardBang @lvyanquan PTAL

lvyanquan:
Thanks @beryllw for this update, since you've added this feature to jdbc common module, it seems like that this config option was not added in DynamicTableSourceFactory and was not passed to Source.

beryllw:
> Thanks @beryllw for this update, since you've added this feature to jdbc common module, it seems like that this config option was not added in DynamicTableSourceFactory and was not passed to Source.

Thanks for pointing out the error; it's fixed. Awaiting CI.
https://github.com/beryllw/flink-cdc/actions/runs/13256583939?pr=7

beryllw:
> Do you want to support it all here, or do you want to split the modification of JDBC base into another PR.

Support it all here is better.Thanks.

beryllw:
https://github.com/beryllw/flink-cdc/actions/runs/13258656371?pr=7
@leonardBang Please help me retrigger the CI.

beryllw:
CI https://github.com/beryllw/flink-cdc/actions/runs/13263762804

beryllw:
CI https://github.com/beryllw/flink-cdc/actions/runs/13263888883.

beryllw:
CI https://github.com/beryllw/flink-cdc/actions/runs/13276446389.

lvyanquan (approved):
Thanks @beryllw for this contribution and it looks good to me.

I think you can create one jira to make the default value of this option to `true` after one or two major versions and testing in product environment.

lvyanquan:
And you might need to do some rebase to fix the ci failure.

beryllw:
> And you might need to do some rebase to fix the ci failure.

Sure.

beryllw:
https://github.com/beryllw/flink-cdc/actions/runs/13387743918
@leonardBang Please help me retrigger the CI.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/splitter/JdbcSourceChunkSplitter.java
```diff
@@ -470,7 +470,13 @@ private List<ChunkRange> splitEvenlySizedChunks(
             }
         }
         // add the ending split
-        splits.add(ChunkRange.of(chunkStart, null));
+        // assign ending split first, both the largest and smallest unbounded chunks are completed
+        // in the first two splits
+        if (sourceConfig.isAssignEndingChunkFirst()) {
+            splits.add(0, ChunkRange.of(chunkStart, null));
```
leonardBang, line 476: Could you consider `  System.arraycopy` for performance ? Currently we need  move n-1 elements.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
```diff
@@ -313,4 +313,12 @@ public class MySqlDataSourceOptions {
                     .defaultValue(true)
                     .withDescription(
                             "Whether to use legacy json format. The default value is true, which means there is no whitespace before value and after comma in json format.");
+
+    @Experimental
+    public static final ConfigOption<Boolean> SCAN_INCREMENTAL_SNAPSHOT_ASSIGN_ENDING_CHUNK_FIRST =
+            ConfigOptions.key("scan.incremental.snapshot.assign-ending-chunk-first.enabled")
```
leonardBang, line 319: How about ```suggestion
            ConfigOptions.key("scan.incremental.snapshot.boundary-chunk-first.enabled")
```? IIUC, this options is used to control the boundary chunks should be assigned first or not?

**Review by leonardBang — COMMENTED** [review]
Thanks @beryllw for the contribution, I left two comments

File: flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java (hunk 4 above)
Comment (beryllw) [reply flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java:319]: IIUC, should it be `ConfigOptions.key("scan.incremental.snapshot.unboundary-chunk-first.enabled")`? The end is the largest and the start is the smallest unbounded chunks.

Comment (leonardBang) [reply flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java:319]: I mean ConfigOptions.key("scan.incremental.snapshot.boundary-chunk-first.enabled") with default value `true`, currently we assign first chunk(unboundary chunk) and then assign boundary chunk, right?

But the two ConfigOptions make sense to me when you use different default value, feel free to choose you flavor.

File: flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/splitter/JdbcSourceChunkSplitter.java (hunk 3 above)
Comment (beryllw) [reply flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/splitter/JdbcSourceChunkSplitter.java:476]: You mentioned using `System.arraycopy` for performance as we need to move n-1 elements. The `ArrayList` function `add(int index, E element)` is implemented as follows, using `System.arraycopy` internally. Is there anything else I need to change?

```java
/**
 * Inserts the specified element at the specified position in this
 * list. Shifts the element currently at that position (if any) and
 * any subsequent elements to the right (adds one to their indices).
 *
 * @param index index at which the specified element is to be inserted
 * @param element element to be inserted
 * @throws IndexOutOfBoundsException {@inheritDoc}
 */
public void add(int index, E element) {
    rangeCheckForAdd(index);

    ensureCapacityInternal(size + 1);  // Increments modCount!!
    System.arraycopy(elementData, index, elementData, index + 1,
                     size - index);
    elementData[index] = element;
    size++;
}
```

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
beryllw: Given that "unboundary" is not a commonly used term and may cause confusion, I have decided to use "unbounded" instead. Therefore, the configuration item will be `scan.incremental.snapshot.unbounded-chunk-first.enabled`.

leonardBang: +1 from my side, please keep go on

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/splitter/JdbcSourceChunkSplitter.java
leonardBang: No, thanks for your explanation

leonardBang (approved):
Thanks @beryllw for the great work, LGTM

lvyanquan:
Hi, @beryllw. Could you check the failed cases? If some are unrelated to the change, please add some description.

beryllw:
> Hi, @beryllw. Could you check the failed cases? If some are unrelated to the change, please add some description.

Looks like unrelated to this change, here is the CI https://github.com/beryllw/flink-cdc/pull/7, and I also tested it manually.

beryllw:
Looks like unrelated to this change, here is the CI https://github.com/beryllw/flink-cdc/pull/7, and I also manually tested the failed tests(except for MySqlConnectorITCase).

beryllw:
@leonardBang Is there anything else that needs to be completed?  I'm concerned about potential conflicts later on.Could you please help me move this forward?

leonardBang:
> @leonardBang Is there anything else that needs to be completed? I'm concerned about potential conflicts later on.Could you please help me move this forward?

Hey, @beryllw Thank you for your reminder, I'll help investigate the failure case and merge your PR.

leonardBang:
> @leonardBang Is there anything else that needs to be completed? I'm concerned about potential conflicts later on.Could you please help me move this forward?

I guess the failed test cases in `SqlServer` and `MongoDB` may be related to our change, as some test case may assume the legacy split order will be assign first, could you check them?

beryllw:
> I guess the failed test cases in `SqlServer` and `MongoDB` may be related to our change, as some test case may assume the legacy split order will be assign first, could you check them?

Sure, I will recheck the failed test cases in `SqlServer` and `MongoDB`.

leonardBang:
Rebase this PR to master to obtain fixed patches, let's see what happens next @beryllw

leonardBang:
CI passed, merging....

beryllw:
> Rebase this PR to master to obtain fixed patches, let's see what happens next @beryllw

Sorry, I've been too busy to check earlier. Thanks for your help.

leonardBang:
> > Rebase this PR to master to obtain fixed patches, let's see what happens next @beryllw
> 
> Sorry, I've been too busy to check earlier. Thanks for your help.

You're welcome @beryllw , no need to say sorry, the community is just our interest, our own work is the first priority ^^

PR #3858, opened 2025-01-14
Fixed the problem of using the same executor to process schema change requests and flush event requests, resulting in blocking timeout.

linjianchang:
When source generate metadata events in parallel for each table,There is a problem of timeout waiting for flush to complete. Exception be like:
959f2253678ad7b7b567e1cb38598970

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/distributed/SchemaCoordinator.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/distributed/SchemaCoordinator.java
@@ -61,6 +61,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
@@ -98,6 +99,9 @@ public class SchemaCoordinator extends SchemaRegistry {
     private transient Multimap<Tuple2<Integer, SchemaChangeEvent>, Integer>
             alreadyHandledSchemaChangeEvents;
 
+    /** Executor service to execute schema change. */
+    private final ExecutorService schemaChangeThreadPool;
+
     public SchemaCoordinator(
             String operatorName,
             OperatorCoordinator.Context context,
@@ -114,6 +118,7 @@ public class SchemaCoordinator extends SchemaRegistry {
                 routingRules,
                 schemaChangeBehavior,
                 rpcTimeout);
+        this.schemaChangeThreadPool = Executors.newSingleThreadExecutor();
     }
 
     // -----------------
@@ -129,6 +134,14 @@ public class SchemaCoordinator extends SchemaRegistry {
         this.alreadyHandledSchemaChangeEvents = HashMultimap.create();
         LOG.info(
                 "Started SchemaRegistry for {}. Parallelism: {}", operatorName, currentParallelism);
+    }
+
+    @Override
+    public void close() throws Exception {
+        super.close();
+        if (schemaChangeThreadPool != null && !schemaChangeThreadPool.isShutdown()) {
+            schemaChangeThreadPool.shutdownNow();
+        }
     }
 
     // --------------------------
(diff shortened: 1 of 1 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
```diff
@@ -96,6 +97,9 @@ public abstract class SchemaRegistry implements OperatorCoordinator, Coordinatio
     protected transient SchemaManager schemaManager;
     protected transient TableIdRouter router;
 
+    /** Executor service to execute handle event from operator. */
+    private final ExecutorService runInEventFromOperatorExecutor;
```
yuxiqian, line 101: Keep its name consistent with regular schema coordinator

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
yuxiqian, line 1: IIUC this problem will only occur if:

* Distributed schema evolution topology is created
* Flush success event takes a while to finish
* The last schema operator initiates request before any flush succeeds

Thus, The handler of FlushSuccessEvent will wait for schema evolution to finish, but the busy-loop is still waiting for collecting all flush success events.

Could you please add a test case to verify this change?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
```diff
@@ -325,6 +337,7 @@ public final void resetToCheckpoint(long checkpointId, @Nullable byte[] checkpoi
      * directly, make sure you're running heavy logics inside, or the entire job might hang!
      */
     protected void runInEventLoop(
+            final ExecutorService coordinatorExecutor,
```
yuxiqian, line 340: It's nice to allow specifying `ExecutorService` when 
 calling `runInEventLoop`. Maybe `regular/SchemaCoordinator` could also invoke this instead of this:

```java
schemaChangeThreadPool.submit(
                () -> {
                    try {
                        applySchemaChange(originalEvent, deducedSchemaChangeEvents);
                    } catch (Throwable t) {
                        failJob(
                                "Schema change applying task",
                                new FlinkRuntimeException(
                                        "Failed to apply schema change event.", t));
                        throw t;
                    }
                });
```

yuxiqian (changes requested):
Thanks for @linjianchang's quick fix! Just left some trivial comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
linjianchang: already modified

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
```diff
@@ -253,6 +260,7 @@ public final CompletableFuture<CoordinationResponse> handleCoordinationRequest(
     public final void handleEventFromOperator(
             int subTaskId, int attemptNumber, OperatorEvent event) {
         runInEventLoop(
+                runInEventFromOperatorExecutor,
```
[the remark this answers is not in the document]
linjianchang: already modified

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
linjianchang: The test case org.apache.flink.cdc.runtime.operators.schema.distributed.SchemaEvolveTest#testLenientSchemaEvolution() has been covered this case.

linjianchang:
Already modified according to comment

linjianchang:
Already modified according to comment @yuxiqian

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
yuxiqian: I'm afraid it's not sufficient because runtime unit tests uses `ValuesDataSink` only. It takes no time to flush and might not be able to verify this.

yuxiqian:
Seems there's something wrong with internal state switching logic. Would @linjianchang like to take a further look?

Screenshot 2025-01-21 09 43 21

lvyanquan:
Met the same problem.

lvyanquan:
@linjianchang could you update the title to add the related jira？

PR #3449, opened 2024-07-02
[FLINK-35743][cdc-runtime] Fix the time zone configuration for temporal functions is not effective

aiwenmo:
This closes [FLINK-35743](https://issues.apache.org/jira/browse/FLINK-35743).

PR #3863, opened 2025-01-16
[FLINK-36973]udf dateformat support LocalZonedTimestampData datatype

hiliuxg:
udf dateformat support LocalZonedTimestampData datatype

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
@@ -105,6 +105,13 @@ public class SystemFunctionUtils {
                 timestamp.getMillisecond(), format, TimeZone.getTimeZone("UTC"));
     }
 
+    public static String dateFormat(LocalZonedTimestampData timestamp, String format) {
+        return DateTimeUtils.formatTimestampMillis(
+                timestamp.getEpochMillisecond(),
+                format,
+                TimeZone.getTimeZone(ZoneId.systemDefault()));
+    }
+
     public static int toDate(String str, String timezone) {
         return toDate(str, "yyyy-MM-dd", timezone);
     }


lvyanquan:
Seems like it was deleted in https://github.com/apache/flink-cdc/pull/3449/files#diff-718e4ed816fe9f6da97bb492b57f1c67121b968b981c175ed4e961901fe25843, could @aiwenmo take a look?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
```diff
@@ -105,6 +105,13 @@ public static String dateFormat(TimestampData timestamp, String format) {
                 timestamp.getMillisecond(), format, TimeZone.getTimeZone("UTC"));
     }
 
+    public static String dateFormat(LocalZonedTimestampData timestamp, String format) {
+        return DateTimeUtils.formatTimestampMillis(
+                timestamp.getEpochMillisecond(),
+                format,
+                TimeZone.getTimeZone(ZoneId.systemDefault()));
+    }
```
aiwenmo, line 113: In the YAML job, the default time zone needs to be fetched from the pipeline configuration instead of the system default time zone.

aiwenmo (changes requested):
The setting method of the default time zone is incorrect. And could you add test to verify this function?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/SystemFunctionUtils.java
hiliuxg: Thank you for taking the time to review my code. Could you please tell me how to fetch the yaml job pipeline configuration  in systemfunctionutils class?

yuxiqian: I believe this could be implemented by this:

### `SystemFunctionUtils` class

Add an extra parameter (`timezone`) in `dateFormat` functions' signature like this:

```java
static String dateFormat(TimestampData timestamp, String format, String timezone)
static String dateFormat(LocalZonedTimestampData timestamp, String format, String timezone)
```

### `JaninoCompiler` class

Move `DATE_FORMAT` from `TIMEZONE_FREE_TEMPORAL_CONVERSION_FUNCTIONS` to `TIMEZONE_REQUIRED_TEMPORAL_CONVERSION_FUNCTIONS`.

hiliuxg:
> Hi @hiliuxg, Is there any progress for this pr

sorry, long time no see this pr, may be next week .

yuxiqian:
Hi @hiliuxg, we're preparing for the next release in the next few days, so I've cherry-picked your commits into #4003. Would appreciate it if you could take a look.

leonardBang:
closed via https://github.com/apache/flink-cdc/pull/4003

PR #3864, opened 2025-01-16
[FLINK-36406][cdc-runtime] Close MetadataApplier when the job stops

leonardBang:
This is a follow-up PR after #3623

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
@@ -45,6 +45,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
+import java.io.IOException;
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/distributed/SchemaCoordinator.java
```diff
@@ -131,6 +132,17 @@ public void start() throws Exception {
                 "Started SchemaRegistry for {}. Parallelism: {}", operatorName, currentParallelism);
     }
 
+    @Override
```
yuxiqian, line 135: Can we escalate Metadata Applier closing code to `SchemaRegistry#close`? Seems `metadataApplier` is also a field in abstract base class anyway.

leonardBang: +1

yuxiqian (approved):
Looks nice!

ruanhang1993 (approved):
LGTM

PR #3865, opened 2025-01-16
[FLINK-37132][transform] validate meta schema in Multi Transform

MOBIN-F:
In Multi Transform, verify that column counts, metadata fields like primaryKeys, partitionKeys, and options are consistent


MOBIN-F:
@yuxiqian Could you help review the code?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaUtils.java
```diff
@@ -452,6 +455,43 @@ public static Schema inferWiderSchema(@Nullable Schema lSchema, Schema rSchema)
...
+        if (schemas.size() > 1) {
+            Schema outputSchema = null;
+            for (Schema schema : schemas) {
+                isMetaSchemaCompatible(outputSchema, schema);
+                outputSchema = schema;
+            }
+        }
+    }
+
+    public static void isMetaSchemaCompatible(
```
yuxiqian, line 468: nit: this is not a pure function returning `Boolean`, so `validateSchemaMetadataCompatibility` might be a better name.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaUtils.java
```diff
@@ -452,6 +455,43 @@ public static Schema inferWiderSchema(@Nullable Schema lSchema, Schema rSchema)
         return lSchema.copy(mergedColumns);
     }
 
+    public static void validateMetaSchemaCompatibility(LinkedHashSet<Schema> schemas) {
```
yuxiqian, line 458: I wonder if is this method really reusable enough to be a utility method?

Seems the argument type is quite specific (it must be a `LinkedHashSet` instead of a generic schemas' collection). Also, it didn't return a "Yes or No" boolean and let the caller to decide what to do, but throws an exception if it's incompatible.

flink-cdc-common/src/test/java/org/apache/flink/cdc/common/utils/SchemaMergingUtilsTest.java
```diff
@@ -1116,6 +1117,27 @@ void testGetLeastCommonType() {
                         MAP));
     }
 
+    @Test
+    void testTransformColumn() {
+        Assertions.assertThatCode(
```
yuxiqian, line 1122: We may use `assertThatThrownBy` as we're surely expecting an error

yuxiqian (commented):
Thanks for @MOBIN-F's nice work, just left some minor comments.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaUtils.java
MOBIN-F: [but throws an exception if it's incompatible]
yeal,I think this kind of check is mandatory and cannot be ignored, and the exception is fatal, so I designed it to return void, but returning boolean seems to be good too. What do you think?

yuxiqian: My concern is if we're throwing an exception in the utility method, it might be less universal. But your implementation is more reasonable, because we can't return any reason message to indicate the incompatible reason by returning a boolean.

Let's just keep it as is.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
```diff
@@ -169,20 +173,52 @@ public static Schema getLeastCommonSchema(
     }
 
     /** Merge compatible schemas. */
-    public static Schema getCommonSchema(List<Schema> schemas) {
-        if (schemas.isEmpty()) {
-            return null;
-        } else if (schemas.size() == 1) {
-            return schemas.get(0);
+    public static Schema getCommonSchema(LinkedHashSet<Schema> schemas) {
```
yuxiqian, line 176: This is a breaking API change since the parameter type constraints are getting stricter. I think changing List => Collection is acceptable since it's getting wider.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaUtils.java
MOBIN-F: I have updated the code, could you please help review it again,thanks~

MOBIN-F:
@yuxiqian @lvyanquan PTAL

lvyanquan:
Thanks for your fix.
But there is now a user scenario where users want to add fields of `__namespace_name__ || '.' || __schema_name__ || '.' || __table_name__ identifier_name` and `__data_event_type__ AS op_type` to certain tables, but they do not need these two fields for other tables(or only one `__data_event_type__ AS op_type` field). and their yaml file may looks like this:
```
transform:
  - source-table: mydb.web_order[0-9]+
    projection: id, order_id, __namespace_name__ || '.' || __schema_name__ || '.' || __table_name__ identifier_name, __data_event_type__ AS op_type
  - source-table: \.*.\.*
    projection: \*, __data_event_type__ AS op_type
```

Because we do not have table.exclude configuration in transform, making it difficult to configure such information for other tables if we require the same schema after each transform rule.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PreTransformOperator.java
```diff
@@ -411,18 +411,16 @@ private void cacheTransformRuleInfo(CreateTableEvent createTableEvent) {
...
-        }
-
-        cachePreTransformProcessor(tableId, createTableEvent.getSchema());
+        Schema originalSchema = createTableEvent.getSchema();
+        LinkedHashSet<Schema> newSchemas =
+                schemaMetadataTransformers.stream()
+                        .filter(transform -> transform.f0.isMatch(tableId))
+                        .map(transform -> transformSchemaMetaData(originalSchema, transform.f1))
+                        .collect(Collectors.toCollection(LinkedHashSet::new));
+        SchemaUtils.validateMetaSchemaCompatibility(newSchemas);
```
lvyanquan, line 420: We should throw an exception to clearly indicate which two rules conflict for which table.

MOBIN-F:
> Thanks for your fix. But there is now a user scenario where users want to add fields of `__namespace_name__ || '.' || __schema_name__ || '.' || __table_name__ identifier_name` and `__data_event_type__ AS op_type` to certain tables, but they do not need these two fields for other tables(or only one `__data_event_type__ AS op_type` field). and their yaml file may looks like this:
> 
> ```
> transform:
>   - source-table: mydb.web_order[0-9]+
>     projection: id, order_id, __namespace_name__ || '.' || __schema_name__ || '.' || __table_name__ identifier_name, __data_event_type__ AS op_type
>   - source-table: \.*.\.*
>     projection: \*, __data_event_type__ AS op_type
> ```
> 
> Because we do not have table.exclude configuration in transform, making it difficult to configure such information for the rest of tables if we require the same schema after each transform rule.


maybe we can use regular expressions to exclude mydb.web_order[0-9]+
```
- source-table: \.*.\.* Change to - source-table: ^(?!mydb\.web_order\d+$).\*
```
	
Suppose we support the following transform (i.e. the first projection is used first when there are multiple projections for the same table)
```
transform:
  - source-table: mydb.web_order[0-9]+
     projection: id, order_id, __namespace_name__ || '.' || __schema_name__ || '.' || __table_name__ identifier_name, __data_event_type__ AS op_type
  - source-table: \.*.\.*
     projection: \*, __data_event_type__ AS op_type
```

Then the following transform will also become legal, but it is difficult to determine the user's intention from the transform rules, which can easily cause confusion
```
transform:
  - source-table: default_namespace.default_schema.mytable2
    projection: \*
    filter: age < 18
  - source-table: default_namespace.default_schema.mytable2
    projection: id,UPPER(name) AS name
    filter: age >= 18
```

PR #3866, opened 2025-01-16
[hotfix][test] Make pipeline migration tests more robust

yuxiqian:
This makes pipeline migration test more robust by checking job submit output correctly, and does not depend on specific output formats.


yuxiqian:
This should fix the CI failure since https://github.com/apache/flink-cdc/actions/runs/12806593233/job/35705386304. cc @leonardBang

leonardBang (approved):
+1

PR #3869, opened 2025-01-16
[FLINK-34865][pipeline-connector/mysql] Support sync table comment in binlog phase

qg-lin:
Support sync table coments in binlog phase, thanks for @yuxiqian nice guidance.


yuxiqian:
Thanks for @qg-lin's nice work, this change looks reasonable to me.

But seems our MySQL CI has been a little unstable for now... Could you please rebase this PR with https://github.com/apache/flink-cdc/commit/e81df946b9b033d52e10d3c1b921d29aeed7d788 commit and see if this problem persists?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlPipelineITCase.java
```diff
@@ -1076,6 +1077,75 @@ public void testIncludeComments() throws Exception {
...
+                (FlinkSourceProvider) dataSource.getEventSourceProvider();
+
+        CloseableIterator<Event> events =
+                env.fromSource(
+                                sourceProvider.getSource(),
+                                WatermarkStrategy.noWatermarks(),
+                                MySqlDataSourceFactory.IDENTIFIER,
+                                new EventTypeInfo())
+                        .executeAndCollect();
+        Thread.sleep(5_000);
```
yuxiqian, line 1113: nit: we can `fetchResults` here to ensure snapshot stage to be over before generating binlog events, instead of sleeping.

lvyanquan:
Looks like we would need to add more event types like AlterTableCommentEvent & AlterColumnCommentEvent.

leonardBang:
> Looks like we would need to add more event types like AlterTableCommentEvent & AlterColumnCommentEvent.

We can support these events in the future

leonardBang (approved):
+1， wait the CI green

PR #3873, opened 2025-01-21
[FLINK-36086] [Sqlserver] Fix Sqlserver restart during one transaction

GOODBOY008:
Test the https://github.com/apache/flink-cdc/pull/3552 and make sure this scenario worked.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-sqlserver-cdc/src/main/java/io/debezium/connector/sqlserver/SqlServerOffsetContext.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-sqlserver-cdc/src/main/java/io/debezium/connector/sqlserver/SqlServerOffsetContext.java
@@ -1,0 +1,240 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.sqlserver;
+
+import io.debezium.connector.SnapshotRecord;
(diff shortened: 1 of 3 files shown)


morozov (approved):
@GOODBOY008, thanks for picking this up! The additional change in the code look good to me, and the test logic checks out. I confirmed that the test covers the scenario in which data was lost.

As for the test quality, I believe the test may cause CI issues and is potentially flaky. I don't think its a blocker. Just FYI.

morozov:
I read more about the semantics of `InputStatus`, and I think the current logic is good. Please validate my understanding below.

There are three possible value of the status:
1. `MORE_AVAILABLE` means that a new record is immediately available (i.e. won't block the reader).
2. `NOTHING_AVAILABLE` means that there's no new record immediately available (so the reader may block).
3. `END_OF_INPUT` means end of input.

In this specific case, here's how we interpret them:
1. `END_OF_INPUT` is impossible in the binlog/streaming mode.
2. There's no difference for us between `MORE_AVAILABLE` and `NOTHING_AVAILABLE` because we don't care about blocking, we want to just read all the changes.

So, what happens in the test is:
1. Initially, the reader returns `NOTHING_AVAILABLE` because the connector hasn't captured anything yet.
2. Since the output is empty, the test keep iterating in the do-while loop.
3. As soon as the change become available, the reader returns `MORE_AVAILABLE` and then returns _all captured changes_ before switching back to `NOTHING_AVAILABLE`.
4. At this point, the output is no longer empty, so the loop breaks.

So if the assumption in #​3 doesn't hold, it may be a reason why the test fail. I don't know of a better simple way to read a fixed expected number of changes from the reader without introducing a timeout or risking a never-terminating test. I think the current approach is fine.

Jiabao-Sun (approved):
+1

PR #3887, opened 2025-01-23
[FLINK-35713][common] Support set parallelism in source and sink

ChaomingZhangCN:
https://issues.apache.org/jira/browse/FLINK-35713


ChaomingZhangCN:
@lvyanquan @leonardBang PTAL.

ChaomingZhangCN:
@yuxiqian I found that the PR-[3438](https://github.com/apache/flink-cdc/pull/3438/files) has been closed, could you please take a look?

ChaomingZhangCN:
ping @leonardBang

leonardBang:
@ChaomingZhangCN Could you rebase this PR to latest master branch as it bumped junit version to junit5 for all tests?

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
```diff
@@ -188,7 +189,14 @@ private SourceDef toSourceDef(JsonNode sourceNode) {
         // "name" field is optional
         String name = sourceMap.remove(NAME_KEY);
 
-        return new SourceDef(type, name, Configuration.fromMap(sourceMap));
+        // "parallelism" field is optional
+        Integer parallelismValue = null;
```
leonardBang, line 193: Could we use `pipeline.parallelism`'s valued as  default parallelism instead of an error-prone null value? then we  can use `int` type and do not need to use `@Nullable ` annotation in related methods.

leonardBang (commented):
Thanks @ChaomingZhangCN for the contribution, generally the PR looks good, but I want to ping @yuxiqian and @lvyanquan take a look as the source parallelism may influence `TransformOperator` and `SchemaOperator` after `SourceOperator`， and pre/post operator topology around `SinkOperator` ?

yuxiqian:
Minor suggestion: shall we pick `MySqlToPaimonE2eITCase` and `FlinkParallelizedPipelineITCase` to be executed with different source / sink parallelism? I believe they could cover most parallelism-aware operators (flush event alignment, 2-PC sinks, and schema operators).

leonardBang:
CI failed...

ChaomingZhangCN:
@leonardBang PTAL~

leonardBang:
@ChaomingZhangCN Could you fix the conflicts?

PR #3890, opened 2025-01-23
[FLINK-37209][transform] Add built-in Qwen model for transform.

lvyanquan:
Add built-in [Qwen](https://github.com/QwenLM/Qwen2.5) model for transform.


flink-cdc-pipeline-model/flink-cdc-pipeline-model-qwen/src/test/java/org/apache/flink/cdc/runtime/model/TestQwenEmbeddingModel.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+/** A test for {@link QwenEmbeddingModel}. */
+public class TestQwenEmbeddingModel {
+
+    @Test
+    @Disabled("For manual test as there is a limit for quota.")
```
lvyanquan, line 32: Testing is disabled here because no publicly available API account was found and there are quota restrictions. 
I will try to find a solution.

PR #3902, opened 2025-02-04
[FLINK-37191][mysql] Avoid back filling if the lowWatermark of BinlogSplit is equal to highWatermark.

lvyanquan:
To fix timeout in test case of PolardbxSourceITCase. 

Avoid extra backfill when there is no binlog event between lowWatermark and highWatermark.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/offset/BinlogOffset.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/offset/BinlogOffset.java
@@ -255,6 +255,11 @@ public class BinlogOffset implements Comparable<BinlogOffset>, Serializable {
             return Long.compare(this.getRestartSkipRows(), that.getRestartSkipRows());
         }
 
+        // serverId of BinlogOffset in snapshot phase is 0.
+        if (serverId == 0) {
+            return 0;
+        }
+
         // The skip rows are the same, so compare the timestamp ...
         return Long.compare(this.getTimestampSec(), that.getTimestampSec());
     }


lvyanquan:
@yuxiqian  @leonardBang PTAL.

leonardBang:
CI failed....

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/offset/BinlogOffset.java
yuxiqian, line 1: BinlogOffset#compareTo is a general-purposed comparison method, and this change might affect more than backfill necessity check. Is it possible to tweak `SnapshotSplitReader#isBackfillRequired` only?

lvyanquan:
Hi, @JNSimba.

I found that the change of https://github.com/apache/flink-cdc/pull/3415 will cause the following problem:
1. Even though there is not record between lowWatermark and highWatermark, we still need a backfill procedure, and the timestamp of highWatermark is greater than lowWatermark.
2. If there is no more record after highWatermark, the backfill procedure will never stop, as we will need to wait for a record that is at or after the highWatermark(as the following code displayed).
https://github.com/apache/flink-cdc/blob/7717779ebff3d5900c3adcd06bc860949d543f97/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlBinlogSplitReadTask.java#L102

I thank that the change of https://github.com/apache/flink-cdc/pull/3415 mainly to solve the problem of the comparson of highWatermark and the last record, but in reality, we don't need to read this record, so we can skip it directly.
https://github.com/apache/flink-cdc/blob/7717779ebff3d5900c3adcd06bc860949d543f97/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlBinlogSplitReadTask.java#L102

Notice that this will not cause data loss, as this record will be read in backfill phase.

What do you think?

JNSimba:
@lvyanquan Is the backfill phase not stopped because the `ts_sec` of `offsetContext` is always 0, which is less than the `ts_sec` of `highwatermark`(binlogSplit.getEndingOffset)?
https://github.com/apache/flink-cdc/blob/7717779ebff3d5900c3adcd06bc860949d543f97/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/task/MySqlBinlogSplitReadTask.java#L99-L102

lvyanquan:
> @lvyanquan Is the backfill phase not stopped because the `ts_sec` of `offsetContext` is always 0, which is less than the `ts_sec` of `highwatermark`(binlogSplit.getEndingOffset)?

Yes.

> and it will also get stuck?

Make sense. So I would like to revert change of adding timestamp to highWaterMark and skip the record that is equal to the offset of highWaterMark.
The test case of [BinlogSplitReaderTest](https://github.com/apache/flink-cdc/blob/c28431832cc0969cd40a60e082e05a53b887ef04/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReaderTest.java#L900) will be remained to verify this.

JNSimba (approved):
LGTM

yuxiqian (commented):
Pulled and tested locally, I could confirm that PolarDBX test case could run without problem.

But seems `BinlogSplitReaderTest#testBinlogOffsetCompareWithSnapshotAndBinlogPhase` (introduced in #3415) will always fail on this assertion:

```java
assertTrue(sourceRecords.isEmpty());
```

```
Expecting empty but was: [SourceRecord{sourcePartition={server=mysql_binlog_source}, sourceOffset={transaction_id=null, ts_sec=1739012804, file=mysql-bin.000003, pos=15691, row=1, server_id=223344, event=2}} ConnectRecord{topic='mysql_binlog_source.customer_tesuwt.customers', kafkaPartition=null, key=Struct{id=999999}, keySchema=Schema{mysql_binlog_source.customer_tesuwt.customers.Key:STRUCT}, value=Struct{after=Struct{id=999999,name=user_22,address=Shanghai,phone_number=123567891234},source=Struct{version=1.9.8.Final,connector=mysql,name=mysql_binlog_source,ts_ms=1739012804000,db=customer_tesuwt,table=customers,server_id=223344,file=mysql-bin.000003,pos=15831,row=0,thread=5},op=c,ts_ms=1739012808026}, valueSchema=Schema{mysql_binlog_source.customer_tesuwt.customers.Envelope:STRUCT}, timestamp=null, headers=ConnectHeaders(headers=)}]
```

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/offset/BinlogOffset.java
```diff
@@ -232,7 +232,9 @@ public int compareTo(BinlogOffset that) {
             // compared ...
             long timestamp = this.getTimestampSec();
             long targetTimestamp = that.getTimestampSec();
-            return Long.compare(timestamp, targetTimestamp);
+            if (timestamp != 0 && targetTimestamp != 0) {
+                return Long.compare(timestamp, targetTimestamp);
```
lvyanquan, line 236: This change is necessary because that if the TimestampSec is 0, the assumption of `These are from different servers, and their binlog coordinates are not related` is not correct, these two BinlogOffsets could come from the same server.

leonardBang (approved):

JohnMa123:
> 2\. If there is no more record after highWatermark, the backfill procedure will never stop, as we will need to wait for a record that is at or after the highWatermark(as the following code displayed).

hello，The fix for this issue has not yet been implemented in the Flink CDC 3.3.0 release — modifying the compare logic leads to the backfill phase failing to terminate.

PR #3906, opened 2025-02-05
[FLINK-37156] [cdc-composer/cli] pipeline supports collecting data once and writing it to multiple sinks.

qyfftf:
I've implemented a solution where data is collected from MySQL once and then sent to multiple target data sources, such as Paimon, Kafka, and StarRocks, which reduces redundant data collection.
pipeline yaml :
```
source:
  type: mysql
  name: MySQL Source
  hostname: 127.0.0.1
  port: 3306
  username:root
  password: 123456
  tables: test.order
  server-id: 5401-5404
  jdbc.properties.useSSL: false

sink:
   - type: paimon
     name: Paimon Sink
     catalog.properties.metastore: filesystem
     catalog.properties.warehouse: /tmp/path/warehouse
   - type: kafka
     name: Kafka Sink
     properties.bootstrap.servers: PLAINTEXT://localhost:9092
   - type: starrocks
     jdbc-url: jdbc:mysql://127.0.0.1:9030
     load-url: 127.0.0.1:8030
     username: root
     password: ""
     table.create.properties.replication_num: 1

route:
  - source-table:  test.order
    sink-table:  test.order

pipeline:
  name: MySQL to Paimon Pipeline
  parallelism: 2
```

image

### DIFF
--- a/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
+++ b/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
@@ -144,14 +144,17 @@ public class YamlPipelineDefinitionParser implements PipelineDefinitionParser {
                                 SOURCE_KEY));
 
         // Sink is required
-        SinkDef sinkDef =
-                toSinkDef(
-                        checkNotNull(
-                                pipelineDefJsonNode.get(SINK_KEY),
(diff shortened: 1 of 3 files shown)


yuxiqian:
Thanks for @qyfftf's contribution! I have some concerns over the topology design.

According to the new topology graph you've posted, seems there's only one `SchemaOperator` in the upstream while there are multiple downstream sinks. How can a single `SchemaOperator` maintain and evolve schemas in multiple sinks? What if downstream schemas differ (presumably caused by tolerated errors)?

PR #3912, opened 2025-02-08
[FLINK-37278] Optimize regular schema evolution topology's performance

yuxiqian:
This closes FLINK-37278.

Currently, regular SE topology uses the following process to drain existing `DataChangeEvent`s in the pipeline:

1. SchemaOperator ("client") emits `FlushEvent` to downstream.
2. The "client" keeps polling the SchemaCoordinator ("server") with 1-second interval.
3. The "server" rejects all requests from clients until it has collected enough `FlushSuccessEvent` notifications from Sink.

As a result, all schema change requests will took at least 1 second to finish, after at least  one polling interval.

---

This PR replaces the polling code with maintaining a pending schema change request queue, where SchemaCoordinator could manage all pending clients and effectively blocking them from handling upstream events. Schema evolution process could start immediately after `FlushSuccessEvent` got reported, needless to wait for polling requests from clients.

---

With this change, time consumption of `testRegularTablesSourceInMultipleParallelism` test case has been reduced from ~6 minutes to ~50 seconds.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.cdc.runtime.operators.schema.regular;
 
+import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.cdc.common.event.CreateTableEvent;
 import org.apache.flink.cdc.common.event.SchemaChangeEvent;
 import org.apache.flink.cdc.common.event.TableId;
@@ -62,6 +63,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
 import static org.apache.flink.cdc.runtime.operators.schema.common.CoordinationResponseUtils.wrap;
@@ -73,16 +75,13 @@ public class SchemaCoordinator extends SchemaRegistry {
     /** Executor service to execute schema change. */
     private final ExecutorService schemaChangeThreadPool;
 
-    /**
-     * Atomic flag indicating if current RequestHandler could accept more schema changes for now.
-     */
-    private transient RequestStatus schemaChangeStatus;
-
     /** Sink writers which have sent flush success events for the request. */
     private transient ConcurrentHashMap<Integer, Set<Integer>> flushedSinkWriters;
 
-    /** Currently handling request's completable future. */
-    private transient CompletableFuture<CoordinationResponse> pendingResponseFuture;
+    /** Currently handling requests' completable future. */
+    private transient Map<
+                    Integer, Tuple2<SchemaChangeRequest, CompletableFuture<CoordinationResponse>>>
+            pendingRequests;
 
     // Static fields
     public SchemaCoordinator(
@@ -108,7 +107,7 @@ public class SchemaCoordinator extends SchemaRegistry {
     public void start() throws Exception {
         super.start();
         this.flushedSinkWriters = new ConcurrentHashMap<>();
-        this.schemaChangeStatus = RequestStatus.IDLE;
+        this.pendingRequests = new ConcurrentHashMap<>();
     }
 
     @Override
@@ -185,7 +184,7 @@ public class SchemaCoordinator extends SchemaRegistry {
     }
 
     @Override
-    protected void handleFlushSuccessEvent(FlushSuccessEvent event) {
+    protected void handleFlushSuccessEvent(FlushSuccessEvent event) throws TimeoutException {
         int sinkSubtask = event.getSinkSubTaskId();
         int sourceSubtask = event.getSourceSubTaskId();
         LOG.info(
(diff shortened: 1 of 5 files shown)


yuxiqian:
Would @hiliuxg like to take a look?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
```diff
@@ -219,73 +227,14 @@ protected void handleUnrecoverableError(String taskDescription, Throwable t) {
...
-
-            LOG.info(
-                    "All sink writers have flushed for subTaskId {}. Switching to APPLYING state and starting schema evolution...",
-                    subTaskId);
-            flushedSinkWriters.remove(subTaskId);
-            schemaChangeStatus = RequestStatus.APPLYING;
-            pendingResponseFuture = responseFuture;
-            startSchemaChangesEvolve(request, responseFuture);
-        } else {
-            responseFuture.complete(wrap(SchemaChangeResponse.busy()));
```
Shawn-Hx, line 253: Method `SchemaChangeResponse.busy()` can be removed. Also for `ResponseCode.BUSY`.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
```diff
@@ -219,73 +227,14 @@ protected void handleUnrecoverableError(String taskDescription, Throwable t) {
...
-                responseFuture.complete(wrap(SchemaChangeResponse.waitingForFlush()));
-                return;
-            }
-
-            if (!activeSinkWriters.equals(flushedSinkWriters.get(subTaskId))) {
-                LOG.info(
-                        "Not all active sink writers have completed flush. Flushed writers: {}, expected: {}.",
-                        flushedSinkWriters.get(subTaskId),
-                        activeSinkWriters);
-                responseFuture.complete(wrap(SchemaChangeResponse.waitingForFlush()));
```
Shawn-Hx, line 241: Method `SchemaChangeResponse.waitingForFlush()` can be removed. Also for `ResponseCode.WAITING_FOR_FLUSH`.

Shawn-Hx (commented):
It seems that `SchemaChangeResponse#ResponseCode` can only be SUCCESS now. Can we remove `SchemaChangeResponse#ResponseCode` and simplify the logic in `SchemaOperator#handleSchemaChangeEvent` ?

yuxiqian:
Thanks for Shawn's kindly review, comments addressed.

Shawn-Hx (approved):
LGTM

gongzexin:
> It seems that `SchemaChangeResponse#ResponseCode` can only be SUCCESS now. Can we remove `SchemaChangeResponse#ResponseCode` and simplify the logic in `SchemaOperator#handleSchemaChangeEvent` ?

@yuxiqian @Shawn-Hx 
Hi, Have you noticed that during fault tolerance, the same table will be flushed multiple times (related to the task parallelism).So I think SchemaChangeResponse#ResponseCode#DUPLICATE should not be deleted, but it should be strengthened.

yuxiqian:
> > It seems that `SchemaChangeResponse#ResponseCode` can only be SUCCESS now. Can we remove `SchemaChangeResponse#ResponseCode` and simplify the logic in `SchemaOperator#handleSchemaChangeEvent` ?
> 
> @yuxiqian @Shawn-Hx Hi, Have you noticed that during fault tolerance, the same table will be flushed multiple times (related to the task parallelism).So I think SchemaChangeResponse#ResponseCode#DUPLICATE should not be deleted, but it should be strengthened.

Thanks for @gongzexin's report. IIUC, the root cause of this problem is `PreTransformOperator` invokes `getUnionListState` to store persistent schemas, all subTasks of `SchemaOperator`s will obtain the same set of table schemas when restoring from state, and `SchemaCoordinator` is expected to receive $N$ duplicate requests ($N$ = parallelism). Worse still, `UnionListState` will block the checkpointing process when some subTasks have entered the FINISHED state (FLINK-37368).

I wonder if we can handle it in another PR, and focus on modifying the schema evolution request queueing logic here?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
```diff
@@ -73,16 +75,13 @@ public class SchemaCoordinator extends SchemaRegistry {
...
-    private transient RequestStatus schemaChangeStatus;
-
     /** Sink writers which have sent flush success events for the request. */
     private transient ConcurrentHashMap<Integer, Set<Integer>> flushedSinkWriters;
 
-    /** Currently handling request's completable future. */
-    private transient CompletableFuture<CoordinationResponse> pendingResponseFuture;
+    /** Currently handling requests' completable future. */
+    private transient Map<
+                    Integer, Tuple2<SchemaChangeRequest, CompletableFuture<CoordinationResponse>>>
```
lvyanquan, line 83: Can you help add a comment that explains the key here as they are Integer type which cannot directly discern the meaning it represents, and update the comment of ` flushedSinkWriters` too.

lvyanquan (approved):
LGTM.

lvyanquan:
Hi, @leonardBang @ruanhang1993, could you take a look at this?

leonardBang (approved):

nihaoya2025:
@yuxiqian I have a table that I want to delete and resynchronize with both full and incremental data. The schema recorded by the Schema Manager will not be automatically deleted when the table is deleted, skipping the new table creation statement. What should I do
```
 // For redundant schema change events (possibly coming from duplicate emitted
        // CreateTableEvents in snapshot stage), we just skip them.
        if (!SchemaUtils.isSchemaChangeEventRedundant(currentUpstreamSchema, originalEvent)) {
            schemaManager.applyOriginalSchemaChange(originalEvent);
            deducedSchemaChangeEvents.addAll(deduceEvolvedSchemaChanges(originalEvent));
        }
```

PR #3876, opened 2025-01-22
[FLINK-36945] MySQL CDC supports parsing RENAME TABLE Statement with multiple tables

yoheimuta:
This closes [FLINK-36945](https://issues.apache.org/jira/browse/FLINK-36945).

[The Vitess schema migration tool](https://vitess.io/docs/user-guides/schema-changes/ddl-strategies/) renames multiple tables within a single statement to perform schema changes, which are not correctly recognized by MySQL CDC.

The issue arises because the schema change event emitted by Debezium includes multiple table IDs, which MySQL CDC does not expect. As a result, changes for tables specified in the table filter configuration may be unintentionally filtered out. This can prevent schema changes from being saved in the state, which is critical for recovering the job from a snapshot.

This PR updates the source info with the correct table name to align with MySQL CDC expectations, ensuring schema changes are saved correctly.

PR #3650, opened 2024-10-18
[FLINK-36564][ci] Running CI in random timezone to expose more time related bugs.

lvyanquan:
Refer to [this comment](https://github.com/apache/flink-cdc/pull/3648#pullrequestreview-2374769012)
Copy from Paimon repo.

How to verify:
image

PR #3298, opened 2024-05-06
[FLINK-35298][cdc] improve fetch delay metric reporter logic

Shawn-Hx:
- In snapshot phase, set -1 as the value of `currentFetchEventTimeLag` metric.
- Support `currentEmitEventTimeLag` metric.

PR #3917, opened 2025-02-12
[FLINK-37306][tests] Fix MySQLSourceITCase failure in non-UTC timezone

yuxiqian:
This closes FLINK-37306.

The failing test cases were introduced in #3298 and #3876 (and more), but has not been exposed until we decided to run CI test cases in random timezones (#3650).

## `MySQLSourceITCase#testSourceMetrics`
MySQL docker container runs in UTC timezone by default, but `MySQLSourceITCase#testSourceMetrics` didn't specify `serverTimeZone` option explicitly like other test cases. As a result, running test cases in non-UTC timezone will fail with the following error:

> org.apache.flink.table.api.ValidationException: The MySQL server has a timezone offset (0 seconds ahead of UTC) which does not match the configured timezone GMT+XX:XX. Specify the right server-time-zone to avoid inconsistencies for time-related fields.

## `MySqlTimezoneITCase`, `MysqlDebeziumTimeConverterITCase`, and `SpecificStartingOffsetITCase`

These cases need to run MySQL container in specific timezones, so they all have a method called `buildMySqlConfigWithTimezone` to create MySQL container with the same timezone as JVM.

However, JVM timezone (fetched via `ZoneId.systemDefault()`) returns UTC-offset timezone in `GMT+04:30` format, while MySQL configuration file requires it in `+04:30` format. This would block the container from starting up.

## `MySqlMultipleTablesRenamingITCase`

This test case uses the same time-zone initialization code as `MySqlTimezoneITCase`. Removed these codes and stick to UTC considering this case is not time-zone sensitive.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/debezium/converters/MysqlDebeziumTimeConverterITCase.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/debezium/converters/MysqlDebeziumTimeConverterITCase.java
@@ -276,6 +276,11 @@ public class MysqlDebeziumTimeConverterITCase {
     }
 
     private String buildMySqlConfigWithTimezone(String timezone) {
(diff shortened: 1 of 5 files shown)


yuxiqian:
@leonardBang @lvyanquan PTAL

leonardBang (approved):
+1

PR #3920, opened 2025-02-13
[FLINK-37313] Fix the problem of reading binlog before the high and low watermarks during the snapshot process

gongzexin:
This closes [FLINK-37313](https://issues.apache.org/jira/browse/FLINK-37313).

# SnapshotSplitReader#isBackfillRequired
If the binlog has not changed during the snapshot split read, the binlog should not be read.

# MySqlStreamingChangeEventSource#handleChange
Even if you read the binlog of a table you don't care about, you should update the timestamp to avoid MySqlBinlogSplitReadTask#handleEvent reading a large amount of binlog


lvyanquan:
Hi, @gongzexin.

I think that the problem that you met was resolved in https://github.com/apache/flink-cdc/pull/3902 as https://github.com/apache/flink-cdc/pull/3902#issuecomment-2639121416 describes, can you check for this?

gongzexin:
> Hi, @gongzexin.
> 
> I think that the problem that you met was resolved in #3902 as [#3902 (comment)](https://github.com/apache/flink-cdc/pull/3902#issuecomment-2639121416) describes, can you check for this?

Hi,

PR #3923, opened 2025-02-18
[FLINK-37300][common] Add RawDataType to represent the type in data sources

whhe:
This closes [FLINK-37300](https://issues.apache.org/jira/browse/FLINK-37300).

In this PR, I added `RawDataType` and `JdbcRawDataType` to represent the original data type in data sources, and made some changes to the base `DataType` class:
1. Added a new copy method `copy(boolean isNullable, RawDataType rawDataType)`, all external calls to copy method should use it directly or indirectly.
2. Added a new toString method `asSerializableStringWithRawDataType`, `asSummaryString` will change to use it  by default.
3. To avoid incorrect calls to the original `copy(boolean isNullable)` and `asSerializableString` method, I changed them to  `protected`.


whhe:
Many test cases are modified because the newly added `rawDataType` field in DataType class,  and now these test cases generally pass. PTAL @leonardBang @yuxiqian

PR #3926, opened 2025-02-20
[FLINK-37358][kafka] Support to pass format options

Shawn-Hx:
This closes [FLINK-37358](https://issues.apache.org/jira/browse/FLINK-37358).

Currently, we cannot pass format options (e.g. `csv.write-null-properties`) to kafka pipeline sink because format options are not defined in `KafkaDataSinkFactory`.

Fix it by validating options except those with format prefix.


ruanhang1993 (approved):
LGTM.

PR #3922, opened 2025-02-16
[FLINK-37332] Support any column as chunk key column (postgres, orcale, db2, sqlserver)

SML0127:
MySQL CDC connector supports this feature from flink cdc v3.2
Now, other connectors that support incremental snapshot allow any columns as chunk key column

apache jira: https://issues.apache.org/jira/browse/FLINK-37332

PR #3924, opened 2025-02-19
[FLINK-37347][DOCS] Support any column as chunk key column (postgres, orcale, db2, sqlserver)

SML0127:
updates documentation for https://github.com/apache/flink-cdc/pull/3922

PR #3928, opened 2025-02-21
[FLINK-37332] Support any column as chunk key column (postgres, orcale, db2, sqlserver) #3922

SML0127:
This close https://github.com/apache/flink-cdc/pull/3922.
MySQL CDC connector supports this feature from flink cdc v3.2
Now, other connectors that support incremental snapshot allow any columns as chunk key column

apache jira: https://issues.apache.org/jira/browse/FLINK-37332
pr for docs: https://github.com/apache/flink-cdc/pull/3924

cc. @gtk96 @leonardBang @lvyanquan

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/test/java/org/apache/flink/cdc/connectors/kafka/json/debezium/DebeziumJsonSerializationSchemaTest.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/test/java/org/apache/flink/cdc/connectors/kafka/json/debezium/DebeziumJsonSerializationSchemaTest.java
@@ -40,7 +40,6 @@ import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.core.JsonGenerator
 import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.JsonNode;
 import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.ObjectMapper;
 
-import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import java.math.BigDecimal;
@@ -48,6 +47,8 @@ import java.time.Instant;
 import java.time.ZoneId;
 import java.util.HashMap;
 import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
 
 /** Tests for {@link DebeziumJsonSerializationSchema}. */
 class DebeziumJsonSerializationSchemaTest {
@@ -74,7 +75,7 @@ class DebeziumJsonSerializationSchemaTest {
                         .primaryKey("col1")
                         .build();
         CreateTableEvent createTableEvent = new CreateTableEvent(TABLE_1, schema);
-        Assertions.assertThat(serializationSchema.serialize(createTableEvent)).isNull();
+        assertThat(serializationSchema.serialize(createTableEvent)).isNull();
         BinaryRecordDataGenerator generator =
                 new BinaryRecordDataGenerator(RowType.of(DataTypes.STRING(), DataTypes.STRING()));
         // insert
@@ -90,7 +91,7 @@ class DebeziumJsonSerializationSchemaTest {
                 mapper.readTree(
                         "{\"before\":null,\"after\":{\"col1\":\"1\",\"col2\":\"1\"},\"op\":\"c\",\"source\":{\"db\":\"default_schema\",\"table\":\"table1\"}}");
         JsonNode actual = mapper.readTree(serializationSchema.serialize(insertEvent1));
-        Assertions.assertThat(actual).isEqualTo(expected);
+        assertThat(actual).isEqualTo(expected);
         DataChangeEvent insertEvent2 =
                 DataChangeEvent.insertEvent(
                         TABLE_1,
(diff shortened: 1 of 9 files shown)


SML0127:
@gtk96 @leonardBang @lvyanquan PTAL.

gtk96:
@GOODBOY008 @Jiabao-Sun PTAL

lvyanquan (approved):
LGTM.

SML0127:
Hi, @gong 
May I ask for your help with the above error?

leonardBang:
@SML0127 Could you rebase to latest master branch which has fixed some unstable mysql case? but for oracle failed case, I think we can file an issue to track.

ruanhang1993:
@SML0127 Thanks for your contribution.

Using a column which is not a primary key column as the chunk key may cause data error. 
Consider a table whose primary key is `id` and chunk key column is `pid`. And there are 2 snapshot splits.
- Split0: 1< pid <= 3
- Split1: 3< pid <= 5

Split 0 and Split 1 are read by different subtasks. An update change happens when reading these snapshot splits and its binlog offset is between the low and high watermark of both splits.
dad

This update will be handled in the backfill task of both split reader and it will not be sent in the incremental phase.
Split0 will have the record [id=0, pid=2]. Split1 will have the record [id=0, pid=4]. 
We cannot control the order of these records with the same id. The final pid of id=0 will be 2 or 4, which depends on the order of records.

Could we add some warnings in docs for this usage in mysql and other connectors ?

SML0127:
@ruanhang1993 Thanks for your kind explanation. 
I understand the situation. As you suggested, I will add warings and some examples to the docs.

I have a  minor question, if we set `scan.incremental.snapshot.backfill.skip` to `true` and using pk as chunk key in the incremental phase, will there be no issue? 
(I means that, first we use a column not in primary key as chunk key in snapshot phase. And then use primary key as chunk key in incremental phase)

yuxiqian:
Hi @SML0127, could you please rebase this PR with latest `master` when available?

* JUnit 4 style test annotations need to be ported to JUnit 5 equivalents

  * `org.junit.Assert`
  * `org.hamcrest.*`

ruanhang1993:
> @ruanhang1993 Thanks for your kind explanation. I understand the situation. As you suggested, I will add warings and some examples to the docs.
> 
> I have a minor question, if we set `scan.incremental.snapshot.backfill.skip` to `true` and using pk as chunk key in the incremental phase, will there be no issue? (I means that, first we use a column not in primary key as chunk key in snapshot phase. And then use primary key as chunk key in incremental phase)

@SML0127 I am not sure whether I understand your question right.

If we set `scan.incremental.snapshot.backfill.skip` to `true`, the source will have the at-least-once semantics. The update (id =0 , pid from 2 to 4) will be sent in incremental phase. 
- If the sink could deduplicate based on the primary key, the final result will be right(For example, Starrocks). 
- If the sink could not deduplicate based on the primary key, the final result could be wrong. 
  - For example, Kafka may contain the change history as +I(id=0,pid=4),+I(id=0,pid=2),-U(id=0,pid=2),+U(id=0,pid=4). I don't think we should treat it as a right result.

SML0127:
@ruanhang1993 Thank you for your response. I asked because I was wondering if there would be no issue using upsert to the target, as you mentioned.

SML0127:
I added the following warning and problem scenario in docs (PR: https://github.com/apache/flink-cdc/pull/3924)

image

SML0127:
All changes have been applied.

ruanhang1993 (approved):
LGTM. Please resolve the conflicts. @SML0127

SML0127:
@ruanhang1993
Resolved conflicts. thank you!

joyCurry30:
Hi, the CI/CD pipeline failed. Could you take a look?

SML0127:
Hi @joyCurry30
There are 3 failures. 
In the mongo7, postgres case, it is shown as a timeout rather than a failure. 
In the case of mongo6, it is successful when testing locally on my mac.
(This PR did not edit any mongodb or related code🥲)

ruanhang1993 (approved):

SML0127:
conflicts resolved

SML0127:
Hi, @leonardBang @ruanhang1993 
Sorry for the late check.
I saw the under error in CI.
```d
Caused by: org.apache.flink.table.api.ValidationException: Chunk key column 'name' doesn't exist in the columns [Id,Name,address,phone_number] of the table customer.Customers.
```
Among the engines that support incremental snapshots, I know of none that are case sensitive unless enclosed in double quotes. So what do you think about the following change in `JdbcChunkUtils.getSplitColumn`?

AS-IS
```java
List<Column> searchColumns = table.columns();
        if (chunkKeyColumn != null) {
            Optional<Column> targetPkColumn =
                    searchColumns.stream()
                            .filter(col -> chunkKeyColumn.equals(col.name()))
                            .findFirst();
```
TO-BE
```java
List<Column> searchColumns = table.columns();
        if (chunkKeyColumn != null) {
            Optional<Column> targetPkColumn =
                    searchColumns.stream()
                            .filter(col -> chunkKeyColumn.equalsIgnoreCase(col.name()))  // changed
                            .findFirst();
```

leonardBang (approved):
Thanks @SML0127 for the update, LGTM

xxntti3n:
# PostgreSQL Incremental Snapshot Configuration Error

## Paraphrased Error Message
When using an incremental snapshot on a PostgreSQL table that *does not have primary keys*
This is my config:  
`scan.incremental.snapshot.chunk.key-column=public.dag_run:updated_at`

Otherwise, the job raise the error:

> "To use incremental snapshot, 'scan.incremental.snapshot.chunk.key-column' must be set when the table doesn't have primary keys."

PR #3929, opened 2025-02-22
[FLINK-37371] Postgres CDC incremental source fails to handle upper-case table and column names

morozov:
## Problem summary

There are two problems in the code:
1. The table name passed to `to_regclass()` isn't quoted, so the upper-case letters get lower-cased.
2. The column names used in snapshot queries aren't quoted, so the upper-case letters get lower-cased.

There's also a minor issue with the `quote()` implementation that it doesn't double the double quotes. So if the table name or column name contains double quote, it will result in a SQL syntax error.

## Connector code changes

1. Instead of quoting the table name passed to `to_regclass()`, pass the schema name and table name individually. This way, they are interpreted as literals don't get automatically lower-cased.
2. Use prepared statement parameters in the query against `pg_class`. Otherwise, the table and schema name would have to be additionally quoted.
3. Quote column names used in snapshot queries.
4. Replace double quote with two double quotes in `quote()` per SQL standard.

## Test changes

I don't think that handling upper-case name should be covered by a special test case. Ideally, it should be tested exhaustively, so I decided to modify an existing test. Prior to that:

1. Building SQL queries in tests has been refactored to enable using upper-case name. The new `TableId` class represents a qualified table name (schema name + table name). Its `toSql()` method generates a valid SQL name that can be used to build queries.
2. `PostgresSourceITCase` has been cleaned up to eliminate some warnings (e.g. raw usage of parameterized types or non-closed `Closable`).
3. Tests have been updated to use upper-case table and column names.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/utils/PostgresQueryUtils.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/utils/PostgresQueryUtils.java
@@ -69,12 +69,17 @@ public class PostgresQueryUtils {
         // https://stackoverflow.com/questions/7943233/fast-way-to-discover-the-row-count-of-a-table-in-postgresql
         // NOTE: it requires ANALYZE or VACUUM to be run first in PostgreSQL.
         final String query =
-                String.format(
-                        "SELECT reltuples::bigint FROM pg_class WHERE oid = to_regclass('%s')",
-                        tableId.toString());
-
-        return jdbc.queryAndMap(
+                "SELECT reltuples::bigint"
+                        + " FROM pg_class c"
+                        + " JOIN pg_namespace n ON n.oid = c.relnamespace"
+                        + " WHERE n.nspname = ?"
+                        + " AND c.relname = ?";
+        return jdbc.prepareQueryAndMap(
                 query,
+                ps -> {
+                    ps.setString(1, tableId.schema());
+                    ps.setString(2, tableId.table());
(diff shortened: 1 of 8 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/utils/PostgresQueryUtils.java
```diff
@@ -69,12 +69,17 @@ public static long queryApproximateRowCnt(JdbcConnection jdbc, TableId tableId)
...
         // NOTE: it requires ANALYZE or VACUUM to be run first in PostgreSQL.
         final String query =
-                String.format(
-                        "SELECT reltuples::bigint FROM pg_class WHERE oid = to_regclass('%s')",
-                        tableId.toString());
-
-        return jdbc.queryAndMap(
+                "SELECT reltuples::bigint"
+                        + " FROM pg_class c"
+                        + " JOIN pg_namespace n ON n.oid = c.relnamespace"
```
loserwang1024, line 74: @morozov , what is the benefit of this modify?

yuxiqian:
Hi @morozov, thanks for your contribution! Could you please rebase this PR with latest `master` when available?

morozov:
> Could you please rebase this PR with latest `master` when available?

Done. FWIW, it took a bit effort to update the new tests to match the changes that I'd made to the test data set. It would be nice to get this merged sooner than later and avoid conflicts in the future.

loserwang1024:
@leonardBang , CC

loserwang1024 (approved):

leonardBang (approved):
Thanks @morozov for the contribution and @loserwang1024 for the review,+1

PR #3932, opened 2025-02-26
[FLINK-37386] Emit CreateTableEvent only when met the related SourceRecord.

lvyanquan:
Emit CreateTableEvent only when met the related SourceRecord to avoid downstream backpressure.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
@@ -110,8 +110,8 @@ public class MySqlPipelineRecordEmitter extends MySqlRecordEmitter<Event> {
                 // New created table in binlog reading phase.
                 if (createTableEvent != null) {
                     output.collect(createTableEvent);
-                    alreadySendCreateTableTables.add(tableId);
                 }
+                alreadySendCreateTableTables.add(tableId);
             }
         }
         super.processElement(element, output, splitState);
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
@@ -60,7 +60,10 @@ import static io.debezium.connector.AbstractSourceInfo.DATABASE_NAME_KEY;
 import static io.debezium.connector.AbstractSourceInfo.SCHEMA_NAME_KEY;
 import static io.debezium.connector.AbstractSourceInfo.TABLE_NAME_KEY;
 import static org.apache.flink.cdc.connectors.mysql.debezium.DebeziumUtils.openJdbcConnection;
+import static org.apache.flink.cdc.connectors.mysql.source.utils.RecordUtils.getTableId;
+import static org.apache.flink.cdc.connectors.mysql.source.utils.RecordUtils.isDataChangeRecord;
 import static org.apache.flink.cdc.connectors.mysql.source.utils.RecordUtils.isLowWatermarkEvent;
+import static org.apache.flink.cdc.connectors.mysql.source.utils.RecordUtils.isSchemaChangeEvent;
 import static org.apache.flink.cdc.connectors.mysql.source.utils.TableDiscoveryUtils.listTables;
 
 /** The {@link RecordEmitter} implementation for pipeline mysql connector. */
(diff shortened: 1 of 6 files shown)


lvyanquan:
Hi @yuxiqian.

Could you help to review this?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
```diff
@@ -99,23 +100,17 @@ protected void processElement(
...
-                                        !alreadySendCreateTableTables.contains(
-                                                MySqlSchemaUtils.toDbzTableId(event.tableId())))
-                        .forEach(output::collect);
-            } else {
-                // In Binlog only mode, we simply emit all schemas at once.
-                createTableEventCache.forEach(output::collect);
+        } else {
+            if (isDataChangeRecord(element) || isSchemaChangeEvent(element)) {
+                TableId tableId = getTableId(element);
+                if (!alreadySendCreateTableTables.contains(tableId)) {
```
yuxiqian, line 106: Seems `alreadySendCreateTableTables` and `createTableEventCache` are not being stored into MySQL source state persistently. Will that block users from recovering from an earlier binlog state where schemas are different from current state?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
```diff
@@ -99,23 +100,17 @@ protected void processElement(
...
+                    CreateTableEvent createTableEvent = createTableEventCache.get(tableId);
+                    // New created table in binlog reading phase.
```
yuxiqian, line 108:
```
--- Create Table A --> Table A Binlog --> Drop Table A --> Startup Offset --->

                    ^
                    |
       What if we start from here?
```

lvyanquan: This is indeed a problem, but it is different from the one I hope to solve here.

For DataChange, perhaps we can derive the schema from SourceRecord instead of using SQL to query the latest schema. But to make this problem more difficult, if we start reading from a position where SchemaChange happened, we cannot derive the original schema.
```
--- Create Table A --> Alter Table A Add a column--> Table A Binlog --> Startup Offset --->

                     ^
                     |
        What if we start from here?
```         

So I think starting from a position where historical schema is different with the current schema is still an unresolved issue for us.

lvyanquan:
The issue is related to https://github.com/apache/flink-cdc/pull/3912#issuecomment-2680469662,CC @gongzexin as you may be interested about it.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlPipelineRecordEmitter.java
yuxiqian: Thanks for clarifying this. Should we throw an exception for such dangling schema change events here?

flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/MysqlE2eITCase.java
```diff
@@ -545,9 +545,7 @@ public void testDanglingDropTableEventInBinlog() throws Exception {
         Path mysqlDriverJar = TestUtils.getResource("mysql-driver.jar");
         submitPipelineJob(pipelineJob, mysqlCdcJar, valuesCdcJar, mysqlDriverJar);
         waitUntilJobRunning(Duration.ofSeconds(30));
-        validateResult(
-                "CreateTableEvent{tableId=%s.customers, schema=columns={`id` INT NOT NULL,`name` VARCHAR(255) NOT NULL 'flink',`address` VARCHAR(1024),`phone_number` VARCHAR(512)}, primaryKeys=id, options=()}",
-                "CreateTableEvent{tableId=%s.products, schema=columns={`id` INT NOT NULL,`name` VARCHAR(255) NOT NULL 'flink',`description` VARCHAR(512),`weight` FLOAT,`enum_c` STRING 'red',`json_c` STRING,`point_c` STRING}, primaryKeys=id, options=()}");
+        validateResult("");
```
yuxiqian, line 548: `validateResult("")` doesn't check anything...

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PreTransformOperator.java
```diff
@@ -270,6 +273,24 @@ public void close() throws Exception {
     @Override
     public void processElement(StreamRecord<Event> element) throws Exception {
         Event event = element.getValue();
+        if (event instanceof ChangeEvent) {
```
yuxiqian, line 276: If I understand correctly, after this change, MySQL source will always emit `CreateTableEvent`  based on current tables' schema, before sending any `DataChangeEvent`s.

As `event instanceof ChangeEvent` (L294) is `true`, we will emit a `CreateTableEvent` restored from `PreTransformOp`'s state.

But the following `event instanceof CreateTableEvent` is also `true`. Are we accidentally emitting duplicate and inconsistent `CreateTableEvent`s to the downstream?

yuxiqian (commented):
Thanks for @lvyanquan's great work, just left some minor comments.

flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/MysqlE2eITCase.java
lvyanquan: Addressed.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PreTransformOperator.java
lvyanquan: > Are we accidentally emitting duplicate and inconsistent CreateTableEvents to the downstream

Yes. Updated and fixed it.

yuxiqian (approved):
LGTM. Would @leonardBang like to take a look?

leonardBang (approved):
+1

PR #3936, opened 2025-03-04
[hotfix][tests] Fix MySQL test case failure in exact UTC timezone

yuxiqian:
As noted in https://github.com/apache/flink-cdc/actions/runs/13623293435/job/38139886612?pr=3912, if JVM timezone is randomly set to GMT+0:00, it will be "smartly" converted to `"GMT"` (w/o offset part). As a result, we're configuring MySQL to start up with an empty timezone string, causing the CI failure.

Extra checking has been added to cover such case.

### DIFF
--- a/.github/workflows/flink_cdc_base.yml
+++ b/.github/workflows/flink_cdc_base.yml
@@ -149,7 +149,7 @@ jobs:
           maven-version: 3.8.6
 
       - name: Compile and test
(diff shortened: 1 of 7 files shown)


leonardBang (approved):

leonardBang:
CI passed finally, merging...

PR #3939, opened 2025-03-06
[FLINK-37429][transform] Map each column name to a new name in generated expression

Shawn-Hx:
This close [FLINK-37429](https://issues.apache.org/jira/browse/FLINK-37429), [FLINK-37326](https://issues.apache.org/jira/browse/FLINK-37326)

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/main/java/org/apache/flink/cdc/connectors/values/source/TimestampTypeMetadataColumn.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/main/java/org/apache/flink/cdc/connectors/values/source/TimestampTypeMetadataColumn.java
@@ -23,7 +23,7 @@ import org.apache.flink.cdc.common.types.DataTypes;
 
 import java.util.Map;
 
-/** A {@link SupportedMetadataColumn} for op_ts. */
+/** A {@link SupportedMetadataColumn} for timestamp-type. */
 public class TimestampTypeMetadataColumn implements SupportedMetadataColumn {
 
     @Override
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/main/java/org/apache/flink/cdc/connectors/values/source/TimestampTypeMetadataColumn.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/main/java/org/apache/flink/cdc/connectors/values/source/TimestampTypeMetadataColumn.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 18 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
```diff
@@ -95,6 +95,8 @@ public class TransformParser {
     private static final Logger LOG = LoggerFactory.getLogger(TransformParser.class);
     private static final String DEFAULT_SCHEMA = "default_schema";
     private static final String DEFAULT_TABLE = "TB";
+    private static final String MAPPED_COLUMN_NAME_PREFIX = "field";
```
yuxiqian, line 98: Just thinking if we can use RegExp-like names (`$1`, `$2`, ...)? Ignore this comment if it doesn't look right.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
yuxiqian, line 1: Seems this fix should close FLINK-37326, too. Adding a test case to verify this would be nice.

yuxiqian (commented):
Thanks for Shawn's quick fix! Just left some minor comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformFilter.java
```diff
@@ -41,11 +42,17 @@ public class TransformFilter implements Serializable {
...
     private final String scriptExpression;
     private final List<String> columnNames;
+    private final Map<String, String> columnNameMap;
 
-    public TransformFilter(String expression, String scriptExpression, List<String> columnNames) {
+    public TransformFilter(
+            String expression,
+            String scriptExpression,
+            List<String> columnNames,
+            Map<String, String> columnNameMap) {
```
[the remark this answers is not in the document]
Shawn-Hx: Every `TransformExpressionKey` has its own `columnNameMap`. If the expression script is not changed, the `columnNameMap` will not change. I have added a test for the schema change scenario.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
```diff
@@ -642,4 +654,16 @@ private static boolean hasAsterisk(SqlNode sqlNode) {
...
+        int i = 0;
+        Map<String, String> columnNameMap = new HashMap<>();
+        for (String columnName : originalColumnNames) {
+            if (!columnNameMap.containsKey(columnName)) {
+                columnNameMap.put(columnName, MAPPED_COLUMN_NAME_PREFIX + i);
+                i++;
+            }
+        }
+        return columnNameMap;
+    }
```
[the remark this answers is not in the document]
Shawn-Hx: I prefer option 3.
For option 1, seems it it not very easy to judge a column name is legal or not considering every java keyword is illegal.
For option 2, a corner case is there are two columns named `a-b` and `a.b` and we still cannot recognize them after mapping.
Have added column name map information to the exception log.

yuxiqian (approved):
Looks good! Covering this case in `FlinkPipelineTransformITCase` would be nice, but it's not a must for now.

leonardBang (approved):
Thanks @Shawn-Hx for the contribution, LGTM

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-values/src/main/java/org/apache/flink/cdc/connectors/values/source/TimestampTypeMetadataColumn.java
```diff
@@ -0,0 +1,48 @@
...
+
+package org.apache.flink.cdc.connectors.values.source;
+
+import org.apache.flink.cdc.common.source.SupportedMetadataColumn;
+import org.apache.flink.cdc.common.types.DataType;
+import org.apache.flink.cdc.common.types.DataTypes;
+
+import java.util.Map;
+
+/** A {@link SupportedMetadataColumn} for op_ts. */
```
leonardBang, line 26:
```suggestion
/** A {@link SupportedMetadataColumn} for timestamp-type. */
```

PR #3943, opened 2025-03-11
[minor] Correct variables name according to option name.

lvyanquan:
Update variables name that related to `scan.incremental.snapshot.unbounded-chunk-first.enabled` option.


lvyanquan:
@leonardBang  PTAL.

leonardBang (approved):
+1, wait the ci pass

lvyanquan:
This failed case should be an unstable case, as it succeeded in the 1.20.1 run and I created [an issue](https://issues.apache.org/jira/browse/FLINK-37461) to track this issue.

PR #3948, opened 2025-03-13
[FLINK-36784][docs] Add op_ts metadata docs in MySQL pipeline connector

ruanhang1993:
Add op_ts metadata docs in MySQL pipeline connector


leonardBang (approved):
+1, nice catch

PR #3952, opened 2025-03-17
[FLINK-35459] Use Incremental Source Framework in Flink tidb CDC Source Connector

Mrart:
Use Incremental Source Framework in Flink CDC TiKV Source Connector


Mrart:
@lvyanquan Please help reopen this pr，I will completed it。

Mrart:
@yuxiqian @lvyanquan  Can this pr  merge into 3.7?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/fetch/TiDBScanFetchTask.java
```diff
@@ -0,0 +1,318 @@
...
+        TiDBSourceFetchTaskContext ctx = (TiDBSourceFetchTaskContext) context;
+        final EventOffset currentOffset =
+                EventOffset.of(
+                        ((TiDBSourceFetchTaskContext) context).getOffsetContext().getOffset());
+        JdbcSourceEventDispatcher dispatcher = ctx.getEventDispatcher();
+        dispatcher.dispatchWatermarkEvent(
+                ctx.getPartition().getSourcePartition(),
+                backfillStreamSplit,
+                currentOffset,
+                WatermarkKind.END);
```
leonardBang, line 78: **Blocking: the backfill read is skipped entirely.** `executeBackfillTask()` never consumes `backfillStreamSplit` — it just takes the current offset and dispatches the END watermark, which discards the whole [LOW, HIGH) interval that the incremental framework relies on for snapshot/stream consistency.

Concretely: a row read as `v1` during the snapshot and updated to `v2` before HIGH will stay `v1` forever, since streaming starts after the boundary. Unless the snapshot reads are proven TSO-consistent at HIGH (which would need to be made explicit), this breaks exactly-once semantics — and the `// just for test` comment suggests this is a known placeholder.

Could you either implement the backfill consumption (via the TiKV CDC reader over the same key range), or document + prove why it's unnecessary for TiDB? Also worth removing the raw `JdbcSourceEventDispatcher` usage in favor of the typed one while here.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/fetch/EventSourceReader.java
```diff
@@ -0,0 +1,499 @@
...
+        if (tableId == null) {
+            LOG.warn("No valid tableId found, skipping log message: {}", row);
+            return;
+        }
+        TableSchema tableSchema = taskContext.getDatabaseSchema().schemaFor(tableId);
+        if (tableSchema == null) {
+            LOG.warn("No table schema found, skipping log message: {}", row);
+            return;
+        }
+        offsetContext.event(tableSchema.id(), Instant.ofEpochMilli(row.getCommitTs()));
```
leonardBang, line 247: **Blocking: TSO vs epoch-millis unit mixing.** `row.getCommitTs()` is a packed TiKV TSO (physical << 18 | logical), not wall-clock milliseconds — the rest of this PR treats it as a TSO (e.g. `EventOffset.getStartTs`, the `TiTimestamp` conversions). Passing a packed TSO into `Instant.ofEpochMilli` yields nonsensical source timestamps that end up in `sourceInfo` and downstream metadata columns.

Also note `EventOffsetContext.event()` only touches `sourceInfo` — it never advances `timestamp`/`commitVersion`, so the checkpointed offset doesn't reliably track the last emitted record; on restore we could replay from a stale offset and re-emit a large window.

I'd suggest keeping the two representations strictly separated: exact `commitVersion` (TSO) for offsets, decoded physical time for display timestamps — and adding a checkpoint → kill → restore ITCase asserting no pre-checkpoint records are re-emitted.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/fetch/EventSourceReader.java
```diff
@@ -0,0 +1,499 @@
...
+                            emitChangeEvent(partition, offsetContext, committedRow);
+                            // use startTs of row as messageTs, use commitTs of row as fetchTs
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
+                            break;
+                        } catch (Exception e) {
+                            if (running && context.isRunning()) {
+                                LOG.error("Read change events error.", e);
+                            }
+                        }
```
leonardBang, line 213: **Blocking: swallowed emission exceptions = silent data loss.** Once `committedEvents.take()` succeeds, the record is dequeued; if `emitChangeEvent` then throws, the catch-all logs and continues — that CDC event is gone permanently while the job stays RUNNING. This PR adds several new conversion paths (RowV2 decoding, value converters, schema lookups) where such failures are realistic.

An emission failure should be fatal: propagate it via the error handler / fail the source so Flink restores from the last successful checkpoint, instead of converting data loss into a log line.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/config/TiDBSourceConfig.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.Properties;
+
+/** The configuration for TiDB source. */
+public class TiDBSourceConfig extends JdbcSourceConfig {
+    private static final long serialVersionUID = 1L;
+    private final String compatibleMode;
+    private final String pdAddresses;
+
+    private final String hostMapping;
+    private TiConfiguration tiConfiguration;
```
leonardBang, line 39: **Blocking: non-serializable field on a serializable config.** `org.tikv.common.TiConfiguration` is not `Serializable` (and not marked `transient` here), but this config crosses the source serialization boundary — it's embedded in the source/fetch-task objects that go JM → TM and into checkpoints. This will blow up with `NotSerializableException` at submission or restore.

Better to store only serializable primitives here (pd addresses, host mapping, TiKV properties, timeouts) and construct `TiConfiguration` lazily on the TM side. A small `InstantiationUtil.serializeObject(config)` unit test would lock this down.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/fetch/EventSourceReader.java
```diff
@@ -0,0 +1,499 @@
...
+                return null;
+            }
+            for (int index : fieldIndex) {
+                serializableObject[index] = tiKVValueAfter[index];
+            }
+            return serializableObject;
+        } catch (Exception e) {
+            LOG.error("decode object error", e);
+            return null;
+        }
```
leonardBang, line 313: **Projection column indexing mixes two index spaces.** `serializableObject` is sized by `fieldIndex.size()`, but the copy loop uses `tiKVValueAfter[index]` where the decoded values are in TiKV physical column order. For projections / non-contiguous column sets these index spaces diverge: either AIOOBE, or silently reading the wrong physical column into the wrong schema slot — and the catch-all at the bottom converts any failure into `return null`, corrupting before/after images silently (see also the swallowed-exceptions comment above).

Could you add an explicit physical-offset → schema-position mapping here, plus a unit test with a projected/non-contiguous column subset?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/fetch/EventSourceReader.java
```diff
@@ -0,0 +1,499 @@
...
+                                "Table %s.%s does not exist.", tableId.catalog(), tableId.table()));
+            }
+            keyRange = TableKeyRangeUtils.getTableKeyRange(tableInfo.getId(), 1, 0);
+            cdcClient = new CDCClient(session, keyRange);
+            prewrites = new TreeMap<>();
+            commits = new TreeMap<>();
+            // cdc event will lose if pull cdc event block when region split
+            // use queue to separate read and write to ensure pull event unblock.
+            // since sink jdbc is slow, 5000W queue size may be safe size.
+            committedEvents = new LinkedBlockingQueue<>();
```
leonardBang, line 145: **Unbounded queue defeats backpressure.** The comment above says "5000W queue size may be safe size", but the `LinkedBlockingQueue` is constructed without a capacity, so nothing enforces it. Under sink backpressure the CDC pull thread keeps ingesting from TiKV and the heap grows without bound → eventual OOM.

A bounded capacity (config-driven) would turn this into normal Flink backpressure instead.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/fetch/EventSourceReader.java
```diff
@@ -0,0 +1,499 @@
...
+        StreamingChangeEventSource.super.init();
+        try {
+            session = TiSession.create(ticonf);
+            Set<TableId> tableIds = this.split.getTableSchemas().keySet();
+            if (tableIds.size() != 1) {
+                throw new IllegalStateException(
+                        "Currently only single table ingest is supported, but found "
+                                + tableIds.size()
+                                + " tables.");
+            }
```
leonardBang, line 130: **Multi-table jobs fail late.** The dialect/config accept multi-table table-lists, and snapshotting works for them — but this check throws at runtime when entering the streaming phase, so such a job would snapshot for hours on a large table and only then die. Either support multi-table ranges in the stream reader, or fail fast at configuration/validation time with a documented limitation.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/io/debezium/connector/tidb/TiDBPartition.java
```diff
@@ -0,0 +1,60 @@
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        final io.debezium.connector.tidb.TiDBPartition other =
+                (io.debezium.connector.tidb.TiDBPartition) obj;
+        return Objects.equals(serverName, other.serverName);
```
leonardBang, line 53: `equals()` is overridden without `hashCode()` — this violates the equals/hashCode contract, and Debezium uses partitions as source identities in maps/sets (offset lookups etc.), where an identity hashCode breaks equal-partition lookups. Also worth giving this a meaningful `toString()` (currently `super.toString()`) — partitions show up in logs and error messages all the time.

leonardBang (commented):
Thanks @Mrart for pushing this migration forward — moving the TiDB connector onto the incremental source framework is definitely the right direction, and the effort in the offset model (`EventOffset` with timestamp + commit_version) and the TiDB-specific value converters shows. I went through the head of your branch (`ac10cc1`; note the PR page seems to serve a cached older view, so some earlier Copilot findings like the chunk-key validation and the retry loop are already fixed — nice).

I left detailed inline comments. The blocking ones from my side:

1. **Snapshot backfill is skipped entirely** in `TiDBScanFetchTask.executeBackfillTask()` — this breaks the snapshot-to-stream consistency guarantee of the incremental framework.
2. **TSO vs epoch-millis unit mixing** in offset/timestamp handling (`Instant.ofEpochMilli(commitTs)` and friends).
3. **Swallowed emission exceptions** in the consumer loop → silent data loss while the job stays RUNNING.
4. **`TiDBSourceConfig` holds a non-serializable `TiConfiguration`** → will fail at job submission/restore.

A few more substantial items are inline (projection column indexing, unbounded queue, multi-table fail-late, `hashCode()` contract, case-sensitivity polarity, offset load/store asymmetry, null prewrite handling).

Lastly, ~9.6k lines across 83 files is very hard to review as one unit — if there's a way to split this into stacked PRs (e.g. Debezium integration layer → offset/schema → fetch tasks → table API + tests), it would significantly speed up the review cycle. I'd also really like to see failure-oriented ITCases: mutations during snapshot scanning, transactions spanning LOW/HIGH, checkpoint → kill → restore, projection columns, and source-config serialization.

Happy to help re-review once the blockers are addressed — thanks again for taking on this migration!

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/fetch/TiDBSourceFetchTaskContext.java
```diff
@@ -0,0 +1,238 @@
...
+        super(sourceConfig, dataSourceDialect);
+        this.connection = connection;
+        this.metadataProvider = new TiDBEventMetadataProvider();
+    }
+
+    @Override
+    public void configure(SourceSplitBase sourceSplitBase) {
+        final TiDBConnectorConfig connectorConfig = getDbzConnectorConfig();
+        final boolean tableIdCaseInsensitive =
+                dataSourceDialect.isDataCollectionIdCaseSensitive(sourceConfig);
```
leonardBang, line 89: **Polarity check please.** The result of `isDataCollectionIdCaseSensitive(...)` is stored into a variable named `tableIdCaseInsensitive` and passed on to `TiDBUtils.newSchema(...)`. The dialect side looks correct in this push, but if `newSchema` really expects a case-*insensitive* flag, the inversion has just moved to the call site — same end behavior. Could you double-check which polarity `newSchema` expects and rename accordingly?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/offset/EventOffsetContext.java
```diff
@@ -0,0 +1,212 @@
...
+                            snapshotCompleted,
+                            TransactionContext.load(offset),
+                            incrementalSnapshotContext,
+                            new TiDBSourceInfo(connectorConfig));
+            String timestamp = (String) offset.get(TIMESTAMP_KEY);
+            offsetContext.setCheckpoint(
+                    timestamp == null
+                            ? Instant.now()
+                            : Instant.ofEpochMilli(Long.parseLong(timestamp)),
+                    (String) offset.get(COMMIT_VERSION_KEY));
```
leonardBang, line 208: **Offset load/store asymmetry.** A missing timestamp silently falls back to `Instant.now()`, so the restored position depends on recovery wall-clock time rather than persisted state — and `getOffset()` never writes the `snapshot_completed` key that this loader reads (L187-189). Stored and loaded fields should be strictly symmetric, and a malformed offset should fail deterministically instead of jumping to "now".

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/fetch/EventSourceReader.java
```diff
@@ -0,0 +1,499 @@
...
+
+    protected void flushRows(final long timestamp) throws Exception {
+        Preconditions.checkState(context != null, "sourceContext shouldn't be null");
+        synchronized (context) {
+            while (!commits.isEmpty() && commits.firstKey().timestamp <= timestamp) {
+                final Cdcpb.Event.Row commitRow = commits.pollFirstEntry().getValue();
+                final Cdcpb.Event.Row prewriteRow =
+                        prewrites.remove(RowKeyWithTs.ofStart(commitRow));
+                // if pull cdc event block when region split, cdc event will lose.
+                committedEvents.offer(prewriteRow);
```
leonardBang, line 365: **`prewrites.remove(...)` can return `null`** — reconnects, region transitions, or a COMMIT arriving without its PREWRITE — and `LinkedBlockingQueue.offer(null)` throws NPE, which would kill the reader inside `flushRows`. Worth defining the expected PREWRITE/COMMIT guarantees across reconnects (the comment above already hints events can be lost during region splits) and handling the missing-prewrite case explicitly (skip + warn, or fail, but not NPE).

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-tidb-cdc/src/main/java/org/apache/flink/cdc/connectors/tidb/source/offset/EventOffset.java
```diff
@@ -0,0 +1,122 @@
...
+        if (a == null && b == null) {
+            return 0;
+        }
+        if (a == null) {
+            return -1;
+        }
+        if (b == null) {
+            return 1;
+        }
+        return Long.compare(Long.parseLong(a), Long.parseLong(b));
```
leonardBang, line 102: `Long.parseLong` here (and in `getCommitVersion()`/`getStartTs()`) throws `NumberFormatException` on malformed or legacy offset strings — `compareTo` is used for split/watermark ordering, so a bad offset blows up deep inside the framework rather than failing at the parse boundary with a clear message. Related unit-mixing: `getCommitVersion()` (L75) feeds an epoch-millis timestamp into `new TiTimestamp(...)`, which is the same TSO-vs-millis conflation as in the reader.

PR #3959, opened 2025-03-19
[minor][pipeline-connectors][doris]Fix deprecated method usage in DorisSchemaChangeManager

jinkachy:
Fix deprecated method usage in DorisSchemaChangeManager

```
    org.apache.doris.flink.catalog.doris.DorisSystem#identifier
    /** @deprecated */
    @Deprecated
    public static String identifier(String name) {
        return DorisSchemaFactory.identifier(name);
    }


yuxiqian (approved):
Thanks for @jinkachy's nice catch!

PR #3960, opened 2025-03-19
[FLINK-37506][transform] Print more descriptive error messages when TransformOperator throws an exception

yuxiqian:
This closes FLINK-37506.

Currently, there's no much contextual information available (like currently handling Event, its tableId and schema, or BinaryRecordData) when transform operator throws an exception, making it very difficult to locating and debugging. Providing extra information would be useful for troubleshooting a crashed pipeline job.

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/event/DataChangeEvent.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/event/DataChangeEvent.java
@@ -24,6 +24,7 @@ import java.io.Serializable;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Function;
 
 /**
  * Class {@code DataChangeEvent} represents the data change events of external systems, such as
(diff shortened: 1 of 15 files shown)


lvyanquan:
What's more, could we add same optimization in DataSinkWriterOperator#processElement method？

yuxiqian:
> What's more, it's better to add same optimization in DataSinkWriterOperator#processElement method.

Seems `DataSinkWriterOperator` does not manipulate data change events in BinaryRecordData level. Will printing corresponding TableId suffice?

lvyanquan:
> Seems `DataSinkWriterOperator` does not manipulate data change events in BinaryRecordData level. Will printing corresponding TableId suffice?

Yeah, corresponding TableId and Schema is enough, and we can get the actual Schema from manual query.

lvyanquan (approved):
Thanks for this contribution, LGTM.

leonardBang (approved):

leonardBang:
@yuxiqian would you like to check the failure tests?

yuxiqian:
Seems some failing test cases have been traced in FLINK-37462, FLINK-35304 and FLINK-37346.

PR #3963, opened 2025-03-21
Postgres CDC Jdbc query should use debezium.snapshot.fetch.size rather than debezium.query.fetch.size.

loserwang1024:
Fix https://issues.apache.org/jira/browse/FLINK-37508


loserwang1024:
Add test waiting for https://github.com/apache/flink-cdc/pull/3964

leonardBang:
> Add test waiting for #3964

Could you fix this in https://github.com/apache/flink-cdc/pull/3964 so that the minor change can be covered by test?

loserwang1024:
included in https://github.com/apache/flink-cdc/pull/3964

PR #3970, opened 2025-03-27
[FLINK-37572] Shuffle Event by bucket and tableid to avoid data skew.

lvyanquan:
BucketWrapper#getBucket is use to shuffle DataChangeEvent to avoid multiple subtasks writting in the same bucket.
However, in many cases, most tables may only have a few buckets, this way, the data will all fall on subtask1, subtask2... even if we have configured a high parallelism.

So I suggest allocating based on bucket+hash of TableId during shuffle, This way, there will be no conflict and data skew can be avoided to a certain extent.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonEventSink.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonEventSink.java
@@ -20,6 +20,7 @@ package org.apache.flink.cdc.connectors.paimon.sink.v2;
 import org.apache.flink.cdc.common.event.Event;
 import org.apache.flink.cdc.connectors.paimon.sink.v2.bucket.BucketAssignOperator;
 import org.apache.flink.cdc.connectors.paimon.sink.v2.bucket.BucketWrapper;
+import org.apache.flink.cdc.connectors.paimon.sink.v2.bucket.BucketWrapperChangeEvent;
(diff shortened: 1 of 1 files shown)


lvyanquan:
Hi, @melin @MOBIN-F PTAL.

MOBIN-F:
ci failed, seems to be related to this PR, can you take a look? @lvyanquan

MOBIN-F (approved):
LGTM.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonEventSink.java
```diff
@@ -63,8 +64,16 @@ public DataStream<Event> addPreWriteTopology(DataStream<Event> dataStream) {
                 .name("Assign Bucket")
                 // All Events after BucketAssignOperator are decorated with BucketWrapper.
                 .partitionCustom(
-                        (bucket, numPartitions) -> bucket % numPartitions,
-                        (event) -> ((BucketWrapper) event).getBucket())
+                        (bucket, numPartitions) -> Math.abs(bucket) % numPartitions,
+                        (event) -> {
+                            if (event instanceof BucketWrapperChangeEvent) {
```
yuxiqian, line 69: It's extremely rare, but `Math.abs(bucket)` still could be negative if `bucket = Integer.MIN_VALUE`. What about using `Math.floorMod(bucket, numPartitions)` to ensure getting a positive mod result here?

lvyanquan: Done.

leonardBang (approved):
+1, thanks @lvyanquan for the contribution and @MOBIN-F for the review

PR #3971, opened 2025-03-28
[FLINK-37577][paimon] Set waitCompaction=true in PaimonWriter only when deletion-vectors.enabled option enabled

lvyanquan:
Set waitCompaction=true in PaimonWriter only when deletion-vectors.enabled of table option is true.


lvyanquan:
@melin @MOBIN-F  PTAL.

MOBIN-F (approved):
LGTM.

leonardBang (approved):
+1, and checked in my local environment

PR #3965, opened 2025-03-21
[WIP][FLINK-37453][tests] Port Ruby migration tests to Java

yuxiqian:
This closes FLINK-37453.

CDC pipeline migration tests were introduced in #3447, originally implemented in Ruby scripting language.

However, as most testing logic was written as script file, it is hard to debug or investigate what's going wrong when migration test fails with no much contextual information or stacktrace available. Existing test cases could not be easily extended, too.

---

This PR tries to extend existing Pipeline E2e test cases to cover such migration tests by:

* Adding utility methods for submitting YAML jobs with previous state & stopping jobs after creating a savepoint
* Cleaning up `PipelineTestEnvironment` and remove duplicate code fragments from sub-classes
* Rewriting migration test cases in Java with JUnit framework
* Removing existing migration test scripts, of course

PR #3972, opened 2025-03-28
[FLINK-37578] Fix distributed schema registry exposes bad internal state accidentally

yuxiqian:
This closes FLINK-37578 and #3858.

There's a subtle sequential bug in both `SchemaCoordinator`s after a schema evolve coordination process finishes. Coordinator may finish operators' blocking state first before restoring internal state properly, which may accidentally expose unwanted internal states or freeze the entire pipeline job.

@linjianchang's optimization in #3858 is actually correct, however it increases the chance of this glitch. Baked the original commit into this PR to test if it works well.


lvyanquan:
Hi, @yuxiqian. This pr looks good to me.
But I believe that we lack the necessary e2e testing to expose such issues. Can you create a Jira to trace it？

yuxiqian:
Thanks for @lvyanquan's suggestion, traced in FLINK-37704. Perhaps more test cases could be added based on changes in #3965.

lvyanquan (approved):

leonardBang (approved):
Thanks @yuxiqian for the contribution, LGTM

PR #3973, opened 2025-03-30
[FLINK-37585] [cdc-connector][mysql] Fixed the data of the newly added table cannot be read

jeff-zou:
Fixed the data of the newly added table cannot be read


gong:
You can view this discussion https://github.com/apache/flink-cdc/discussions/3338.

PR #2571, opened 2023-10-21
[postgres] Remove unnecessary schema fresh to improve performance.

lzshlzsh:
PR of #2570 

### Motivation
It's very time-consuming for postgres to refresh schema if there are many tables to read. According to our testing, refreshing 400 tables takes 15 minutes. Because it takes a long time to refresh the current table's schema when reading a chunk in the scan stage, the data rate shows the following sawtooth pattern. Therefore, we need to minimize unnecessary shema refreshes as much as possible。

image

### Solution
[...]

PR #3983, opened 2025-04-10
[FLINK-34820][postgressql] Not recycle but reuse fetcher for all data to improve performant

Mrart:
PR from https://github.com/apache/flink-cdc/pull/2571

Ignore unnecessary fetcher shutdowns
To hide the element queue from the connector developers and make SplitFetcherManager the only owner
class of the queue
It's enough to refresh the schema when [PostgresSourceFetchTaskContext#configure](https://github.com/ververica/flink-cdc-connectors/blob/e0fd6f965b702cc2876372dc068379dafe066277/flink-connector-postgres-cdc/src/main/java/com/ververica/cdc/connectors/postgres/source/fetch/PostgresSourceFetchTaskContext.java#L176) is called, and there is no need to refresh the schema afterwards.


hql0312:
@Mrart i have a question. i think the performance issue is that, when add new table to config, the stream split will change to snapshot split , and when every snapshot split finished , it will add stream split once for every snapshot,  and the stream split reader will be set null, the stream split  reader will be recreated. the schema will reload for the config tablefilter. and the change code how to performance the logic ?

Mrart:
> @Mrart @loserwang1024 i have a question. i think the performance issue is that, when add new table to config, the stream split will change to snapshot split , and when every snapshot split finished , it will add stream split once for every snapshot, and the stream split reader will be set null, so the stream split reader will be recreated. the schema will reload for the config tablefilter. and the change code how to performance the logic ? in IncrementalSourceSplitReader.java , method pollSplitRecords have follow code :
> 
> ```java
> private ChangeEventRecords pollSplitRecords() throws InterruptedException {
>         Iterator<SourceRecords> dataIt = null;
>         if (currentFetcher == null) {
>             // (1) Reads stream split firstly and then read snapshot split
>             if (streamSplits.size() > 0) {
>                 // the stream split may come from:
>                 // (a) the initial stream split
>                 // (b) added back stream-split in newly added table process
>                 StreamSplit nextSplit = streamSplits.poll();
>                 submitStreamSplit(nextSplit);
>             } else if (snapshotSplits.size() > 0) {
>                 submitSnapshotSplit(snapshotSplits.poll());
>             } else {
>                 LOG.info("No available split to read.");
>             }
> 
>             if (currentFetcher != null) {
>                 dataIt = currentFetcher.pollSplitRecords();
>             } else {
>                 currentSplitId = null;
>             }
>             return dataIt == null ? finishedSplit() : forRecords(dataIt);
>         } else if (currentFetcher instanceof IncrementalSourceScanFetcher) {
>             // (2) try to switch to stream split reading util current snapshot split finished
>             dataIt = currentFetcher.pollSplitRecords();
>             if (dataIt != null) {
>                 // first fetch data of snapshot split, return and emit the records of snapshot split
>                 ChangeEventRecords records;
>                 if (context.isHasAssignedStreamSplit()) {
>                     records = forNewAddedTableFinishedSplit(currentSplitId, dataIt);
>                     closeScanFetcher();
>                     closeStreamFetcher();
>                 } else {
>                     records = forRecords(dataIt);
>                     SnapshotSplit nextSplit = snapshotSplits.poll();
>                     if (nextSplit != null) {
>                         checkState(reusedScanFetcher != null);
>                         submitSnapshotSplit(nextSplit);
>                     } else {
>                         closeScanFetcher();
>                     }
>                 }
>                 return records;
>             } else {
>                 return finishedSplit();
>             }
>         } else if (currentFetcher instanceof IncrementalSourceStreamFetcher) {
>             // (3) switch to snapshot split reading if there are newly added snapshot splits
>             dataIt = currentFetcher.pollSplitRecords();
>             if (dataIt != null) {
>                 // try to switch to read snapshot split if there are new added snapshot
>                 SnapshotSplit nextSplit = snapshotSplits.poll();
>                 if (nextSplit != null) {
>                     closeStreamFetcher();
>                     LOG.info("It's turn to switch next fetch reader to snapshot split reader");
>                     submitSnapshotSplit(nextSplit);
>                 }
>                 return ChangeEventRecords.forRecords(STREAM_SPLIT_ID, dataIt);
>             } else {
>                 // null will be returned after receiving suspend stream event
>                 // finish current stream split reading
>                 closeStreamFetcher();
>                 return finishedSplit();
>             }
>         } else {
>             throw new IllegalStateException("Unsupported reader type.");
>         }
>     }
> 
>     private void closeStreamFetcher() {
>         if (reusedStreamFetcher != null) {
>             LOG.debug("Close stream reader {}", reusedStreamFetcher.getClass().getCanonicalName());
>             reusedStreamFetcher.close();
>             if (currentFetcher == reusedStreamFetcher) {
>                 currentFetcher = null;
>             }
>             reusedStreamFetcher = null;
>         }
>     }
> ```
This pr is preventing the fetcher from reclaiming when it is idle

loserwang1024:
> and the change code how to performance the logic ?
@Mrart this pr is mainly focused on most situations. We can improve it later when newly added table.

To be honest, it's not a good idea for that a reader fequently switch from streaming split and snapshot split（unless just 1 perrellism). The switch costs   
 a lot. I have too ideas:
1. the first idea: do not assign newly added snapshot split to the reader which handling binlog. I will recently push forward a FLIP to let enumerator knows currently split distributions.
2. the second idea: The splitReader can reading snapshot split and stream split into a queue in the same time, is no need to just read one in a time. It will need to change current thread model.

But this PR still can improve a lot, we can let it in at first.

hql0312:
> > and the change code how to performance the logic ?
> 
> Hi, @Mrart , @hql0312 . this pr is mainly focused on most situations. As I discussed in #2571, In snapshot split phase, after each snapshot split is finished, the fetcher will be closed. And newly added snapshot will open a new fetcher, and need lookup schema again. Snipaste_2025-04-10_19-56-51
> 
> What you say is a special case when enabling newly added table. To be honest, it's not a good idea for that a reader fequently switch between streaming split and snapshot split（unless just 1 parrellism). The switch costs a lot. I have too ideas:
> 
> 1. the first idea: do not assign newly added snapshot split to the reader which handling binlog. I will recently push forward a FLIP to let enumerator knows currently split distributions.
> 2. the second idea: The splitReader can reading snapshot split and stream split into a queue in the same time, is no need to just read one in a time. It will need to change current thread model.
> 
> But this PR still can improve a lot, we can let it in at first.

if can hold streamsplit context when snapshot come ,and not recycle . the performance can improve.

in stream phase, the pg connector will costs too much time for load schema. if we can hold schema in cache ,the cost will reduce.

the logic is right?
@loserwang1024

Mrart:
> > > and the change code how to performance the logic ?
> > 
> > 
> > Hi, @Mrart , @hql0312 . this pr is mainly focused on most situations. As I discussed in #2571, In snapshot split phase, after each snapshot split is finished, the fetcher will be closed. And newly added snapshot will open a new fetcher, and need lookup schema again. Snipaste_2025-04-10_19-56-51
> > What you say is a special case when enabling newly added table. To be honest, it's not a good idea for that a reader fequently switch between streaming split and snapshot split（unless just 1 parrellism). The switch costs a lot. I have too ideas:
> > 
> > 1. the first idea: do not assign newly added snapshot split to the reader which handling binlog. I will recently push forward a FLIP to let enumerator knows currently split distributions.
> > 2. the second idea: The splitReader can reading snapshot split and stream split into a queue in the same time, is no need to just read one in a time. It will need to change current thread model.
> > 
> > But this PR still can improve a lot, we can let it in at first.
> 
> if can hold streamsplit context when snapshot come ,and not recycle . the performance can improve.
> 
> in stream phase, the pg connector will costs too much time for load schema. if we can hold schema in cache ,the cost will reduce.
> 
> the logic is right? @loserwang1024
Looking forward to your flip, we can discuss it in dev email list. @hql0312

hql0312:
> > i add a static object to hold the schema
> 
> @hql0312 To be honest, I doesn't recommanded that use static object in cdc connectot. It will influence between tasks(when in session mode), and also will influence multiple slots in same jvm. @leonardBang , CC, WDYT?

@loserwang1024 because source operator, in stream phase, only one task will hold the stream split , and the static object Will be dynamically created , even the source operator is assigned to other task.

loserwang1024:
LGTM now, please fix test.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/fetch/PostgresSourceFetchTaskContext.java
```diff
@@ -174,13 +174,15 @@ public void configure(SourceSplitBase sourceSplitBase) {
...
 
         try {
-            this.schema =
-                    PostgresObjectUtils.newSchema(
-                            jdbcConnection,
-                            dbzConfig,
-                            jdbcConnection.getTypeRegistry(),
-                            topicSelector,
-                            valueConverterBuilder.build(jdbcConnection.getTypeRegistry()));
+            if (this.schema == null) {
```
loserwang1024, line 177: Would you like to take apart this into another pr? It seems this pr do too much things, which is not easy to review and trace.

This PR please just do one thing:Not recycle but reuse fetcher for all data to improve performant

loserwang1024 (commented):
I have left some comment

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinarySegmentUtils.java
```diff
@@ -88,23 +88,6 @@ public static byte[] allocateReuseBytes(int length) {
         return bytes;
     }
 
-    public static char[] allocateReuseChars(int length) {
```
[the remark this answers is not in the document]
Mrart: It unused.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/fetch/PostgresSourceFetchTaskContext.java
Mrart: I think it's a good idea

leonardBang:
@Mrart @loserwang1024 Do we still need this PR ?

PR #3988, opened 2025-04-11
[FLINK-37653] Refactor incremental snapshot framework to support multiple stream splits

yuxiqian:
This closes FLINK-37653.

Currently, `HybridSplitAssigner`s and `StreamSplitAssigner`s in incremental snapshot framework implicitly assume that there will be at most one unbounded stream split, which isn't true for some data sources that supports multiple change stream splits, like MongoDB and PolarDB-X.

This change extends assigners' API without changing existing data sources' behavior.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/HybridSplitAssigner.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/HybridSplitAssigner.java
@@ -19,6 +19,7 @@ package org.apache.flink.cdc.connectors.base.source.assigner;
 
 import org.apache.flink.api.connector.source.SourceSplit;
 import org.apache.flink.api.connector.source.SplitEnumeratorContext;
+import org.apache.flink.cdc.common.utils.Preconditions;
 import org.apache.flink.cdc.connectors.base.config.SourceConfig;
 import org.apache.flink.cdc.connectors.base.dialect.DataSourceDialect;
 import org.apache.flink.cdc.connectors.base.source.assigner.state.HybridPendingSplitsState;
@@ -32,12 +33,14 @@ import org.apache.flink.cdc.connectors.base.source.meta.split.StreamSplit;
 import org.apache.flink.cdc.connectors.base.source.metrics.SourceEnumeratorMetrics;
 
 import io.debezium.relational.TableId;
+import io.debezium.relational.history.TableChanges;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
@@ -58,14 +61,15 @@ public class HybridSplitAssigner<C extends SourceConfig> implements SplitAssigne
     private final int splitMetaGroupSize;
     private final C sourceConfig;
 
-    private boolean isStreamSplitAssigned;
-
     private final SnapshotSplitAssigner<C> snapshotSplitAssigner;
 
     private final OffsetFactory offsetFactory;
 
     private final SplitEnumeratorContext<? extends SourceSplit> enumeratorContext;
     private SourceEnumeratorMetrics enumeratorMetrics;
+
+    private final int numberOfStreamSplits;
+    private transient int updatedStreamSplitCount;
 
     public HybridSplitAssigner(
             C sourceConfig,
@@ -87,7 +91,8 @@ public class HybridSplitAssigner<C extends SourceConfig> implements SplitAssigne
                 false,
                 sourceConfig.getSplitMetaGroupSize(),
                 offsetFactory,
-                enumeratorContext);
+                enumeratorContext,
+                dialect.getNumberOfStreamSplits(sourceConfig));
     }
 
     public HybridSplitAssigner(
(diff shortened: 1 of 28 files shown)


loserwang1024 (commented):
1. Is there any test?
2. I wonder what is the assign status now?
3. If parrelism is 1, what is the split reader behavior with two stream splits?

yuxiqian:
Thanks for @loserwang1024's kindly review. My answers:

* We don't have any incremental snapshot source that supports multiple StreamSplits for now, so I'll add some mocking tests here.

* The transition of assigning status should not change, the only difference would be the amount of streaming splits assigned.

* An exception will be thrown since there won't be enough splits. Extra checking could also be added from the source side to fail in the config validation phase.

```java
Preconditions.checkArgument(
     pendingStreamSplits.size() <= enumeratorContext.currentParallelism(),
     "%s stream splits generated, which is greater than current parallelism %s. Some splits might never be assigned.",
     pendingStreamSplits.size(),
     enumeratorContext.currentParallelism());
```

loserwang1024:
> The transition of assigning status should not change, the only difference would be the amount of streaming splits assigned.

Please check each assign status(focus on the switch condition ). For example, in your pr, MySqlSnapshotSplitAssigner#onBinlogSplitUpdated is no change. If one binlog split is update, the status will switch to final status. However, only after all the binlog splits are update , this can be update. 

Another question, is newly added table case taken into consideration? (for example, StreamSplitUpdateRequestEvent, LatestFinishedSplitsNumberRequestEvent and LatestFinishedSplitsNumberEvent)

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/StreamSplitAssigner.java
```diff
@@ -165,39 +168,82 @@ public void close() throws IOException {
...
+            Offset minOffset,
+            Offset stoppingOffset,
+            List<FinishedSnapshotSplitInfo> finishedSnapshotSplitInfos,
+            Map<TableId, TableChanges.TableChange> tableSchemas,
+            int totalFinishedSplitSize,
+            boolean isSuspended,
+            boolean isSnapshotCompleted) {
+        return Collections.singletonList(
+                new StreamSplit(
+                        STREAM_SPLIT_ID,
```
loserwang1024, line 240: maybe we can provide a name like STREAM_SPLIT_ID_{number} to distinguish different stream split?

yuxiqian: Good idea, applied it in the mocked incremental source. Shall we leave the split name unchanged for the default implementation since it has only one stream split?

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/enumerator/IncrementalSourceEnumerator.java
```diff
@@ -172,7 +181,7 @@ public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
             LOG.info(
                     "The enumerator receives notice from subtask {} for the stream split assignment. ",
                     subtaskId);
-            this.streamSplitTaskId = subtaskId;
```
loserwang1024, line 175: 1. Use set rather than list?   2. What if the stream split's taskId changed because addSplitBack or something else? I remember @ruanhang use to mention it when I intruduce newly added framework.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/HybridSplitAssigner.java
```diff
@@ -252,56 +260,106 @@ public void close() throws IOException {
...
-        Map<String, Offset> splitFinishedOffsets = snapshotSplitAssigner.getSplitFinishedOffsets();
-        final List<FinishedSnapshotSplitInfo> finishedSnapshotSplitInfos = new ArrayList<>();
-
-        Offset minOffset = null, maxOffset = null;
-        for (SchemalessSnapshotSplit split : assignedSnapshotSplit) {
-            // find the min and max offset of change log
-            Offset changeLogOffset = splitFinishedOffsets.get(split.splitId());
-            if (minOffset == null || changeLogOffset.isBefore(minOffset)) {
-                minOffset = changeLogOffset;
+    // Overridable methods
```
loserwang1024, line 263: Why variable is not put into begining? And I don't unstand what's Overridable methods meaning hear?

loserwang1024 (commented):
I have left some comment

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/HybridSplitAssigner.java
```diff
@@ -159,16 +168,11 @@ public Optional<SourceSplitBase> getNext() {
                 // we need to wait snapshot-assigner to be finished before
                 // assigning the stream split. Otherwise, records emitted from stream split
                 // might be out-of-order in terms of same primary key with snapshot splits.
-                isStreamSplitAssigned = true;
```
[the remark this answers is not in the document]
yuxiqian: `getNextStreamSplit` will set the `isStreamSplitAllAssigned` flag if all stream splits have been assigned. Seems we only need to wake up existing stream readers after newly added assigning finished.

yuxiqian:
As discussed with @loserwang1024 offline, modifying current framework seems to be a breaking change that might break backwards compatibility. Close it for now.

PR #3992, opened 2025-04-18
[FLINK-37696][test][cdc-connector][postgres] replace postgres image to official pg 14 and adjust test to use pgoutput plugin

loserwang1024:
As discussed in http://issues.apache.org/jira/browse/FLINK-37696, the postgres 14 is most used version of postgres, thus this Jira will replace test image and use pgoutput.


loserwang1024:
@leonardBang @ruanhang1993 , CC

loserwang1024:
Oceanbase test failed while this pr just modify postgres test.

leonardBang (approved):

PR #3994, opened 2025-04-18
[FLINK-37676][cdc-common] Add caching mechanism to Selectors for improved performance

joyCurry30:
Refer to https://issues.apache.org/jira/browse/FLINK-37676
Adding a cache for better performance.

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/schema/Selectors.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/schema/Selectors.java
@@ -20,6 +20,10 @@ package org.apache.flink.cdc.common.schema;
 import org.apache.flink.cdc.common.event.TableId;
 import org.apache.flink.cdc.common.utils.Predicates;
 
+import org.apache.flink.shaded.guava31.com.google.common.cache.Cache;
+import org.apache.flink.shaded.guava31.com.google.common.cache.CacheBuilder;
+
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
@@ -29,7 +33,15 @@ import java.util.function.Predicate;
 /** Selectors for filtering tables. */
 public class Selectors {
 
+    private static final Duration CACHE_EXPIRE_DURATION = Duration.ofHours(1);
+
     private List<Selector> selectors;
+
+    private final Cache<TableId, Boolean> cache =
+            CacheBuilder.newBuilder()
+                    .expireAfterAccess(CACHE_EXPIRE_DURATION)
+                    .maximumSize(1024)
+                    .build();
 
     private Selectors() {}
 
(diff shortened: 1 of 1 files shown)


lvyanquan:
I think a JMH test for this will be better.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/schema/Selectors.java
```diff
@@ -29,8 +33,16 @@
...
 
+    private static final Duration CACHE_EXPIRE_DURATION = Duration.ofHours(1);
+
     private List<Selector> selectors;
 
+    private final Cache<TableId, Boolean> cache =
+            CacheBuilder.newBuilder()
+                    .expireAfterAccess(CACHE_EXPIRE_DURATION)
+                    .maximumSize(1024)
+                    .build();
```
yuxiqian, line 44: Is it correct to share this cache among all `Selector`s? IIUC it depends on the effective `Selector` to determine if a `TableId` should be matched.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/schema/Selectors.java
```diff
@@ -29,8 +33,16 @@
 /** Selectors for filtering tables. */
 public class Selectors {
 
+    private static final Duration CACHE_EXPIRE_DURATION = Duration.ofHours(1);
+
     private List<Selector> selectors;
 
+    private final Cache<TableId, Boolean> cache =
```
yuxiqian, line 40: Maybe we can use `LoadingCache` here instead of checking / updating caches manually.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/schema/Selectors.java
```diff
@@ -71,8 +83,20 @@ public boolean isMatch(TableId tableId) {
...
+    /** Match the {@link TableId} against the {@link Selector}s. */
     public boolean isMatch(TableId tableId) {
+        Boolean cachedResult = cache.getIfPresent(tableId);
+        if (cachedResult != null) {
+            return cachedResult;
+        }
+
+        boolean match = computeIsMatch(tableId);
+        cache.put(tableId, match);
+        return match;
```
yuxiqian, line 95: This could be simplified with `LoadingCache`:

```suggestion
        return cache.getUnchecked(tableId);
```

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/schema/Selectors.java
yuxiqian: IIUC both `Selector` and `Selectors` are meant to be effectively `final` and should not have any mutable states. Any comments @GOODBOY008?

yuxiqian:
Could @joyCurry30 please rebase this PR with latest `master` branch?

joyCurry30:
@yuxiqian done. cc plz.

yuxiqian (approved):
Looks good, thank you for the nice work.

PR #3996, opened 2025-04-18
[FLINK-37634][pipeline-connector/mysql]When the table name of mysql s…

linjianchang:
When the table name of mysql supports size sensitive, an error occurred when the table name of alter capitalization will not be found.
The Exception like:

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSource.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSource.java
@@ -29,6 +29,7 @@ import org.apache.flink.cdc.connectors.mysql.source.config.MySqlSourceConfig;
 import org.apache.flink.cdc.connectors.mysql.source.config.MySqlSourceConfigFactory;
 import org.apache.flink.cdc.connectors.mysql.source.reader.MySqlPipelineRecordEmitter;
 import org.apache.flink.cdc.connectors.mysql.table.MySqlReadableMetadata;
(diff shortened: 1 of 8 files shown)


joyCurry30:
As documented in the official MySQL 8.4 reference manual: https://dev.mysql.com/doc/refman/8.4/en/identifier-case-sensitivity.html
The case sensitivity behavior of MySQL identifiers is governed by the system variable "lower_case_table_names", which defines three distinct operational modes:
| mode | Behavior |	Scope |
| ----------- | ----------- | ----------- |
| 0 | Case-sensitive comparisons with case-preserving storage | Affects database names and table names |
| 1 | Case-insensitive comparisons with lowercase conversion | Applies to database names and table names |
| 2 | Case-insensitive comparisons with case-preserving storage| Impacts database names and table names |

 Should we introduce a validation mechanism to ensure configuration parity between the connector and MySQL server (e.g., rejecting mixed-case identifiers when lower_case_table_names=1)?

linjianchang:
> As documented in the official MySQL 8.4 reference manual: https://dev.mysql.com/doc/refman/8.4/en/identifier-case-sensitivity.html The case sensitivity behavior of MySQL identifiers is governed by the system variable "lower_case_table_names", which defines three distinct operational modes:
> 
> mode	Behavior	Scope
> 0	Case-sensitive comparisons with case-preserving storage	Affects database names and table names
> 1	Case-insensitive comparisons with lowercase conversion	Applies to database names and table names
> 2	Case-insensitive comparisons with case-preserving storage	Impacts database names and table names
> Should we introduce a validation mechanism to ensure configuration parity between the connector and MySQL server (e.g., rejecting mixed-case identifiers when lower_case_table_names=1)?



> rejecting mixed-case identifiers when lower_case_table_names=1

thanks for review，I have some questions about this“rejecting mixed-case identifiers when lower_case_table_names=1”，so I don't understand how to modify it

joyCurry30:
@linjianchang please rebase master.

joyCurry30 (approved):
LGTM.

PR #4004, opened 2025-04-26
[FLINK-37479][postgres] Add support for PARTITIONED TABLE

phamvinh1712:


phamvinh1712:
Hello @loserwang1024 , could you help to review when you have time. Thank you

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/test/java/org/apache/flink/cdc/connectors/postgres/table/PostgreSQLConnectorITCase.java
```diff
@@ -232,6 +232,118 @@ void testConsumingAllEvents(boolean parallelismSnapshot)
...
+                Statement statement = connection.createStatement()) {
+            statement.execute(
+                    String.format(
+                            "CREATE PUBLICATION %s FOR TABLE inventory_partitioned.products "
+                                    + " WITH (publish_via_partition_root=true)",
+                            publicationName));
+            statement.execute(
+                    String.format(
+                            "select pg_create_logical_replication_slot('%s','pgoutput');",
+                            slotName));
```
loserwang1024, line 253: We can also create a new sub-partition table and insert data here. Newly added partition table will also be read.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/test/java/org/apache/flink/cdc/connectors/postgres/table/PostgreSQLConnectorITCase.java
```diff
@@ -232,6 +232,118 @@ void testConsumingAllEvents(boolean parallelismSnapshot)
...
+
+        waitForSnapshotStarted("sink");
+
+        // wait a bit to make sure the replication slot is ready
+        Thread.sleep(5000);
+
+        // generate WAL
+        try (Connection connection = getJdbcConnection(POSTGRES_CONTAINER);
+                Statement statement = connection.createStatement()) {
+            statement.execute(
```
loserwang1024, line 317: We can also create a new sub-partition table and insert data here. Newly added partition table will also be read.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/test/java/org/apache/flink/cdc/connectors/postgres/table/PostgreSQLConnectorITCase.java
```diff
@@ -232,6 +232,118 @@ void testConsumingAllEvents(boolean parallelismSnapshot)
...
+                        "INSERT INTO sink SELECT id, name, description, weight, country FROM debezium_source");
+
+        waitForSnapshotStarted("sink");
+
+        // wait a bit to make sure the replication slot is ready
+        Thread.sleep(5000);
+
+        // generate WAL
+        try (Connection connection = getJdbcConnection(POSTGRES_CONTAINER);
+                Statement statement = connection.createStatement()) {
```
loserwang1024, line 316: We can also create a new sub-partition table and insert data here. Newly added partition table will also be read.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/utils/TableDiscoveryUtils.java
```diff
@@ -38,7 +38,8 @@ public static List<TableId> listTables(
             throws SQLException {
 
         Set<TableId> allTableIds =
-                jdbc.readTableNames(database, null, null, new String[] {"TABLE"});
```
loserwang1024, line 41: I hope we can add a param such as partitioned(maybe we can check whether debezium has), when this param is enabled:
1. discovery table here.
2. add publish_via_partition_root=true when create publication in io.debezium.connector.postgresql.connection.PostgresReplicationConnection#initPublication.

To be honest, I hope debezium helps do it. But if it lacks it, we can do it to make easier for user.

loserwang1024 (commented):
This Pr looks good to me in general, just some minor advice.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/utils/TableDiscoveryUtils.java
phamvinh1712: I know that debezium doesn't create publication with `publish_via_partition_root=true` and users will need to create it themselves in advance. 
I agree with you that if we do it it will help users, but i'm afraid it will make us hard to upgrade debezium version (which i believe we need to do at some points since flink-cdc is using an old version of debezium 1.9.6)

loserwang1024: Theres is also another concern. If user use regex expression as table name, both parent table and subtable will be captured. **Then snapshot data will be read twice.** 

Thus, a option can also tell user that. If they enable partition table, they should pay attention to it.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/test/java/org/apache/flink/cdc/connectors/postgres/table/PostgreSQLConnectorITCase.java
phamvinh1712: @loserwang1024 i've updated test for this.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/utils/TableDiscoveryUtils.java
phamvinh1712: > Theres is also another concern. If user use regex expression as table name, both parent table and subtable will be captured. **Then snapshot data will be read twice.**
> 
> Thus, a option can also tell user that. If they enable partition table, they should pay attention to it.

do you have any suggestion for this? should we add a section on flink-cdc doc for this?

loserwang1024:
I have dedug this modification, it turns out.

`jdbc.readTableNames(database, null, null, new String[] {"TABLE"})` will show sub partition tables:
image

`jdbc.readTableNames( database, null, null, new String[] {"PARTITIONED TABLE"})` will show parents partition table:
image

Thus, if set `table-name` = 'inventory_partitioned.*',  both leaf tables and parents table are captured.
image

Thus, even though we can not publish_via_partition_root in this pr, a option to enable partition table is needed:
```java
 Set<TableId> allTableIds ;
        if(!partition) {
            jdbc.readTableNames(
                    database, null, null, new String[]{ "TABLE"});
        }else{
            jdbc.readTableNames(
                    database, null, null, new String[]{ "PARTITIONED TABLE"});
        }


```

phamvinh1712:
hi @loserwang1024 , I've updated the PR to add this feature via a boolean flag (default to false) so it won't affect existing users. I also added a note on the docs that users who opt-in to use this feature should create PUBLICATION beforehand and avoid listing parent/child tables twice. Please let me know what you think.

loserwang1024:
I agree with you in general. Maybe only a minor discuss on option name.  What about `scan.publish-via-partition-root.enabled`? @leonardBang @ruanhang1993 , WDYT?

```xml
      <td>include-partitioned-table</td>
      <td>optional</td>
      <td style="word-wrap: break-word;">false</td>
      <td>Boolean</td>
      <td>
        Whether to enable reading partitioned table.<br>
        If enabled:
          (1) PUBLICATION must be created beforehand with parameter publish_via_partition_root=true
          (2) Table list (regex or predefined list) should only match the parent table name, if table list matches both parent and child tables, snapshot data will be read twice.
      </td>
    </tr>
```

phamvinh1712:
hi @loserwang1024 , i've updated the option name based on your suggestion

ruanhang1993:
> I agree with you in general. Maybe only a minor discuss on option name. What about `scan.publish-via-partition-root.enabled`? @leonardBang @ruanhang1993 , WDYT?
> 
> ```
>       <td>include-partitioned-table</td>
>       <td>optional</td>
>       <td style="word-wrap: break-word;">false</td>
>       <td>Boolean</td>
>       <td>
>         Whether to enable reading partitioned table.<br>
>         If enabled:
>           (1) PUBLICATION must be created beforehand with parameter publish_via_partition_root=true
>           (2) Table list (regex or predefined list) should only match the parent table name, if table list matches both parent and child tables, snapshot data will be read twice.
>       </td>
>     </tr>
> ```

Hi, @phamvinh1712 @loserwang1024 .Sorry for my late response. 

IMO, I think this new option name `scan.publish-via-partition-root.enabled` is difficult to understand for users who are not familiar with postgres. Actually I am one of them. ;)

The first time I saw this option name, I was confused and do not know when I should open this option. 
I think the option name could be like `scan.include-partitioned-tables.enabled`. WDYT?

leonardBang:
> `scan.include-partitioned-tables.enabled`.

+1 for this option name, I think it's better to not expose to many underlying concept to users, especially many Flink CDC users are not familiar with PG internal.

loserwang1024:
I agree with you guys. It seems that I was considered from a postgres connector developer rather than a norman user.

phamvinh1712:
hi @loserwang1024 @leonardBang @ruanhang1993 , thanks all for the feedbacks 🙌  and sorry for the late response. 
I've updated the configuration like suggested, could you help to review the PR again.

loserwang1024 (approved):
LGTM!

lvyanquan:
CC @Mrart, as you may be interested in this.

Mrart:
> CC @Mrart, as you may be interested in this.

Yes, thanks. Then can we solve the UT test first? I will also take a close look at this pr.

Mrart:
> > CC @Mrart, as you may be interested in this.
> 
> Yes, thanks. @phamvinh1712  Then can we solve the UT test first? I will also take a close look at this pr.

Mrart (approved):
LGTM， and let a problem,  It will not supported pgsql 10 version，need version greater then 10.

phamvinh1712:
hi @Mrart , thanks for reviewing. I merge master to my branch and the test is fixed now.
Could any committer help to review/approve the PR to trigger UT tests again, thanks in advance

PR #3918, opened 2025-02-12
[FLINK-36581][cli] Allow passing Flink configuration by yaml job file

MOBIN-F:
At present, FlinkCDC only supports reading Flink configurations from Flink con files, but this approach is not user-friendly for multiple jobs that require different configurations. Allow passing Flink configuration by yaml job file
such as :
```
source:
  type: mysql
  ...

sink:
  type: value

pipeline:
  name: MySQL to Kafka Pipeline
  parallelism: 1
  flink-conf:
    execution.checkpointing.interval: 6min
```

PR #4006, opened 2025-04-28
[cli] fix flink-conf is not obtained in local mode

MOBIN-F:
current local mode lacks loading additionalJars and flinkConf
image

### DIFF
--- a/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
+++ b/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
@@ -18,6 +18,7 @@
 package org.apache.flink.cdc.composer.flink;
 
 import org.apache.flink.cdc.common.annotation.Internal;
+import org.apache.flink.cdc.common.annotation.VisibleForTesting;
 import org.apache.flink.cdc.common.configuration.Configuration;
 import org.apache.flink.cdc.common.event.Event;
 import org.apache.flink.cdc.common.pipeline.PipelineOptions;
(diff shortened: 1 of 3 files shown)


flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
```diff
@@ -59,32 +60,28 @@ public class FlinkPipelineComposer implements PipelineComposer {
...
 
     public static FlinkPipelineComposer ofApplicationCluster(StreamExecutionEnvironment env) {
         return new FlinkPipelineComposer(env, false);
     }
 
+    @VisibleForTesting
     public static FlinkPipelineComposer ofMiniCluster() {
         return new FlinkPipelineComposer(
                 StreamExecutionEnvironment.getExecutionEnvironment(), true);
     }
```
joyCurry30, line 75: Is it still necessary to keep this method?

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
```diff
@@ -254,4 +251,27 @@ private Optional<URL> getContainingJar(Class<?> clazz) throws Exception {
...
+                                        jarPath),
+                                e);
+                    }
+                });
+    }
+
+    @VisibleForTesting
+    public StreamExecutionEnvironment getEnv() {
+        return env;
+    }
```
joyCurry30, line 276: Would it be possible to use Java reflection in the test to obtain this object, instead of creating a separate method for it?

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
MOBIN-F: There are still many test cases that use this method. Since they do not use flinkConf and additionalJars, I chose to keep this method.

flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposer.java
MOBIN-F: Reflection makes code difficult to maintain

lvyanquan:
Would you like to rebase master to solve the conflict?

MOBIN-F:
This bug has been fixed in  #3918.

PR #3986, opened 2025-04-10
[FLINK-37104] Refactor transform module, code cleanup

yuxiqian:
This closes FLINK-37104.

...by refactoring transform module to improve code readability and maintainability. No behavior changes are expected for all valid use cases.

Kindly ping @aiwenmo as he is the main developer of flink-runtime package.

PR #4007, opened 2025-04-28
[FLINK-37741] Fix transform operator performance degradation

yuxiqian:
This closes FLINK-37741.

PR #3986 refactored transform operator by cleaning up code and simplifying logic. However, `Schema#getColumnDataTypes` are repeatedly invoked for every `DataChangeEvent`, which isn't cached and severely degrades the performance.

Tested locally with 10000000 records grabbed from MySQL, 40% ~ 60% throughput degradation could be noticed. Applied this patch appears to resolve this issue.

Version | Transform | Time Elapsed | RPS | RPS +- |
-- | -- | -- | -- | -- |
3.4-patch | Complex | 1m 44s 580ms | 96,153.85 | ▲ 1.92% |
3.4-patch | Simple | 1m 26s 621ms | 116,279.07 | ▲ 9.30% |
3.4-patch | Empty | 1m 20s 909ms | 125,000.00 | ▲ 1.25% |
3.4-patch | Off | 1m 18s 674ms | 128,205.13 | ▲ 20.51% |
3.4 | Complex | 4m 21s 86ms | 38,314.18 | ▼ 59.39% |
3.4 | Simple | 2m 45s 488ms | 60,606.06 | ▼ 43.03% |
3.4 | Empty | 1m 31s 869ms | 109,890.11 | ▼ 10.99% |
3.4 | Off | 1m 45s 587ms | 95,238.10 | ▼ 10.48% |
3.3 | Complex | 1m 46s 54ms | 94,339.62 | Baseline |
3.3 | Simple | 1m 34s 971ms | 106,382.98 | Baseline |
3.3 | Empty | 1m 21s 546ms | 123,456.79 | Baseline |
3.3 | Off | 1m 34s 5ms | 106,382.98 | Baseline |

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
@@ -30,6 +30,7 @@ import org.apache.flink.cdc.common.event.SchemaChangeEvent;
 import org.apache.flink.cdc.common.event.TableId;
 import org.apache.flink.cdc.common.schema.Schema;
 import org.apache.flink.cdc.common.schema.Selectors;
+import org.apache.flink.cdc.common.types.DataType;
 import org.apache.flink.cdc.common.udf.UserDefinedFunctionContext;
 import org.apache.flink.cdc.common.utils.SchemaMergingUtils;
 import org.apache.flink.cdc.common.utils.SchemaUtils;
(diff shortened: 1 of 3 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperator.java
```diff
@@ -362,16 +363,17 @@ private Tuple2<BinaryRecordData, Boolean> transformRecord(
...
+        List<DataType> postDataTypes = info.getPostTransformedDataTypes();
+
         BinaryRecordDataGenerator postGenerator = info.getPostTransformedRecordDataGenerator();
 
         Object[] preRow = new Object[preFieldGetters.length];
         for (int i = 0; i < preFieldGetters.length; i++) {
             preRow[i] =
                     DataTypeConverter.convertToOriginal(
-                            preFieldGetters[i].getFieldOrNull(recordData),
-                            preSchema.getColumnDataTypes().get(i));
```
lvyanquan: I think reducing the time consumption of this `Schema.getColumnDataTypes()` method would be more appropriate, as the caller may not be clear about the logic behind it, and this is an immutable class.

yuxiqian:
Thanks for Kunni's kindly review. I've modified the `Schema` class to cache these calculated fields.

One potential breaking change is that since we're returning always returning a cached version instead of calculating it every time, we have to make it unmodifiable to avoid it from being changed accidentally. Any callers that need the previous behavior must create a copy explicitly.

lvyanquan:
> One potential breaking change is that since we're always returning a cached version instead of calculating it every time, we have to make it unmodifiable to avoid it from being changed accidentally. Any callers that need the previous behavior must create a copy explicitly.

As you have returned an immutable list, do we still need to worry about the caller modifying it?

yuxiqian:
> > One potential breaking change is that since we're always returning a cached version instead of calculating it every time, we have to make it unmodifiable to avoid it from being changed accidentally. Any callers that need the previous behavior must create a copy explicitly.
> 
> As you have returned an immutable list, do we still need to worry about the caller modifying it?

Yes, this could be prohibited by wrapping it with `Collections.unmodifiableList`, but some client codes like `schema.getColumnNames().add("...")` have to be changed to make a copy manually.

lvyanquan:
Yes, this is expected. 
If such code existed before, we can detect such usage through testing as such a call would fail.

lvyanquan (approved):
LGTM.

PR #4000, opened 2025-04-22
[hotfix][cdc-base] Fix split not serialize isSnapshotCompleted

loserwang1024:
This Pr do two things:
1. [hotfix][cdc-base] Fix split not serialize isSnapshotCompleted
2. Cp https://github.com/apache/flink-cdc/pull/2705.

PR #4014, opened 2025-05-10
FLINK-37774: SQL Server CDC connector incorrectly handles special characters in database and table names

morozov:
There are two problems:
1. Invalid SQL query for listing tables.
2. Invalid serialization of `TableId`s.

The latter is similar to #4000, which wasn't fixed consistently across the codebase.

Besides making the database name quoted in SQL, I modified the implementation of `SqlServerUtils#quote()` – it should not only enclose the name in square brackets but also escape the closing square bracket.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-sqlserver-cdc/src/main/java/org/apache/flink/cdc/connectors/sqlserver/source/utils/SqlServerUtils.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-sqlserver-cdc/src/main/java/org/apache/flink/cdc/connectors/sqlserver/source/utils/SqlServerUtils.java
@@ -361,12 +361,16 @@ public class SqlServerUtils {
         return quoted.toString();
     }
 
-    public static String quote(String dbOrTableName) {
-        return "[" + dbOrTableName + "]";
+    /**
+     * @link <a
+     *     href="https://learn.microsoft.com/en-us/sql/t-sql/functions/quotename-transact-sql">QUOTENAME</a>
+     */
+    public static String quote(String name) {
+        return "[" + name.replace("]", "]]") + "]";
     }
 
     public static String quote(TableId tableId) {
-        return "[" + tableId.schema() + "].[" + tableId.table() + "]";
+        return String.format("%s.%s", quote(tableId.schema()), quote(tableId.table()));
     }
 
     private static void addPrimaryKeyColumnsToCondition(
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-sqlserver-cdc/src/main/java/org/apache/flink/cdc/connectors/sqlserver/source/utils/SqlServerUtils.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-sqlserver-cdc/src/main/java/org/apache/flink/cdc/connectors/sqlserver/source/utils/SqlServerUtils.java
@@ -366,7 +366,7 @@ public class SqlServerUtils {
      *     href="https://learn.microsoft.com/en-us/sql/t-sql/functions/quotename-transact-sql">QUOTENAME</a>
      */
     public static String quote(String name) {
-        return "[" + name.replace("]", "]]") + "]";
+        return "[" + name + "]";
     }
 
     public static String quote(TableId tableId) {
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/state/PendingSplitsStateSerializer.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/assigner/state/PendingSplitsStateSerializer.java
@@ -187,7 +187,7 @@ public class PendingSplitsStateSerializer implements SimpleVersionedSerializer<P
         out.writeBoolean(hasTableIsSplitting);
         if (hasTableIsSplitting) {
             ChunkSplitterState chunkSplitterState = state.getChunkSplitterState();
-            out.writeUTF(chunkSplitterState.getCurrentSplittingTableId().toString());
+            out.writeUTF(chunkSplitterState.getCurrentSplittingTableId().toDoubleQuotedString());
             out.writeUTF(
                     SerializerUtils.rowToSerializedString(
                             new Object[] {chunkSplitterState.getNextChunkStart().getValue()}));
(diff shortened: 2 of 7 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-sqlserver-cdc/src/main/java/org/apache/flink/cdc/connectors/sqlserver/source/utils/SqlServerUtils.java
```diff
@@ -361,12 +361,16 @@ private static String quoteSchemaAndTable(TableId tableId) {
...
     }
 
-    public static String quote(String dbOrTableName) {
-        return "[" + dbOrTableName + "]";
+    /**
+     * @link <a
+     *     href="https://learn.microsoft.com/en-us/sql/t-sql/functions/quotename-transact-sql">QUOTENAME</a>
+     */
+    public static String quote(String name) {
+        return "[" + name.replace("]", "]]") + "]";
```
lvyanquan, line 369: It's better to verify this change in testing.

morozov: I could write a separate test for this specific function (e.g. create a table, introspect it and confirm that the column name is what's expected), but I cannot make Flink CDC correctly capture changes from such a table (e.g. `customers [1]` (see the code changes: https://github.com/morozov/flink-cdc/compare/FLINK-37774-sqlserver-special-characters...morozov:flink-cdc:FLINK-37774-sqlserver-special-characters-testing).

If I rename the test table from `customers` to `customers [1]`, the task will fail with the following error:
> To use incremental snapshot, 'scan.incremental.snapshot.chunk.key-column' must be set when the table doesn't have primary keys

The reason is that Debezium's `JdbcConnection#readSchema()` incorrectly introspects the schema of the table. Specifically, it cannot find its columns. A similar problem was fixed with introspecting table names with the backslash in them (https://github.com/debezium/debezium/pull/4206), but introspection of tables with other special characters in the name (which is I believe `%`, `_`, `[` and similar) is still broken.

Alternatively, I can revert this change (although, I'm quite confident that the modified logic is correct).

morozov: Here's how the quotation logic can be validated manually:

1. Create the table. Note that the closing square bracket is doubled: 
   ```sql
   CREATE TABLE [test [1]]] (id INT);
   ```
2. Introspect databas tables:
   ```sql
   SELECT *
   FROM INFORMATION_SCHEMA.TABLES
   WHERE TABLE_TYPE = 'BASE TABLE';
   ```
   Note that the actual table name is `test [1]`:
   | TABLE\_CATALOG | TABLE\_SCHEMA | TABLE\_NAME | TABLE\_TYPE |
   | :--- | :--- | :--- | :--- |
   | customer | dbo | test \[1\] | BASE TABLE |
3. Introspect table columns. Note that the opening square bracket is wrapped in square brackets. This is what Debezium doesn't do, and this is why Flink CDC cannot find the table PK in the reworked test:
   ```sql
   EXEC sp_columns_100 'test [[]1]'
   ``` 
   | TABLE\_QUALIFIER | TABLE\_OWNER | TABLE\_NAME | COLUMN\_NAME | DATA\_TYPE | TYPE\_NAME | PRECISION | LENGTH | SCALE | RADIX | NULLABLE | REMARKS | COLUMN\_DEF | SQL\_DATA\_TYPE | SQL\_DATETIME\_SUB | CHAR\_OCTET\_LENGTH | ORDINAL\_POSITION | IS\_NULLABLE | SS\_IS\_SPARSE | SS\_IS\_COLUMN\_SET | SS\_IS\_COMPUTED | SS\_IS\_IDENTITY | SS\_UDT\_CATALOG\_NAME | SS\_UDT\_SCHEMA\_NAME | SS\_UDT\_ASSEMBLY\_TYPE\_NAME | SS\_XML\_SCHEMACOLLECTION\_CATALOG\_NAME | SS\_XML\_SCHEMACOLLECTION\_SCHEMA\_NAME | SS\_XML\_SCHEMACOLLECTION\_NAME | SS\_DATA\_TYPE |
   | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
   | customer | dbo | test \[1\] | id | 4 | int | 10 | 4 | 0 | 10 | 1 | null | null | 4 | null | null | 1 | YES | 0 | 0 | 0 | 0 | null | null | null | null | null | null | 38 |

lvyanquan: > To use incremental snapshot, 'scan.incremental.snapshot.chunk.key-column' must be set when the table doesn't have primary keys

Does this error mean that we cannot handle such table names through FlinkCDC in actual environments too(not just e2e test)?

I tend to revert this change first, although I believe your change is correct. We should verify and fix this issue in another PR.

morozov: > Does this error mean that we cannot handle such table names through FlinkCDC in actual environments too(not just e2e test)?

It won't work in any environment.

> I tend to revert this change first, although I believe your change is correct. We should verify and fix this issue in another PR.

Sounds good.

lvyanquan: We can copy the dependent debezium code and make changes, just annotate the changes.
Here is an example: https://github.com/apache/flink-cdc/blob/a91f8b7b273bdbb08c451c833ecd5838f8615a70/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlDefaultValueConverter.java#L41

morozov: For the record, the patches that enable capturing tables with special character names in Debezium ended up being quite significant: https://github.com/debezium/debezium/pull/6464, https://github.com/debezium/debezium/pull/6486.

PR #4022, opened 2025-05-21
[FLINK-37577][pipeline-connector/paimon] Fix missing column comment

qg-lin:
JIRA: https://issues.apache.org/jira/browse/FLINK-37824


qg-lin:
@lvyanquan @xiqian PTAL

yuxiqian (commented):
Thanks for @qg-lin's patch! Just one trivial comment on assertion style.

yuxiqian (approved):

PR #4032, opened 2025-05-30
[FLINK-37855-][cdc-connector-mysql] Reduce the time cost in MySqlSchemaUtils#listTables

zhyayun:
[https://issues.apache.org/jira/browse/FLINK-37855](url)
MySqlSchemaUtils#listTables and TableDiscoveryUtils#listTables is called to get the table names of all tables that need to be processed, However, in scenarios involving many databases and tables, this method call will require querying the database many times, resulting in a long processing time.
The optimization idea is to use mysql's metadata information, a connection to get all the library table information.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/utils/MySqlSchemaUtils.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/utils/MySqlSchemaUtils.java
@@ -33,9 +33,12 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -45,6 +48,9 @@ import static org.apache.flink.cdc.connectors.mysql.source.utils.StatementUtils.
 /** Utilities for converting from debezium {@link Table} types to {@link Schema}. */
 public class MySqlSchemaUtils {
 
+    private static final String[] TABLE_QUERY = {"TABLE"};
+    private static final List<String> DB_LIST =
+            Arrays.asList("information_schema", "mysql", "sys", "performance_schema");
     private static final Logger LOG = LoggerFactory.getLogger(MySqlSchemaUtils.class);
 
     public static List<String> listDatabases(MySqlSourceConfig sourceConfig) {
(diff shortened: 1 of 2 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/utils/MySqlSchemaUtils.java
```diff
@@ -45,6 +48,9 @@
 /** Utilities for converting from debezium {@link Table} types to {@link Schema}. */
 public class MySqlSchemaUtils {
 
+    private static final String[] TABLE_QUERY = {"TABLE"};
+    private static final List<String> DB_LIST =
```
lvyanquan: SYSTEM_DB_LIST.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/utils/MySqlSchemaUtils.java
```diff
@@ -58,16 +64,23 @@ public static List<String> listDatabases(MySqlSourceConfig sourceConfig) {
...
-
             List<TableId> tableIds = new ArrayList<>();
-            for (String database : databases) {
-                tableIds.addAll(listTables(jdbc, database));
+            try (Connection connection = jdbc.connection()) {
+                DatabaseMetaData metaData = connection.getMetaData();
+                try (ResultSet resultSet = metaData.getTables(dbName, null, "%", TABLE_QUERY)) {
+                    while (resultSet.next()) {
+                        String database = resultSet.getString("TABLE_CAT");
+                        if (dbName == null && DB_LIST.contains(database)) {
```
lvyanquan: DB_Set will have better performance than DB_LIST.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/TableDiscoveryUtils.java
```diff
@@ -50,54 +54,26 @@ public static List<TableId> listTables(
...
-                        String databaseName = rs.getString(1);
-                        if (databaseFilter.test(databaseName)) {
-                            databaseNames.add(databaseName);
-                        }
+        try (Connection connection = jdbc.connection()) {
+            DatabaseMetaData databaseMetaData = connection.getMetaData();
+            try (ResultSet tableResult = databaseMetaData.getTables(null, null, "%", TABLE_QUERY)) {
+                while (tableResult.next()) {
+                    String dbName = tableResult.getString("TABLE_CAT");
+                    if (!databaseFilter.test(dbName)) {
```
lvyanquan, line 62: We should skip the whole tables of this database directly.

lvyanquan: You can ignore this comment, reducing the number of interactions with the database is more appropriate than reducing calculations.

lvyanquan:
I think the key point is reusing the connection, instead of changing the query sql.
So passing Connection to listTables() method will be better.

PR #4038, opened 2025-06-05
[FLINK-37905] Fix transform failure with non-ascii string literals

yuxiqian:
This closes FLINK-37905.

As reported by @ruanhang1993, any non-ascii string literals in transform expressions will be encoded into something like `"&'\xxxx"`, which is unexpected since Unicodes are fully supported in Flink SQL expressions.

+@aiwenmo

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
@@ -34,6 +34,7 @@ import org.apache.calcite.sql.SqlNodeList;
 import org.apache.calcite.sql.SqlNumericLiteral;
 import org.apache.calcite.sql.fun.SqlCase;
 import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.util.NlsString;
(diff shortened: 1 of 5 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -158,10 +159,10 @@ private static Java.Rvalue translateSqlSqlLiteral(SqlLiteral sqlLiteral) {
         if (sqlLiteral.getValue() == null) {
             return new Java.NullLiteral(Location.NOWHERE);
         }
-        String value = sqlLiteral.getValue().toString();
+        Object value = sqlLiteral.getValue();
         if (sqlLiteral instanceof SqlCharStringLiteral) {
             // Double quotation marks represent strings in Janino.
-            value = "\"" + value.substring(1, value.length() - 1) + "\"";
+            value = "\"" + ((NlsString) value).getValue() + "\"";
```
lvyanquan: value = "\"" + ((SqlCharStringLiteral) sqlLiteral).getValueAs(NlsString.class).getValue() + "\"";

lvyanquan (approved):
LGTM.

PR #4039, opened 2025-06-06
[FLINK-37909-][cdc-connector-mongodb] Update MongoDBStreamFetchTask to renew heartbeatManager

sjmittal:
https://issues.apache.org/jira/browse/FLINK-37909

Update `heartbeatManager` so it has latest `changeStreamCursor`, when it changes in the running while loop.
Right now `heartbeatManager` is created with orginal  `changeStreamCursor` and not renewed whenever `changeStreamCursor` is updated upon Change stream cursor expiry.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mongodb-cdc/src/main/java/org/apache/flink/cdc/connectors/mongodb/source/reader/fetch/MongoDBStreamFetchTask.java
@@ -121,6 +121,7 @@ public class MongoDBStreamFetchTask implements FetchTask<SourceSplitBase> {
                                     "Resume token has expired, fallback to timestamp restart mode");
                         }
                         changeStreamCursor = openChangeStreamCursor(descriptor, resumeTokenExpires);
+                        heartbeatManager = openHeartbeatManagerIfNeeded(changeStreamCursor);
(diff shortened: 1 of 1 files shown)


yuxiqian:
Thanks for @sjmittal's nice investigation, failing CI stages are irrelevant to this change.

Can we add a test case that simulates resume token expiration to verify this change?

sjmittal:
> Thanks for @sjmittal's nice investigation, failing CI stages are irrelevant to this change.
> 
> Can we add a test case that simulates resume token expiration to verify this change?

Currently there are no test cases for `MongoDBStreamFetchTask`. Also this problem usually surfaces after the streaming job is run for days. Not sure how to simulate that.
Looking at the code fix seems obvious that `changeStreamCursor` is recreated but `heartbeatManager` still uses the older one, which essentially causes the `ChangeStreamHistoryLost` exception on the mongodb side.

yuxiqian (approved):
Thanks for clarifying, this change looks reasonable. Will try to add test coverages later.

sjmittal:
> Thanks for clarifying, this change looks reasonable. Will try to add test coverages later.

@yuxiqian Please let me know if anything else needs to be done to get this PR merged.
We would like this fix to be part of whatever next release we are doing for Flink CDC.

yuxiqian:
There will be a patch release (3.4.1) later with bug-fixes soon. Could @sjmittal please backport it to `release-3.4` branch, too?

sjmittal:
> There will be a patch release (3.4.1) later with bug-fixes soon. Could @sjmittal please backport it to `release-3.4` branch, too?

@yuxiqian Please find another PR to backport it to `release-3.4` branch.

https://github.com/apache/flink-cdc/pull/4040

PR #4044, opened 2025-06-17
[FLINK-37963] Fix potential NPE when triggering JobManager failover prematurely

yuxiqian:
This closes FLINK-37963.

Occasionally, test cases related to triggering JM failover might fail, with an NPE thrown when manipulating jobManagerLeaderServices (like https://github.com/apache/flink-cdc/actions/runs/15509047609/job/43667549280).

It is likely to be caused by uninitialised JobManagerService and could be circumvented by invoking `EmbeddedHaServices#getOrCreateJobManagerService` manually.


yuxiqian:
Would @lvyanquan like to take a look?

lvyanquan (approved):
LGTM.

PR #4056, opened 2025-07-08
[FLINK-38045] Build createTableEventCache using TableSchemas from split and Make transform operator stateless

proletarians:
For historical reasons, the current implementation does not utilize the Source to build a schema cache. This requires the TransformOperator to maintain its own complex state.

An optimization is proposed: to build a reusable schema cache at the Source (or in an operator immediately following it).

The primary benefit of this approach is that the transform operator can be made completely stateless. This simplifies its design, improves architectural modularity, and makes the entire pipeline more robust and maintainable.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlSourceReader.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlSourceReader.java
@@ -43,7 +43,6 @@ import org.apache.flink.cdc.connectors.mysql.source.split.SourceRecords;
 import org.apache.flink.cdc.connectors.mysql.source.utils.ChunkUtils;
 import org.apache.flink.cdc.connectors.mysql.source.utils.TableDiscoveryUtils;
 import org.apache.flink.configuration.Configuration;
-import org.apache.flink.connector.base.source.reader.RecordEmitter;
 import org.apache.flink.connector.base.source.reader.RecordsWithSplitIds;
 import org.apache.flink.connector.base.source.reader.SingleThreadMultiplexSourceReaderBase;
 import org.apache.flink.connector.base.source.reader.fetcher.SingleThreadFetcherManager;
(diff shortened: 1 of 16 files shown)


lvyanquan:
Please rebase master to solve the conflict.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlSourceReader.java
```diff
@@ -377,7 +379,7 @@ private void reportFinishedSnapshotSplitsIfNeed() {
             FinishedSnapshotSplitsReportEvent reportEvent =
                     new FinishedSnapshotSplitsReportEvent(finishedOffsets);
             context.sendSourceEventToCoordinator(reportEvent);
-            LOG.debug(
+            LOG.info(
```
lvyanquan, line 382: Avoid unnecessary change.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PreTransformOperator.java
```diff
@@ -144,54 +129,17 @@ public void setup(
...
-                                stateTableChangeInfo.getTableId(),
-                                stateTableChangeInfo.getPreTransformedSchema());
-                // hasAsteriskMap needs to be recalculated after restoring from a checkpoint.
-                cacheTransformRuleInfo(restoredCreateTableEvent);
-            }
-        }
+        // Historically, transform operator maintains internal state to help transforming schemas
+        // correctly after restart.
+        // However, it is not required after
+        // https://code.alibaba-inc.com/ververica/flink-cdc/codereview/21946168 got merged, since
```
lvyanquan: Avoid using non-public links.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlSourceReaderTest.java
```diff
@@ -766,7 +766,7 @@ public void emitRecord(
             }
         }
 
-        private void processElement(
+        protected void processElement(
```
lvyanquan, line 769: Avoid unnecessary change.

PR #4058, opened 2025-07-08
[FLINK-38065] Use curly braces in PostgresOffset#toString()

morozov:
Additionally, `PostgresOffset#getOffset()` is being removed since it's identical to the implementation in its super class.


leonardBang (approved):
Thanks @morozov for the fix, LGTM

PR #4060, opened 2025-07-10
[FLINK-38079] Add Pipeline support for DateType and TimeType

proletarians:
Pipeline currently treats DateType and TimeType as plain INT, so date/time columns lose precision (TIME only to milliseconds, DATE range limited) and cannot be used with built-in temporal functions.


flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
```diff
@@ -85,6 +93,37 @@
  */
 @PublicEvolving
 public class SchemaMergingUtils {
+
```
yuxiqian: We may handle these changes in another PR and focus on supporting `DATE` and `TIME` formats here

flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineTransformITCase.java
```diff
@@ -135,11 +141,21 @@ void cleanup() {
...
+                Arguments.of(ValuesDataSink.SinkApi.SINK_V2, false));
+    }
+
     /** This tests if we can append calculated columns based on existing columns. */
-    @ParameterizedTest
-    @EnumSource
-    void testCalculatedColumns(ValuesDataSink.SinkApi sinkApi) throws Exception {
+    @ParameterizedTest(name = "API version: {0}, initializeMode: {1}")
+    @MethodSource(value = "parameterProvider")
+    void testCalculatedColumns(ValuesDataSink.SinkApi sinkApi, boolean initializeMode)
```
yuxiqian: What is `initializeMode`? Seems it doesn't affect test cases at all.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlFullTypesITCase.java
```diff
@@ -141,6 +143,7 @@ public void testMySql8JsonDataTypesWithUseLegacyJsonFormat() throws Throwable {
 
     @Test
     void testMysql57TimeDataTypes() throws Throwable {
+        UniqueDatabase usedDd = fullTypesMySql57Database;
```
yuxiqian: Please double check if these changes are necessary, and revert these to keep this PR minimum.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-debezium/src/main/java/org/apache/flink/cdc/debezium/event/DebeziumSchemaDataTypeInference.java
```diff
@@ -48,6 +48,7 @@ public class DebeziumSchemaDataTypeInference implements SchemaDataTypeInference,
 
     private static final long serialVersionUID = 1L;
 
+    public static final String DEBEZIUM_DATE_SCHEMA_NAME = "io.debezium.time.Date";
```
yuxiqian: Seems redundant as it's an alias of `io.debezium.time.Date.SCHEMA_NAME`?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/writer/BinaryWriter.java
```diff
@@ -98,9 +105,21 @@ static void write(
...
             case INTEGER:
+                writer.writeInt(pos, (int) o);
+                break;
             case DATE:
+                if (o instanceof DateData) {
+                    writer.writeDate(pos, (DateData) o);
+                } else {
+                    writer.writeInt(pos, (int) o);
+                }
+                break;
```
yuxiqian: We may add some comments here, clarifying that the `writeInt` paths are kept for compatibility.

yuxiqian (changes requested):
Thanks for @proletarians' nice work, just left some trivial comments.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-debezium/src/main/java/org/apache/flink/cdc/debezium/event/DebeziumEventDeserializationSchema.java
```diff
@@ -284,22 +286,22 @@ protected Object convertToFloat(Object dbzObj, Schema schema) {
...
-        return (int) TemporalConversions.toLocalDate(dbzObj).toEpochDay();
+        return DateData.fromLocalDate(TemporalConversions.toLocalDate(dbzObj));
     }
 
     protected Object convertToTime(Object dbzObj, Schema schema) {
         if (dbzObj instanceof Long) {
             switch (schema.name()) {
                 case MicroTime.SCHEMA_NAME:
-                    return (int) ((long) dbzObj / 1000);
+                    return TimeData.fromNanoOfDay((long) dbzObj * 1000);
```
Mrart: TimeData implement From MillOfDay replace *1000 here?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-debezium/src/main/java/org/apache/flink/cdc/debezium/event/DebeziumEventDeserializationSchema.java
```diff
@@ -284,22 +286,22 @@ protected Object convertToFloat(Object dbzObj, Schema schema) {
...
                 case NanoTime.SCHEMA_NAME:
-                    return (int) ((long) dbzObj / 1000_000);
+                    return TimeData.fromNanoOfDay((long) dbzObj);
             }
         } else if (dbzObj instanceof Integer) {
-            return dbzObj;
+            return TimeData.fromNanoOfDay((int) dbzObj * 1_000_000L);
```
Mrart: TimeData implement fromSecondOfDay?

yuxiqian:
> DebeziumEventDeserializationSchema#convertToTime Should we also change the implementation of this class etc？

Updated.

Mrart (commented):
I let some comment.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-debezium/src/main/java/org/apache/flink/cdc/debezium/event/DebeziumEventDeserializationSchema.java
yuxiqian: 1 second = 1,000 milliSeconds = 1,000,000 microSeconds = 1,000,000,000 nanoSeconds.

So we have to use `fromMillisOfDay(x / 1000)`, which isn't better than `fromNanoOfDay(x * 1000)`

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-debezium/src/main/java/org/apache/flink/cdc/debezium/event/DebeziumEventDeserializationSchema.java
yuxiqian: Addressed, seems it should be `fromMillisOfDay`.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-debezium/src/main/java/org/apache/flink/cdc/debezium/event/DebeziumEventDeserializationSchema.java
Mrart: It's possible to use these schemes in other places as well, so that you don't have to x*1000 everywhere, which will reduce the number of low-level errors when writing code

yuxiqian: Makes sense, added `fromMicroOfDay` convenience init method.

Mrart:
LGTM

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/TimeData.java
```diff
@@ -0,0 +1,95 @@
...
+        return new TimeData(millisOfDay);
+    }
+
+    public static TimeData fromMicroOfDay(long microOfDay) {
+        return new TimeData((int) (microOfDay / MILLIS_TO_MICRO));
+    }
+
+    public static TimeData fromNanoOfDay(long nanoOfDay) {
+        // millisOfDay should not exceed 86400000, which is safe to fit into INT.
+        return new TimeData((int) (nanoOfDay / MILLIS_TO_NANO));
```
Mrart, line 53: Why only down to ms?not ns？

yuxiqian: We'd better keep the same memory layout as previous BinaryRecordData.

Mrart: It's much better to have better timing accuracy for downstream processing where sorting is required

Mrart: But I seem ok like this.

Mrart:
LGTM

yuxiqian:
Just reverted an accidental change that breaks Postgres Pipeline CI, PTAL @lvyanquan

lvyanquan (approved):
LGTM.

yuxiqian (approved):
+1, thanks to all involved.

PR #4069, opened 2025-07-31
[minor] Fix potential sql connection statement issue

lvyanquan:
The vulnerability exists in the `quote()` function implementations that fail to properly escape special characters:

**Vulnerable Implementations:**

1. **MySQL** (Line 226):
```java
public static String quote(String dbOrTableName) {
    return "`" + dbOrTableName + "`";  // No backtick escaping
}
```

2. **SQL Server** (Line 367):
```java
public static String quote(String name) {
    return "[" + name + "]";  // No bracket escaping
}
```

3. **Oracle** (Line 232):
```java
public static String quote(String dbOrTableName) {
    return "\"" + dbOrTableName + "\"";  // No quote escaping
}
```

4. **DB2** (Line 358):
```java
public static String quote(String dbOrTableName) {
    return "\"" + dbOrTableName + "\"";  // No quote escaping
}
```

**Correct Implementation (PostgreSQL only):**
```java
public static String quote(String name) {
    return "\"" + name.replace("\"", "\"\"") + "\"";  // Proper escaping
}
```

However, this issue should not occur in FlinkCDC, as FlinkCDC will first retrieve the table names of all tables and then use regular expressions for matching.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlEventDeserializer.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlEventDeserializer.java
@@ -48,6 +48,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import static io.debezium.connector.AbstractSourceInfo.DATABASE_NAME_KEY;
+import static io.debezium.connector.AbstractSourceInfo.TABLE_NAME_KEY;
 import static org.apache.flink.cdc.connectors.mysql.source.utils.RecordUtils.getHistoryRecord;
 
 /** Event deserializer for {@link MySqlDataSource}. */
@@ -146,8 +148,11 @@ public class MySqlEventDeserializer extends DebeziumEventDeserializationSchema {
 
     @Override
     protected TableId getTableId(SourceRecord record) {
-        String[] parts = record.topic().split("\\.");
-        return TableId.tableId(parts[1], parts[2]);
+        Struct value = (Struct) record.value();
+        Struct source = value.getStruct(Envelope.FieldName.SOURCE);
+        String dbName = source.getString(DATABASE_NAME_KEY);
+        String tableName = source.getString(TABLE_NAME_KEY);
+        return TableId.tableId(dbName, tableName);
     }
 
     @Override
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlPipelineITCase.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlPipelineITCase.java
@@ -45,6 +45,7 @@ import org.apache.flink.cdc.common.types.RowType;
 import org.apache.flink.cdc.common.utils.Preconditions;
 import org.apache.flink.cdc.connectors.mysql.factory.MySqlDataSourceFactory;
 import org.apache.flink.cdc.connectors.mysql.source.config.MySqlSourceConfigFactory;
+import org.apache.flink.cdc.connectors.mysql.source.utils.StatementUtils;
 import org.apache.flink.cdc.connectors.mysql.table.StartupOptions;
 import org.apache.flink.cdc.connectors.mysql.testutils.MySqlContainer;
 import org.apache.flink.cdc.connectors.mysql.testutils.MySqlVersion;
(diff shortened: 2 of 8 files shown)


lvyanquan:
Hi @yuxiqian, PTAL.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/StatementUtils.java
```diff
@@ -225,7 +225,7 @@ public static PreparedStatement readTableSplitDataStatement(
     }
 
     public static String quote(String dbOrTableName) {
-        return "`" + dbOrTableName + "`";
+        return "`" + dbOrTableName.replace("`", "``") + "`";
```
yuxiqian, line 228: It would be nice to link docs URL ([like this](https://dev.mysql.com/doc/refman/8.4/en/identifiers.html)) here to clarify the quoting rules.

yuxiqian (commented):
Thanks for @lvyanquan's quick fix, the fix looks reasonable. However a quick local test revealed that some code doesn't use `quote` function properly, like the following snippet:

```java
// MySqlPipelineRecordEmitter.java
private String showCreateTable(JdbcConnection jdbc, TableId tableId) {
        final String showCreateTableQuery =
                String.format("SHOW CREATE TABLE `%s`.`%s`", tableId.catalog(), tableId.table());
        ...
```

Do you think we need some tests covering SQL injection vulnerabilities?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/StatementUtils.java
lvyanquan: In fact, we do not directly use the table names filled in by users for SQL concatenation.
1. We will use the table name filled in by the user for regular matching, and then find the `correct table names` in the database.
2. Then we concatenate the sql using the `correct table names`.

So the SQL we concatenate must use the actual table names that already existed.
This is just to avoid potential vulnerability code here (which would not actually occur).

yuxiqian: Seems reasonable. This issue, if we can't capture tables with backtick characters, is more like a limitation of function instead of a vulnerability issue.

lvyanquan: The user's table name itself may contain the risk of SQL injection, and testing has been updated to verify the correctness of SQL and pipeline jobs.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlEventDeserializer.java
```diff
@@ -146,8 +148,11 @@ protected boolean isSchemaChangeRecord(SourceRecord record) {
 
     @Override
     protected TableId getTableId(SourceRecord record) {
-        String[] parts = record.topic().split("\\.");
-        return TableId.tableId(parts[1], parts[2]);
+        Struct value = (Struct) record.value();
+        Struct source = value.getStruct(Envelope.FieldName.SOURCE);
+        String dbName = source.getString(DATABASE_NAME_KEY);
+        String tableName = source.getString(TABLE_NAME_KEY);
+        return TableId.tableId(dbName, tableName);
```
lvyanquan, line 155: Copy from [SourceRecordUtils#getTableId](https://github.com/apache/flink-cdc/blob/5eaf9eb86b81a1fe80a06abe7c39a85cadc9b35e/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/utils/SourceRecordUtils.java#L119).
There may be issues when the table name contains special characters in record.topic() method, for example, `sqlInjection; DROP TABLE important_data;` will be converted to `sqlInjection__DROP_TABLE_important_data_`.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlPipelineITCase.java
```diff
@@ -254,6 +254,148 @@ void testInitialStartupMode() throws Exception {
                 .isEqualTo(expectedBinlog);
     }
 
+    @Test
+    void testSqlInjection() throws Exception {
+        inventoryDatabase.createAndInitialize();
+        env.setParallelism(1);
+        String sqlInjectionTable = "sqlInjection; DROP TABLE important_data;";
```
yuxiqian:
```suggestion
        String sqlInjectionTable = "sqlInjection`; DROP TABLE important_data; --";
```

lvyanquan: Addressed.

PR #4070, opened 2025-07-31
[FLINK-38141][pipeline-connector/iceberg] Fix iceberg connector incorrect type mapping

doovvv:
According to the Iceberg documentation https://iceberg.apache.org/docs/nightly/flink/#flink-to-iceberg, the CDC type TINYINT should be mapped to the Iceberg type INTEGER. However, in the code implementation of the Iceberg sink, I noticed that TINYINT is actually mapped to the Iceberg type BOOLEAN.


lvyanquan (approved):
+1. Good catch.

PR #4071, opened 2025-08-02
[minor][improvement] Add subdirectory scrapping to add all .jar in classpath

CatOrLeader:
## Rationale

I was stuck with this issue for a while when deploying CDC pipelines on my cluster. It seems that it is not convenient to get .jar files only at the first level of depth, so I want to fix it and add a more convenient search for files.

## Solution

Add all-depth search for .jar files

### DIFF
--- a/flink-cdc-dist/src/main/flink-cdc-bin/bin/flink-cdc.sh
+++ b/flink-cdc-dist/src/main/flink-cdc-bin/bin/flink-cdc.sh
@@ -64,11 +64,11 @@ FLINK_CDC_LOG="$FLINK_CDC_HOME"/log
 # Build Java classpath
 CLASSPATH=""
 # Add Flink libraries to the classpath
-for jar in "$FLINK_HOME"/lib/*.jar; do
+for jar in $(find "$FLINK_HOME"/lib -name "*.jar" -type f); do
   CLASSPATH=$CLASSPATH:$jar
 done
 # Add Flink CDC libraries to classpath
-for jar in "$FLINK_CDC_LIB"/*.jar; do
+for jar in $(find "$FLINK_CDC_LIB" -name "*.jar" -type f); do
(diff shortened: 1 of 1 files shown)


lvyanquan:
Hi @CatOrLeader, could you elaborate more on your use cases and the issues you've encountered?

CatOrLeader:
Hello, @lvyanquan! Let me explain a little bit of the rationale behind this: for the engineer, it is common sense to structure the environment he is working in, so there occur situations like the following

A developer/engineer/DevOps creates a custom Flink CDC image and puts multiple .jar files into different folders, for example:

```bash
.
├── lib
|   ├── some-custom-code.jar
│   ├── metric
│   │   ├── oshi-core-6.1.5.jar
│   │   ├── jna-5.10.0.jar
│   │   └── jna-platform-5.10.0.jar
│   ├── cdc
│   │   ├── flink-cdc-common-3.4.0.jar
│   │   ├── flink-cdc-runtime-3.4.0.jar
│   │   ├── flink-sql-connector-postgres-cdc-3.4.0.jar
│   │   ├── ojdbc8-19.3.0.0.jar
│   │   └── xdb-19.3.0.0.jar
│   └── utils
│       ├── jackson-databind-2.18.2.jar
│       ├── jackson-core-2.18.2.jar
│       ├── jackson-annotations-2.18.2.jar
│       ├── sshj-0.40.0.jar
│       └── bcprov-jdk18on-1.80.jar
```

Then, we experience a small inconvenience when someone tries to run a Flink CDC job and an exception occurs due to not finding any dependencies or even a CDC connector. It can take (and has taken) a significant amount of time to debug and identify the issue, even if the developer sees that all dependencies are present. Additionally, Flink does not prevent the placement of JAR files in subdirectories, and it is not clear when the CDC library requires exactly one level of depth for storing dependencies.

lvyanquan (approved):
+1 for this.

PR #4072, opened 2025-08-03
[FLINK-38183] Fix the problem of data loss in the GTID disordered scenario

lzshlzsh:
As the design of [https://github.com/apache/flink-cdc/pull/2220](http://example.com/)，CDC only cares about the maximum GTID position and starts from it. For example, if reading from gtid offset 1-7:9-10, it will automatically adjust to read from 1-10, which causes an error in skipping gitd site 8, thus losing data. In particular, when gtid bit 8 is a large transaction, it will cause more serious data loss. We have encountered this problem many times in the production environment.

More details can be found in [FLINK-38183].


leonardBang:
@ruanhang1993 would you take a look this PR?

mielientiev:
This is a duplicate PR for the same issue that I opened and proposed fix awhile ago
https://github.com/apache/flink-cdc/pull/3845

cc @leonardBang

lvyanquan:
Hi @lzshlzsh, this can be closed as https://github.com/apache/flink-cdc/pull/3845 was merged.

PR #4074, opened 2025-08-04
[FLINK-38185][pipeline-connector][iceberg] Correctly handle the type conversion of TIMESTAMP_TITH_TIME_ZONE.

lvyanquan:
Fix TIMESTAMP_TITH_TIME_ZONE type conversion and failed case of IcebergWriterTest.testWriteWithAllSupportedTypes.


lvyanquan:
Hi @yuxiqian, maybe you can help to review this.

yuxiqian (commented):
+1. Since we're mapping CDC `TIMESTAMP_TZ` to Iceberg timestamps with timezone, it's reasonable to convert based on the same global instant.

As suggested by @Jzjsnow, We may support the alternative conversion path from `TIMESTAMP_TZ` to Iceberg timestamps w/o timezone, where we should keep the same calendar literals.

PR #4078, opened 2025-08-06
Paimon connector - support table specific table.properties config

nickdelnano:
This PR adds a new format for defining table properties as `specific.table.properties.<database_name>.<table_name>.<property>`. This allows defining properties that are specific to a table. The table specific property takes precedence.

Paimon sink supports [defining table properties](https://nightlies.apache.org/flink/flink-cdc-docs-master/docs/connectors/pipeline-connectors/paimon/#pipeline-connector-options) as `table.properties.*`. They apply to ALL tables. In some cases it is warranted to set table properties based on the table use case. Some examples are compaction settings, Iceberg compatibility settings, etc. This PR allows this.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/PaimonDataSinkFactory.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/PaimonDataSinkFactory.java
@@ -39,6 +39,7 @@ import java.util.Objects;
 import java.util.Set;
 
 import static org.apache.flink.cdc.connectors.paimon.sink.PaimonDataSinkOptions.PREFIX_CATALOG_PROPERTIES;
+import static org.apache.flink.cdc.connectors.paimon.sink.PaimonDataSinkOptions.PREFIX_SPECIFIC_TABLE_PROPERTIES;
 import static org.apache.flink.cdc.connectors.paimon.sink.PaimonDataSinkOptions.PREFIX_TABLE_PROPERTIES;
 
 /** A {@link DataSinkFactory} to create {@link PaimonDataSink}. */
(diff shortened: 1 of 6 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/PaimonDataSinkFactory.java
```diff
@@ -49,15 +50,21 @@ public class PaimonDataSinkFactory implements DataSinkFactory {
...
         Map<String, String> tableOptions = new HashMap<>();
+        Map<String, String> tableSpecificOptions = new HashMap<>();
         allOptions.forEach(
                 (key, value) -> {
                     if (key.startsWith(PREFIX_TABLE_PROPERTIES)) {
-                        tableOptions.put(key.substring(PREFIX_TABLE_PROPERTIES.length()), value);
+                        String keyWithoutPrefix = key.substring(PREFIX_TABLE_PROPERTIES.length());
+                        // General table option: table.properties.<property>
+                        tableOptions.put(keyWithoutPrefix, value);
+                    } else if (key.startsWith(PREFIX_SPECIFIC_TABLE_PROPERTIES)) {
```
tub, line 65: Shouldn't we flip these so that PREFIX_SPECIFIC_TABLE_PROPERTIES take priority of PREFIX_TABLE_PROPERTIES?

tub: Actually… I just re-read it and realised it's in the for-each in parsing, not when they're actually applied.

nickdelnano: That is the behavior. Test case `testTableSpecificOptionsPrecedenceOverTableOptions` demonstrates it.

The function here parses the configuration. Then [here](https://github.com/apache/flink-cdc/pull/4078/files#diff-5b829cedf9c430c9efc3a14629963c7fa7f3070eb6fbc470f4b0612a8ed63af3) the table specific overrides are applied when a table is created and its properties are set.

nickdelnano:
@lvyanquan can you review this change?

lvyanquan:
We can modify the table options through transform, but I agree to provide more configurations in the sink as alternative solutions, refer to https://nightlies.apache.org/flink/flink-cdc-docs-release-3.4/docs/core-concept/transform/#specify-table-creation-configuration. 

Compared to transform, the current configuration requires too many parameters to be set for situations where many tables need to be parameterized(users usually synchronize a large number of tables in one job). Can we provide a solution for parameter configuration based on regular expressions (Selectors)?

PR #4080, opened 2025-08-06
[FLINK-38194][pipeline-connector/iceberg] Iceberg connector supports auto-creating namespace

doovvv:
Currently, Icerberg connector does not support auto-creating namaspace. When using hive matastore, it will throw a exception if a user tries write data to iceberg which does not have a namespace.
In the Iceberg sink, when a createTableEvent is encountered, we can first create the namespace and then create the table.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/IcebergMetadataApplier.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/IcebergMetadataApplier.java
@@ -171,8 +171,8 @@ public class IcebergMetadataApplier implements MetadataApplier {
             if (!catalog.tableExists(tableIdentifier)) {
                 catalog.createTable(tableIdentifier, icebergSchema, partitionSpec, tableOptions);
                 LOG.info(
-                        "Spend {} ms to create iceberg table",
-                        System.currentTimeMillis() - startTimestamp);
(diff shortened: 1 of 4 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/IcebergMetadataApplier.java
```diff
@@ -159,6 +170,9 @@ private void applyCreateTable(CreateTableEvent event) {
             PartitionSpec partitionSpec = builder.build();
             if (!catalog.tableExists(tableIdentifier)) {
                 catalog.createTable(tableIdentifier, icebergSchema, partitionSpec, tableOptions);
+                LOG.info(
+                        "Spend {} ms to create iceberg table",
+                        System.currentTimeMillis() - startTimestamp);
```
lvyanquan: It's better to add table name here too.

lvyanquan:

doovvv:
> Run 'mvn spotless:apply' to fix these violations.

Done.

lvyanquan:
Error:  src/test/java/org/apache/flink/cdc/connectors/iceberg/sink/utils/HiveContainer.java:[17] (whitespace) EmptyLineSeparator: 'package' should be separated from previous statement.

doovvv:
> Error: src/test/java/org/apache/flink/cdc/connectors/iceberg/sink/utils/HiveContainer.java:[17] (whitespace) EmptyLineSeparator: 'package' should be separated from previous statement.

resolved.

lvyanquan (approved):
LGTM.

PR #4094, opened 2025-08-14
[FLINK-38204]use getLatestEvolvedSchema to get Schema in SessionManageOperator in case of using route.

proletarians:
changing the method of retrieving the schema from getLatestOriginalSchema to getLatestEvolvedSchema


lvyanquan (approved):
LGTM.

PR #4095, opened 2025-08-14
[FLINK-38244] When the column name of mysql supports case-insensitive, fix the error that column name not being found.

loserwang1024:
Though in [FLINK-37634](https://issues.apache.org/jira/browse/FLINK-37634), the table name of mysql supports case insensitive, fix the error that table name not being found. When mysql supports case insensitive, column names also doesn't matches(one ddl upper, while another not.


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
```diff
@@ -319,15 +319,21 @@ public void exitAlterByChangeColumn(MySqlParser.AlterByChangeColumnContext ctx)
         parser.runIfNotNull(
                 () -> {
                     Column column = columnDefinitionListener.getColumn();
+                    String oldColumnName =
+                            isTableIdCaseInsensitive ? column.name().toLowerCase() : column.name();
```
lvyanquan: It's better to use `toLowerCase (Locale. ROOT)` uniformly.

PR #4096, opened 2025-08-15
[FLINK-38248] The default value of '0000-00-00 00:00:00' for MySQL TIMESTAMP fields is not supported in downstream systems.

suhwan-cheon:
issue: https://issues.apache.org/jira/browse/FLINK-38248

description
---
This PR resolves the schema evolution error that occurs when MySQL's default timestamp default value '0000-00-00 00:00:00' is processed by the Doris sink connector.
The fix automatically converts MySQL's invalid timestamp to Doris-compatible default value during table creation and column addition operations. ('0000-00-00 00:00:00' -> '1970-01-01 00:00:00')

Future Work
---
If this is the correct approach, I will proceed with the following.
Similar MySQL timestamp compatibility fixes will be applied to other downstream connectors (Paimon, StarRocks, etc.) in follow-up PRs to ensure consistent behavior across all pipeline connectors.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/utils/DorisSchemaUtils.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/utils/DorisSchemaUtils.java
@@ -46,6 +46,8 @@ public class DorisSchemaUtils {
     public static final String DEFAULT_DATE = "1970-01-01";
     public static final String DEFAULT_DATETIME = "1970-01-01 00:00:00";
 
+    public static final String MYSQL_INVALID_TIMESTAMP = "0000-00-00 00:00:00";
+    
     /**
      * Get partition info by config. Currently only supports DATE/TIMESTAMP AUTO RANGE PARTITION and
      * doris version should greater than 2.1.6
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/test/java/org/apache/flink/cdc/connectors/doris/sink/DorisMetadataApplierTest.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/test/java/org/apache/flink/cdc/connectors/doris/sink/DorisMetadataApplierTest.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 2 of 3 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/test/java/org/apache/flink/cdc/connectors/doris/sink/DorisMetadataApplierTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.lang.reflect.Method;
+
+/** Unit tests for {@link DorisMetadataApplier}. */
+class DorisMetadataApplierTest {
+
+    @Test
+    public void testMysqlTimestampDefaultValueConversion() throws Exception {
```
yuxiqian: It's a little suspicious to test an internal method via reflection. How about testing this end-to-end in `DorisMetadataApplierITCase`?

suhwan-cheon: I agree. I will transfer the code to DorisMetadataApplierITCase and proceed with end-to-end testing.

suhwan-cheon: @yuxiqian 
hello! I attached the test results to the description after modifying the code!

lvyanquan:
Hi @suhwan-cheon, would you like to check if this problem existed in StarRocks/Paimon connector too?
If you are not familiar with these two connectors(or you don't have much time), we can do this in another pr.

suhwan-cheon:
> Hi @suhwan-cheon, would you like to check if this problem existed in StarRocks/Paimon connector too? If you are not familiar with these two connectors(or you don't have much time), we can do this in another pr.

@lvyanquan
Sure! I want do this 🙌
I have modified it to fit the existing code of StarRocks/Paimon.
However, please note that the default time criteria for each DB have been modified to follow unix epoch time. (1970-01-01 00:00:00)

lvyanquan:
> However, please note that the default time criteria for each DB have been modified to follow unix epoch time. (1970-01-01 00:00:00)

Yes, otherwise it will lead to job failure.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-doris/src/main/java/org/apache/flink/cdc/connectors/doris/utils/DorisSchemaUtils.java
```diff
@@ -46,6 +46,8 @@ public class DorisSchemaUtils {
     public static final String DEFAULT_DATE = "1970-01-01";
     public static final String DEFAULT_DATETIME = "1970-01-01 00:00:00";
 
+    public static final String MYSQL_DEFAULT_TIMESTAMP_VALUE = "0000-00-00 00:00:00";
```
lvyanquan: Maybe we can use `INVALID_OR_MISSING_DATATIME` to avoid binding to a specific upstream database?

suhwan-cheon: I agree, the same thing could happen with other upstream DBs.
I changed about that.

suhwan-cheon:
@lvyanquan 
Hello!
The test seems to fail intermittently regardless of this PR.
What's the problem?

lvyanquan:
This is caused by the instability of the SQLServer test case and has nothing to do with this PR, we will try to fix it later.

lvyanquan:
CI passed, merged.

patrickw1025:
we still encounter this issue, if the default datetime is '0000-00-00 00:00:00.000000'

PR #4097, opened 2025-08-15
[Flink-38238] Support varchar(0) and map to cdc string type

proletarians:
The main changes in this code review include adding support for MySQL's VARCHAR(0) type in the test cases.


lvyanquan (approved):
LGTM.

PR #4104, opened 2025-08-21
[FLINK-36165][source-connector/mysql] Support capturing snapshot data with conditions

Mrart:
issue link: https://issues.apache.org/jira/browse/FLINK-36165

fixed https://github.com/apache/flink-cdc/pull/3776


docs/content/docs/connectors/pipeline-connectors/mysql.md
```diff
@@ -293,6 +293,17 @@ pipeline:
...
+      <td>scan.snapshot.filters</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">(none)</td>
+      <td>String</td>
+      <td>When reading a table snapshot, the rows of captured tables will be filtered using the specified filter expression (AKA a SQL WHERE clause). <br>
+          By default, no filter is applied, meaning the entire table will be synchronized. <br>
+          A colon (:) separates table name and filter expression, while a semicolon (;) separate multiple filters, 
+          e.g. `db1.user_table_[0-9]+:id > 100;db[1-2].[app|web]_order_\\.*:id < 0;`.
+      </td>
+    </tr>
```
SML0127, line 306: How about explicitly stating that the filter conditions are combined using `AND` and that it has nothing to do with the binlog step?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/SnapshotFilterUtils.java
```diff
@@ -0,0 +1,71 @@
...
+        Map<Selectors, String> snapshotFilters = toSelector(filters);
+
+        String filter = null;
+        for (Selectors selector : snapshotFilters.keySet()) {
+            if (selector.isMatch(
+                    org.apache.flink.cdc.common.event.TableId.tableId(
+                            tableId.catalog(), tableId.table()))) {
+                filter = snapshotFilters.get(selector);
+                break;
+            }
```
SML0127, line 67: Can we fail fast on unknown/nonexiststent columns in filters?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/utils/StatementUtils.java
```diff
@@ -56,6 +61,31 @@ public static Object[] queryMinMax(JdbcConnection jdbc, TableId tableId, String
...
+                    if (!rs.next()) {
+                        // this should never happen
+                        throw new SQLException(
+                                String.format(
+                                        "No result returned after running query [%s]", cntQuery));
+                    }
+                    return rs.getLong(1);
+                });
+    }
+
```
SML0127, line 88: Because the filter is applied during split planning, the distribution factor may fall outside the configured bounds(0.05 ~1,000), affecting snapshot performance. I’d appreciate your thoughts on this.

SML0127 (commented):
Thanks for the PR. I've left a few minor comments.

yuxiqian:
It seems the author of #3776 is actively working on their PR, and there's some duplicated work... It would be nice if we can discuss & implement this nice feature at one place.

PR #4105, opened 2025-08-22
[FLINK-38275][Pipeline-connectors][doris]Cannot create a table when the upstream has no primary key and the first column is String

JNSimba:
If the upstream model lacks PrimaryKeys and the first column is a String, the Doris Sink will identify the model as a duplicate and will not set keys. In this case, an error message will be reported when creating the table: `errCode = 2, detailMessage = String Type should not be used in key column[name]`.

Doris-Flink-Connector handles table creation. https://github.com/apache/doris-flink-connector/blob/master/flink-doris-connector/src/main/java/org/apache/doris/flink/catalog/doris/DorisSchemaFactory.java#L226-L227 If the key is a String, the type is changed to Varchar(65535). Therefore, the first key set here can be used as the table key.


lvyanquan (approved):
LGTM.

PR #4106, opened 2025-08-22
[FLINK-38276][cdc-common] PaimonWriter does not invalidate cache when…

tbpure:
invalidateTable when receiving SchemaChangeEvent to prevent retrieving outdated schema


lvyanquan:
Please add more description about the necessity of this change.

PR #4112, opened 2025-09-02
[FLINK-38310] Fix handling binary boundaries in FinishedSnapshotSplitInfo

morozov:
I believe this change is backward-compatible as `FinishedSnapshotSplitInfo` hashes are not used anywhere. Even if it was, the BC break would only impact the cases where the keys are binary (which are broken anyways).


lvyanquan:
+1 for this.

I checked and found that there are currently no places using the FinishedSnapshotSplitInfo#equals and FinishedSnapshotSplitInfo#hashcode methods, so this change is safe.

lvyanquan (approved):

PR #4113, opened 2025-09-03
[Flink-38277] [postgres]drop slot when startup.mode=snapshot

Mrart:
[Flink-38277] [postgres]drop slot when startup.mode=snapshot

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresSourceReader.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresSourceReader.java
@@ -144,7 +144,7 @@ public class PostgresSourceReader extends IncrementalSourceReaderWithCommit {
             if (sourceSplit.isStreamSplit()) {
                 StreamSplit streamSplit = sourceSplit.asStreamSplit();
                 if (this.sourceConfig.getStartupOptions().isSnapshotOnly()
-                        || streamSplit
+                        && streamSplit
                                 .getStartingOffset()
                                 .isAtOrAfter(streamSplit.getEndingOffset())) {
                     PostgresDialect dialect = (PostgresDialect) this.dialect;


Mrart:
@loserwang1024 PTLK

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/fetch/PostgresSourceFetchTaskContext.java
```diff
@@ -148,8 +148,11 @@ public void configure(SourceSplitBase sourceSplitBase) {
                             dbzConfig
                                     .getConfig()
                                     .edit()
-                                    // never drop slot for stream split, which is also global split
-                                    .with(DROP_SLOT_ON_STOP.name(), false)
+                                    // When startup.mode='snapshot' drop slot for stream split,
+                                    // which is also global split
```
loserwang1024: What if a exception is thrown, then job fails. If drop in this case, the job cannot restart! I do think we should do it in org.apache.flink.connector.base.source.reader.SourceReaderBase#onSplitFinished. We should determine whether it's is finished: stream split and endingOffset >= startingOffset

Mrart: In our current implementation of deleting backfill slots, in snapshot mode, there is no need to restart. You just have to run it again with overwrite mode？ Also Streaming job，It's the same thing.

loserwang1024: I don't think so, If the job success to finish the stream split, it does no need to restart. However, If fails before complete stream split, it means that the job is not finished, and need to restart with checkpoint .

For batch mode, it has no checkpoint, thus no need to restart with checkpont. But streaming job, it do need.  Otherwise, cost is too much.
@LYanquan , WDYT?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/fetch/PostgresScanFetchTask.java
```diff
@@ -193,16 +191,16 @@ private void maybeCreateSlotForBackFillReadTask(
...
-            PostgresReplicationConnection replicationConnection, boolean skipSnapshotBackfill) {
+            PostgresConnection connection, String slotName, boolean skipSnapshotBackfill) {
         // if skip backfill, no need to create slot here
         if (skipSnapshotBackfill) {
             return;
         }
 
         try {
-            replicationConnection.close(true);
+            connection.dropReplicationSlot(slotName);
```
loserwang1024: Why needs change here?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresSourceReader.java
```diff
@@ -133,6 +135,17 @@ public void notifyCheckpointComplete(long checkpointId) throws Exception {
...
+    protected void onSplitFinished(Map finishedSplitIds) {
+        super.onSplitFinished(finishedSplitIds);
+
+        if (this.sourceConfig.getStartupOptions().isSnapshotOnly()) {
+            PostgresDialect dialect = (PostgresDialect) this.dialect;
+            boolean removed = dialect.removeSlot(dialect.getSlotName());
+            LOG.info("Remove slot '{}' result is {}.", dialect.getSlotName(), removed);
+        }
+    }
+
```
loserwang1024:
```suggestion
@Override
    protected void onSplitFinished(Map<String, SourceSplitState> finishedSplitIds) {
        super.onSplitFinished(finishedSplitIds);

        for (SourceSplitState splitState : finishedSplitIds.values()) {
            SourceSplitBase sourceSplit = splitState.toSourceSplit();
            if (sourceSplit.isStreamSplit()  ) {
                StreamSplit streamSplit = sourceSplit.asStreamSplit();
                if(streamSplit.getStartingOffset().isAtOrAfter(streamSplit.getEndingOffset())){
                    PostgresDialect dialect = (PostgresDialect) this.dialect;
                    boolean removed = dialect.removeSlot(dialect.getSlotName());
                    LOG.info("Remove slot '{}' result is {}.", dialect.getSlotName(), removed);
                }

            }
        }
    }

```

It seems that even a snapshot split is finished, the global slot name will be removed.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresSourceReader.java
```diff
@@ -133,6 +136,25 @@ public void notifyCheckpointComplete(long checkpointId) throws Exception {
...
+    protected void onSplitFinished(Map finishedSplitIds) {
+        super.onSplitFinished(finishedSplitIds);
+        for (Object splitState : finishedSplitIds.values()) {
+            SourceSplitBase sourceSplit = ((SourceSplitState) splitState).toSourceSplit();
+            if (sourceSplit.isStreamSplit()) {
+                StreamSplit streamSplit = sourceSplit.asStreamSplit();
+                if (this.sourceConfig.getStartupOptions().isSnapshotOnly()
+                        || streamSplit
+                                .getStartingOffset()
+                                .isAtOrAfter(streamSplit.getEndingOffset())) {
```
loserwang1024:
```suggestion
               this.sourceConfig.getStartupOptions().isSnapshotOnly()
                        && streamSplit
                                .getStartingOffset()
                                .isAtOrAfter(streamSplit.getEndingOffset()
```

loserwang1024 (approved):
LGTM

PR #4119, opened 2025-09-16
[FLINK-38358][pipeline-connector][mysql] Ignore trailing spaces in table capturing rules.

lvyanquan:
Ignore trailing spaces in table capturing rules like `tables:  foo.bar , foo.baz `.
This is not actually completely forward compatible, but we believe that using spaces in table names is very dangerous(could not guarantee that all downstream systems support this configuration).

For those who want to capture a table with space in table name, they can use `\s` to replace the ` `.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlTablePatternMatchingTest.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlTablePatternMatchingTest.java
@@ -167,6 +167,23 @@ class MySqlTablePatternMatchingTest extends MySqlSourceTestBase {
     }
 
     @Test
+    void testMatchingTablesWithSpacedRules() {
+        List<String> spacedRules =
+                Arrays.asList(
+                        "db.tbl1, db2.tbl\\.*, db3.tbl3",
+                        "db.tbl1 ,db2.tbl\\.* ,db3.tbl3",
+                        "db.tbl1 , db2.tbl\\.* , db3.tbl3");
+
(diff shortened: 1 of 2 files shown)


lvyanquan:
Hi @yuxiqian PTAL.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/test/java/org/apache/flink/cdc/connectors/mysql/source/MySqlTablePatternMatchingTest.java
```diff
@@ -280,6 +297,29 @@ void testMultipleRulesWithRealTables() throws Exception {
...
+                    "CreateTableEvent{tableId=db2.tbl2, schema=columns={`id` INT NOT NULL}, primaryKeys=id, options=()}",
+                    "DataChangeEvent{tableId=db2.tbl2, before=[], after=[17], op=INSERT, meta=()}",
+                    "CreateTableEvent{tableId=db3.tbl3, schema=columns={`id` INT NOT NULL}, primaryKeys=id, options=()}",
+                    "DataChangeEvent{tableId=db3.tbl3, before=[], after=[17], op=INSERT, meta=()}"
+                };
+
+        Assertions.assertThat(
+                        getRealWorldMatchedTables(
+                                "db.tbl1 , db2.tbl\\.* , db3.tbl3", null, false, expected.length))
+                .containsExactly(expected);
```
yuxiqian: minor: `containsExactlyInAnyOrder` might be more suitable, as we don't guarantee tables will be handled in alphabetical order.

lvyanquan: Done.

yuxiqian (approved):
LGTM.

PR #4120, opened 2025-09-16
[FLINK-38359][pipeline-connector][paimon] Avoid exception when writing update event to Append only table.

lvyanquan:
Avoid exception when writing update/delete data change events to Paimon append only table.


lvyanquan:
Hi @yuxiqian, PTAL.

PR #4126, opened 2025-09-19
[FLINK-38393][runtime/transform] Loosen schema consistency check for unreachable rules

yuxiqian:
This closes FLINK-38393.

Consider this transform block:

```yaml
transform:
  - { source-table: db.tbl1, filter: 'id > 0', projection: 'id, col1' }
  - { source-table: db.tbl1, filter: 'id <= 0', projection: 'id, col2, col3' }
```

This should be correctly rejected, since there might be two incompatible schemas deduced, depending on if the filter condition is satisfied or not.

However, consider a slightly different example:

```yaml
transform:
  - { source-table: db.tbl1, projection: 'id, col1' }
  - { source-table: db.tbl2, projection: 'id, col2' }
  - ...
  - { source-table: db.\.*, projection: 'id, fallback' }
```

One may write a wildcard matching rule at the end, as a fallback rule. However, this will be rejected in CDC 3.4.0, because Table "db.tbl1" matches two rules with incompatible schema, though the latter one is unreachable.

We may loose the validation check here to avoid breaking existing use cases.


lvyanquan (approved):
LGTM.

lvyanquan:
I suggest picking this change to release-3.5, what do you think?

yuxiqian:
Good idea, addressed in #4130.

PR #4127, opened 2025-09-19
[FLINK-38279] Add SnapshotPendingSplitsState to MySQL checkpoint stat…

dwash96:
…e restoration

[FLINK-38279]

This issue currently prevents restoring snapshot jobs from being restart-able from save points/checkpoints which means that large table and database replication must be done in one shot. Any thing that crashes the snapshotting process creates an entirely non-recoverable situation which is highly inconvenient and un-workable in many set ups.


lvyanquan:
Nice catch! 
And could you add a test for this?

yuxiqian:
Hi @dwash96, do you have time to add an E2e test for this change? YamlJobMigrationITCase might be suitable for such a test. Also, please run `mvn spotless:apply` to fix code style violations.

dwash96:
Oh man, sorry, I will do this by saturday at the latest

yuxiqian:
Hi @dwash96 @lvyanquan, I've pushed commits to add corresponding test cases and resolve conflicts. Would you like to take another look?

leonardBang (approved):
Thanks @dwash96 and @yuxiqian for the contribution, LGTM

PR #4134, opened 2025-09-23
[build] Automate property replacement in pom.xml and exclude the release directory from license checks.

lvyanquan:
Avoid the following manual operation mentioned in this release document https://cwiki.apache.org/confluence/display/FLINK/Creating+a+Flink+CDC+Release#CreatingaFlinkCDCRelease-Buildareleasecandidate, which is because the mvn versions:set command cannot modify the <properties> section in the pom.xml file.
```
# NOTE: Here manually change the value of <revision> property to $RELEASE_VERSION in root POM, and include the change into the latest commit using "--amend"
```

and some build artifacts are stored in the release directory. To continue with Maven builds, skip license checks for these files.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -421,6 +421,8 @@ limitations under the License.
                     </licenseFamilies>
                     <excludes>
                         <!-- Additional files like .gitignore etc.-->
+                        <!-- Release files -->
+                        <exclude>**/release/**</exclude>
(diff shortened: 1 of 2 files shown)


lvyanquan:
Hi @leonardBang @yuxiqian, PTAL.

PR #4137, opened 2025-09-23
[FLINK-38413] Adding support for append only tables to MySQL Pipeline connector

junmuz:
### Description

In some cases, it is nice to have an append only table created for keeping infinite history of the MySQL CDC changes. With [FLINK-38413](https://issues.apache.org/jira/browse/FLINK-38413), I am proposing the capability to create append only tables. This PR allows enabling append only table creation from mysql pipeline connector. I have an added an integration test as well that creates out an Append only Paimon table from MySQL. In order to keep it consistent with the mysql-cdc-source connector, I have used the same property name here.


junmuz:
@lvyanquan Can I get a review on the above PR? Thanks.

lvyanquan:
Please create a jira issue for this.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
```diff
@@ -330,4 +330,12 @@ public class MySqlDataSourceOptions {
                     .defaultValue(false)
                     .withDescription(
                             "Whether to skip backfill in snapshot reading phase. If backfill is skipped, changes on captured tables during snapshot phase will be consumed later in change log reading phase instead of being merged into the snapshot.WARNING: Skipping backfill might lead to data inconsistency because some change log events happened within the snapshot phase might be replayed (only at-least-once semantic is promised). For example updating an already updated value in snapshot, or deleting an already deleted entry in snapshot. These replayed change log events should be handled specially.");
+
+    @Experimental
+    public static final ConfigOption<Boolean> SCAN_READ_CHANGELOG_AS_APPEND_ONLY_ENABLED =
+            ConfigOptions.key("scan.read-changelog-as-append-only.enabled")
```
lvyanquan: I think this change would affect too many tables. Could we set a regular expression (e.g., `scan.read-changelog-as-append-only.tables: adb.*`) to modify the primary key only for tables that match this pattern?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
```diff
@@ -76,11 +77,22 @@ public CustomAlterTableParserListener(
             LinkedList<SchemaChangeEvent> changes,
             boolean tinyInt1isBit,
             boolean isTableIdCaseInsensitive) {
+        this(parser, listeners, changes, tinyInt1isBit, isTableIdCaseInsensitive, false);
```
lvyanquan: Perhaps we can make this change in the `MySqlPipelineRecordEmitter` instead of modifying the `CustomAlterTableParserListener`(since this logic is not related to parsing).

junmuz:
> Please create a jira issue for this.

There is already a ticket created mentioned in the description. [FLINK-38413](https://issues.apache.org/jira/browse/FLINK-38413) for reference.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
junmuz: That's a good suggestion, I will try incorporating it.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/parser/CustomAlterTableParserListener.java
junmuz: Sure, I can do that. I was curious around the schema evolution side of things if there are any workflows from where primary key might get added accidentally.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-mysql/src/main/java/org/apache/flink/cdc/connectors/mysql/source/MySqlDataSourceOptions.java
junmuz: @lvyanquan On a second thought, one can have a dedicated Flink instance for creating append only tables. It seems having a common configuration `scan.read-changelog-as-append-only.enabled` seems to better as it brings consistency with the[ mysql-cdc-connector.](https://nightlies.apache.org/flink/flink-cdc-docs-master/docs/connectors/flink-sources/mysql-cdc/)

junmuz:
@lvyanquan can you trigger the tests again and review the changes? I have tried responding to the comments raised from your side. Thanks in advance.

leonardBang:
@lvyanquan Would you like to help review this PR when you have time？

PR #4146, opened 2025-09-26
[FLINK-38440][pipeline-connector][iceberg] fixed iceberg write SMALLINT/TINYINT data wrong

fcfangcc:
flink-cdc run with mysql -> iceberg, column type in mysql is TINYINT and the value is
negative( like -1), write to iceberg table the value change to 65535.
 
Because -1 pack to "\xff\xff" (short, 2bytes). But in code uppack by int, is get "\xff\xff\x00\x00" (int, 4bytes, got more \x00\x00 from data binary), then get 65535.


lvyanquan (approved):
Nice catch, LGTM.

PR #4156, opened 2025-10-14
[FLINK-38522][cdc connector mysql] use ssl for BinaryLogClient when searching for binlog offset for starting mysql cdc from timestamp

watsonjo737:
https://issues.apache.org/jira/browse/FLINK-38522

# What is the purpose of the change
When using StartupOptions.timestamp(), the MySQL CDC connector calls DebeziumUtils.findBinlogOffset() to locate the appropriate binlog position. This method creates a short-lived BinaryLogClient to scan binlog files, but does not configure SSL mode on the client.

If the MySQL server requires SSL connections (e.g., require_secure_transport=ON or SSL mode configured in connection parameters), the connection fails because the BinaryLogClient attempts an unencrypted connection.

This Pull Request uses the ssl mode provided as part of the connection config for the BinaryLogClient as well to prevent this issue in cases where mysql is configured for require_secure_transport.

# Changes
Set SSLMode to match what was provided in the connectionConfig for BinaryLogClient
Add unit test to ensure SSLMode is converted correctly.

# Verifying this change
  1. Configure MySQL with require_secure_transport=ON or SSL mode requirements
  2. Create a MySQL CDC source with StartupOptions.timestamp(timestampMs)
  3. Start the Flink job
  4. Ensure that the source is able to read records from the specified timestamp.


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtils.java
```diff
@@ -242,12 +243,30 @@ private static Map<String, String> querySystemVariables(
...
+                return SSLMode.PREFERRED;
+            case REQUIRED:
+                return SSLMode.REQUIRED;
+            case VERIFY_CA:
+                return SSLMode.VERIFY_CA;
+            case VERIFY_IDENTITY:
+                return SSLMode.VERIFY_IDENTITY;
+        }
+        return null;
+    }
```
gguptp, line 290: could we simplify this to: 

```
static SSLMode sslModeFor(MySqlConnectorConfig.SecureConnectionMode mode) {
    return mode == null ? null : SSLMode.valueOf(mode.name());
}
```

gguptp (approved):

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtils.java
watsonjo737: Thanks @gguptp , I have updated according to your suggestion

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtilsTest.java
```diff
@@ -83,6 +89,26 @@ private MySqlSourceConfig getConfig(Properties jdbcProperties) {
                 .createConfig(0);
     }
 
+    @ParameterizedTest
+    @MethodSource("sslModeProvider")
+    void testSslModeConversion(MySqlConnectorConfig.SecureConnectionMode input, SSLMode expected) {
```
ruanhang1993, line 94: @watsonjo737 Could you add a test for mysql cdc source in Flink SQL about how to use ssl mode?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtils.java
```diff
@@ -242,12 +243,23 @@ private static Map<String, String> querySystemVariables(
...
+
     public static BinlogOffset findBinlogOffset(
             long targetMs, MySqlConnection connection, MySqlSourceConfig mySqlSourceConfig) {
         MySqlConnection.MySqlConnectionConfiguration config = connection.connectionConfig();
         BinaryLogClient client =
                 new BinaryLogClient(
                         config.hostname(), config.port(), config.username(), config.password());
+        client.setSSLMode(sslModeFor(config.sslMode()));
```
ruanhang1993: If sslMode is null, method `setSSLMode` will throw an exception.
It is better to change as following:
```java
if (sslMode != null) {
  client.setSSLMode(sslMode);
}
```

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtilsTest.java
watsonjo737: Thanks @ruanhang1993 , I have added a new test there MySqlSourceSSLConnectionITCase.java which shows how it can be configured, and confirmed that without this fix in place it fails to run.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtils.java
```diff
@@ -242,12 +243,26 @@ private static Map<String, String> querySystemVariables(
         return variables;
     }
 
+    static SSLMode sslModeFor(MySqlConnectorConfig.SecureConnectionMode mode) {
+        try {
+            return mode == null ? null : SSLMode.valueOf(mode.name());
```
gguptp: i guess this cannot be null right? should we remove this null check?

watsonjo737: While unlikely, I think it is still possible for it to be null. In the most simple case `sslModeFor(null);` for example.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtils.java
```diff
@@ -242,12 +243,26 @@ private static Map<String, String> querySystemVariables(
...
+        } catch (IllegalArgumentException e) {
+            LOG.error("Invalid SecureConnectionMode provided {}", mode.name(), e);
```
ruanhang1993: I think we should throw an exception here.

watsonjo737: Thanks @ruanhang1993 , I have thrown an exception here now

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtils.java
```diff
@@ -242,12 +243,26 @@ private static Map<String, String> querySystemVariables(
...
+        } catch (IllegalArgumentException e) {
+            throw new FlinkRuntimeException(
+                    String.format("Invalid SecureConnectionMode provided: %s ", mode.name()), e);
+        }
+    }
+
     public static BinlogOffset findBinlogOffset(
             long targetMs, MySqlConnection connection, MySqlSourceConfig mySqlSourceConfig) {
         MySqlConnection.MySqlConnectionConfiguration config = connection.connectionConfig();
         BinaryLogClient client =
```
lvyanquan, line 376: We can centralize the SSL configuration in the DebeziumUtils#createBinaryClient method,
https://github.com/apache/flink-cdc/blob/001cba06f58014a627931ac81fb64934d34fa191/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtils.java#L96
 and based on the implementation of MySqlStreamingChangeEventSource, we also need to add the following code to set SslSocketFactory:
https://github.com/apache/flink-cdc/blob/001cba06f58014a627931ac81fb64934d34fa191/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java#L224-L231

watsonjo737: Thanks for the review and recommendation @lvyanquan !
I have now updated this PR:

- ssl config now centralised under DebeziumUtils#createBinaryClient
- updated BinlogSplitReader and SnapshotSplitReader to use the new version of createBinaryClient
- added sslSocketFactory config, similar to what  is used in flink-cdc/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
- updated iTest to include testing sslSocketFactory.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/DebeziumUtils.java
```diff
@@ -93,13 +111,27 @@ public static MySqlConnection createMySqlConnection(
...
-                connectorConfig.username(),
-                connectorConfig.password());
+        BinaryLogClient client =
+                new BinaryLogClient(
+                        connectorConfig.hostname(),
+                        connectorConfig.port(),
+                        connectorConfig.username(),
+                        connectorConfig.password());
+        SSLMode sslMode = sslModeFor(connectorConfig.sslMode());
+        if (sslMode != null) {
```
ruanhang1993: Hi, @watsonjo737 .

I previously left a review comment about the null pointer issue here. However, I now noticed through the comments that the same code exists in `MySqlStreamingChangeEventSource` — when it creates the client, it does not perform an additional null check. It expects an error to be thrown for invalid values, and `sslMode` has a default value. Should we revert this to the same logic as `MySqlStreamingChangeEventSource`, so that it also throws an error when encountering null? WDYT.

CC @lvyanquan

watsonjo737: Hi @ruanhang1993 , Makes sense I have updated the function. It now throws in the default case when it is any unexpected value

wloverine:
Has this been fixed in the latest release?

watsonjo737:
> Has this been fixed in the latest release?

Hi @wloverine , this one did not make it into the latest release unfortunately, needs final approval

PR #4161, opened 2025-10-23
Fix for FLINK-38550

sjmittal:
Added more error codes handling and updated check for `checkIfResumeTokenExpires` method.


sjmittal:
@yuxiqian Could you please review it. I think checking for error code 286 alongside error code 280 makes sense as per mngodb KB.
Refer: https://support.mongodb.com/article/000022220

PR #4170, opened 2025-11-04
[FLINK-38691] [cdc connector mysql] Support for MySQL Transaction Boundary Events in Flink CDC Connector

tejanshrana:
Transaction BEGIN/END events (when provide.transaction.metadata is enabled) are currently treated as unknown and skipped, causing loss of transaction boundaries.

Solution

- Detect transaction metadata events via key schema name
- Update binlog offset when processing these events
- Emit events to downstream consumers if includeTransactionMetadataEvents() is set to true (default = false)

Logs from before the change:
`2025-11-04 14:52:28 2025-11-04 14:52:28,827 INFO  org.apache.flink.cdc.connectors.mysql.source.reader.MySqlRecordEmitter [] - Meet unknown element SourceRecord{sourcePartition={server=mysql_binlog_source}, sourceOffset={transaction_id=4541146d-b988-11f0-87f6-0242ac140006:33, ts_sec=1762267948, file=mysql-bin.000003, pos=13381, gtids=4541146d-b988-11f0-87f6-0242ac140006:1-32, server_id=1}} ConnectRecord{topic='mysql_binlog_source.transaction', kafkaPartition=null, key=Struct{id=4541146d-b988-11f0-87f6-0242ac140006:33}, keySchema=Schema{io.debezium.connector.common.TransactionMetadataKey:STRUCT}, value=Struct{status=BEGIN,id=4541146d-b988-11f0-87f6-0242ac140006:33}, valueSchema=Schema{io.debezium.connector.common.TransactionMetadataValue:STRUCT}, timestamp=null, headers=ConnectHeaders(headers=)}, just skip.`

After the change, the transaction start and end events look like below:
`{"status":"BEGIN","id":"4541146d-b988-11f0-87f6-0242ac140006:47","event_count":null,"data_collections":null}
{"status":"END","id":"4541146d-b988-11f0-87f6-0242ac140006:47","event_count":1,"data_collections":[{"data_collection":"sample.sample","event_count":1}]}`

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
@@ -23,6 +23,8 @@ import org.apache.flink.cdc.connectors.mysql.source.MySqlSource;
 import org.apache.flink.cdc.connectors.mysql.table.StartupOptions;
 import org.apache.flink.table.catalog.ObjectPath;
 
+import io.debezium.config.CommonConnectorConfig;
+
 import java.io.Serializable;
 import java.time.Duration;
 import java.time.ZoneId;
@@ -376,7 +378,8 @@ public class MySqlSourceConfigFactory implements Serializable {
         props.setProperty("include.schema.changes", String.valueOf(true));
         // enable transaction metadata if includeTransactionMetadataEvents is true
         props.setProperty(
-                "provide.transaction.metadata", String.valueOf(includeTransactionMetadataEvents));
+                CommonConnectorConfig.PROVIDE_TRANSACTION_METADATA.name(),
+                String.valueOf(includeTransactionMetadataEvents));
         // disable the offset flush totally
         props.setProperty("offset.flush.interval.ms", String.valueOf(Long.MAX_VALUE));
         // disable tombstones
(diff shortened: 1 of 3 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/reader/MySqlRecordEmitter.java
```diff
@@ -102,6 +102,9 @@ protected void processElement(
             emitElement(element, output);
         } else if (RecordUtils.isHeartbeatEvent(element)) {
             updateStartingOffsetForSplit(splitState, element);
+        } else if (RecordUtils.isTransactionMetadataEvent(element)) {
```
Amareswar, line 114: Shouldn't the emit of events be configurable for backward compatibility?

tejanshrana: That's correct. This is being discussed at the moment and I will introduce a new configuration (in addition to the existing debezium config for transaction metadata) to make this a non-breaking change.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
```diff
@@ -366,6 +374,9 @@ public MySqlSourceConfig createConfig(int subtaskId, String serverName) {
         // Note: the includeSchemaChanges parameter is used to control emitting the schema record,
         // only DataStream API program need to emit the schema record, the Table API need not
         props.setProperty("include.schema.changes", String.valueOf(true));
+        // enable transaction metadata if includeTransactionMetadataEvents is true
+        props.setProperty(
+                "provide.transaction.metadata", String.valueOf(includeTransactionMetadataEvents));
```
lvyanquan: nit:
Please reference the original variable name to facilitate tracking.
```
props.setProperty(CommonConnectorConfig.PROVIDE_TRANSACTION_METADATA.name(), String.valueOf(includeTransactionMetadataEvents));
```

tejanshrana: Done. Could you please take another look and merge if all good? Could this feature also be included in v3.6?

lvyanquan:
Thanks @tejanshrana's contribution, merged.

PR #4179, opened 2025-11-25
[FLINK-38723][common/composer] Add getFlinkConf method to Context.

lvyanquan:
Add getFlinkConf method to Context to be consistent with Flink table Source/Sink.
As we only provided additional information, this did not introduce any risks.

### DIFF
--- a/flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposerTest.java
+++ b/flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineComposerTest.java
@@ -50,8 +50,8 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


yuxiqian (commented):
LGTM overall, but I'm a little confused about how the `testGettingFlinkConfiguration` case works.

What about storing Flink configurations from context in the mocked factory and validating it later?

lvyanquan:
> but I'm a little confused about how the `testGettingFlinkConfiguration` case works.
> What about storing Flink configurations from context in the mocked factory and validating it later?

When invoking createDataSource/createDataSink method, the `execution.target` parameter in FlinkConf is validated. If the value does not meet expectations (this parameter has no default value and will return `null` if not properly assigned), an exception will be thrown.

I prefer not to store the reference of `context.getFlinkConf()` for subsequent validation, as this would require introducing a static variable, which slightly reduces the testability and scalability of the code.

PR #4183, opened 2025-11-27
[FLINK-38741][mysql-cdc] Fix MODIFY COLUMN position change events are…

qyfftf:
When using MySQL CDC Pipeline to synchronize data to Paimon, ALTER TABLE MODIFY COLUMN operations that change column order are ignored and not propagated to the downstream sink, causing schema evolution failure and potential data inconsistency.


leonardBang:
@LYanquan Would you like to help review this PR when you have time ?

PR #4185, opened 2025-12-02
[FLINK-38737][common] Fix potential backwards incompatibility with fallback keys.

lvyanquan:
Fix the issue where the fallback key causes a validation error.


lvyanquan:
Hi @yuxiqian, PTAL.

lvyanquan:
> A unit test for `FactoryHelper` would be nice.

Added.

PR #4187, opened 2025-12-03
[FLINK-38730][build] Bump jdk version from 8 to 11.

lvyanquan:
Bump jdk version from 8 to 11.
This will allow us to gradually start the compatibility work with Flink 2.x, Debezium 2.x, and Iceberg 1.10+.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresPipelineRecordEmitter.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresPipelineRecordEmitter.java
@@ -21,8 +21,11 @@ import org.apache.flink.api.connector.source.SourceOutput;
 import org.apache.flink.cdc.common.event.CreateTableEvent;
 import org.apache.flink.cdc.common.event.Event;
 import org.apache.flink.cdc.common.schema.Schema;
+import org.apache.flink.cdc.common.types.DataType;
 import org.apache.flink.cdc.connectors.base.options.StartupOptions;
 import org.apache.flink.cdc.connectors.base.source.meta.offset.OffsetFactory;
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -773,17 +804,6 @@ limitations under the License.
                 </plugins>
             </build>
         </profile>
-        <profile>
```
yuxiqian, line 776: As we're running tests against Java 17 in nightly builds, shall we create a `java-17-target` profile and activate it with `[17,)`?

lvyanquan: Done.

pom.xml
```diff
@@ -774,23 +805,23 @@ limitations under the License.
             </build>
         </profile>
         <profile>
-            <id>java-8-target</id>
+            <id>java-11-target</id>
             <activation>
-                <jdk>[1.8,11)</jdk>
+                <jdk>[11,)</jdk>
```
yuxiqian:
```suggestion
                <jdk>[11,17)</jdk>
```

to avoid activating both profiles with Java 17

yuxiqian (commented):
Looks good except one minor comment.

pom.xml
lvyanquan: Addressed.

yuxiqian:
Tested this PR with Java 17 in https://github.com/yuxiqian/flink-cdc/pull/123, seems some cases are failing because we need to explicitly `--add-opens` for specific modules.

lvyanquan:
> Tested this PR with Java 17 in [yuxiqian#123](https://github.com/yuxiqian/flink-cdc/pull/123), seems some cases are failing because we need to explicitly `--add-opens` for specific modules.

Done.
And test for jdk17 ran as expected in https://github.com/lvyanquan/flink-cdc/pull/22.

lvyanquan:
Hi @leonardBang, PTAL.

pom.xml
```diff
@@ -465,7 +473,7 @@ limitations under the License.
                 <configuration>
                     <java>
                         <googleJavaFormat>
-                            <version>1.7</version>
+                            <version>1.10.0</version>
```
yuxiqian: nit: It's ideal to follow Flink code format style (AOSP 1.24.0). 
As that would result in changes in many files, we could do it later.

lvyanquan: Traced in https://issues.apache.org/jira/browse/FLINK-38838.

PR #4194, opened 2025-12-18
[FLINK-38818] [posgtresql]Build createTableEventCache using TableSchemas from split 

Mrart:
As https://github.com/apache/flink-cdc/pull/4056 only support mysql.
We add postgres implement.
`
For historical reasons, the current implementation does not utilize the Source to build a schema cache. This requires the TransformOperator to maintain its own complex state.

An optimization is proposed: to build a reusable schema cache at the Source (or in an operator immediately following it).

The primary benefit of this approach is that the transform operator can be made completely stateless. This simplifies its design, improves architectural modularity, and makes the entire pipeline more robust and maintainable.
`

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/test/java/org/apache/flink/cdc/connectors/postgres/source/PostgresPipelineITCaseTest.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/test/java/org/apache/flink/cdc/connectors/postgres/source/PostgresPipelineITCaseTest.java
@@ -19,7 +19,7 @@ package org.apache.flink.cdc.connectors.postgres.source;
 
 import org.apache.flink.api.common.eventtime.WatermarkStrategy;
 import org.apache.flink.api.common.restartstrategy.RestartStrategies;
-import org.apache.flink.cdc.common.configuration.Configuration;
+import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.cdc.common.data.binary.BinaryStringData;
 import org.apache.flink.cdc.common.event.CreateTableEvent;
 import org.apache.flink.cdc.common.event.DataChangeEvent;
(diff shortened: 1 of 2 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/IncrementalSourceRecordEmitter.java
```diff
@@ -160,6 +161,8 @@ protected void emitElement(SourceRecord element, SourceOutput<T> output) throws
         debeziumDeserializationSchema.deserialize(element, outputCollector);
     }
 
+    public void applySplit(SourceSplitBase split) {}
```
lvyanquan, line 174: It's better to add some java doc for this newly added public method.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/test/java/org/apache/flink/cdc/connectors/postgres/source/PostgresPipelineITCaseTest.java
```diff
@@ -135,6 +135,127 @@ public void testInitialStartupMode() throws Exception {
...
+                fetchResultsExcept(events, expectedBinlog.size(), createTableEvent);
+
+        // Filter out schema change events and keep only data change events
+        List<Event> actualDataChangeEvents =
+                actualEvents.stream()
+                        .filter(event -> event instanceof DataChangeEvent)
+                        .collect(Collectors.toList());
+
+        // Verify that we captured the expected number of data change events
+        assertThat(actualDataChangeEvents.size()).isGreaterThanOrEqualTo(expectedBinlog.size());
```
lvyanquan: It's better to add a scenario for creating a savepoint and restoring reads from the savepoint here too.

PR #4196, opened 2025-12-19
[cdc-connector-mysql] Add MySQL 8.4+ compatibility for binlog status command

sunlicp:
MySQL 8.4+ removed the SHOW MASTER STATUS command and replaced it with
  SHOW BINARY LOG STATUS. This change adds version detection to automatically
  use the appropriate command based on the MySQL server version.

  Changes:
  - Added getBinlogStatusCommand() method in DebeziumUtils to detect MySQL
    version and return the correct binlog status command
  - MySQL 8.4+ uses SHOW BINARY LOG STATUS
  - Earlier versions continue to use SHOW MASTER STATUS
  - Updated MySqlConnection to use the new version-aware method

  This ensures compatibility across MySQL 5.7, 8.0, 8.4+, and future versions.


yuxiqian:
Closed in #3666.

PR #4197, opened 2025-12-23
[FLINK-38829][pipline-connector/starrocks] Support rename column DDL for starrocks sink connector

zml1206:
From v3.3.2 onwards, StarRocks supports renaming columns.


zml1206:
Thanks for review. @lvyanquan

zml1206:
I think alter column type can also be supported. Although we don't know what the original type is, most types are already supported, we can throw exception if the conversion fails. What do you think? @lvyanquan

lvyanquan:
> I think alter column type can also be supported. Although we don't know what the original type is, most types are already supported, we can throw exception if the conversion fails. What do you think? @lvyanquan

I agree to support this ability.
For the data types before the change, perhaps we can obtain them by querying the downstream table structure (if necessary).

PR #4199, opened 2025-12-25
[FLINK-38839][runtime] Support semicolon delimiter for table-options

tchivs:
## Summary
Allow `table-options` in transform rules to use semicolon `;` as the key/value pair delimiter so option values can safely contain commas (e.g. `sequence.field=gxsj,jjsj`). Keep comma delimiter for backward compatibility.

## Motivation
Some downstream table options are multi-value and use comma inside the value. The existing `key1=value1,key2=value2` syntax makes those options impossible to express reliably.

## Changes
- `SchemaMetadataTransform` parses `table-options` pairs using:
  - `;` when present in the string, otherwise `,`
  - `split("=", 2)` to avoid breaking values containing `=`
- Docs updated to mention semicolon delimiter for comma-in-value cases.
- New unit tests for the parsing behavior.

## Behavior
- Still supports legacy format:
  - `table-options: key1=value1,key2=value2`
- Supports semicolon format (recommended when values contain commas):
  - `table-options: sequence.field=gxsj,jjsj;file-index.range-bitmap.columns=jjsj;file-index.bloom-filter.columns=jjdbh`
- Note: delimiter is chosen by presence of `;` (do not mix `,` and `;` in the same string).

### DIFF
--- a/flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/resources/rules/unexpected.yaml
+++ b/flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/resources/rules/unexpected.yaml
@@ -48,7 +48,7 @@ steps:
               sink: uuid
     error: |
       Unexpected key `mapping` in YAML transform block.
-      Allowed keys in this context are: [source-table, projection, filter, primary-keys, partition-keys, table-options, description, converter-after-transform]
+      Allowed keys in this context are: [source-table, projection, filter, primary-keys, partition-keys, table-options, table-options.delimiter, description, converter-after-transform]
       Note: option mapping: [{"source":"userUuid","sink":"uuid"}] is unexpected. It was silently ignored in previous versions, and probably should be removed.
   # Unexpected route block keys
   - type: submit


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/SchemaMetadataTransform.java
```diff
@@ -58,13 +58,17 @@ public SchemaMetadataTransform(
             partitionKeys = Arrays.asList(partitionKeyArr);
         }
         if (!StringUtils.isNullOrWhitespaceOnly(tableOptionString)) {
-            for (String tableOption : tableOptionString.split(",")) {
-                String[] kv = tableOption.split("=");
+            // Use semicolon as delimiter to support multi-value options like
+            // sequence.field=jjsj,gxsj.
+            // Keep comma delimiter for backward compatibility.
+            String delimiter = tableOptionString.contains(";") ? ";" : ",";
```
leonardBang: Could expand this PR scope to support custom delimiter like `;`, `|`, `$`... so that the feature will more powerful?

leonardBang: and the default value of optional transform config option `table-options.delimiter` could be `,` for backforward compatibility consideration.

tchivs:
Thanks for the suggestion @leonardBang\! I've updated the PR to support custom delimiters as you recommended.

**Changes made:**
1. Added a new optional configuration `table-options.delimiter` that allows users to specify any custom delimiter (e.g., `;`, `|`, `$`, etc.)
2. The default value is `,` for backward compatibility
3. Updated documentation with examples showing how to use custom delimiters

**Example usage:**
```yaml
transform:
  - source-table: mydb.mytable
    table-options: sequence.field=gxsj,jjsj;file-index.bloom-filter.columns=jjdbh
    table-options.delimiter: ";"
```

This approach is more flexible and powerful as users can now choose any delimiter that suits their needs.

docs/content.zh/docs/core-concept/transform.md
```diff
@@ -467,8 +474,8 @@ transform:
 
 ## Embedding AI 模型
 
-Embedding AI 模型可以在 transform 规则中使用。
-为了使用 Embedding AI 模型，你需要下载内置模型的 jar，然后在 `flink-cdc.sh` 命令中添加 `--jar {$BUILT_IN_MODEL_PATH}`。
+内置AI模型可以在transform规则中使用。
+为了使用内置AI模型，你需要下载内置模型的jar，然后在flink-cdc.sh命令中添加`--jar {$BUILT_IN_MODEL_PATH}`。
```
yuxiqian:
```suggestion
内置 AI 模型可以在 transform 规则中使用。
为了使用内置 AI 模型，你需要下载内置模型的 jar ，然后在 `flink-cdc.sh` 命令中添加 `--jar {$BUILT_IN_MODEL_PATH}`。
```

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/PostTransformOperatorBuilder.java
```diff
@@ -67,6 +90,7 @@ public PostTransformOperatorBuilder addTransform(
                         "",
                         "",
                         "",
+                        "",
```
yuxiqian: The default field for `tableOptionsDelimiter` is inconsistent. It is `,` in TransformDef, `null` in an `addTransform` method and `""` in another `addTransform` method.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/SchemaMetadataTransform.java
```diff
@@ -58,13 +61,19 @@ public SchemaMetadataTransform(
             partitionKeys = Arrays.asList(partitionKeyArr);
         }
         if (!StringUtils.isNullOrWhitespaceOnly(tableOptionString)) {
-            for (String tableOption : tableOptionString.split(",")) {
-                String[] kv = tableOption.split("=");
+            // Use custom delimiter if provided, otherwise default to comma for backward
+            // compatibility.
+            String delimiter =
+                    StringUtils.isNullOrWhitespaceOnly(tableOptionsDelimiter)
```
yuxiqian: I'm not quite sure about the `isNullOrWhitespaceOnly` check. Sometimes using `\n` as delimiter might make sense, allowing such clearer syntax:

```
table-options: |-
  k1 = v1
  k2 = v2

table-options.delimiter: "\n"
```

Anyway, adding logs here would be nice as it's error-prone.

PR #4201, opened 2025-12-29
[hotfix] Fix build_docs GHA workflow not working after Java 11 upgrade

yuxiqian:
Specify default JDK version to resolve compilation failure like https://github.com/apache/flink-cdc/actions/runs/20563581361/job/59057922998?pr=4190.


leonardBang (approved):
Thanks for the quick fix, +1

PR #4203, opened 2025-12-29
[common] Add PatternCache to optimize regex compilation performance

tchivs:
This PR introduces a **PatternCache** utility class to optimize regex pattern compilation performance by caching compiled `Pattern` instances with LRU eviction strategy.

  ## Brief change log

  - **feat(common)**: Add `PatternCache` class with LRU-based caching mechanism
  - **perf(common)**: Integrate `PatternCache` into `Predicates.setOfRegex()` method
  - **test**: Add comprehensive unit tests including LRU eviction behavior test

  ## Technical Details

  ### PatternCache Implementation
  - Uses `LinkedHashMap` with access-order (`true`) for LRU behavior
  - Maximum cache size: 100 patterns
  - Thread-safe with `synchronized` methods
  - Only caches patterns without flags (`regexFlags == 0`)

  ### Performance Benefits
  - **Avoids repeated `Pattern.compile()` calls** for the same regex string
  - **Reduces CPU overhead** in pattern-heavy operations
  - **Improves throughput** for table pattern matching scenarios

  ### Integration Points
  - `Predicates.setOfRegex()`: Uses cache when `regexFlags == 0`
  - Patterns with flags are compiled directly without caching

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/PatternCache.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/PatternCache.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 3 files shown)


yuxiqian (commented):
Seems the only place `Predicates` got used is in the Table ID `Selectors`.

Could you please do some benchmark tests on this? I wonder if adding caches for the `Selectors` matching method (#3994) resolve this, too?

PR #4209, opened 2026-01-04
[FLINK-38840][postgres] Postgres YAML connector supports emitting complete Table ID

Mrart:
Currently, Postgres YAML connector maps (schema, table) as Pipeline event Table ID, and the database part is missing.

We may add a new option to generate complete Table ID (database, schema, table). This could be useful when records are written into Kafka as JSON messages.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/test/java/org/apache/flink/cdc/connectors/postgres/source/PostgresPipelineITCaseTest.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/test/java/org/apache/flink/cdc/connectors/postgres/source/PostgresPipelineITCaseTest.java
@@ -57,7 +57,8 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.ValueSource;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


Mrart:
@yuxiqian PTAL

loserwang1024:
@Mrart 
It seems that even we don't have `table-id.include-database` option but directly emitting complete Table ID. The compability  of currently pipeline won't be broke. And kafka sink can add a new option to determine whether to print database. @yuxiqian , WDYT?

yuxiqian:
> @Mrart It seems that even we don't have `table-id.include-database` option but directly emitting complete Table ID. The compability of currently pipeline won't be broke. And kafka sink can add a new option to determine whether to print database. @yuxiqian , WDYT?

I'm a little concerned about changing default Table ID format. It may affect transform & route table matching rules and schema registry states where Table IDs are used as key.

Mrart:
> @Mrart It seems that even we don't have `table-id.include-database` option but directly emitting complete Table ID. The compability of currently pipeline won't be broke. And kafka sink can add a new option to determine whether to print database. @yuxiqian , WDYT?

We still need to control the output of the full tableID through configuration to ensure that it is compatible with production

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/test/java/org/apache/flink/cdc/connectors/postgres/source/PostgresPipelineITCaseTest.java
```diff
@@ -365,6 +365,7 @@ public void testInitialStartupModeWithOpts(boolean unboundedChunkFirst) throws E
         sourceConfiguration.set(
                 PostgresDataSourceOptions.SCAN_INCREMENTAL_SNAPSHOT_UNBOUNDED_CHUNK_FIRST_ENABLED,
                 unboundedChunkFirst);
+        sourceConfiguration.set(PostgresDataSourceOptions.TABLE_ID_INCLUDE_DATABASE, true);
```
yuxiqian: nit: Better add a ParameterizedTest here to cover both `TABLE_ID_INCLUDE_DATABASE` configurations.

Mrart:
@loserwang1024  PTLK

loserwang1024:
LGTM

PR #4212, opened 2026-01-05
[FLINK-38831][route] The route configuration supports matching the first one

linguoxuan:
### Purpose
This PR adds support for configurable route matching modes in Flink CDC pipelines. Previously, all route rules were evaluated and applied (all-match behavior). With this change, users can now choose between two matching strategies:

- all-match (default): Apply all matching route rules to a table
- first-match: Apply only the first matching route rule and stop evaluation

### Core Components Changes
1. RouteRule.java: Added MatchMode enum with two modes:
- ALL_MATCH: Matches all applicable routing rules (default behavior)
- FIRST_MATCH: Matches only the first applicable routing rule

2. PipelineDef.java:
- Added routeMode field to store the global route matching mode
- Updated constructors to support the new field
- Added getRouteMode() method
3. YamlPipelineDefinitionParser.java:
- Added parsing logic for route-mode configuration from the pipeline section
- Defaults to "all-match" if not specified
- Validates configuration values using RouteRule.MatchMode.fromConfigValue()
4. SchemaOperatorTranslator.java:
- Updated to pass routeMode from PipelineDef to SchemaRegistry
5. SchemaRegistry.java:
- Added routeMatchMode field
- Updated constructor to accept match mode parameter
- Passes match mode to TableIdRouter
6. TableIdRouter.java:
- Implemented match mode logic in the route() method
- For FIRST_MATCH: Returns immediately after finding the first matching rule
- For ALL_MATCH: Continues to evaluate all rules (existing behavior)

### Tests
1. TableIdRouterMatchModeTest.java
2. YamlPipelineDefinitionParserTest.java

### Usage
```
pipeline:
  name: mysql_to_doris_pipeline
  parallelism: 2
  route-mode: FIRST-MATCH

route:
  - source-table: mydb.order_.*
    sink-table: ods_db.ods_orders
    description: "Merge all order sharded tables"
  - source-table: mydb.product_.*
    sink-table: ods_db.ods_products
    description: "Merge all product sharded tables"
  - source-table: mydb.*
    sink-table: ods_db.ods_<>
    replace-symbol: <>
    description: "One-to-one mapping for other tables"
```


linguoxuan:
@yuxiqian Could you please take a look at this PR?

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
```diff
@@ -143,6 +149,13 @@ private PipelineDef parse(JsonNode pipelineDefJsonNode, Configuration globalPipe
                             ((ObjectNode) pipelineDefJsonNode.get(PIPELINE_KEY)).remove(MODEL_KEY))
                     .map(node -> validateArray("model", node))
                     .ifPresent(node -> modelDefs.addAll(parseModels(node)));
+
+            // Extract route-mode from pipeline if present
+            JsonNode pipelineNode = pipelineDefJsonNode.get(PIPELINE_KEY);
+            if (pipelineNode.has(ROUTE_MODE_KEY)) {
+                routeMode = pipelineNode.get(ROUTE_MODE_KEY).asText();
+                ((ObjectNode) pipelineNode).remove(ROUTE_MODE_KEY);
+            }
```
yuxiqian: We may define `ROUTE_MODE_KEY` as an EnumOption in `PipelineOptions` and retrieve it later without modifying the parser:

```java
RouteMode routeMode = pipelineConfig.get(PIPELINE_ROUTE_MODE_BEHAVIOR)
```

similar to `PIPELINE_SCHEMA_CHANGE_BEHAVIOR`.

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/parser/YamlPipelineDefinitionParser.java
```diff
@@ -189,7 +202,14 @@ private PipelineDef parse(JsonNode pipelineDefJsonNode, Configuration globalPipe
...
         return new PipelineDef(
-                sourceDef, sinkDef, routeDefs, transformDefs, udfDefs, modelDefs, pipelineConfig);
+                sourceDef,
+                sinkDef,
+                routeDefs,
+                routeMode,
+                transformDefs,
+                udfDefs,
+                modelDefs,
+                pipelineConfig);
```
yuxiqian: Putting `routeMode` in `pipelineConfig` would be neater

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/route/RouteRule.java
```diff
@@ -24,6 +24,43 @@ public class RouteRule implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
+    /** Match mode for routing rules. */
+    public enum MatchMode {
```
yuxiqian: This enum should be part of public API. What about put it in `org.apache.flink.cdc.common.pipeline` and mark it with `@PublicEvolving`?

yuxiqian (commented):
Thanks for @linguoxuan's nice work, just left some minor comments. 

Adding more IT / E2e tests in `FlinkPipelineComposerITCase` and `RouteE2eITCase` would be nice, too.

linguoxuan:
> Thanks for @linguoxuan's nice work, just left some minor comments.
> 
> Adding more IT / E2e tests in `FlinkPipelineComposerITCase` and `RouteE2eITCase` would be nice, too.

@yuxiqian Thanks for the review! I've addressed your comments and add more IT/E2E case in a new commit. 

Please take another look when you have time and let me know if anything else needs adjustment!

linguoxuan:
I've checked the error logs. The CI failure appears to be an environment issue.  cc @yuxiqian

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/pipeline/PipelineOptions.java
```diff
@@ -72,6 +72,22 @@ public class PipelineOptions {
                                                             "EXCEPTION: Throw an exception to terminate the sync pipeline.")))
                                     .build());
 
+    public static final ConfigOption<String> PIPELINE_ROUTE_MODE =
+            ConfigOptions.key("route-mode")
+                    .stringType()
```
yuxiqian: Is it possible to declare it like this?

```suggestion
    public static final ConfigOption<RouteMode> PIPELINE_ROUTE_MODE =
            ConfigOptions.key("route-mode")
                    .enumType(RouteMode.class)
```

linguoxuan: In that case, we can only pass this parameter via route-mode: FIRST_MATCH. What I'm thinking is to allow users to pass it in ways like route-mode: first-match, route-mode: First-Match, or route-mode: FIRST_MATCH. What do you think? Or have I missed some methods? @yuxiqian

yuxiqian: Just checked other enumType options and most of them require keys in SCREAMING_SNAKE_CASE.

We may modify `ConfigurationUtils#convertToEnum` to allow barbeque namings for enum values later in another PR.

linguoxuan: Got it. Thanks to clarify! I have pushed a new commit. Please take another look when you have time.

sd4324530:
Will this feature be available in version 3.6.0?

yuxiqian:
Hi @linguoxuan, sorry for the late reply. Could you please rebase this PR with master so we can push this forward?

flink-cdc-cli/src/test/resources/definitions/pipeline-definition-with-route-mode.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+route:
+  - source-table: mydb.order_.*
+    sink-table: ods_db.ods_orders
+    description: "Merge all order sharded tables"
+  - source-table: mydb.product_.*
+    sink-table: ods_db.ods_products
+    description: "Merge all product sharded tables"
+  - source-table: mydb.*
+    sink-table: ods_db.ods_<>
```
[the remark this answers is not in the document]
yuxiqian: Irrelevant to changes in this PR.

yuxiqian:
Would @leonardBang @lvyanquan like to take a look?

linguoxuan:
> Hi @linguoxuan, sorry for the late reply. Could you please rebase this PR with master so we can push this forward?

Sorry, I missed your message.

yuxiqian:
Never mind, thanks for your contribution!

PR #4213, opened 2026-01-06
[FLINK-38601][mongodb] Catch throwable errors when consuming from unbounded streams

linguoxuan:
This PR fixes a issue in the Flink MongoDB CDC connector where certain fatal errors (like OutOfMemoryError, StackOverflowError) were silently swallowed, causing the CDC pipeline to stop running without propagating the error or triggering any recovery mechanism.


lvyanquan (approved):
+1.

PR #4215, opened 2026-01-09
[FLINK-38596][pipeline][maxcompute] Fix the column comment of CreateTableEvent and AddColumnEvent

linguoxuan:
### Summary
Fix column comment handling in MaxCompute connector for CreateTableEvent and AddColumnEvent. Previously, column comments were not properly preserved when creating tables or adding columns, resulting in type descriptions being used as comments instead of the actual column comments from the source schema.
### Changes
1. TypeConvertUtils.java
Added logic to preserve column comments when converting Flink CDC columns to MaxCompute columns
2. SchemaEvolutionUtils.java
Fixed addColumns() method to use the actual column comment from addColumn.getAddColumn().getComment() instead of always using the type description

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-maxcompute/src/main/java/org/apache/flink/cdc/connectors/maxcompute/utils/SchemaEvolutionUtils.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-maxcompute/src/main/java/org/apache/flink/cdc/connectors/maxcompute/utils/SchemaEvolutionUtils.java
@@ -122,15 +122,15 @@ public class SchemaEvolutionUtils {
                 sqlBuilder
                         .append(addColumn.getAddColumn().getName())
                         .append(" ")
-                        .append(string(addColumn.getAddColumn().getType()))
-                        .append(" comment '")
-                        // Use column comment if available, otherwise use type description as
-                        // fallback
-                        .append(
-                                addColumn.getAddColumn().getComment() != null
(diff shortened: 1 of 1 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-maxcompute/src/main/java/org/apache/flink/cdc/connectors/maxcompute/utils/SchemaEvolutionUtils.java
```diff
@@ -124,7 +124,12 @@ public static void addColumns(
                         .append(" ")
                         .append(string(addColumn.getAddColumn().getType()))
                         .append(" comment '")
-                        .append(addColumn.getAddColumn().getType().asSummaryString())
+                        // Use column comment if available, otherwise use type description as
+                        // fallback
+                        .append(
+                                addColumn.getAddColumn().getComment() != null
+                                        ? addColumn.getAddColumn().getComment()
+                                        : addColumn.getAddColumn().getType().asSummaryString())
```
lvyanquan: I believe the original comment was caused by a coding error, and we do not need to retain the previous configuration scheme. 
We can directly use the new comment (or null) instead.

linguoxuan: Yes, you're right. I'm a little unsure about this part too. I have pushed a new commit.

PR #4218, opened 2026-01-12
[FLINK-38886] Introduce `GenericRecordData` and Internal / External converters in flink-cdc-common

yuxiqian:
This closes FLINK-38886.

Currently, there's no portable way to construct a CDC RecordData without depending on heavy flink-cdc-runtime packages. Also BinaryRecordData is not easy to parse and inspect.

Providing such classes and converters between CDC Internal objects / classes and Java objects / classes would simplify connectors implementation.


yuxiqian:
Updated this PR to reflect newly introduced Variant type, PTAL. @lvyanquan

PR #4221, opened 2026-01-12
[FLINK-38889][pipeline][kafka] Support serializing complex types(MAP, ARRAY, ROW) to JSON (Debezium / Canal)

linguoxuan:
This closes [FLINK-38889](https://issues.apache.org/jira/browse/FLINK-38889).
### Purpose
This PR fixes the issue where YAML Kafka sink connector does not support serializing complex types (MAP, ARRAY, ROW) to JSON format (Debezium / Canal), while Kafka SQL connector handles them without problem.
### Root Cause
The issue was in the TableSchemaInfo class, which is responsible for converting CDC's RecordData format to Flink's RowData format before JSON serialization. The createFieldGetter() method lacked the necessary conversion logic for complex types.
### Changes
1. Added complex type conversion methods in TableSchemaInfo.java: support for ARRAY, MAP, and ROW types
### Testing

1. TableSchemaInfoTest.java:
2. DebeziumJsonSerializationSchemaTest.java:
3. CanalJsonSerializationSchemaTest.java:
4. KafkaDataSinkITCase.java

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/test/java/org/apache/flink/cdc/connectors/kafka/json/utils/RecordDataConverterTest.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/test/java/org/apache/flink/cdc/connectors/kafka/json/utils/RecordDataConverterTest.java
@@ -1,0 +1,445 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.connectors.kafka.json.utils;
+
+import org.apache.flink.cdc.common.data.DateData;
+import org.apache.flink.cdc.common.data.DecimalData;
+import org.apache.flink.cdc.common.data.GenericArrayData;
+import org.apache.flink.cdc.common.data.GenericMapData;
+import org.apache.flink.cdc.common.data.LocalZonedTimestampData;
+import org.apache.flink.cdc.common.data.RecordData;
+import org.apache.flink.cdc.common.data.TimeData;
+import org.apache.flink.cdc.common.data.TimestampData;
+import org.apache.flink.cdc.common.data.binary.BinaryRecordData;
+import org.apache.flink.cdc.common.data.binary.BinaryStringData;
+import org.apache.flink.cdc.common.schema.Schema;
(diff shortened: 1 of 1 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/test/java/org/apache/flink/cdc/connectors/kafka/json/canal/CanalJsonSerializationSchemaTest.java
```diff
@@ -49,7 +49,7 @@ class CanalJsonSerializationSchemaTest {
             TableId.tableId("default_namespace", "default_schema", "table1");
 
     @Test
-    void testSerialize() throws Exception {
+    public void testSerializeComplexTypes() throws Exception {
```
yuxiqian: It would be nice to keep existing cases unchanged and add new IT cases in `KafkaDataSinkITCase`.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/test/java/org/apache/flink/cdc/connectors/kafka/json/ComplexTypesEdgeCasesTest.java
```diff
@@ -0,0 +1,491 @@
...
+
+        ArrayData resultInnerArray1 = resultOuterArray.getArray(0);
+        assertThat(resultInnerArray1.size()).isEqualTo(2);
+        assertThat(resultInnerArray1.getString(0).toString()).isEqualTo("a");
+        assertThat(resultInnerArray1.getString(1).toString()).isEqualTo("b");
+
+        ArrayData resultInnerArray2 = resultOuterArray.getArray(1);
+        assertThat(resultInnerArray2.size()).isEqualTo(2);
+        assertThat(resultInnerArray2.getString(0).toString()).isEqualTo("c");
+        assertThat(resultInnerArray2.getString(1).toString()).isEqualTo("d");
```
yuxiqian: It's not straightforward to validate results like this. What about put complex type serialization tests in `KafkaDataSinkITCase` where we can validate result JSON directly?

Also, a parameterized test is welcome so we don't have to repeat similar codes.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka/src/main/java/org/apache/flink/cdc/connectors/kafka/json/TableSchemaInfo.java
```diff
@@ -195,4 +224,231 @@ public Schema getSchema() {
...
+        DataType elementType = arrayType.getElementType();
+        int size = arrayData.size();
+        Object[] result = new Object[size];
+
+        for (int i = 0; i < size; i++) {
+            result[i] = convertElement(arrayData, i, elementType, zoneId);
+        }
+
+        return new org.apache.flink.table.data.GenericArrayData(result);
+    }
```
yuxiqian: nit: move CDC RecordData => Flink SQL converter to utils package?

linguoxuan:
Thanks @yuxiqian for the review!  I have made the changes as suggested. Since the suggestions focused on code formatting, I force-pushed the code to make it clearer. PTAL.

yuxiqian:
Thanks for the quick response! Just pushed another commit to simplify IT case and docs style.

Would @lvyanquan like to take another look?

lvyanquan:
Merged. Feel free to open a follow-up PR to update the documentation.

linguoxuan:
> LGTM. Would @linguoxuan like to append docs in this PR, or open an individual PR for it?

I will do it.

PR #4228, opened 2026-01-20
[FLINK-38879][pipeline-connector][paimon] Add support for creating and writing table with variant type.

lvyanquan:
Add support for creating and writing table with variant type.

Why not add end-to-end (e2e) tests?
Because the Flink version we're currently using is 1.20, and Flink SQL in this version does not yet support querying the VARIANT type. Therefore, we cannot use the catalog in Flink SQL to query and verify the results.
Support for this will only be available after we upgrade to Flink 2.2.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/utils/TypeUtils.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/utils/TypeUtils.java
@@ -35,6 +35,7 @@ public class TypeUtils {
      * @return Paimon DataType
      */
     public static org.apache.paimon.types.DataType toPaimonDataType(DataType dataType) {
+        // TODO remove this branch after bumping Flink version to 2.2
         if (dataType.is(DataTypeRoot.VARIANT)) {
             return new VariantType(dataType.isNullable());
         } else {
@@ -50,6 +51,7 @@ public class TypeUtils {
      * @return Flink CDC DataType
      */
     public static DataType toCDCDataType(org.apache.paimon.types.DataType dataType) {
+        // TODO remove this branch after bumping Flink version to 2.2
         if (dataType.is(org.apache.paimon.types.DataTypeRoot.VARIANT)) {
             return new org.apache.flink.cdc.common.types.VariantType(dataType.isNullable());
         } else {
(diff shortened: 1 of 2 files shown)


lvyanquan:
Hi @yunfengzhou-hub, maybe you can help to review this as you are more familiar with Paimon Sink.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/utils/TypeUtils.java
```diff
@@ -0,0 +1,61 @@
...
+     * Convert Flink CDC DataType to Paimon DataType.
+     *
+     * @param dataType Flink CDC DataType
+     * @return Paimon DataType
+     */
+    public static org.apache.paimon.types.DataType toPaimonDataType(DataType dataType) {
+        if (dataType.is(DataTypeRoot.VARIANT)) {
+            return new VariantType(dataType.isNullable());
+        } else {
+            return LogicalTypeConversion.toDataType(
```
yunfengzhou-hub, line 42: Let's add a comment or todo to explain that after Flink 2.2, we won't need a special case for Variant type.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonWriterHelper.java
```diff
@@ -176,6 +176,14 @@ private static RecordData.FieldGetter createFieldGetter(
             case MAP:
                 fieldGetter = new BinaryFieldDataGetter(fieldPos, fieldType.getTypeRoot());
                 break;
+            case VARIANT:
+                fieldGetter =
+                        row -> {
+                            BinaryVariant binaryVariant = (BinaryVariant) row.getVariant(fieldPos);
```
yunfengzhou-hub: It might be better to get `Variant` first, then check whether the variant can be cast to `BinaryVariant`.

lvyanquan: Done.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/utils/TypeUtils.java
lvyanquan: Done.

PR #4233, opened 2026-01-22
[FLINK-38959][pipeline-connector][postgres]Added support for schema change in the postgres pipeline connector

linjianchang:
Added support for schema change in the postgres pipeline connector

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresDataSource.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresDataSource.java
@@ -20,6 +20,7 @@ package org.apache.flink.cdc.connectors.postgres.source;
 import org.apache.flink.cdc.common.annotation.Internal;
 import org.apache.flink.cdc.common.annotation.VisibleForTesting;
 import org.apache.flink.cdc.common.event.Event;
+import org.apache.flink.cdc.common.event.TableId;
 import org.apache.flink.cdc.common.source.DataSource;
 import org.apache.flink.cdc.common.source.EventSourceProvider;
 import org.apache.flink.cdc.common.source.FlinkSourceProvider;
@@ -34,13 +35,16 @@ import org.apache.flink.cdc.connectors.postgres.source.config.PostgresSourceConf
 import org.apache.flink.cdc.connectors.postgres.source.offset.PostgresOffsetFactory;
 import org.apache.flink.cdc.connectors.postgres.source.reader.PostgresPipelineRecordEmitter;
 import org.apache.flink.cdc.connectors.postgres.table.PostgreSQLReadableMetadata;
+import org.apache.flink.cdc.connectors.postgres.utils.PostgresSchemaUtils;
 import org.apache.flink.cdc.debezium.DebeziumDeserializationSchema;
 import org.apache.flink.cdc.debezium.event.DebeziumEventDeserializationSchema;
 import org.apache.flink.cdc.debezium.table.DebeziumChangelogMode;
 import org.apache.flink.connector.base.source.reader.RecordEmitter;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /** A {@link DataSource} for Postgres cdc connector. */
 @Internal
@@ -50,6 +54,7 @@ public class PostgresDataSource implements DataSource {
     private final PostgresSourceConfig postgresSourceConfig;
 
     private final List<PostgreSQLReadableMetadata> readableMetadataList;
+    Map<TableId, Map<String, Integer>> beforeTableColumnsOidMaps = new HashMap<>();
 
     public PostgresDataSource(PostgresSourceConfigFactory configFactory) {
         this(configFactory, new ArrayList<>());
@@ -67,12 +72,17 @@ public class PostgresDataSource implements DataSource {
     public EventSourceProvider getEventSourceProvider() {
         String databaseName = postgresSourceConfig.getDatabaseList().get(0);
         boolean includeDatabaseInTableId = postgresSourceConfig.isIncludeDatabaseInTableId();
+        beforeTableColumnsOidMaps =
+                PostgresSchemaUtils.getAllTablesColumnOids(
+                        postgresSourceConfig, postgresSourceConfig.getTableList());
         DebeziumEventDeserializationSchema deserializer =
                 new PostgresEventDeserializer(
                         DebeziumChangelogMode.ALL,
                         readableMetadataList,
                         includeDatabaseInTableId,
-                        databaseName);
+                        databaseName,
+                        postgresSourceConfig,
+                        beforeTableColumnsOidMaps);
 
         PostgresOffsetFactory postgresOffsetFactory = new PostgresOffsetFactory();
         PostgresDialect postgresDialect = new PostgresDialect(postgresSourceConfig);
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresEventDeserializer.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresEventDeserializer.java
@@ -19,9 +19,20 @@ package org.apache.flink.cdc.connectors.postgres.source;
 
 import org.apache.flink.cdc.common.annotation.Internal;
 import org.apache.flink.cdc.common.data.binary.BinaryStringData;
+import org.apache.flink.cdc.common.event.AddColumnEvent;
+import org.apache.flink.cdc.common.event.AlterColumnTypeEvent;
+import org.apache.flink.cdc.common.event.DropColumnEvent;
+import org.apache.flink.cdc.common.event.Event;
+import org.apache.flink.cdc.common.event.RenameColumnEvent;
 import org.apache.flink.cdc.common.event.SchemaChangeEvent;
 import org.apache.flink.cdc.common.event.TableId;
+import org.apache.flink.cdc.common.schema.Column;
+import org.apache.flink.cdc.common.schema.PhysicalColumn;
+import org.apache.flink.cdc.common.types.DataType;
+import org.apache.flink.cdc.common.types.DataTypes;
+import org.apache.flink.cdc.connectors.postgres.source.config.PostgresSourceConfig;
 import org.apache.flink.cdc.connectors.postgres.table.PostgreSQLReadableMetadata;
+import org.apache.flink.cdc.connectors.postgres.utils.PostgresSchemaUtils;
 import org.apache.flink.cdc.debezium.event.DebeziumEventDeserializationSchema;
 import org.apache.flink.cdc.debezium.table.DebeziumChangelogMode;
 import org.apache.flink.table.data.TimestampData;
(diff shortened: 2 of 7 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/utils/PostgresSchemaUtils.java
```diff
@@ -100,6 +102,55 @@ public static Schema getTableSchema(PostgresSourceConfig sourceConfig, TableId t
...
+        try (PostgresConnection jdbc = getPostgresDialect(sourceConfig).openJdbcConnection()) {
+            Map<String,Integer> oidMaps = new HashMap<>();
+            String inClause = columns.stream()
+                    .map(column -> "'" + column.replace("'", "''") + "'")
+                    .collect(Collectors.joining(",", "in (", ")"));
+            try {
+                jdbc.query(
+                        "SELECT a.attname attname,a.attnum AS age_oid \n" +
+                                "FROM pg_attribute a\n" +
+                                "JOIN pg_class c ON a.attrelid = c.oid\n" +
```
leonardBang: Is the official recommended way to obtain history schema of a Postgres table ?

zml1206:
Could a configuration option be added to control the on/off state? Also, I'm implementing a way to parse DDL SQL to support DDL, which requires adding an event table in Postgres. The two methods don't seem to conflict, is this approach acceptable to the community?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/utils/PostgresSchemaUtils.java
linjianchang: > Is the official recommended way to obtain history schema of a Postgres table ?

Here, we are only obtaining the OID of the schema field in the current state of the Postgres table, and determining whether the field has undergone a rename operation based on whether the OID remains consistent

linjianchang:
> added to control the on/off state

Already added to control the on/off state，thanks for suggestion！

lvyanquan:
Hi @Mrart CC as you may be interested in this.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresDataSource.java
```diff
@@ -67,12 +72,17 @@ public PostgresDataSource(
     public EventSourceProvider getEventSourceProvider() {
         String databaseName = postgresSourceConfig.getDatabaseList().get(0);
         boolean includeDatabaseInTableId = postgresSourceConfig.isIncludeDatabaseInTableId();
+        beforeTableColumnsOidMaps =
+                PostgresSchemaUtils.getAllTablesColumnOids(
+                        postgresSourceConfig, postgresSourceConfig.getTableList());
```
JNSimba, line 77: If the parallelism of PostgresDataSource is greater than 1, will there be any inconsistency issues here?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresEventDeserializer.java
```diff
@@ -31,63 +42,375 @@
...
+    private Map<String, String> getRenameColumnMaps(
+            TableId tableId,
+            List<String> newAddColumnNames,
+            List<String> newDelColumnNames,
+            Schema oldSchema,
+            Schema afterSchema,
+            Map<String, Integer> beforeColumnsOidMaps) {
+        Map<String, String> renameColumnMaps = new HashMap<>();
+        Map<String, Integer> newAddColumnsOidMaps =
+                PostgresSchemaUtils.getColumnOids(postgresSourceConfig, tableId, newAddColumnNames);
```
JNSimba, line 365: Since DDL is driven by DML, consider this scenario:
1. DDL is executed at T1
2. DML is executed at T2
3. DDL is executed at T3

If a query is performed at T2, the obtained OID might be from T3, which could cause problems.

By the way, it seems difficult to distinguish between DROP+ADD and Rename？

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresDataSource.java
```diff
@@ -50,6 +54,7 @@ public class PostgresDataSource implements DataSource {
     private final PostgresSourceConfig postgresSourceConfig;
 
     private final List<PostgreSQLReadableMetadata> readableMetadataList;
+    Map<TableId, Map<String, Integer>> beforeTableColumnsOidMaps = new HashMap<>();
```
loserwang1024, line 57: typo: private final

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresEventDeserializer.java
```diff
@@ -31,63 +42,375 @@
...
     @Override
     protected List<SchemaChangeEvent> deserializeSchemaChangeRecord(SourceRecord record) {
         return Collections.emptyList();
     }
 
+    @Override
+    public List<? extends Event> deserialize(SourceRecord record) throws Exception {
+        List<Event> result = new ArrayList<>();
+        if (postgresSourceConfig.isIncludeSchemaChanges()) {
+            handleSchemaChange(record, result);
```
loserwang1024, line 119: the record is not schema change , we have isSchemaChangeRecord(record) in the bellow. I think method named `inferSchemaChangeByRecord` will be better

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresEventDeserializer.java
```diff
@@ -31,63 +42,375 @@
...
 /** Event deserializer for {@link PostgresDataSource}. */
 @Internal
 public class PostgresEventDeserializer extends DebeziumEventDeserializationSchema {
-
+    private static final Logger LOG = LoggerFactory.getLogger(PostgresEventDeserializer.class);
     private static final long serialVersionUID = 1L;
     private List<PostgreSQLReadableMetadata> readableMetadataList;
     private final boolean includeDatabaseInTableId;
     private final String databaseName;
+    private Map<TableId, Schema> schemaMap = new HashMap<>();
```
loserwang1024, line 72: ---
The `schemaMap` can be restored from the `StreamSplit`, specifically from its `tableSchemas` field upon job restart. To ensure correctness, the `tableSchemas` in the `StreamSplit` should be updated during each checkpoint to reflect any changes in the schema.

Otherwise, after a job restart, the `schemaMap` will be empty, and the previous schema will be `null`. In your current design, it seems that even if the schema of newly arriving data changes, no schema change event will be generated.

You can address this by passing the schema information through  `org.apache.flink.cdc.debezium.event.DebeziumEventDeserializationSchema#applyChangeEvent`.

---
To be honest, I believe that `DebeziumEventDeserializationSchema#createTableEventCache` is not necessary for PostgreSQL CDC. We can achieve the same — and more powerful — functionality using the `schemaMap`.

### ✅ My Proposed Design for CreateTableEvent

1. **Replace `PostgresPipelineRecordEmitter#createTableEventCache` with `schemaMap`**:  
   Use `schemaMap` as the source of truth for table schemas instead of maintaining a separate cache.

2. **Remove the logic in `PostgresPipelineRecordEmitter` that generates `CreateTableEvent` from the table ID**:
   ```java
   // In rare cases, we may miss some CreateTableEvents before DataChangeEvents.
   // Don't send CreateTableEvent for SchemaChangeEvents as it's the latest schema.
   if (isDataChangeRecord && !createTableEventCache.containsKey(tableId)) {
       CreateTableEvent createTableEvent = getCreateTableEvent(sourceConfig, tableId);
       output.collect((T) createTableEvent);
       createTableEventCache.put(tableId, createTableEvent);
   }
   ```

3. **Move schema change handling to `PostgresEventDeserializer#handleSchemaChange`**:  
   If the previous schema is `null`, emit a `CreateTableEvent`. This ensures that all schema change events are centralized within the deserializer rather than being scattered across different components.

---

### ✅ Summary

By centralizing schema change handling in `PostgresEventDeserializer` and leveraging `schemaMap` via `StreamSplit`, we can simplify the architecture, reduce redundancy, and improve the robustness of schema tracking in PostgreSQL CDC scenarios.

Let me know if you'd like help writing this as a pull request description, or need further clarification on any part!

@leonardBang , WDYT?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresEventDeserializer.java
```diff
@@ -31,63 +42,375 @@
...
+
+    private void handleSchemaChange(SourceRecord record, List<Event> result) {
+        TableId tableId = getTableId(record);
+        Schema valueSchema = record.valueSchema();
+        Schema beforeSchema = schemaMap.get(tableId);
+        List<String> beforeColumnNames;
+        List<String> afterColumnNames;
+        Schema afterSchema = fieldSchema(valueSchema, Envelope.FieldName.AFTER);
+        List<Field> afterFields = afterSchema.fields();
+        org.apache.flink.cdc.common.schema.Schema schema =
```
loserwang1024, line 141:
```java
 org.apache.flink.cdc.common.schema.Schema schema =
                PostgresSchemaUtils.getTableSchema(postgresSourceConfig, tableId);
```

It's too heavy that lookup current schema from database for each record. In production, it cannot work.

Why not just just afterSchema and beforeSchema? why stiil need to lookup current schema?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresEventDeserializer.java
```diff
@@ -31,63 +42,375 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(PostgresEventDeserializer.class);
     private static final long serialVersionUID = 1L;
     private List<PostgreSQLReadableMetadata> readableMetadataList;
     private final boolean includeDatabaseInTableId;
     private final String databaseName;
+    private Map<TableId, Schema> schemaMap = new HashMap<>();
+    private final PostgresSourceConfig postgresSourceConfig;
 
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private Map<TableId, Map<String, Integer>> beforeTableColumnsOidMaps;
```
loserwang1024, line 76: ### problem
The beforeTableColumnsOidMaps is only populated during the YAML compilation phase. When the job is restarted without recompilation—such as in the case of a failover or other non-recompile-based restarts—any modifications to this map within PostgresEventDeserializer will not take effect.

As a result, two potential issues may arise:
* If the job has significant read lag, and it is later recompiled and restarted, the beforeTableColumnsOidMaps could be newer than the actual data being processed.
* If the job is simply restarted without recompilation, PostgresPipelineSource recovers from its serialized state and retrieves an outdated initial value that lags significantly behind the current log position.

### suggestion
However, I question whether comparing OIDs is strictly necessary to determine if a table has been renamed or dropped and recreated. In the default lenient mode:

* Column Rename: In downstream systems, a renamed column is typically handled by omitting the old column and introducing a new one. This is functionally similar to removing the column (which is ignored) and then adding a new one with the same name.
* Drop and Recreate Same Column Name: In lenient mode, this scenario behaves identically to no change at all.

Therefore, I recommend not using beforeTableColumnsOidMaps for now, especially since the evolve mode is still immature and large-scale data scenarios generally avoid dropping or modifying downstream data structures. This would significantly simplify the current design.

@leonardBang , WDYT?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/test/java/org/apache/flink/cdc/connectors/postgres/source/PostgresPipelineITCaseTest.java
```diff
@@ -157,6 +163,278 @@ public void testInitialStartupMode() throws Exception {
...
+                        getJdbcConnection(POSTGRES_CONTAINER, inventoryDatabase.getDatabaseName());
+                Statement stmt = conn.createStatement()) {
+            stmt.execute(
+                    "INSERT INTO inventory.products (name, description, weight) "
+                            + "VALUES ('scooter', 'Small 2-wheel scooter', 3.14)");
+            List<Event> actual;
+            List<Event> expected;
+            Map<String, String> renameMap = new HashMap<>();
+            switch (testType) {
+                case "modifyType":
```
loserwang1024, line 227: In your code,also includes change of `oldField.schema().defaultValue()` and `oldField.schema().parameters()`, please also test them.

loserwang1024 (commented):
Thanks for your job which mean a lot for CDC. I have left some comments:

loserwang1024:
@linjianchang if you don't have time, could I help to modified based on your CR? And I will reserve your commit.

loserwang1024:
I have reviewd this PR(by @linjianchang )and https://github.com/apache/flink-cdc/pull/4259 by @zml1206 . Thanks for both hard work！

I have also done some research on the Debezium codebase and PostgreSQL replication, and I have another design idea that prefers to leverage PostgreSQL’s native capabilities more fully. So I’d like to discuss it with you.

------------------------------------------------------

### PostgreSQL Pgoutput's relation message in logical replication
As document of  PostgreSQL 16 says:  [Section 55.5.3 - Logical Replication Protocol Message Flow](https://www.postgresql.org/docs/16/protocol-logical-replication.html)：

>  "Every DML message contains a relation OID, identifying the publisher's relation that was acted on. Before the first DML message for a given relation OID, a Relation message will be sent, describing the schema of that relation. Subsequently, a new Relation message will be sent if the relation's definition has changed since the last Relation message was sent for it."


When DDL changes are executed in PostgreSQL, no corresponding logs are generated. However, if the pgoutput sender is about to send the first DML message for a new schema, it will first send a Relation message.

The relation message will include the schema of table.

### how Debezium use it?
Depending on The decoder plug-in, schema updates take two completely different paths:
● pgoutput: Sends a correlation message before each DML event,You can use the `applySchemaChangesForTable` to actively update schema in advance. `shouldSchemaBeSynchronized()` returns false, so `synchronizeTableSchema() `is an empty operation for DML events.
● decoderbufs: If no RELATION message is sent, `shouldSchemaBeSynchronized()` returns true (the default value). The schema is synchronized by comparing the message column with the in-memory schema in a reactive manner.
Thus, if it is pgoutput, we can send schema change events on demand without comparing each message?

### What cdc need to do?
Therefore, my personal opinion:
1. Extend the PostgresSchema and pass in a dispatcher. When a correlation message is received,Put the schema in the event queue as a special message
2. When Postgres RecordEmitter receives the schema event, it:
3. Update schemas in split,For persistence (in the current master branch, since pg cdc does not have a schema change event, the information in schemas will never be updated)
4. Compare table changes and send scheme ddl in yaml.
image


In this way, we can avoid intrusive changes to PostgreSQL through triggers, and there is also no need to compare schemas for every message. More importantly, it updates the schemas stored in the split, so schema consistency can still be guaranteed after state recovery or restart.

Though only useful for pgoutput, but I think it's enough. Because, the pgoutput is the only official replication plugin. 


@linjianchang @zml1206 @leonardBang , WDYT?

here is my poc code: https://github.com/loserwang1024/flink-cdc-connectors/tree/poc-schema-change-event

leonardBang:
> I have reviewd this PR(by @linjianchang )and #4259 by @zml1206 . Thanks for both hard work！
> 
> I have also done some research on the Debezium codebase and PostgreSQL replication, and I have another design idea that prefers to leverage PostgreSQL’s native capabilities more fully. So I’d like to discuss it with you.
> 
> ### PostgreSQL Pgoutput's relation message in logical replication
> As document of PostgreSQL 16 says: [Section 55.5.3 - Logical Replication Protocol Message Flow](https://www.postgresql.org/docs/16/protocol-logical-replication.html)：
> 
> > "Every DML message contains a relation OID, identifying the publisher's relation that was acted on. Before the first DML message for a given relation OID, a Relation message will be sent, describing the schema of that relation. Subsequently, a new Relation message will be sent if the relation's definition has changed since the last Relation message was sent for it."
> 
> When DDL changes are executed in PostgreSQL, no corresponding logs are generated. However, if the pgoutput sender is about to send the first DML message for a new schema, it will first send a Relation message.
> 
> The relation message will include the schema of table.
> 
> ### how Debezium use it?
> Depending on The decoder plug-in, schema updates take two completely different paths: ● pgoutput: Sends a correlation message before each DML event,You can use the `applySchemaChangesForTable` to actively update schema in advance. `shouldSchemaBeSynchronized()` returns false, so `synchronizeTableSchema() `is an empty operation for DML events. ● decoderbufs: If no RELATION message is sent, `shouldSchemaBeSynchronized()` returns true (the default value). The schema is synchronized by comparing the message column with the in-memory schema in a reactive manner. Thus, if it is pgoutput, we can send schema change events on demand without comparing each message?
> 
> ### What cdc need to do?
> Therefore, my personal opinion:
> 
> 1. Extend the PostgresSchema and pass in a dispatcher. When a correlation message is received,Put the schema in the event queue as a special message
> 2. When Postgres RecordEmitter receives the schema event, it:
> 3. Update schemas in split,For persistence (in the current master branch, since pg cdc does not have a schema change event, the information in schemas will never be updated)
> 4. Compare table changes and send scheme ddl in yaml.
> 
> image
> In this way, we can avoid intrusive changes to PostgreSQL through triggers, and there is also no need to compare schemas for every message. More importantly, it updates the schemas stored in the split, so schema consistency can still be guaranteed after state recovery or restart.
> 
> Though only useful for pgoutput, but I think it's enough. Because, the pgoutput is the only official replication plugin.
> 
> @linjianchang @zml1206 @leonardBang , WDYT?
> 
> here is my poc code: https://github.com/loserwang1024/flink-cdc-connectors/tree/poc-schema-change-event

Thanks @loserwang1024 for the detailed investigation！Here are my opinions: 

(1) Querying the PostgreSQL database table for every CDC event is not feasible — this approach will directly overwhelm and crash the business database once the data volume grows even slightly.

(2) The trigger-based solution is also unacceptable, as it directly intrudes into the business write workflow on the database. Downstream CDC subscribers may not have the permission to implement this, and requiring all business teams to accept such triggers is not realistic.

The approach of subscribing to the latest table schema based on OID in real-time is conceptually sound. It is non-intrusive to the business database, avoids frequent polling of the upstream DB as traffic increases, and ensures the ordering between schema events and data events. Debezium also uses this mechanism — +1 to implementing our solution based on this approach.

zml1206:
> The approach of subscribing to the latest table schema based on OID in real-time is conceptually sound. It is non-intrusive to the business database, avoids frequent polling of the upstream DB as traffic increases, and ensures the ordering between schema events and data events. Debezium also uses this mechanism — +1 to implementing our solution based on this approach.

+1

linjianchang:
> @linjianchang if you don't have time, could I help to modified based on your CR? And I will reserve your commit.

Sure, I haven't had time to make the revision recently because I've been busy with other things，thanks for review！

lvyanquan:
Hi @linjianchang, this can be closed as https://github.com/apache/flink-cdc/pull/4316 is merged.

Issue #2105, opened 2023-04-26
[Bug][Mysql] Starting flink with a savepoint and adding a new table for synchronization is problematic


hiSandog:
### Search before asking

- [X] I searched in the [issues](https://github.com/ververica/flink-cdc-connectors/issues) and found nothing similar.

### Flink version

1.14.2

### Flink CDC version

2.2.1

### Database and its version

mysql 8.0.22

### Minimal reproduce step

1. 使用startupOptions : initial 方式启动 flink cdc，同步若干张表，比如 (10 张)，这时候没有问题，很顺利，不仅快照数据全部进入kafka，后续的binlog 数据也都能进入
    Starting the flink cdc with startupOptions: initial and syncing a few tables, say (10), worked fine，Not only the snapshot data will all enter kafka, but also the subsequent binlog data
2. 通过执行 `flink stop -t yarn-application --savepointPath xxx  -Dyarn.application.id={flink_yarn_id} {flink_id} ` 停止flink 任务，并且设置保存点
stop the flink job by running 'flink stop -t yarn-application --savepointPath xxx-Dyarn.application. id={flink_yarn_id} {flink_id}', And set the savepoints
3. 有些新增需要同步的表加到了启动参数里面，这些表的数据量大概 100多万，使用 `flink run --detached  --fromSavepoint xxx`，通过刚才的保存点恢复任务。
`flink run --detached --fromSavepoint xxx `is used to retrieve a few new tables that need to be synchronized.

### What did you expect to see?

flink cdc 在同步完这些新增表的快照后，会连同之前的表一起同步新的binlog

After the flink cdc syncs the snapshots of these new tables, it syncs the new binlog along with the previous tables

### What did you see instead?

flink cdc 的日志确实显示了有新的表加入，也确实开始同步了新的表的快照数据，可是在同步新的表的快照时候同步了一部分之后就似乎停止同步了，然后也没有同步任何的binlog数据，仿佛所有的任务全部卡死，也没有任何的报错。
The flink cdc logs do show that new tables have been added, and it does start syncing snapshots of new tables, but it seems to stop syncing after a few syncs of the new table snapshot, and then it doesn't sync any binlog data, as if all tasks are stuck without any errors.

![image](https://user-images.githubusercontent.com/24449799/234463648-2781e895-03a8-4f0e-9fd0-0fad044b15f3.png)
[...]

PR #4246, opened 2026-01-30
[Flink-38911][cdc connector mysql] mysql-cdc-connector datastream support scan.binlog.newly-added-table.enabled 

ThorneANN:
This commit implements the binlog-only newly added table capture feature for MySQL CDC DataStream connector, allowing dynamic table discovery without snapshot phase.

Key changes:

Add new config option 'scan.binlog.newly-added-table.enabled' in MySqlSourceOptions
Add scanBinlogNewlyAddedTableEnabled field and getter in MySqlSourceConfig
Implement table pattern conversion from Flink CDC style to Debezium style in MySqlSourceConfigFactory
Add validation logic to ensure binlog-only mode works only with stream-only startup modes
Enhance BinlogSplitReader to auto-capture newly added tables matching the pattern
Add logging in MySqlSnapshotSplitAssigner for binlog-only mode
Expose scanBinlogNewlyAddedTableEnabled() API in MySqlSourceBuilder
The feature converts table patterns (e.g., "db.table_.") to Debezium regex style (e.g., "db.table_.")
and enables dynamic table discovery during binlog reading phase without triggering snapshots.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
@@ -18,6 +18,7 @@
 package org.apache.flink.cdc.connectors.mysql.source.config;
 
 import org.apache.flink.cdc.common.annotation.Internal;
+import org.apache.flink.cdc.common.route.TableIdRouter;
 import org.apache.flink.cdc.connectors.mysql.debezium.EmbeddedFlinkDatabaseHistory;
 import org.apache.flink.cdc.connectors.mysql.source.MySqlSource;
 import org.apache.flink.cdc.connectors.mysql.table.StartupOptions;
@@ -68,6 +69,7 @@ public class MySqlSourceConfigFactory implements Serializable {
     private boolean includeHeartbeatEvents = false;
     private boolean includeTransactionMetadataEvents = false;
     private boolean scanNewlyAddedTableEnabled = false;
+    private boolean scanBinlogNewlyAddedTableEnabled = false;
     private boolean closeIdleReaders = false;
     private Properties jdbcProperties;
     private Duration heartbeatInterval = MySqlSourceOptions.HEARTBEAT_INTERVAL.defaultValue();
@@ -258,6 +260,17 @@ public class MySqlSourceConfigFactory implements Serializable {
         return this;
     }
 
+    /**
+     * Whether to capture newly added tables in binlog reading phase without snapshot. This option
+     * can only be used with stream-only startup modes. Cannot be enabled together with {@link
+     * #scanNewlyAddedTableEnabled(boolean)}.
+     */
+    public MySqlSourceConfigFactory scanBinlogNewlyAddedTableEnabled(
+            boolean scanBinlogNewlyAddedTableEnabled) {
+        this.scanBinlogNewlyAddedTableEnabled = scanBinlogNewlyAddedTableEnabled;
+        return this;
+    }
+
     /** Custom properties that will overwrite the default JDBC connection URL. */
     public MySqlSourceConfigFactory jdbcProperties(Properties jdbcProperties) {
         this.jdbcProperties = jdbcProperties;
@@ -397,8 +410,21 @@ public class MySqlSourceConfigFactory implements Serializable {
         if (databaseList != null) {
             props.setProperty("database.include.list", String.join(",", databaseList));
         }
+        // Validate: Two modes are mutually exclusive
+        if (scanBinlogNewlyAddedTableEnabled && scanNewlyAddedTableEnabled) {
+            throw new IllegalArgumentException(
+                    "Cannot enable both 'scan.binlog.newly-added-table.enabled' and "
+                            + "'scan.newly-added-table.enabled' as they may cause duplicate data");
+        }
+
         if (tableList != null) {
-            props.setProperty("table.include.list", String.join(",", tableList));
+            if (scanBinlogNewlyAddedTableEnabled) {
+                props.setProperty(
+                        "table.include.list",
+                        TableIdRouter.convertTableListToRegExpPattern(String.join(",", tableList)));
+            } else {
+                props.setProperty("table.include.list", String.join(",", tableList));
+            }
         }
         if (serverTimeZone != null) {
             props.setProperty("database.serverTimezone", serverTimeZone);
@@ -436,6 +462,7 @@ public class MySqlSourceConfigFactory implements Serializable {
                 includeHeartbeatEvents,
                 includeTransactionMetadataEvents,
                 scanNewlyAddedTableEnabled,
+                scanBinlogNewlyAddedTableEnabled,
                 closeIdleReaders,
                 props,
                 jdbcProperties,
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceOptions.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceOptions.java
@@ -244,6 +244,22 @@ public class MySqlSourceOptions {
                             "Whether capture the scan the newly added tables or not, by default is false. This option is only useful when we start the job from a savepoint/checkpoint.");
 
     @Experimental
+    public static final ConfigOption<Boolean> SCAN_BINLOG_NEWLY_ADDED_TABLE_ENABLED =
+            ConfigOptions.key("scan.binlog.newly-added-table.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "In binlog reading stage, whether to capture newly added tables "
+                                    + "that match the table patterns. When enabled, new tables will be "
+                                    + "captured without snapshot, only binlog events will be emitted. "
+                                    + "Cannot be enabled together with 'scan.newly-added-table.enabled'. "
+                                    + "table-name pattern examples: "
+                                    + "'db.\\.*' (all tables in database 'db'), "
+                                    + "'db.user_\\.*' (tables like 'user_orders', 'user_profiles'), "
+                                    + "'db.order_[0-9]+' (tables like 'order_1', 'order_2'), "
+                                    + "'db1.\\.*,db2.user_\\.*' (all tables in 'db1' and 'user_*' tables in 'db2').");
+
+    @Experimental
     public static final ConfigOption<String> SCAN_INCREMENTAL_SNAPSHOT_CHUNK_KEY_COLUMN =
             ConfigOptions.key("scan.incremental.snapshot.chunk.key-column")
                     .stringType()
--- a/docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md
+++ b/docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md
@@ -397,6 +397,21 @@ Flink SQL> SELECT * FROM orders;
       <td>是否在快照结束后关闭空闲的 Reader。 此特性需要 flink 版本大于等于 1.14 并且 'execution.checkpointing.checkpoints-after-tasks-finish.enabled' 需要设置为 true。</td>
     </tr>
     <tr>
+      <td>scan.binlog.newly-added-table.enabled</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">false</td>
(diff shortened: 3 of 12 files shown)


ThorneANN:
@copilot review

ThorneANN:
@Copilot review

ThorneANN:
help retrigger ci @lvyanquan

ThorneANN:
can u help me retrigger ci once  @lvyanquan

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
```diff
@@ -222,6 +222,16 @@ else if (!isRemainingTablesCheckpointed
...
+        // Binlog-only mode: no action needed in Assigner, BinlogSplitReader handles it
+        if (sourceConfig.isScanBinlogNewlyAddedTableEnabled()) {
+            LOG.info(
+                    "Binlog-only newly added table capture is enabled. "
+                            + "New tables matching the pattern will be automatically captured "
+                            + "in binlog phase without snapshot.");
+            // No action needed here, BinlogSplitReader will handle the auto-capture
+            return;
+        }
+
```
yuxiqian: Seems this check is unnecessary since `sourceConfig.isScanNewlyAddedTableEnabled` and `sourceConfig.isScanBinlogNewlyAddedTableEnabled` will not be enabled simultaneously?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
```diff
@@ -446,4 +481,47 @@ public MySqlSourceConfig createConfig(int subtaskId, String serverName) {
...
+     *
+     * <ul>
+     *   <li>{@code "db1.table_\.*,db2.user_\.*"} -> {@code "db1\.table_.*|db2\.user_.*"}
+     *   <li>{@code "test_db.orders"} -> {@code "test_db\.orders"}
+     * </ul>
+     *
+     * @param tables Flink CDC style table pattern
+     * @return Debezium style table pattern
+     */
+    private static String convertToDebeziumStyle(String tables) {
```
yuxiqian: Avoid rewriting `TableIdRouter#convertTableListToRegExpPattern`?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceOptions.java
```diff
@@ -243,6 +243,22 @@ public class MySqlSourceOptions {
...
+                    .withDescription(
+                            "In binlog reading stage, whether to capture newly added tables "
+                                    + "that match the table patterns. When enabled, new tables will be "
+                                    + "captured without snapshot, only binlog events will be emitted. "
+                                    + "Cannot be enabled together with 'scan.newly-added-table.enabled'. "
+                                    + "table-name pattern examples: "
+                                    + "'db\\.*' (all tables in database 'db'), "
+                                    + "'db\\.user_\\.*' (tables like 'user_orders', 'user_profiles'), "
+                                    + "'db\\.order_[0-9]+' (tables like 'order_1', 'order_2'), "
+                                    + "'db1\\.*,db2\\.user_\\.*' (all tables in 'db1' and 'user_*' tables in 'db2').");
```
yuxiqian: The example seems incorrect, should be `db.\\.*`, `db.user_\\.*` ...

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
```diff
@@ -397,8 +413,26 @@ public MySqlSourceConfig createConfig(int subtaskId, String serverName) {
...
+                            + "'scan.newly-added-table.enabled' as they may cause duplicate data");
+        }
+
         if (tableList != null) {
-            props.setProperty("table.include.list", String.join(",", tableList));
+            // Convert table patterns to Debezium style if binlog auto-capture is enabled
+            if (scanBinlogNewlyAddedTableEnabled) {
+                String originalPattern = String.join(",", tableList);
+                String debeziumPattern = convertToDebeziumStyle(originalPattern);
+                props.setProperty("table.include.list", debeziumPattern);
```
yuxiqian: IIUC MySQL DataStream connector uses standard RegEx (`.` is wildcard matcher and `\.` is the dot character). But after your modification and binlog newly tables capturing enabled, `.` will be the dot character and db/table delimiter while `\.` is the meta character, which is strange. Keeping the same behavior in binlog capturing mode unchanged would make more sense.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlSnapshotSplitAssigner.java
ThorneANN: yes

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
ThorneANN: may be batter

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceOptions.java
ThorneANN: U are right and i  forget it , fix it soon

yuxiqian:
Kindly reminder: you may run `git rebase master` to keep commit history clean.

ThorneANN:
@yuxiqian  Please help retrigger ci,tks

ThorneANN:
> @yuxiqian Please help retrigger ci,tks

@yuxiqian Can u help to review the ci fail,because i run success on local env.

lvyanquan:
Hi @ThorneANN maybe you can try to rebase master to fix these failures.

ThorneANN:
> Hi @ThorneANN maybe you can try to rebase master to fix these failures.

Tks,and i will try it.

ThorneANN:
@lvyanquan Please help retrigger ci,tks

J9527H:
Hi @ThorneANN @yuxiqian, thanks for working on this — we'd really benefit from this feature.

We're running a production MySQL CDC pipeline using the DataStream API (`MySqlSource` + custom `DebeziumDeserializationSchema`) on Apache Flink (Flink 2.2 / flink-cdc 3.6.0-2.2). Our job relies on custom parsing, filtering, and DLQ routing logic that's tightly coupled to the DataStream API — migrating to the Pipeline (YAML) connector isn't an option for us without losing that flexibility.

Our real-world use case: we periodically need to add new tables to a long-running job, but **we don't need historical/snapshot data for those new tables** — only incremental binlog events going forward. Today our only options are:

1. `scanNewlyAddedTableEnabled(true)` — always triggers a snapshot phase, which doesn't match our requirement, and per #2105 has also been reported to occasionally hang during the snapshot phase.
2. Run a separate, independent job per newly-added table using `StartupOptions.latest()` — works, but doesn't scale operationally.
3. Manually extract the last committed binlog offset from TaskManager checkpoint logs and cold-start with `StartupOptions.specificOffset(...)` — works, but is manual, error-prone, and not officially documented for this use case (adding tables vs. failure recovery).

Having `scanBinlogNewlyAddedTableEnabled` on `MySqlSourceBuilder`, consistent with what's already available in `MySqlDataSourceFactory` for the Pipeline connector, would let us avoid all three workarounds above.

Happy to test against a patched build if that helps move this forward. Let me know if there's anything I can do to help (testing, providing more context on our use case, etc.).

bigdataMS:
Hi @ThorneANN, thanks for the great work on this feature — it addresses a real gap for DataStream users who need to add tables to a long-running job **without** a snapshot phase (the same use case @J9527H described above, and one we hit in production too).

Since the PR has been idle for a few months, would you be OK with me helping push it forward? I'd rebase onto the latest master to get CI green and address the outstanding review feedback. Of course I'll keep your authorship via `Co-authored-by` and credit this as the original work.

@yuxiqian @lvyanquan — if that's acceptable, my plan would be:
1. Rebase on master, fix CI.
2. Address @yuxiqian's point on regex semantics: drop the custom `.`/`\.` remapping and keep binlog-capture mode consistent with the connector's standard pattern semantics (this also removes the need to rewrite `TableIdRouter#convertTableListToRegExpPattern`).
3. Add the missing stream-only startup-mode validation for `scan.binlog.newly-added-table.enabled`.
4. Remove the redundant mutual-exclusion check in `MySqlSnapshotSplitAssigner` and fix the doc regex examples.
5. Validate end-to-end on a real DataStream pipeline (@J9527H kindly offered to help test).

Happy to do this as a new PR that supersedes #4246, or by pushing to @ThorneANN's branch if preferred — whichever the maintainers think is cleaner. Please let me know.

ThorneANN:
@J9527H @bigdataMS  Hi, thank you for your suggestions and i find that a new pr https://github.com/apache/flink-cdc/pull/4409 to capture new tables without restarting flink's job.
 I think we need focus on that pr may be this is not a good ides at present.

bigdataMS:
Thanks for the update @ThorneANN, and thanks for cleaning up the commit history.

I took a close look at #4409 before replying, and I believe the two changes are orthogonal rather than overlapping — I'd hate to see this one dropped on that basis.

**#4409 answers "which tables should I subscribe to?"** It introduces the `ObjectIdDiscoverer<T>` / `TableDiscoverer` SPI plus a JDBC implementation, so the subscription list can be read dynamically from a metadata table or a custom query. All 11 files land under `flink-cdc-common/.../source/discover/`.

**#4246 answers "a new table showed up — how do we start reading it?"** Specifically: pick it up mid-binlog *without* running a snapshot, so only forward increments are emitted.

Concretely:

- As far as I can tell from searching master, `TableDiscoverer` currently has **no callers outside its own package** — the SPI is in place but not yet wired into any source, so it doesn't change runtime behaviour for MySQL CDC today.
- Its configuration surface (`table.discoverer.jdbc.*` under `source:`) targets the YAML pipeline. The use case @J9527H and I have is the DataStream API with a custom deserializer.
- It doesn't define snapshot semantics at all. Even once a source consumes it, newly discovered tables would presumably go through the existing `scan.newly-added-table.enabled` path, which *does* snapshot. Skipping the snapshot is exactly the point of this PR: for a table that has been live for months, we only want increments from now on, and we backfill history separately through a bounded job.

If anything the two compose nicely later: a discoverer decides *what* to add, and binlog-only mode decides *how* it gets read.

On the current state of this PR — I have a branch rebased onto master (no conflicts) that addresses the outstanding review feedback:

1. Dropped the custom `.` / `\.` remapping and the `TableIdRouter.convertTableListToRegExpPattern` call, so `table.include.list` keeps the connector's standard regex semantics in both modes (@yuxiqian's point). This also removes the need to touch `TableIdRouter` at all.
2. Added the missing stream-only startup-mode validation for `scan.binlog.newly-added-table.enabled`.
3. Fixed the regex examples in `MySqlSourceOptions` and the EN/ZH docs (`db\.*` → `db\..*`, `db\.user_\.*` → `db\.user_.*`), plus three non-standard patterns in the ITCase.
4. Re: Copilot's comment about a redundant check in `MySqlSnapshotSplitAssigner` — the current version of this PR doesn't touch that file, so there's nothing to remove.

`spotless`, compile and `MySqlTableSourceFactoryTest` pass locally. The ITCase needs Docker, which I can't run in my current environment, so I'd rely on CI for that.

@ThorneANN if you'd still rather not continue, I'm happy to either open a PR against your branch so the changes land under #4246 with your authorship, or open a superseding PR keeping you as the commit author — whichever you prefer.

@yuxiqian @lvyanquan could you weigh in on whether this feature is still wanted independently of #4409? If the maintainers' view is that the discoverer SPI is the intended long-term direction for this use case, I'd rather know now than keep pushing on it.

PR #4267, opened 2026-02-09
[FLINK-38236][mongodb] Add metadata support for full document in MongoDB CDC connector

Hisoka-X:
This close https://issues.apache.org/jira/browse/FLINK-38236

This pull request adds support for exposing the full document of MongoDB change events as a metadata column in the Flink MongoDB CDC connector.


Jiabao-Sun (commented):
Thanks @Hisoka-X for this contribution, but I believe using the full document as a metadata column does not align well with the original intention of the table API's strong typing for data processing. For flexibly reading full document JSON data, I would recommend using the Flink Stream API instead. Additionally, using the Variant type introduced by FLIP-521 for mapping MongoDB's dynamic types might be a better choice.

Hisoka-X:
> Additionally, using the Variant type introduced by FLIP-521 for mapping MongoDB's dynamic types might be a better choice.

Ok, we should wait flink dependency version upgrade to 2.1.0.

PR #4297, opened 2026-03-03
[FLINK-39198] Fix data loss during restoration from checkpoint in TIMESTAMP startup mode.

danzhewuju:
Fix: https://issues.apache.org/jira/browse/FLINK-39198

Flink CDC creates an eventFilter inside the MySqlBinlogSplitReadTask. This filter is produced through the following method. In timestamp startOptions mode, the eventFilter will exclude events whose timestamps are earlier than the user-defined timestamp. However, if the job resumes from a savepoint or checkpoint and the system automatically sets the timestamp parameter, this approach can result in data loss.

In this scenario, the user configures a startOption with timestamp mode (which may be the default behavior, tied to the Flink job’s submission time) and restarts the Flink CDC job from a savepoint. If the savepoint contains MySQL binlog information including a ts_sec value, and that ts_sec is earlier than the user-defined timestamp, data loss may occur.

### Brief change log

Prevent filtering out binlog events when resuming from savepoints/checkpoints with older ts_sec

### Verifying this change

Test result:  
Adding split(s) to reader: [MySqlBinlogSplit{splitId='binlog-split', offset={transaction_id=null, ts_sec=1772531889, file=***.577482, pos=883008898, kind=SPECIFIC, gtids=***, row=1, event=3, server_id=5212, timestamp_strictly=false}, endOffset={ts_sec=0, file=, pos=-9223372036854775808, kind=NON_STOPPING, row=0, event=0, timestamp_strictly=false}, isSuspended=false}]
Creating an event filter that drops row mutation events occurring before the destination timestamp in seconds 1772531889

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
@@ -408,6 +408,9 @@ public class BinlogSplitReader implements DebeziumReader<SourceRecords, MySqlSpl
                                 + "configuration. ");
             }
             if (startingOffset.getOffsetKind() != BinlogOffsetKind.TIMESTAMP) {
+                LOG.info(
+                        "The starting offset kind is {}, no need to apply timestamp filter",
+                        startingOffset.getOffsetKind());
                 return event -> true;
             }
 
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
@@ -22,6 +22,7 @@ import org.apache.flink.cdc.connectors.mysql.debezium.task.MySqlBinlogSplitReadT
 import org.apache.flink.cdc.connectors.mysql.debezium.task.context.StatefulTaskContext;
 import org.apache.flink.cdc.connectors.mysql.source.config.MySqlSourceConfig;
 import org.apache.flink.cdc.connectors.mysql.source.offset.BinlogOffset;
+import org.apache.flink.cdc.connectors.mysql.source.offset.BinlogOffsetKind;
 import org.apache.flink.cdc.connectors.mysql.source.split.FinishedSnapshotSplitInfo;
 import org.apache.flink.cdc.connectors.mysql.source.split.MySqlBinlogSplit;
 import org.apache.flink.cdc.connectors.mysql.source.split.MySqlSplit;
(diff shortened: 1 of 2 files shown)


haruki-830 (approved):
+1. It might be helpful to add a brief code comment explaining the submission-time-based timestamp scenario for future readers.

danzhewuju:
> +1. It might be helpful to add a brief code comment explaining the submission-time-based timestamp scenario for future readers.

Thank you. I’ve added a comment to clarify the recovery scenario: a TIMESTAMP-startup job may restore using a concrete, ⁠SPECIFIC offset; therefore, we should honor the restored split offset rather than reapplying the original submission-time timestamp filter.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
```diff
@@ -421,6 +424,41 @@ private Predicate<Event> createEventFilter() {
...
+                "Creating an event filter that drops row mutation events occurring before the destination"
+                        + " timestamp in seconds {}",
+                startTimestampSec);
+        return event -> {
+            if (!EventType.isRowMutation(getEventType(event))) {
+                return true;
+            }
+            return event.getHeader().getTimestamp() >= startTimestampSec * 1000;
+        };
+    }
```
lvyanquan: Thanks for working on this. I think the fix can be smaller and also avoid an edge case.

In the original code, `createEventFilter()` uses `startupOptions.binlogOffset.getTimestampSec()` from the current source config. However, after restoring from checkpoint/savepoint, the source config still keeps `startupMode = TIMESTAMP`, while the restored split may already contain the actual resume offset in `currentBinlogSplit.getStartingOffset()`.

So I think we can fix this by changing only the TIMESTAMP branch in `createEventFilter()`:

```java
StartupOptions startupOptions = statefulTaskContext.getSourceConfig().getStartupOptions();
if (startupOptions.startupMode.equals(StartupMode.TIMESTAMP)) {
    BinlogOffset startingOffset = currentBinlogSplit.getStartingOffset();
    if (startingOffset == null) {
        throw new NullPointerException(
                "The startup option was set to TIMESTAMP "
                        + "but unable to find starting binlog offset. Please check if the timestamp is specified in "
                        + "configuration. ");
    }
    if (startingOffset.getOffsetKind() != BinlogOffsetKind.TIMESTAMP) {
        return event -> true;
    }

    long startTimestampSec = startingOffset.getTimestampSec();
    ...
}

Comment (danzhewuju) [reply flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java]: You're right, this can indeed be made smaller. I'll limit the scope of changes to the TIMESTAMP branch in createEventFilter(), and use the restored split's getStartingOffset() method instead of the source configuration, which also avoids that edge case. I've already pushed the latest version~ Thanks!

File: flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReaderTest.java [hunk 2]
```diff
@@ -1131,15 +1131,16 @@ void testRestoreFromCheckpointWithTimestampStartingOffset() throws Exception {
...
+        // based on the original TIMESTAMP startup setting.
         EventHeaderV4 header = new EventHeaderV4();
         header.setEventType(EventType.WRITE_ROWS);
         header.setTimestamp(1L);
         Event event = new Event(header, new WriteRowsEventData());
 
-        // Check if the filter works
+        // Check that the restored reader keeps the event.
         Predicate<Event> eventFilter = binlogReader.getBinlogSplitReadTask().getEventFilter();
-        assertThat(eventFilter.test(event)).isFalse();
```
lvyanquan, line 1142: Could we add one more test for the early checkpoint/savepoint case?

Scenario:
1. The original job starts with TIMESTAMP mode and timestamp T1.
2. A checkpoint/savepoint is taken before the binlog split processes any record, so the restored split starting offset is still `BinlogOffsetKind.TIMESTAMP` with timestamp T1.
3. The job is submitted again from that state, but the source config now contains a different TIMESTAMP startup option, e.g. T2 > T1.
4. The event filter should use the restored split timestamp T1, not the newly submitted config timestamp T2.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
```diff
@@ -391,21 +392,26 @@ private void configureFilter() {
...
+            // TIMESTAMP, otherwise valid row events after recovery would be incorrectly dropped.
+            BinlogOffset startingOffset = currentBinlogSplit.getStartingOffset();
+            if (startingOffset == null) {
                 throw new NullPointerException(
                         "The startup option was set to TIMESTAMP "
                                 + "but unable to find starting binlog offset. Please check if the timestamp is specified in "
                                 + "configuration. ");
             }
-            long startTimestampSec = startupOptions.binlogOffset.getTimestampSec();
+            if (startingOffset.getOffsetKind() != BinlogOffsetKind.TIMESTAMP) {
```
lvyanquan, line 410: minor: Add one more log here.

lvyanquan (commented):
The production code looks good to me now.

Left some minor comments.

lvyanquan:
CC @ruanhang1993.
Because this change touches the MySQL connector, which is one of the most commonly used connectors, could you also help review it?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/test/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReaderTest.java
danzhewuju: Thanks, I've added a unit test for this scenario.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/debezium/reader/BinlogSplitReader.java
danzhewuju: done

lvyanquan:

danzhewuju:
> Run 'mvn spotless:apply' to fix these violations.

Sorry. Done~

danzhewuju:
@lvyanquan Hello, thank you very much for your review. Are there any other problems with this PR? If there are still issues, please let me know, and I will continue to fix them.

danzhewuju:
@ruanhang1993 Can you help review this PR when you are free? Thanks!

PR #4299, opened 2026-03-04
[FLINK-38828][cdc-common][cdc-runtime] Handle schema evolution upon projection updates via SchemaCoordinator

VinaySagarGonabavi:
This is a v2 approach following feedback on PR #4280 — schema evolution responsibility stays in `SchemaCoordinator` rather than adding state to `PostTransformOperator`, keeping the transform operator stateless as intended by PR #4056.

When a Flink CDC pipeline uses transform projections and the pipeline is restarted from a savepoint/checkpoint with an updated projection, the downstream sink never receives schema change events for the new/removed columns, leaving the sink schema stale and out of sync with the actual projected data.

This PR fixes the issue by teaching SchemaCoordinator to detect schema differences when a `CreateTableEvent` arrives for an already-known table. Instead of treating it as redundant, the coordinator computes the schema diff and emits appropriate `AddColumnEvent`/`DropColumnEvent`/`AlterColumnTypeEvent` downstream. `PostTransformOperator` remains stateless - it's a pure mapping node.

## Brief change log

- `SchemaUtils.isSchemaChangeEventRedundant()`: CreateTableEvent is now redundant only if the table exists AND the schema is identical (was: just existence check)
- `SchemaCoordinator.deduceEvolvedSchemaChanges()`: When a CreateTableEvent arrives for an existing table in 1:1 routing, computes schema diff via SchemaMergingUtils.getSchemaDifference() and returns diff events instead of forwarding the raw CreateTableEvent

**Note**: `isSchemaChangeEventRedundant` is also called by Hudi and Paimon sink connectors. The behavioral change is a bugfix for those paths as well (they were silently dropping CreateTableEvents with changed schemas), but no connector-specific code was modified.

## Verifying this change
3 new test methods across 3 files:

- `SchemaUtilsTest`: Redundancy checks for CreateTableEvent - non-existent table, same schema, extra columns, type diff, fewer columns
- `SchemaMergingUtilsTest`: Diff scenarios for projection changes - add column, drop column, type change, swap columns, identical schemas
- `SchemaEvolveTest`: End-to-end through SchemaOperator + SchemaCoordinator - add column, drop column, type change, identical schema (no-op)

- Dependencies: no
- The public API: no
- The runtime per-record code path: no (only affects schema change event processing)
- Anything that affects deployment or recovery: yes (changes how CreateTableEvent is handled after checkpoint restore)

## Documentation

- Does this pull request introduce a new feature? No (fixes existing behavior gap)

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
@@ -288,12 +288,29 @@ public class SchemaCoordinator extends SchemaRegistry {
 
             List<SchemaChangeEvent> rawSchemaChangeEvents = new ArrayList<>();
             if (upstreamDependencies.size() == 1) {
-                // If it's a one-by-one routing rule, we can simply forward it to downstream sink.
-                SchemaChangeEvent rawEvent = event.copy(evolvedTableId);
-                rawSchemaChangeEvents.add(rawEvent);
-                LOG.info(
-                        "Step 3.3 - It's an one-by-one routing and could be forwarded as {}.",
-                        rawEvent);
+                // If it's a one-by-one routing rule, we can simply forward it to downstream
+                // sink. However, if the incoming event is a CreateTableEvent for an
+                // already-known evolved table (e.g. after restart with changed projections),
+                // we must compute the schema diff instead of forwarding the raw
+                // CreateTableEvent, which would fail at the sink.
+                if (event instanceof CreateTableEvent && currentEvolvedSchema != null) {
+                    CreateTableEvent createTableEvent = (CreateTableEvent) event;
+                    List<SchemaChangeEvent> diffEvents =
+                            SchemaMergingUtils.getSchemaDifference(
+                                    evolvedTableId,
+                                    currentEvolvedSchema,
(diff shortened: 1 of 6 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
```diff
@@ -288,12 +288,29 @@ private List<SchemaChangeEvent> deduceEvolvedSchemaChanges(SchemaChangeEvent eve
...
+                    LOG.info(
+                            "Step 3.3 - It's a one-by-one routing but CreateTableEvent for existing table. Computed diff events: {}.",
+                            diffEvents);
+                } else {
+                    SchemaChangeEvent rawEvent = event.copy(evolvedTableId);
+                    rawSchemaChangeEvents.add(rawEvent);
+                    LOG.info(
+                            "Step 3.3 - It's an one-by-one routing and could be forwarded as {}.",
+                            rawEvent);
+                }
```
yuxiqian, line 313: Can you add E2e test cases in org.apache.flink.cdc.pipeline.tests.migration to cover these scenarios?

VinaySagarGonabavi: Thank you! I've gone ahead and added test cases for the scenarios

PR #4302, opened 2026-03-05
[FLINK-39200][mysql] Skip insert/update/delete binlog deserialization of unsubscribed tables in MySql CDC binary log client.

chengcongchina:
This closes [FLINK-39200](https://issues.apache.org/jira/browse/FLINK-39200).

Currently, the MySQL row event deserializers in Flink CDC always fully deserialize WRITE/UPDATE/DELETE binlog events for all tables, regardless of whether the tables are actually subscribed by the connector. This leads to unnecessary CPU and memory overhead, especially when only a small portion of tables are monitored.

This PR implements early skipping of deserialization for unsubscribed tables in binary log client to improve performance. This optimization is applied by default. **No table option is added to allow turning off this improvement, since there is no valid use case to deserialize events for unsubscribed tables.**

### Brief change log

- Introduce `TableIdFilter` to support filtering tables during event deserialization.
- Add optimized event data deserializers (`WriteRowsEventDataDeserializer`, `UpdateRowsEventDataDeserializer`, `DeleteRowsEventDataDeserializer`) that support skipping deserialization for excluded tables.
- Update `MySqlStreamingChangeEventSource` to use the `TableIdFilter` based on the connector's table subscription configuration.

### Verifying this change

~~This change is verified by existing tests:~~
~~- `MySqlPipelineITCase`~~
~~- `BinlogSplitReaderTest`~~

**The second commit add unit tests.**

**Performance benchmarks can be added later to quantify the CPU/memory savings, which could be discussed in [FLINK-37743](https://issues.apache.org/jira/browse/FLINK-37743)**

### Does this pull request potentially affect one of the following parts:

- The public API, i.e., is any changed class annotated with @Public(Evolving): no
- The serializers: no
- The runtime per-record code paths (performance sensitive): yes

### Documentation

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/com/github/shyiko/mysql/binlog/event/deserialization/DeleteRowsEventDataDeserializer.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/com/github/shyiko/mysql/binlog/event/deserialization/DeleteRowsEventDataDeserializer.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.shyiko.mysql.binlog.event.deserialization;
+
+import com.github.shyiko.mysql.binlog.event.DeleteRowsEventData;
+import com.github.shyiko.mysql.binlog.event.TableMapEventData;
+import com.github.shyiko.mysql.binlog.io.ByteArrayInputStream;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.BitSet;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Copied from mysql-binlog-connector 0.27.2 to add a {@link TableIdFilter}.
+ *
+ * <p>Line 65-70: Use a {@link TableIdFilter} to skip the binlog deserialization of unwanted tables.
+ */
+public class DeleteRowsEventDataDeserializer
+        extends AbstractRowsEventDataDeserializer<DeleteRowsEventData> {
+
+    private boolean mayContainExtraInformation;
+
+    /** the table id filter to skip further deserialization of unsubscribed table ids. */
+    private TableIdFilter tableIdFilter = TableIdFilter.all();
+
+    public DeleteRowsEventDataDeserializer(Map<Long, TableMapEventData> tableMapEventByTableId) {
+        super(tableMapEventByTableId);
+    }
+
+    public DeleteRowsEventDataDeserializer setMayContainExtraInformation(
+            boolean mayContainExtraInformation) {
+        this.mayContainExtraInformation = mayContainExtraInformation;
+        return this;
+    }
+
+    public DeleteRowsEventDataDeserializer setTableIdFilter(TableIdFilter tableIdFilter) {
+        this.tableIdFilter = tableIdFilter;
+        return this;
+    }
+
+    @Override
+    public DeleteRowsEventData deserialize(ByteArrayInputStream inputStream) throws IOException {
+        DeleteRowsEventData eventData = new DeleteRowsEventData();
+        eventData.setTableId(inputStream.readLong(6));
+
+        // skip further deserialization if the table id is unsubscribed
+        if (!tableIdFilter.test(eventData.getTableId())) {
+            eventData.setIncludedColumns(null);
+            eventData.setRows(Collections.emptyList());
+            return eventData;
+        }
+
+        inputStream.readInteger(2); // reserved
+        if (mayContainExtraInformation) {
+            int extraInfoLength = inputStream.readInteger(2);
+            inputStream.skip(extraInfoLength - 2);
+        }
+        int numberOfColumns = inputStream.readPackedInteger();
+        eventData.setIncludedColumns(inputStream.readBitSet(numberOfColumns, true));
+        eventData.setRows(
+                deserializeRows(
+                        eventData.getTableId(), eventData.getIncludedColumns(), inputStream));
+        return eventData;
+    }
+
+    private List<Serializable[]> deserializeRows(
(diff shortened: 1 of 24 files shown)


ThorneANN:
Thank u for your contributor and i left some comment.
Why  not filter tableId instand of event actions?

chengcongchina:
@ThorneANN Thank you for your comment.
In the MySQL binlog protocol, apart from events like GTID events, there are mainly two types of events related to table names: DDL changes and row data changes (INSERT/UPDATE/DELETE).
Currently, DDL changes are captured for all tables, partly because downstream DDL capture for online DDL tools may require DDL parsing for shadow tables. Aside from that, the main deserialization workload in the binlog client lies in parsing row data changes. Therefore, adding the table ID filtering logic specifically within the deserialization of these row change events is sufficient to achieve the optimization.

ThorneANN:
> @ThorneANN Thank you for your comment. In the MySQL binlog protocol, apart from events like GTID events, there are mainly two types of events related to table names: DDL changes and row data changes (INSERT/UPDATE/DELETE). Currently, DDL changes are captured for all tables, partly because downstream DDL capture for online DDL tools may require DDL parsing for shadow tables. Aside from that, the main deserialization workload in the binlog client lies in parsing row data changes. Therefore, adding the table ID filtering logic specifically within the deserialization of these row change events is sufficient to achieve the optimization.

Yes, the point that confuses me is that I noticed you added a lot of event serialization classes because I think you only need to filter tableids

ThorneANN:
May be we modify `TableMapEventDataDeserializer` and `MySqlStreamingChangeEventSource` 's classes  to achieve this feature
﻿?

ThorneANN:
```
 eventDeserializer.setEventDataDeserializer(
                EventType.TABLE_MAP,
                new com.github.shyiko.mysql.binlog.event.deserialization
                        .TableMapEventDataDeserializer(
                        connectorConfig.getTableFilters().dataCollectionFilter()::isIncluded));
```

As the code show , flink-cdc also skip the data without capture table ,but always deserializer dataEvent's  mata

ThorneANN:
This pr i edit for last month ,u can cc .https://github.com/ThorneANN/flink-cdc/tree/serrizer-only-tablels

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/com/github/shyiko/mysql/binlog/event/deserialization/DeleteRowsEventDataDeserializer.java
```diff
@@ -0,0 +1,94 @@
...
+        super(tableMapEventByTableId);
+    }
+
+    public DeleteRowsEventDataDeserializer setMayContainExtraInformation(
+            boolean mayContainExtraInformation) {
+        this.mayContainExtraInformation = mayContainExtraInformation;
+        return this;
+    }
+
+    public DeleteRowsEventDataDeserializer setTableIdFilter(TableIdFilter tableIdFilter) {
```
lvyanquan: Because there will be no further operations to modify TableIdFilter in the future, I tend to remove this method and pass TableIdFilter by providing additional constructors.

lvyanquan (dismissed):

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/com/github/shyiko/mysql/binlog/event/deserialization/DeleteRowsEventDataDeserializer.java
chengcongchina: Resolved.

chengcongchina:
> This pr i edit for last month ,u can cc .https://github.com/ThorneANN/flink-cdc/tree/serrizer-only-tablels

During binlog parsing, filtering out the parsing of fields for irrelevant tables in TableMapEventMetadata is indeed a great optimization opportunity. However, doing only that is not enough: if we don’t also apply filtering during the subsequent parsing of insert/update/delete row change events, parsing can still fail because the metadata column types are actually needed there. Therefore, we can first consider merging the parsing optimization in the current PR, and then continue optimizing TableMapEventMetadata afterward.

chengcongchina:
After discussing with @lvyanquan , we agreed not to enable this optimization feature too aggressively right away. Instead, we will introduce a configuration option that is disabled by default, and turn it on in a future release after validating it through real-world usage.

ThorneANN:
> After discussing with @lvyanquan , we agreed not to enable this optimization feature too aggressively right away. Instead, we will introduce a configuration option that is disabled by default, and turn it on in a future release after validating it through real-world usage.

May be and i agree with this point

chengcongchina:
I’ve added a configuration option to keep this optimization disabled by default. Users can choose to enable it manually, and I’ve also updated the corresponding documentation. Could @lvyanquan @ThorneANN please take another look?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
```diff
@@ -1350,6 +1363,15 @@ protected void initSSLContext(SSLContext sc) throws GeneralSecurityException {
         return null;
     }
 
+    private TableIdFilter getTableIdDeserializationFilter() {
+        return tableId -> {
+            // since only subscribed table is recording schema, the result could be null
+            TableId table = taskContext.getSchema().getTableId(tableId);
+            return table != null
+                    && connectorConfig.getTableFilters().dataCollectionFilter().isIncluded(table);
```
lvyanquan, line 1371: I'm wondering if this filter is necessary. The validation should have already been done in [MySqlStreamingChangeEventSource#handleUpdateTableMetadata](https://github.com/apache/flink-cdc/blob/ec244f195b2f7f7f7dc63da4514780b015df0424/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java#L726).

chengcongchina: @lvyanquan Good point. With the current Debezium MySqlDatabaseSchema behavior, `taskContext.getSchema().getTableId(tableNumber)` returns non-null only when the
  table mapping was assigned for a known table schema, so `table != null` is effectively enough to exclude unsubscribed tables today.
That said, I’m keeping the additional `connectorConfig.getTableFilters().dataCollectionFilter().isIncluded(table)` check intentionally as a defensive guard. The `getTableId(tableNumber)` API itself doesn’t formally guarantee that null always means “unsubscribed/unmatched”, it’s just how the current implementation behaves (it depends on how schema/history is populated). If schema/history handling changes in the future (e.g., more tables’ schemas are present/loaded than the subscribed set), getTableId() could become non-null for tables we still don’t want to deserialize. The explicit `isIncluded(table)` keeps the intent (“only deserialize captured tables”) robust against such evolution.

tub:
Any thoughts on merging this? For our setup this would really help as we're only streaming a subset of a very large binlog.

yuxiqian:
Due to inactivity, I'll removing this ticket from the 3.7 release plan. Feel free to keep working on this PR later.

PR #4323, opened 2026-03-18
[FLINK-39265][postgres] PostgreSQL CDC intermittently drops INSERT records after checkpoint recovery due to WalPositionLocator filtering

JNSimba:
https://issues.apache.org/jira/browse/FLINK-39265

Fix data loss in PostgreSQL CDC source where INSERT/UPDATE/DELETE records                                                             
  are silently dropped after checkpoint recovery.                                                                                       
                                                                                                                                        
  ## Root Cause   
                                                                                                                                        
  When recovering from a checkpoint, the stored offset contains `lsn_proc`                                                              
  and `lsn_commit` (both equal to the last commit LSN `Y`). These are used
  to construct `WalPositionLocator(lastCommitLsn=Y, lastEventStoredLsn=Y)`.                                                             
                                                                                                                                        
  In pgoutput non-streaming mode, BEGIN and DML messages of the first new                                                               
  transaction after `Y` share the same `data_start=Y` as the previous                                                                   
  COMMIT. The find phase puts `Y` into `lsnSeen` but sets                                                                               
  `startStreamingLsn` to the subsequent COMMIT LSN `Z`. In the actual                                                                   
  streaming phase, DML messages with `data_start=Y` satisfy                                                                             
  `Y ∈ lsnSeen && Y ≠ Z` and are permanently filtered out.                                                                              
                                                                                                                                        
  ## Fix                                                                                                                                
                                                                                                                                        
  Remove `lsn_proc` and `lsn_commit` from `PostgresOffset` when                                                                         
  constructed from a Debezium source offset. With `lastCommitStoredLsn=null`,
  `WalPositionLocator` immediately sets `startStreamingLsn` to the first                                                                
  received LSN and switch-off happens before any DML is filtered.                                                                       
                                                                                                                                        
  ## Changes                                                                                                                            
                                                                                                                                        
  - `PostgresOffset.java`: filter out `lsn_proc` and `lsn_commit` in the                                                                
    `Map` constructor

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/offset/PostgresOffset.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/org/apache/flink/cdc/connectors/postgres/source/offset/PostgresOffset.java
@@ -45,27 +45,29 @@ public class PostgresOffset extends Offset {
     // used by PostgresOffsetFactory
     PostgresOffset(Map<String, String> offset) {
         Map<String, String> filtered = new HashMap<>(offset);
-        // When lsn == lsn_proc, WalPositionLocator is constructed with
-        // (lastCommitStoredLsn=Y, lastEventStoredLsn=Y). In pgoutput non-streaming mode,
-        // BEGIN and DML of the first new transaction after recovery share the same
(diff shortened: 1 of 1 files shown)


JNSimba:
Debezium has already fixed this issue; simply removing these two will cause other problems.
https://github.com/debezium/debezium/commit/3b5740f1a836c8b438888f2458ebb1554320bac7

However, he introduced a `lastProcessedMessageType`, which may involve several files, resulting in too many changes to Debeizum and hindering future upgrades.

PR #4358, opened 2026-04-01
[FLINK-39372][connect/mysql] Fix comparison logic for binlog filename with various digits

yuxiqian:
This closes FLINK-39372.

MySQL CDC uses Java's String.compareTo() for comparing Binlog filenames. This works correctly for files like mysql.000001 through mysql.999999 since they share the same length.

However, when the sequence rolls over to mysql.1000000, string comparison breaks — it treats '1' < '9' and incorrectly considers mysql.1000000 smaller than mysql.999999.


lvyanquan (approved):
LGTM, Thank you for identifying and fixing this issue.

PR #4359, opened 2026-04-02
[FLINK-39391][cdc-connector] Propagate scan.snapshot.fetch.size to Debezium properties in Oracle, SqlServer, DB2, and Postgres connectors

paulo-t:
This closes [FLINK-39391](https://issues.apache.org/jira/browse/FLINK-39391).

The user-configured `scan.snapshot.fetch.size` option has no effect on the actual JDBC `fetchSize` used during the incremental snapshot phase for Oracle, SqlServer, DB2, and Postgres connectors.

The value is correctly parsed and stored in `JdbcSourceConfig.fetchSize`, but is never written into the Debezium `Properties` object that the snapshot execution path actually reads.

For Oracle/SqlServer/DB2, the snapshot task calls `connectorConfig.getQueryFetchSize()`, which reads Debezium's `query.fetch.size` — defaulting to `0`. This causes the JDBC driver to fall back to its own small default (e.g., Oracle defaults to `fetchSize=10`). On high-latency networks this results in order-of-magnitude performance degradation: ~440 rows/s instead of ~3,500 rows/s with 4 parallel readers on ~31ms RTT.

For Postgres, the snapshot task reads `snapshot.fetch.size` which defaults to `2000`, so the impact is less severe but the user-configured value is still silently ignored.

MySQL is NOT affected — `MySqlSourceConfigFactory` already correctly sets `props.setProperty("database.fetchSize", ...)`.

The fix adds a `props.setProperty()` call in each affected `SourceConfigFactory.create()` before the `dbzProperties.putAll()` block, so that explicit user overrides via `debezium.query.fetch.size` / `debezium.snapshot.fetch.size` still take precedence.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-db2-cdc/src/main/java/org/apache/flink/cdc/connectors/db2/source/config/Db2SourceConfigFactory.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-db2-cdc/src/main/java/org/apache/flink/cdc/connectors/db2/source/config/Db2SourceConfigFactory.java
@@ -75,6 +75,8 @@ public class Db2SourceConfigFactory extends JdbcSourceConfigFactory {
                 throw new UnsupportedOperationException();
         }
 
+        props.setProperty("query.fetch.size", String.valueOf(fetchSize));
+
         if (dbzProperties != null) {
             props.putAll(dbzProperties);
         }
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-db2-cdc/src/main/java/org/apache/flink/cdc/connectors/db2/source/config/Db2SourceConfigFactory.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-db2-cdc/src/main/java/org/apache/flink/cdc/connectors/db2/source/config/Db2SourceConfigFactory.java
@@ -75,7 +75,7 @@ public class Db2SourceConfigFactory extends JdbcSourceConfigFactory {
                 throw new UnsupportedOperationException();
         }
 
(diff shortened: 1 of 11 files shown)


ThorneANN:
Thank you for your contribution and i see that [snapshot.fetch.size](https://debezium.io/documentation/reference/1.9/connectors/mysql.html#mysql-property-snapshot-fetch-size) instand of `query.fecth.size` ，may be  the `query.fecth.size`config also can the same effect ?

paulo-t:
Thanks for pointing this out.

I checked the latest upstream master before updating the PR. The issue has not been fixed there yet: Oracle, SQLServer, and DB2 scan fetch tasks still read `connectorConfig.getQueryFetchSize()`, while Postgres reads `connectorConfig.getSnapshotFetchSize()`.

You are right that `scan.snapshot.fetch.size` should map to Debezium snapshot fetch-size semantics. Mapping it to `query.fetch.size` would make the current Oracle/SQLServer/DB2 code path effective, but it is not semantically correct.

I have updated the PR accordingly:

- Oracle / SQLServer / DB2 scan fetch tasks now use `connectorConfig.getSnapshotFetchSize()`.
- Oracle / SQLServer / DB2 config factories now propagate Flink CDC `scan.snapshot.fetch.size` to Debezium `snapshot.fetch.size`.
- The related unit tests now assert `snapshot.fetch.size` and `getSnapshotFetchSize()`.
- Postgres already used `snapshot.fetch.size` / `getSnapshotFetchSize()`, so it remains consistent with the updated behavior.

I also ran the related factory/config tests locally and they passed.

ThorneANN:
@yuxiqian  cc

ThorneANN (approved):
LGTM

PR #4360, opened 2026-04-02
[FLINK-38450][iceberg] Fix duplicate records when schema change splits writes within a checkpoint

spoorthibasu:
**TL;DR:** Fix duplicate rows caused by same-snapshot equality-deletes by committing per-batch snapshots with increasing sequence numbers.

## Root Cause

When a schema-change event arrives mid-checkpoint, the writer flushes the affected table before applying the new schema.

This produces two batches within one checkpoint:
- batch 0: pre-schema-change data  
- batch 1: post-schema-change data (including equality-delete files targeting batch 0 rows)

Previously, all batches for a table were merged into a single Iceberg `RowDelta` and committed as one snapshot.

Iceberg equality-delete files only suppress data with strictly lower sequence numbers. When data files and equality-deletes are committed in the same snapshot, they share the same sequence number, so deletes are ineffective. As a result, both versions of a row remain visible, causing duplicates.

A secondary issue was that `flush(boolean)` rotated all table writers globally, unnecessarily splitting unrelated tables into multiple batches.

### Writer-side (scope reduction)

- `flush(boolean)` no longer rotates task writers globally and becomes a no-op for non-schema-change paths
- Schema-change events call `flushTableWriter(tableId)`, flushing only the affected table
- A per-table `batchIndex` increments on each flush and is propagated via `WriteResultWrapper`

### Committer-side (primary correctness fix)

- Each batch is committed as a separate Iceberg snapshot, giving later batches strictly higher sequence numbers
- Equality-delete files in batch N correctly supersede data files from batch M (M < N)
- Two snapshot summary properties are introduced:
  - `flink-cdc.batch-index`
  - `flink-cdc.checkpoint-id`
- On retry, the committer scans snapshot history, finds the highest committed batch index for the checkpoint, and resumes from the next uncommitted batch
- This ensures idempotent recovery after partial commits
- `MAX_COMMITTED_CHECKPOINT_ID` is written only on the final non-empty batch, preserving compatibility with Flink checkpoint semantics

## Why `getLastCommittedBatchIndex()` is safe

- Batches are committed in ascending order
- Scanning snapshot history from newest to oldest always encounters the highest committed batch first
- The scan stops when a snapshot with  
  `MAX_COMMITTED_CHECKPOINT_ID < checkpointId` is encountered  
  which marks the boundary of the previously completed checkpoint
- No intermediate batch snapshots for the current checkpoint can appear before this boundary

## Tests Added

- **testNoDuplicateWhenFlushSplitsSamePkUpdatesWithinCheckpoint**  
  Verifies `flush(false)` is a no-op and same-PK updates produce a single correct row

- **testNoDuplicateWhenSchemaChangeFlushSplitsSamePkUpdates**  
  Verifies schema-change split produces correct dedup via batch ordering

- **testRetryAfterPartialBatchCommit**  
  Verifies idempotent recovery when batch 0 is already committed and retry occurs

- **testNoDuplicateWithMultipleSchemaChangesInOneCheckpoint**  
  Verifies correctness across multiple sequential batches

- **testSchemaChangeFlushDoesNotAffectOtherTable**  
  Verifies schema-change flush is scoped to the affected table only

- **MySqlToIcebergE2eITCase** (end-to-end)
  Extended with same-key updates spanning a schema-change flush, an update followed by a delete, and a delete followed by a re-insert, validating no duplicate or stale rows in the Iceberg table

## Notes

- The common (non-schema-change) path remains unchanged: one checkpoint to one batch to one snapshot
- Additional snapshots occur only when a table is split into multiple batches within a checkpoint, currently triggered by schema changes
- Existing constructor paths default `batchIndex` to `0`, preserving compatibility with current usage

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/v2/IcebergCommitter.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/v2/IcebergCommitter.java
@@ -43,11 +43,11 @@ import org.slf4j.LoggerFactory;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.TreeMap;
 
 import static java.util.stream.Collectors.toList;
 import static org.apache.flink.runtime.checkpoint.CheckpointIDCounter.INITIAL_CHECKPOINT_ID;
@@ -110,17 +110,18 @@ public class IcebergCommitter implements Committer<WriteResultWrapper> {
         Map<TableId, List<WriteResultWrapper>> tableMap = new HashMap<>();
         for (WriteResultWrapper w : writeResultWrappers) {
             tableMap.computeIfAbsent(w.getTableId(), k -> new ArrayList<>()).add(w);
-            LOGGER.info(w.buildDescription());
         }
 
         for (Map.Entry<TableId, List<WriteResultWrapper>> entry : tableMap.entrySet()) {
             TableId tableId = entry.getKey();
 
-            // Sort ascending by batch index to guarantee correct Iceberg sequence number ordering.
-            // Equality-delete files in batch N will have sequence number > batch M (M < N), so
-            // they correctly supersede stale data written by earlier same-checkpoint batches.
-            List<WriteResultWrapper> batches = entry.getValue();
-            batches.sort(Comparator.comparingInt(WriteResultWrapper::getBatchIndex));
+            // Group by batchIndex so wrappers from different subtasks for the same batch
+            // are merged into one snapshot, not committed separately.
+            TreeMap<Integer, List<WriteResultWrapper>> batchGroups = new TreeMap<>();
+            for (WriteResultWrapper w : entry.getValue()) {
+                batchGroups.computeIfAbsent(w.getBatchIndex(), k -> new ArrayList<>()).add(w);
+                LOGGER.info(w.buildDescription());
+            }
 
             Table table =
                     catalog.loadTable(
@@ -150,31 +151,38 @@ public class IcebergCommitter implements Committer<WriteResultWrapper> {
             Optional<TableMetric> tableMetric = getTableMetric(tableId);
             tableMetric.ifPresent(TableMetric::increaseCommitTimes);
 
-            // Find the last non-empty batch so we know where to write MAX_COMMITTED_CHECKPOINT_ID.
-            int lastNonEmptyBatchPos = -1;
-            for (int i = batches.size() - 1; i >= startBatchIndex; i--) {
-                if (!isBatchEmpty(batches.get(i))) {
-                    lastNonEmptyBatchPos = i;
-                    break;
+            int lastNonEmptyBatchIndex = -1;
+            for (Map.Entry<Integer, List<WriteResultWrapper>> g : batchGroups.entrySet()) {
+                List<DataFile> df = collectDataFilesFromGroup(g.getValue());
+                List<DeleteFile> del = collectDeleteFilesFromGroup(g.getValue());
+                if (!df.isEmpty() || !del.isEmpty()) {
+                    lastNonEmptyBatchIndex = g.getKey();
                 }
             }
 
-            // Commit each batch as a separate Iceberg snapshot to get distinct sequence numbers.
-            for (int i = startBatchIndex; i < batches.size(); i++) {
-                WriteResultWrapper batch = batches.get(i);
-                List<DataFile> dataFiles = collectDataFiles(batch.getWriteResult());
-                List<DeleteFile> deleteFiles = collectDeleteFiles(batch.getWriteResult());
-
-                if (dataFiles.isEmpty() && deleteFiles.isEmpty()) {
+            // Commit each batch as a separate snapshot so sequence numbers increase per batch.
+            for (Map.Entry<Integer, List<WriteResultWrapper>> g : batchGroups.entrySet()) {
+                int batchIdx = g.getKey();
+                if (batchIdx < startBatchIndex) {
                     LOGGER.info(
-                            "Batch {} for checkpoint {} of table {} has nothing to commit, skipping",
-                            batch.getBatchIndex(),
+                            "Batch {} for checkpoint {} of table {} already committed, skipping",
+                            batchIdx,
                             checkpointId,
                             tableId.identifier());
                     continue;
                 }
 
-                boolean isLastNonEmptyBatch = (i == lastNonEmptyBatchPos);
+                List<DataFile> dataFiles = collectDataFilesFromGroup(g.getValue());
+                List<DeleteFile> deleteFiles = collectDeleteFilesFromGroup(g.getValue());
+
+                if (dataFiles.isEmpty() && deleteFiles.isEmpty()) {
+                    LOGGER.info(
+                            "Batch {} for checkpoint {} of table {} has nothing to commit, skipping",
+                            batchIdx,
+                            checkpointId,
+                            tableId.identifier());
+                    continue;
+                }
 
                 SnapshotUpdate<?> operation;
                 if (deleteFiles.isEmpty()) {
@@ -190,9 +198,9 @@ public class IcebergCommitter implements Committer<WriteResultWrapper> {
 
                 operation.set(SinkUtil.FLINK_JOB_ID, newFlinkJobId);
                 operation.set(SinkUtil.OPERATOR_ID, operatorId);
-                operation.set(FLINK_BATCH_INDEX, String.valueOf(batch.getBatchIndex()));
+                operation.set(FLINK_BATCH_INDEX, String.valueOf(batchIdx));
                 operation.set(FLINK_CHECKPOINT_ID_PROP, String.valueOf(checkpointId));
-                if (isLastNonEmptyBatch) {
+                if (batchIdx == lastNonEmptyBatchIndex) {
                     operation.set(
                             SinkUtil.MAX_COMMITTED_CHECKPOINT_ID, String.valueOf(checkpointId));
                 }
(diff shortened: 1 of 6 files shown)


fcfangcc:
@spoorthibasu 

  The current fix uses a per-writer batchIndex counter to track flush batches within a checkpoint. However, when the IcebergWriter runs   
  with parallelism > 1, each subtask maintains its own independent tableBatchIndexMap. This causes a critical issue when multiple subtasks
   receive the same SchemaChangeEvent for the same table.


Here is my implementation(there is no merge upstream, only refer to this part of logic)

[flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/v2/IcebergCommitter.java](https://github.com/fcfangcc/flink-cdc/blob/aa8b6fb48fb2d94c95f3fc9c348f055a66e34f09/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/v2/IcebergCommitter.java)

* The eventId of the same SchemaChangeEvent is the same on all subtasks
* Data from different subtasks but belonging to the same schema change will be merged into a snapshot
* LinkedHashMap guarantees to process according to the sequence of eventId insertion (that is, chronological order)

spoorthibasu:
Thanks for the detailed review, @fcfangcc, this was a very helpful catch.

You’re right there were two issues here:

1. Original bug:
Schema changes could split writes within a checkpoint, but all batches were committed into a single Iceberg snapshot with the same sequence number, so equality-deletes could not suppress earlier data.
Fixed by flushing per-table on schema change and committing each batch as a separate snapshot.

2. Parallelism:
	•	Writer: flushTableWriter could return early and cause batchIndex drift across subtasks. Fixed by advancing the counter before the early return.
	•	Committer: Wrappers were committed individually, so same batchIndex data from different subtasks could produce multiple snapshots. Fixed by grouping by batchIndex and merging into a single snapshot.

Retry safety:
Only the final non-empty batch writes flink-max-committed-checkpoint-id. Intermediate batches carry flink.batch-index and flink.checkpoint-id, so retries resume from the next uncommitted batch instead of skipping.

The changes are in this commit for reference:
https://github.com/apache/flink-cdc/pull/4360/changes/a6183525619eb2b48be4d251813dd4062498cb0e

I also looked at your implementation. Using explicit per-snapshot markers to track batch boundaries makes sense. One concern is retry behavior: if the committer crashes after batch 0 but before batch 1, the retry would see the checkpoint id property on the batch 0 snapshot and skip the entire checkpoint, leaving batch 1’s data uncommitted.

In this approach, only the final non-empty batch sets that property, so retries resume from the next uncommitted batch instead of skipping the checkpoint entirely.

Happy to discuss further, and I would appreciate your feedback on this approach.

spoorthibasu:
Hi @fcfangcc , just wanted to follow up on this thread when you get a chance.

Happy to clarify or adjust anything based on your feedback.

fcfangcc:
> Hi @fcfangcc , just wanted to follow up on this thread when you get a chance.
> 
> Happy to clarify or adjust anything based on your feedback.

😂 I am not a maintainer, but I also encountered the same problem to be solved by the main branch

spoorthibasu:
Thanks @fcfangcc for the clarification. Hi @lvyanquan, would appreciate your review on this PR when you get a chance. Happy to make any adjustments based on your feedback.

spoorthibasu:
Hi @lvyanquan , could you please re-trigger the 3 failing CI checks?  I looked into each one and none of them are related to this PR's changes (the PR only touches 4 files inside flink-cdc-pipeline-connector-iceberg). Here's what I found:
  - Source Unit Tests 2.x (oracle, sqlserver): column_type_test.sql has a comment about this exact race condition: "Cannot perform this operation while SQLServerAgent is starting, we simply wait for 3 seconds to improve test stability."
  - Pipeline E2E Tests (1-Parallelism) / OceanBase: The container has a 4-minute startup timeout (withStartupTimeout(Duration.ofMinutes(4))). It looks like it just missed the window on this run.
  - Pipeline E2E Tests (4-Parallelism) / MySqlToDoris: The pipeline E2E job compiles all modules before running tests. SchemaEvolveE2eITCase ran first and took 252 seconds, which left no time for MySqlToDorisE2eITCase before the 90-minute budget ran out.

16 of 19 checks passed, including all the Iceberg pipeline unit tests. Happy to provide more detail if helpful.

lvyanquan:
@spoorthibasu, Thank you for raising and fixing this issue. Since this code involves many edge cases (schema changes, job restarts), I need more time to review it. We will complete this PR in this version.

spoorthibasu:
Thank you @lvyanquan ! Happy to answer any questions or make changes based on your feedback.

spoorthibasu:
Hi @lvyanquan , just checking in. Happy to address any review feedback or answer questions.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/v2/WriteResultWrapper.java
```diff
@@ -40,17 +40,31 @@ public class WriteResultWrapper implements Serializable {
 
     private final String operatorId;
 
+    /** Batch index within the checkpoint for this table; increments on each schema-change flush. */
+    private final int batchIndex;
```
lvyanquan, line 44: `WriteResultWrapper` now includes a new `batchIndex` field, which changes the serialized content. I'd suggest bumping `WriteResultWrapperSerializer.VERSION` to 2 and handling version 1 compatibility in deserialize. 

While the current Java native serialization approach won't cause any functional issues (since `serialVersionUID` remains unchanged and Java serialization gracefully handles added fields with default values), bumping the serializer version is a more disciplined practice — it explicitly documents when the serialized format changed and makes future maintenance easier (e.g., if someone later replaces Java serialization with a custom binary format, they'll know which version introduced `batchIndex`).

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/v2/IcebergCommitter.java
```diff
@@ -62,6 +63,12 @@ public class IcebergCommitter implements Committer<WriteResultWrapper> {
 
     public static final String TABLE_GROUP_KEY = "table";
 
+    /** Snapshot summary key for the batch index; used to resume partial commits on retry. */
+    static final String FLINK_BATCH_INDEX = "flink.batch-index";
+
+    /** Snapshot summary key for the checkpoint ID on intermediate batch commits. */
+    static final String FLINK_CHECKPOINT_ID_PROP = "flink.checkpoint-id";
```
lvyanquan: The two new snapshot summary keys `flink.batch-index` and `flink.checkpoint-id` use the `flink.` prefix, which is the namespace reserved by the official Iceberg Flink Connector (e.g., `flink.job-id`, `flink.operator-id`, `flink.max-committed-checkpoint-id` in SinkUtil). These are custom properties introduced by this PR with no upstream precedent.

If the Iceberg community later introduces properties with the same names but different semantics, it would cause a conflict — especially `flink.checkpoint-id`, which is a very generic name that upstream might reasonably adopt in the future.

I'd suggest using a `flink-cdc.` prefix to clearly scope these as Flink CDC-specific properties

lvyanquan (commented):
Thanks for the fix! The overall approach — splitting schema-change-induced batches into separate snapshots to leverage Iceberg's sequence number semantics — looks sound to me. Left a few minor comments inline.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/v2/WriteResultWrapper.java
spoorthibasu: Done, bumped VERSION to 2 and deserialize now accepts both 1 and 2. They use the same byte layout, so old v1 state still reads back with batchIndex defaulting to 0.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-iceberg/src/main/java/org/apache/flink/cdc/connectors/iceberg/sink/v2/IcebergCommitter.java
spoorthibasu: Agreed, moved both keys under the `flink-cdc.` prefix (flink-cdc.batch-index, flink-cdc.checkpoint-id) so they stay out of the connector's `flink.` namespace.

spoorthibasu:
Thanks for the review @lvyanquan ! Pushed both fixes. Also, since I worked on this one, could you please assign FLINK-38450 to me on JIRA? My ASF JIRA username is spoorthibasu.

lvyanquan (approved):
+1.

Looking at `MySqlToIcebergE2eITCase`, the incremental phase interleaves UPDATE and DELETE events within the same checkpoint — both produce equality-delete files via RowDelta, potentially across multiple batches when a schema change triggers a writer flush. Before this fix, the committer had no mechanism to skip already-committed batches on retry, so a partial re-delivery would re-commit stale delete files and corrupt results. This likely explains the flaky failures we've seen. 

Feel free to add more test scenarios on this E2E test to further validate edge cases.

lvyanquan:
> could you please assign [FLINK-38450](https://issues.apache.org/jira/browse/FLINK-38450) to me on JIRA? My ASF JIRA username is spoorthibasu.

Done.

spoorthibasu:
Thanks so much for the review and approval @lvyanquan, and for assigning the ticket.

On the E2E edge cases: happy to add a few more interleaved UPDATE/DELETE plus schema-change scenarios. I can push them to this PR if you'd like, or keep this one as-is and do them as a follow-up, whichever you prefer.

Heads up on the two remaining red checks, both unrelated to this change (it only touches the Iceberg sink), and the same two suites pass on the 2.x runs:
- **Source Unit Tests / test (11, generic, oracle, sqlserver)**: Oracle passes; the SQL Server tests can't pull their Docker image (`mcr.microsoft.com/mssql/server:2019-latest` returns `Status 404, "The request is blocked"`), so `SqlServerParallelSourceExampleTest`, `SqlserverSourceReaderTest`, and `SqlServerScanFetchTaskTest` error out before running.
- **Source Unit Tests / test (11, generic, postgres-source, oceanbase-source, tidb, vitess)**: the flaky `OceanBaseFailoverITCase.testTaskManagerFailoverSingleParallelism[3]` (a data-comparison assertion in a failover test).

All the Iceberg pipeline unit and E2E suites pass. Could you please re-run these two when you get a chance? Thanks again!

lvyanquan:
> On the E2E edge cases: happy to add a few more interleaved UPDATE/DELETE plus schema-change scenarios. I can push them to this PR if you'd like, or keep this one as-is and do them as a follow-up, whichever you prefer.

It depends on your schedule. I'll observe the e2e stability for a period of time after this PR is merged. 
If there are no other objections, I'll merge this code in a few days.

spoorthibasu:
@lvyanquan I've added edge-case scenarios to `MySqlToIcebergE2eITCase`, verified them locally at parallelism 1 and 4, and updated the PR description to use the `flink-cdc.` property names. The scenarios cover same-key updates spanning a schema-change flush, an update followed by a delete, and a delete followed by a re-insert, each checking the Iceberg table ends up with no duplicate or stale rows.

spoorthibasu:
Could you please re-run the Pipeline E2E (1-Parallelism) job when you get a chance? The single failure is unrelated to this change: `MysqlToKafkaE2eITCase.testSyncWholeDatabaseWithDebeziumJson` hit `TimeoutException: failed to get specific event count: 20 from stdout: 13` (the Kafka consumer read only 13 of the 20 expected events before timing out), and the same test passed in the 4-Parallelism run. Thanks!

lvyanquan:
Merged. Thanks to @spoorthibasu  for the effort.

PR #4361, opened 2026-04-03
[FLINK-38729] Add flink-cdc-flink*-compat to output t jar for source connector.

sd4324530:
Add flink-cdc-flink*-compat to output t jar for source connector.


lvyanquan (approved):
+1

PR #4363, opened 2026-04-07
[FLINK-39373][pipeline][starrocks] Fix the error when modifying the default value of a column

linguoxuan:
### Purpose
Fix a bug in the StarRocks pipeline sink where executing AlterColumnTypeEvent would cause the existing default value of a column to be lost.

### Bug
When AlterColumnTypeEvent is applied to modify a column's data type in StarRocks, the event does not carry the column's original default value. As a result, the generated ALTER TABLE ... MODIFY COLUMN SQL statement omits the DEFAULT clause, causing StarRocks to throw an error. 
### Changes
Added getColumnDefaultValue() method in StarRocksEnrichedCatalog to query the current default value of a column from information_schema.COLUMNS.
Added getConnection() helper method to obtain a JDBC connection from the parent StarRocksCatalog via reflection.
Modified alterColumnType() to check and preserve the existing default value when the new column definition does not specify one.
Added integration test testAlterColumnTypePreservesDefaultValue to verify that the default value is retained after altering the column type.


linguoxuan:
@lvyanquan @rmetzger  PTAL

haruki-830 (approved):
+1.
The fix looks good to me. Querying information_schema.COLUMNS for the default is a reasonable approach here.

PR #2150, opened 2023-05-22
[mysql] Support tables which do not contain a primary key (#1827)

ruanhang1993:
This PR implements #1827.

PR #4367, opened 2026-04-08
[FLINK-39409][connect/postgres] Support tables without primary key for incremental snapshot

JNSimba:
## Summary
                                                                                                                                                  
  This PR adds support for PostgreSQL tables without primary keys in incremental snapshot mode,
  following the same approach as MySQL (PR #2150).                                                                                                
                                                                                                                                                  
  ### Changes
                                                                                                                                                  
  **JDBC Base Layer (flink-cdc-base):**                                                                                                           
  - `SplitKeyUtils.getSplitKey()`: Handle null record key by extracting chunk key from value's
    after/before struct instead of record key                                                                                                     
  - `IncrementalSourceScanFetcher.pollWithBuffer()`: Use after struct as buffer key when
    record key is null (no-PK table)                                                                                                              
  - `JdbcSourceFetchTaskContext.rewriteOutputBuffer()`: Support no-PK merge logic — for tables                                                    
    without primary key, use before/after struct as buffer key for CREATE/UPDATE/DELETE operations                                                
                                                                                                                                                  
  **PostgreSQL Connector (flink-connector-postgres-cdc):**                                                                                        
  - `PostgresDialect`: Validate that tables without primary key must have `REPLICA IDENTITY FULL`                                                 
    set, querying `pg_class.relreplident` and failing fast with a clear error message if not                                                      
                                                                                                                                                  
  **Tests:**
  - Added `products_no_pk` table DDL (without PK, with `REPLICA IDENTITY FULL`)                                                                   
  - `testNoPKTableWithChunkKey`: End-to-end test for no-PK table with `scan.incremental.snapshot.chunk.key-column` configured                     
  - `testNoPKTableWithoutChunkKey`: Verify `ValidationException` is thrown when chunk key column is not specified                                 
                                                                                                                                                  
  ### Design Decisions                                                                                                                            
                                                                                                                                                  
  - **No interface change**: Kept `Map<Struct, SourceRecord>` in `FetchTask.Context.rewriteOutputBuffer()`                                        
    unchanged to avoid breaking MongoDB and other connectors. For no-PK tables, the full row struct
    (all columns via REPLICA IDENTITY FULL) serves as the buffer key.                                                                             
  - **At-least-once for chunk key updates**: When an UPDATE changes the chunk key column value and                                                
    crosses split boundaries, only at-least-once semantics can be guaranteed (consistent with MySQL behavior).                                    
  - **REPLICA IDENTITY FULL required**: PostgreSQL requires this setting on no-PK tables to provide                                               
    full before image in WAL for UPDATE/DELETE events. The connector validates this at startup.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/JdbcSourceFetchTaskContext.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/JdbcSourceFetchTaskContext.java
@@ -49,7 +49,6 @@ import java.util.stream.Collectors;
 @Internal
 public abstract class JdbcSourceFetchTaskContext implements FetchTask.Context {
 
-
     protected final JdbcSourceConfig sourceConfig;
     protected final JdbcDataSourceDialect dataSourceDialect;
     protected CommonConnectorConfig dbzConnectorConfig;
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/JdbcSourceFetchTaskContext.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/JdbcSourceFetchTaskContext.java
@@ -38,6 +38,8 @@ import io.debezium.relational.TableId;
 import io.debezium.util.SchemaNameAdjuster;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.source.SourceRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.time.Instant;
 import java.util.Collection;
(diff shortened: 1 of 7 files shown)


JNSimba:
@lvyanquan  Could you help me trigger the ci? This MySqlToIcebergE2eITCase and UdfE2eITCase.testFlinkCompatibleScalarFunctions failure seems unrelated to the changes. Thanks

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/JdbcSourceFetchTaskContext.java
```diff
@@ -92,43 +92,71 @@ public Object[] getSplitKey(SourceRecord record) {
...
-                                    changeRecord.keySchema(),
-                                    changeRecord.key(),
-                                    changeRecord.valueSchema(),
-                                    envelope.read(after, source, fetchTs));
-                    outputBuffer.put(key, record);
+                    if (hasPrimaryKey) {
+                        outputBuffer.put(
+                                (Struct) changeRecord.key(), buildReadRecord(changeRecord));
+                    } else {
+                        // For no-PK table: remove the before image, insert the after image
```
Mrart: Is there something wrong with the log message?

Mrart (commented):
Can we add e2e test case to cover this change？

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/JdbcSourceFetchTaskContext.java
JNSimba: Thanks, I’ll clarify it to describe the buffer entry  replacement.

PR #4369, opened 2026-04-09
[hotfix] Fix the value of the document parameter `IsStable`.

sd4324530:
Fix the error message "This documentation is for an unreleased version of Apache Flink CDC. We recommend you use the latest stable version." on the website.
image


lvyanquan (approved):
+1.

PR #4374, opened 2026-04-10
[FLINK-39417] Fix GenericRecordData could not be [de]serialized in pipeline

yuxiqian:
This closes FLINK-39417.

Previously in FLINK-38886, we introduced a GenericRecordData as a replacement of BinaryRecordData so connector and internal operator developers could get rid of the heavy `flink-cdc-runtime` operator and construct binary-unaware records.

However, lacking corresponding serializers, it could not be used in complete pipelines but only in specific test cases.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/GenericRecordDataSerializer.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/GenericRecordDataSerializer.java
@@ -133,7 +133,7 @@ public class GenericRecordDataSerializer {
             target.writeDouble((Double) field);
         } else if (field instanceof StringData) {
             target.writeByte(TAG_STRING);
-            byte[] bytes = field.toString().getBytes(StandardCharsets.UTF_8);
+            byte[] bytes = ((StringData) field).toBytes();
             target.writeInt(bytes.length);
             target.write(bytes);
         } else if (field instanceof byte[]) {
@@ -219,7 +219,7 @@ public class GenericRecordDataSerializer {
                     int len = source.readInt();
                     byte[] bytes = new byte[len];
                     source.readFully(bytes);
-                    return BinaryStringData.fromString(new String(bytes, StandardCharsets.UTF_8));
+                    return BinaryStringData.fromBytes(bytes);
                 }
             case TAG_BINARY:
                 {
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/GenericRecordDataSerializer.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/GenericRecordDataSerializer.java
@@ -281,14 +281,6 @@ public class GenericRecordDataSerializer {
                 serializeField(element, target);
             }
         } else {
-            // For BinaryArrayData or other implementations, convert to object array via getters
-            int size = arrayData.size();
-            target.writeInt(size);
-            // We serialize each element as a generic object; since we don't know element types,
-            // we attempt to read them generically. For binary array data, the safest approach is
-            // to serialize the raw binary data.
-            // However, we don't have direct access to the underlying binary data in a generic way.
-            // So we fall back to reading elements as objects.
             throw new IOException(
                     "Serialization of non-generic ArrayData is not supported in GenericRecordDataSerializer. "
                             + "Actual type: "
@@ -312,18 +304,18 @@ public class GenericRecordDataSerializer {
         if (mapData instanceof GenericMapData) {
             ArrayData keyArray = mapData.keyArray();
             ArrayData valueArray = mapData.valueArray();
-            int size = mapData.size();
-            target.writeInt(size);
-            if (keyArray instanceof GenericArrayData && valueArray instanceof GenericArrayData) {
-                Object[] keys = ((GenericArrayData) keyArray).toObjectArray();
-                Object[] values = ((GenericArrayData) valueArray).toObjectArray();
-                for (int i = 0; i < size; i++) {
-                    serializeField(keys[i], target);
-                    serializeField(values[i], target);
-                }
-            } else {
+            if (!(keyArray instanceof GenericArrayData)
+                    || !(valueArray instanceof GenericArrayData)) {
                 throw new IOException(
                         "MapData with non-generic key/value arrays is not supported in GenericRecordDataSerializer.");
+            }
+            int size = mapData.size();
+            target.writeInt(size);
+            Object[] keys = ((GenericArrayData) keyArray).toObjectArray();
+            Object[] values = ((GenericArrayData) valueArray).toObjectArray();
+            for (int i = 0; i < size; i++) {
+                serializeField(keys[i], target);
+                serializeField(values[i], target);
             }
         } else {
             throw new IOException(
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/GenericRecordDataSerializer.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/GenericRecordDataSerializer.java
@@ -1,0 +1,417 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.runtime.serializer.data;
+
+import org.apache.flink.cdc.common.annotation.Internal;
+import org.apache.flink.cdc.common.data.ArrayData;
+import org.apache.flink.cdc.common.data.DateData;
+import org.apache.flink.cdc.common.data.DecimalData;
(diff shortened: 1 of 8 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/RecordDataSerializer.java
```diff
@@ -33,43 +34,86 @@ public class RecordDataSerializer extends TypeSerializerSingleton<RecordData> {
...
-        // BinaryRecordData is the only implementation of RecordData
-        return binarySerializer.deserialize(source);
+        byte type = source.readByte();
+        if (type == BINARY_RECORD_TYPE) {
+            return binarySerializer.deserialize(source);
+        } else if (type == GENERIC_RECORD_TYPE) {
+            return GenericRecordDataSerializer.deserialize(source);
+        } else {
+            throw new IOException("Unknown RecordData type tag: " + type);
+        }
```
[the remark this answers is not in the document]
yuxiqian: Should be fine, as `DataChangeEvent`s are not persisted in states. They're just [de]serialized for network transferring.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/writer/BinaryWriter.java
```diff
@@ -171,7 +176,19 @@ static void write(
...
+                    RowType rowType = (RowType) type;
+                    List<DataType> childTypes = rowType.getChildren();
+                    Object[] fields = new Object[recordData.getArity()];
+                    for (int i = 0; i < fields.length; i++) {
+                        fields[i] =
+                                RecordData.createFieldGetter(childTypes.get(i), i)
+                                        .getFieldOrNull(recordData);
+                    }
+                    recordData = new BinaryRecordDataGenerator(rowType).generate(fields);
+                }
```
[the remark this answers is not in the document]
yuxiqian: That shouldn't differ. Added assertions to ensure that claim.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/GenericRecordDataSerializer.java
```diff
@@ -0,0 +1,406 @@
...
+
+/**
+ * Serializer for {@link GenericRecordData}. Uses a self-describing format where each field is
+ * prefixed with a type tag, so no schema information is needed at serialization time.
+ */
+@Internal
+public class GenericRecordDataSerializer {
+
+    // Type tags for self-describing format
+    static final byte TAG_NULL = 0;
```
wenshao: **[Suggestion]** The TAG constants (`TAG_NULL`, `TAG_BOOLEAN`, etc.) use package-private visibility (`static final byte`). Since they are only consumed internally by `serializeField`/`deserializeField`, they should be `private static final byte` to match the project convention (e.g., `RecordDataSerializer.java` uses `private static final byte` for its type constants).

_— qwen3.6-plus via Qwen Code /review_

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/GenericRecordDataSerializer.java
```diff
@@ -0,0 +1,406 @@
...
+            ArrayData valueArray = mapData.valueArray();
+            if (keyArray instanceof GenericArrayData && valueArray instanceof GenericArrayData) {
+                Object[] keys = ((GenericArrayData) keyArray).toObjectArray();
+                Object[] values = ((GenericArrayData) valueArray).toObjectArray();
+                Map<Object, Object> newMap = new LinkedHashMap<>(keys.length);
+                for (int i = 0; i < keys.length; i++) {
+                    newMap.put(copyField(keys[i]), copyField(values[i]));
+                }
+                return new GenericMapData(newMap);
+            }
```
wenshao, line 408: **[Suggestion]** The `return field;` fallback here is dead code. `GenericMapData.keyArray()` always returns `new GenericArrayData(map.keySet().toArray())` and `valueArray()` always returns `new GenericArrayData(map.values().toArray())`, so the `instanceof GenericArrayData` check is always true. Consider replacing the fallback with `throw new IllegalStateException(...)` to make the invariant explicit and fail-fast if the assumption ever breaks.

_— qwen3.6-plus via Qwen Code /review_

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/writer/BinaryWriter.java
```diff
@@ -171,7 +177,25 @@ static void write(
                 writer.writeMap(pos, (MapData) o, (MapDataSerializer) serializer);
                 break;
             case ROW:
-                writer.writeRecord(pos, (RecordData) o, (TypeSerializer<RecordData>) serializer);
+                RecordData recordData = (RecordData) o;
```
wenshao, line 180: **[Suggestion]** The ROW branch creates `new Object[arity]` and `new BinaryRecordDataGenerator(rowType)` (which runs a stream pipeline via `Arrays.stream(...).map(InternalSerializers::create)`) on every non-`BinaryRecordData` row write. In high-throughput pipelines with frequent `GenericRecordData`, this adds significant GC pressure and CPU overhead. Consider caching the `BinaryRecordDataGenerator` at a higher level (e.g., per-table) or adding a direct `GenericRecordData` write path.

_— qwen3.6-plus via Qwen Code /review_

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/RecordDataSerializer.java
```diff
@@ -33,43 +34,86 @@ public class RecordDataSerializer extends TypeSerializerSingleton<RecordData> {
...
-        // BinaryRecordData is the only implementation of RecordData
-        return ((BinaryRecordData) from).copy((BinaryRecordData) reuse);
+        if (from instanceof BinaryRecordData) {
+            BinaryRecordData reuseRecord =
+                    (reuse instanceof BinaryRecordData)
+                            ? (BinaryRecordData) reuse
+                            : new BinaryRecordData(from.getArity());
+            return ((BinaryRecordData) from).copy(reuseRecord);
+        } else if (from instanceof GenericRecordData) {
+            return GenericRecordDataSerializer.copy((GenericRecordData) from);
```
wenshao: **[Suggestion]** The `copy(from, reuse)` method ignores the `reuse` parameter for `GenericRecordData`, always delegating to `GenericRecordDataSerializer.copy()` which allocates a new instance. This differs from the `BinaryRecordData` branch which actively reuses the provided object. In Flink's execution model, `copy(from, reuse)` is called in hot paths — consistently ignoring reuse for one branch creates unnecessary GC pressure when `GenericRecordData` flows through the pipeline.

_— qwen3.6-plus via Qwen Code /review_

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/serializer/data/RecordDataSerializerTest.java
```diff
@@ -49,7 +69,295 @@ protected RecordData[] getTestData() {
...
+            GenericRecordData.of(1L, BinaryStringData.fromString("test1")),
+            GenericRecordData.of(2L, BinaryStringData.fromString("test2")),
+            GenericRecordData.of(3L, null)
         };
     }
+
+    @Test
+    void testGenericRecordDataWithVariousTypes() throws Exception {
+        RecordDataSerializer serializer = RecordDataSerializer.INSTANCE;
+
```
wenshao, line 82: **[Suggestion]** `testGenericRecordDataWithVariousTypes` does not cover `DateData`, `TimeData`, or `ZonedTimestampData` field types, even though the serializer supports `TAG_DATE`, `TAG_TIME`, and `TAG_ZONED_TIMESTAMP`. The corresponding `testBinaryRecordDataWithVariousTypes` test includes all three. Consider adding these to ensure full type coverage for the `GenericRecordData` serialization path.

_— qwen3.6-plus via Qwen Code /review_

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/serializer/data/RecordDataSerializerTest.java
```diff
@@ -49,7 +69,295 @@ protected RecordData[] getTestData() {
...
+        assertThat(dIntArray.size()).isEqualTo(5);
+        assertThat(dIntArray.getInt(0)).isEqualTo(1);
+        assertThat(dIntArray.getInt(4)).isEqualTo(5);
+
+        ArrayData dStringArray = deserialized.getArray(3);
+        assertThat(dStringArray.size()).isEqualTo(2);
+        assertThat(dStringArray.getString(0).toString()).isEqualTo("a");
+        assertThat(dStringArray.getString(1).toString()).isEqualTo("b");
+
+        MapData dMap = deserialized.getMap(4);
```
wenshao, line 261: **[Suggestion]** `testGenericRecordDataWithNestedTypes` checks `dMap.size() == 2` but does not verify actual key-value content (e.g., `"k1"→100`, `"k2"→200`). If map key-value pairs were corrupted or mismatched during serialization/deserialization, this test would not catch it. Consider adding assertions like `assertThat(dMap.get(BinaryStringData.fromString("k1"))).isEqualTo(100)`.

_— qwen3.6-plus via Qwen Code /review_

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/writer/BinaryWriter.java
yuxiqian: I think it's OK as GenericRecordData are used merely for tests and PoC. And keeping `BinaryWriter` stateless seems more appropriate.

flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/serializer/data/RecordDataSerializerTest.java
yuxiqian: It's not possible as `MapData` interface does not provide `get` lookup method.

lvyanquan (approved):
+1.

PR #4377, opened 2026-04-15
[FLINK-39438][MaxCompute] Add support for sink operation types in MaxCompute options

dion-ricky:
## Problem
- Sink operation mode cannot be explicitly configured via pipeline YAML.
- Behavior is tightly coupled with MaxCompute table type.
- Schema evolution auto-creation enforces transactional tables → always upsert.
- No way to override this behavior for append use cases.
## Expected Behavior
- Users can explicitly define sink behavior (append or upsert) via configuration.
- Writer selection logic respects the configured option instead of table type.
## Proposal
Example YAML Configuration:
```yaml
sink:
  type: maxcompute
  name: maxcompute
  access-id: ${secret_values.maxcompute_access_id}
  access-key: ${secret_values.maxcompute_access_key}
  endpoint: ${secret_values.maxcompute_endpoint}
  project: maxcompute_project
  quota.name: res_grp
  sink.operation: append 
```

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-maxcompute/src/main/java/org/apache/flink/cdc/connectors/maxcompute/MaxComputeDataSinkOptions.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-maxcompute/src/main/java/org/apache/flink/cdc/connectors/maxcompute/MaxComputeDataSinkOptions.java
@@ -21,6 +21,7 @@ package org.apache.flink.cdc.connectors.maxcompute;
 import org.apache.flink.cdc.common.configuration.ConfigOption;
 import org.apache.flink.cdc.common.configuration.ConfigOptions;
 import org.apache.flink.cdc.connectors.maxcompute.options.CompressAlgorithm;
+import org.apache.flink.cdc.connectors.maxcompute.options.MaxComputeOptions;
 
 /** Options for MaxCompute Data Sink. */
 public class MaxComputeDataSinkOptions {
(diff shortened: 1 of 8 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-maxcompute/src/main/java/org/apache/flink/cdc/connectors/maxcompute/MaxComputeDataSinkOptions.java
```diff
@@ -107,4 +108,11 @@ public class MaxComputeDataSinkOptions {
                     .intType()
                     .defaultValue(4)
                     .withDescription("The number of concurrent with flush bucket data.");
+
+    public static final ConfigOption<MaxComputeOptions.SinkOperation> SINK_OPERATION =
```
lvyanquan, line 112: Please add documentation for the newly introduced sink.operation option. This should explain:
- **upsert (default)**: Requires primary keys in schema. Creates a transactional table and supports update/delete semantics.
- **append**: Creates a regular table regardless of primary keys. Only supports insert operations, suitable for append-only scenarios.

This helps users understand the configuration impact and choose the appropriate mode for their use case.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-maxcompute/src/main/java/org/apache/flink/cdc/connectors/maxcompute/options/MaxComputeOptions.java
```diff
@@ -140,8 +147,46 @@ public Builder withSchemaOperatorUid(String schemaOperatorUid) {
...
+        public String getValue() {
+            return value;
+        }
+
+        @Override
+        public String toString() {
+            return value;
+        }
+
+        public static SinkOperation fromValue(String value) {
```
lvyanquan, line 180: This method is unused and could be removed.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-maxcompute/src/main/java/org/apache/flink/cdc/connectors/maxcompute/options/MaxComputeOptions.java
```diff
@@ -140,8 +147,46 @@ public Builder withSchemaOperatorUid(String schemaOperatorUid) {
...
+        APPEND("append"),
+        UPSERT("upsert");
+
+        private final String value;
+
+        SinkOperation(String value) {
+            this.value = value;
+        }
+
+        public String getValue() {
```
lvyanquan, line 171: This method is unused and could be removed.

PR #4385, opened 2026-04-29
[FLINK-39573][cdc][docs] Fix Build documentation CI startup_failure by replacing third-party rsync action

yuanoOo:
## Problem

The `Build documentation` GitHub Actions workflow has been failing with `startup_failure` since April 18, 2026. All scheduled and PR-triggered documentation builds are broken, meaning updated documentation cannot be published to the Apache Nightlies website.

**Error message:**
```
The action burnett01/rsync-deployments@0dc935cdecc5f5e571865e60d2a6cdc673704823 is not allowed in apache/flink-cdc
because all actions must be from a repository owned by your enterprise, created by GitHub,
or match one of the patterns in the allowed list.
```

**CI failure history:** All runs since April 18 show `startup_failure` — [Build documentation workflow runs](https://github.com/apache/flink-cdc/actions/workflows/build_docs.yml)

## Root Cause

The workflow uses `burnett01/rsync-deployments@0dc935cd` which is pinned to **v5.2** (from 2022). This version is **not** in the [Apache Infra approved actions list](https://github.com/apache/infrastructure-actions/blob/main/actions.yml). Apache enforces an allowlist for third-party GitHub Actions, and only the following versions of `burnett01/rsync-deployments` are approved:

- `7659d600d8bdd035bb5cdfba1d4bd0dd4a307ca6` (v8.0.3, expires 2026-05-22)
- `dc0d5d44c4728aad3f02154a87309809e62a960f` (v8.0.4)

### Option 1: Upgrade to an approved version (v8.0.4)

Simply replace the SHA with the approved `dc0d5d44c4728aad3f02154a87309809e62a960f` (v8.0.4).

**Pros:** Minimal change, single-line fix.
**Cons:** Still depends on a third-party action. The approved versions have **expiration dates** (e.g., v8.0.3 expires 2026-05-22). When they expire, the workflow will break again with the same `startup_failure`. This was exactly the experience reported in [FLINK-38448](https://issues.apache.org/jira/browse/FLINK-38448) — upgrading to the latest version did not help at that time.

### Option 2: Replace with a local composite action (Chosen ✅)

Create a local composite action at `.github/actions/rsync-deployments/` that uses native `rsync` + `ssh-agent` (both pre-installed on `ubuntu-latest` runners), eliminating the dependency on any third-party action entirely.

**Pros:**
- No dependency on third-party actions — immune to future allowlist changes or version expirations
- No Docker overhead — the original action runs inside a Docker container, the composite action runs natively
- Same approach already proven in [FLINK-38448](https://issues.apache.org/jira/browse/FLINK-38448) for flink-kubernetes-operator ([commit 3e3cb584](https://github.com/apache/flink-kubernetes-operator/commit/3e3cb584))

**Cons:**
- Slightly more files to maintain
- After merging, the action files need to be backported to `release-3.5` and `release-3.6` branches (the workflow checks out these branches via matrix and requires the local action to be present)

## Changes

- **New:** `.github/actions/rsync-deployments/action.yml` — Local composite action that uses `ssh-agent` + `rsync` to deploy documentation, with the same interface as the original third-party action
- **New:** `.github/actions/rsync-deployments/LICENSE` — MIT License from the original [burnett01/rsync-deployments](https://github.com/Burnett01/rsync-deployments)
- **Modified:** `.github/workflows/build_docs.yml` — Replace `burnett01/rsync-deployments@0dc935cd` with `./.github/actions/rsync-deployments` in both upload steps

## Post-merge Steps

After this PR is merged to `master`, the `.github/actions/rsync-deployments/` directory must be backported to the following branches so that the matrix builds for those versions continue to work:

- `release-3.6`
- `release-3.5`

## References

- JIRA: [FLINK-39573](https://issues.apache.org/jira/browse/FLINK-39573)
- Similar issue: [FLINK-38448](https://issues.apache.org/jira/browse/FLINK-38448) (same fix in flink-kubernetes-operator)
- Apache Actions policy: https://infra.apache.org/github-actions-policy.html
- Apache approved actions list: https://github.com/apache/infrastructure-actions/blob/main/actions.yml

### DIFF
--- a/.github/actions/rsync-deployments/LICENSE
+++ b/.github/actions/rsync-deployments/LICENSE
@@ -1,0 +1,22 @@
+MIT License
+
+Copyright (c) 2019-2022 Contention
+Copyright (c) 2019-2026 Burnett01
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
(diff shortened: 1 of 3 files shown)


yuanoOo:
@lvyanquan PTAL

lvyanquan (approved):
LGTM

PR #4388, opened 2026-05-07
[FLINK-39621] Fix incremental-snapshot based sources may get stuck in binlog-backfill stage

yuxiqian:
This closes FLINK-39621.

As pointed out by @loserwang1024 , for the same reason in FLINK-39207, all incremental-snapshot-based sources are prone to this bug, causing the source operator to get stuck forever during streaming log backfilling.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceStreamFetcher.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceStreamFetcher.java
@@ -90,6 +90,7 @@ public class IncrementalSourceStreamFetcher implements Fetcher<SourceRecords, So
         configureFilter();
         taskContext.configure(currentStreamSplit);
         this.queue = taskContext.getQueue();
+        startReadTask();
         executorService.submit(
                 () -> {
                     try {
@@ -280,6 +281,10 @@ public class IncrementalSourceStreamFetcher implements Fetcher<SourceRecords, So
         this.finishedSplitsInfo = splitsInfoMap;
         this.maxSplitHighWatermarkMap = tableIdOffsetPositionMap;
         this.pureStreamPhaseTables.clear();
+    }
+
+    public void startReadTask() {
+        this.currentTaskRunning = true;
     }
 
     public void stopReadTask() throws Exception {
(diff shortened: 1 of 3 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceStreamFetcher.java
```diff
@@ -282,6 +283,10 @@ private void configureFilter() {
         this.pureStreamPhaseTables.clear();
     }
 
+    public void startReadTask() {
+        this.currentTaskRunning = true;
```
loserwang1024: Why org.apache.flink.cdc.connectors.postgres.source.fetch.StoppableChangeEventSourceContext is not changed? If StoppableChangeEventSourceContext#isRunning  become false, the second split won't be read.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceStreamFetcher.java
```diff
@@ -282,6 +283,10 @@ private void configureFilter() {
...
+    }
+
     public void stopReadTask() throws Exception {
         this.currentTaskRunning = false;
```
loserwang1024: I think a  `private volatile AtomicInteger counter = new AtomicInteger(0)` to count the running job will be better for the thread model.

Though currentTaskRunning is also work because executorService is SingleThreadExecutor, use counter will be better for later multiple thread pools.

WDYT? @leonardBang @yuxiqian

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceStreamFetcher.java
loserwang1024: StoppableChangeEventSourceContext in postgres is not shared.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceStreamFetcher.java
leonardBang: +1

loserwang1024 (approved):
LGTM

PR #4390, opened 2026-05-08
[FLINK-39633][postgres] Fix NPE in incremental snapshot backfill when WAL carries records for other captured tables

JNSimba:
This closes https://issues.apache.org/jira/browse/FLINK-39633.

## Problem

During the snapshot backfill phase of PostgreSQL CDC (incremental snapshot), the WAL stream from the logical replication slot may carry change records for captured tables other than the one whose chunk is currently being snapshotted. `IncrementalSourceScanFetcher#isChangeRecordInChunkRange` does not filter records by tableId before delegating to `JdbcSourceFetchTaskContext#isRecordBetween`, which calls `getDatabaseSchema().tableFor(record.tableId)`. If the foreign table's schema is not yet present in `RelationAwarePostgresSchema`'s lazy cache, the lookup returns null and `ChunkUtils.getSplitColumn` throws NPE on `null.primaryKeyColumns()`:

```
java.lang.NullPointerException: Cannot invoke "io.debezium.relational.Table.primaryKeyColumns()" because "table" is null
    at org.apache.flink.cdc.connectors.postgres.source.utils.ChunkUtils.getSplitColumn(ChunkUtils.java:45)
    at org.apache.flink.cdc.connectors.postgres.source.fetch.PostgresSourceFetchTaskContext.getSplitType(PostgresSourceFetchTaskContext.java:291)
    at org.apache.flink.cdc.connectors.base.source.reader.external.JdbcSourceFetchTaskContext.isRecordBetween(JdbcSourceFetchTaskContext.java:76)
    at org.apache.flink.cdc.connectors.base.source.reader.external.IncrementalSourceScanFetcher.isChangeRecordInChunkRange(IncrementalSourceScanFetcher.java:265)
```

The streaming reader (`IncrementalSourceStreamFetcher#shouldEmit`) already guards against this case via `finishedSplitsInfo.containsKey(tableId)` before invoking `isRecordBetween`. The scan fetcher was missing the symmetric guard.

## Fix

Filter change records by tableId in `IncrementalSourceScanFetcher#isChangeRecordInChunkRange` before delegating to `isRecordBetween`. Records whose tableId does not equal `currentSnapshotSplit.getTableId()` are skipped.

## Test plan

- [x] Added `IncrementalSourceScanFetcherTest` with three cases:
  - foreign-table change record is filtered out (no NPE, `isRecordBetween` is never invoked)
  - same-table change record is still delegated to `isRecordBetween`
  - non-data change records (watermark / signal) short-circuit without calling `getTableId`
- [x] All existing `flink-cdc-base` unit tests still pass (16/16)

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceScanFetcher.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceScanFetcher.java
@@ -269,14 +269,11 @@ public class IncrementalSourceScanFetcher implements Fetcher<SourceRecords, Sour
         // Skip records of other captured tables; their schema may not be loaded yet
         // and their PKs do not align with this chunk's bounds.
         TableId recordTableId = taskContext.getTableId(record);
-        if (recordTableId == null
-                || !recordTableId.equals(currentSnapshotSplit.getTableId())) {
+        if (recordTableId == null || !recordTableId.equals(currentSnapshotSplit.getTableId())) {
             return false;
         }
         return taskContext.isRecordBetween(
-                record,
-                currentSnapshotSplit.getSplitStart(),
-                currentSnapshotSplit.getSplitEnd());
+                record, currentSnapshotSplit.getSplitStart(), currentSnapshotSplit.getSplitEnd());
     }
 
     @VisibleForTesting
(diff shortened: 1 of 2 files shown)


yuxiqian:
@Mrart Would you like to take a look?

Mrart:
It seems ok, And Can add test case in postgres connector,for e2e test.

JNSimba:
> It seems ok, And Can add test case in postgres connector,for e2e test.

@Mrart It feels like the e2e test might not be reliably reproducible, doesn't it? After all, it requires ensuring that data for other tables is being written during the backfill phase.

Mrart:
> > It seems ok, And Can add test case in postgres connector,for e2e test.
> 
> @Mrart It feels like the e2e test might not be reliably reproducible, doesn't it? After all, it requires ensuring that data for other tables is being written during the backfill phase.

LGTM @yuxiqian

Mrart (approved):
LGTM

PR #4393, opened 2026-05-11
[FLINK-39610] Mongo CDC supports specifying individual SSL context

yuxiqian:
This closes FLINK-39610.

Currently, MongoDB CDC relies on connection option params (`tls=true`) to enable TLS verification, but not in an independent TLS context. Global JVM certificate file path config will apply, and there's no way to override that.

It is suggested to introduce new connector options, and pass them into `MongoClientSettings#applyToSslSettings`.


ruanhang1993 (commented):
@yuxiqian LGTM. Please add docs about the new options. The new options only can be used when `scan.incremental.snapshot.enabled=true`, this should be included in the docs.

ruanhang1993 (approved):

PR #4397, opened 2026-05-15
[FLINK-39682] Fix unable to invoke TO_DATE with TIMESTAMP types

yuxiqian:
This closes FLINK-39682.

The behavior of `TO_DATE` built-in function is silently changed after FLINK-38888. Previously, `TO_DATE(timestamp)` is equivalent to `TO_DATE(CAST(timestamp AS VARCHAR))` or `TO_DATE('2026-01-01 21:48:25')`, results in `2026-01-01`.

However, after rewriting `TO_DATE` strictly checks if input matches `yyyy-MM-dd` format, and returns `null` then.


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/metadata/TransformSqlOperatorTable.java
```diff
@@ -285,8 +285,10 @@ public SqlSyntax getSyntax() {
                             SqlTypeTransforms.FORCE_NULLABLE),
                     null,
                     OperandTypes.or(
+                            // Only "from string" mode supports specifying formatter.
                             OperandTypes.family(SqlTypeFamily.STRING),
-                            OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING)),
+                            OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING),
+                            OperandTypes.family(SqlTypeFamily.TIMESTAMP)),
```
lvyanquan, line 291: It would be better if you could update the related document to specify the supported parameter types.

lvyanquan (approved):
+1.
Leave a comment.

PR #4398, opened 2026-05-15
[FLINK-39684] Fix PaimonWriterHelper NullPointerException when handling REPLACE events

yuxiqian:
This closes FLINK-39684.

Pipeline YAML ReplaceEvent carries `after()` payload only, but PaimonWriterHelper attempts to access its `before()` field, causing the following exception:

```
java.lang.NullPointerException
	at org.apache.flink.cdc.connectors.paimon.sink.v2.PaimonWriterHelper.convertRecordDataToGenericRow(PaimonWriterHelper.java:307)
	at org.apache.flink.cdc.connectors.paimon.sink.v2.PaimonWriterHelper.convertEventToFullGenericRows(PaimonWriterHelper.java:257)
```


lvyanquan (approved):
LGTM.

PR #4403, opened 2026-05-20
[FLINK-39622] [postgres] Fix O(N²) JDBC metadata lookups in CustomPostgresSchema d…

ThorneANN:
、 CustomPostgresSchema#readTableSchema invokes jdbcConnection.readSchema with
  the full captured-table filter, so a single call already loads metadata for
  every captured table. However the cache-population loop only iterates the
  requested subset, discarding the rest. As a result, snapshot startup performs
  one full pg_catalog scan per split, scaling as O(N²) with the number of
  captured tables and causing severe latency on multi-tenant Postgres deployments
  that capture hundreds of tables across schemas.
  
  This change caches every table discovered by readSchema into schemasByTableId,
  while the returned tableChanges still contains only the originally-requested
  subset. Subsequent splits are served entirely from the cache.
  
  Also fixes a related issue where getTableSchema(List<TableId>) re-fetched
  already-cached tables by passing the full tableIds list to readTableSchema
  instead of the unmatched subset.


tomasmiguez:
Hi there,
is there any blocker to merging this?

We are using a custom patch to fix this and would love to be able to use official releases again.

Thanks!

ThorneANN:
@loserwang1024 PTAL,if u have free time.tks.

matiasbertani:
I would love to see this merged!!

PR #4404, opened 2026-05-20
[FLINK-39492][starrocks] Fix microsecond precision loss for TIMESTAMP types in StarRocks sink

ThorneANN:
TIMESTAMP_WITHOUT_TIME_ZONE and TIMESTAMP_WITH_LOCAL_TIME_ZONE fields were
  serialized using a hardcoded "yyyy-MM-dd HH:mm:ss" formatter, silently
  dropping sub-second precision regardless of the declared column precision.

  Add datetimeFormatter(int precision), mirroring the existing timeFormatter,
  which appends fractional seconds only when precision > 0. The formatter
  instances are cached per precision to avoid repeated allocation.

  For StarRocks < 3.1, the fractional part is truncated by the server on
  ingestion — no behavioral regression. For StarRocks >= 3.1 (which supports
  microsecond DATETIME storage), full precision is now preserved end-to-end.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
@@ -137,6 +137,31 @@ public class StarRocksUtils {
     private static final DateTimeFormatter DATETIME_FORMATTER =
             DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
 
+    private static final DateTimeFormatter[] DATETIME_FORMATTERS = new DateTimeFormatter[10];
+
+    private static DateTimeFormatter datetimeFormatter(int precision) {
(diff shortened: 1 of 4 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
```diff
@@ -137,6 +137,31 @@ public static void toStarRocksDataType(
     private static final DateTimeFormatter DATETIME_FORMATTER =
             DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
 
+    private static final DateTimeFormatter[] DATETIME_FORMATTERS = new DateTimeFormatter[10];
+
+    private static DateTimeFormatter datetimeFormatter(int precision) {
+        if (precision <= 0) {
+            return DATETIME_FORMATTER;
+        }
+        if (precision < DATETIME_FORMATTERS.length) {
```
yuxiqian: The maximal precision for temporal types is 9. We may add a precondition guard here, and remove the non-caching branch.

yuxiqian (commented):
Adding integrated tests / E2e cases would be nice.

PR #4405, opened 2026-05-20
[FLINK-39383][Docs] Improve scan.incremental.snapshot.chunk.key-column documentation for MySQL/PostgreSQL/Oracle/DB2/SQLServer connectors

ThorneANN:
The `scan.incremental.snapshot.chunk.key-column` option had several gaps
  and inaccuracies in its documentation across multiple connectors:

1. NOT NULL / no-primary-key constraint was undocumented.
The option is required when a captured table has no primary key and the chosen column must be NOT NULL. None of the connector docs mentioned this.

2. Multi-table format was undocumented (MySQL).
The MySQL connector supports `<table-pattern>:<column>[;...]` for assigning different chunk key columns to different tables. This was not documented in either the flink-source or pipeline connector docs.

3. Option was entirely missing from PostgreSQL pipeline connector docs. 

4. Stale option description wording across all connectors.

### DIFF
--- a/docs/content.zh/docs/connectors/flink-sources/db2-cdc.md
+++ b/docs/content.zh/docs/connectors/flink-sources/db2-cdc.md
@@ -240,8 +240,13 @@ for more detailed information.</td>
           <td>optional</td>
           <td style="word-wrap: break-word;">(none)</td>
           <td>String</td>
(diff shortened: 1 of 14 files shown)


ThorneANN:
@yux PTAL

ThorneANN:
@yuxiqian  PTAL,if u have  free time,tks.

PR #4407, opened 2026-05-21
[FLINK-39721] Support flink 1.18 compatibility for cdc runtime.

loserwang1024:
fix https://issues.apache.org/jira/browse/FLINK-39721

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/v2/FlussSink.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/v2/FlussSink.java
@@ -19,7 +19,6 @@ package org.apache.flink.cdc.connectors.fluss.sink.v2;
 
 import org.apache.flink.api.connector.sink2.Sink;
 import org.apache.flink.api.connector.sink2.SinkWriter;
-import org.apache.flink.api.connector.sink2.WriterInitContext;
 import org.apache.flink.runtime.metrics.groups.InternalSinkWriterMetricGroup;
 
 import org.apache.fluss.config.Configuration;
@@ -46,14 +45,6 @@ public class FlussSink<InputT> implements Sink<InputT> {
 
     @Deprecated
     public SinkWriter<InputT> createWriter(InitContext context) throws IOException {
-        FlussSinkWriter<InputT> flinkSinkWriter =
-                new FlussSinkWriter<>(flussConfig, context.getMailboxExecutor(), serializer);
-        flinkSinkWriter.initialize(InternalSinkWriterMetricGroup.wrap(context.metricGroup()));
-        return flinkSinkWriter;
-    }
-
-    @Override
-    public SinkWriter<InputT> createWriter(WriterInitContext context) throws IOException {
         FlussSinkWriter<InputT> flinkSinkWriter =
                 new FlussSinkWriter<>(flussConfig, context.getMailboxExecutor(), serializer);
         flinkSinkWriter.initialize(InternalSinkWriterMetricGroup.wrap(context.metricGroup()));
--- a/flink-cdc-flink1-compat/src/main/java/org/apache/flink/api/connector/sink2/SupportsCommitter.java
+++ b/flink-cdc-flink1-compat/src/main/java/org/apache/flink/api/connector/sink2/SupportsCommitter.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.api.connector.sink2;
+
+import org.apache.flink.core.io.SimpleVersionedSerializer;
+
+import java.io.IOException;
+
+/**
+ * Compatibility adapter for Flink 1.18. This class is part of the multi-version compatibility layer
+ * that allows Flink CDC to work across different Flink versions.
+ */
+public interface SupportsCommitter<CommittableT> {
+
+    Committer<CommittableT> createCommitter(CommitterInitContext context) throws IOException;
+
+    SimpleVersionedSerializer<CommittableT> getCommittableSerializer();
+}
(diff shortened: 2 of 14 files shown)


sd4324530:
@loserwang1024 
Hi, Flinkcdc has been no longer compatible with flink 1.18 since version 3.3.

loserwang1024:
@sd4324530 yes, in our document, flink cdc is only support flink 1.20 an 2.x.
image


 This pr is just a minor inproment, won't expose to document.  If someone need to use cdc in flink 1.18, he can do it.

leonardBang (commented):
## Review: [FLINK-39721] Support flink 1.18 compatibility for cdc runtime

### Strengths

- **Compile-time adapters instead of reflection is the right design direction for multi-version compatibility.** It avoids fragile approaches like `NoSuchMethodException` handling or `VersionUtil` string parsing: `RuntimeContextAdapter` uses `runtimeContext.getIndexOfThisSubtask()` in `flink1-compat` (present in 1.18/1.19/1.20) and `runtimeContext.getTaskInfo().getIndexOfThisSubtask()` in `flink2-compat` (the old method was removed in 2.x). Type-safe, zero reflection.
- **The source reader adapter constructor choices are correct.** In `flink-cdc-flink1-compat`, `SingleThreadFetcherManagerAdapter` calls `super(elementsQueue, splitReaderSupplier)` and `SingleThreadMultiplexSourceReaderBaseAdapter` calls `super(elementsQueue, splitFetcherManager, recordEmitter, config, context)` — I verified these `FutureCompletingBlockingQueue` (FCBQ) constructors exist in `flink-connector-base` 1.18.1/1.19.3/1.20.3. The flink2 versions drop FCBQ and match `flink-connector-base` 2.2.0.
- **Core data-plane call sites in the runtime module are uniformly routed through the adapter**: `distributed/SchemaOperator`, `regular/SchemaOperator`, `DataSinkWriterOperator` (both call sites), and `DistributedPrePartitionOperator` no longer call `getTaskInfo()` directly.
- **`FlussSink` correctly drops the `createWriter(WriterInitContext)` overload** (the type does not exist in 1.18), keeping only `createWriter(InitContext)` — the least common denominator of the `Sink` interface between 1.18 and 1.20.

### Issues

#### Critical (must fix)

- **[DataSinkFunctionOperator.java, lines ~81 and ~137] The migration is incomplete: these two call sites still invoke `getRuntimeContext().getTaskInfo().getIndexOfThisSubtask()` directly instead of going through the new `RuntimeContextAdapter`.** `RuntimeContext.getTaskInfo()` does not exist in Flink **1.18** (it was introduced in 1.19), and `DataSinkFunctionOperator` is a real runtime path used by `DataSinkTranslator` for the SinkFunction branch → this would throw `NoSuchMethodError` on 1.18, directly contradicting the PR's goal. Fix: replace both call sites with `RuntimeContextAdapter.getIndexOfThisSubtask(getRuntimeContext())`.
- **[flink-cdc-flink1-compat .../api/connector/sink2/SupportsPreCommitTopology.java and SupportsPreWriteTopology.java] These shims are in the wrong package.** Real Flink (1.19/1.20) defines `SupportsPreCommitTopology`/`SupportsPreWriteTopology` in `org.apache.flink.streaming.api.connector.sink2` (the streaming package), but the shims here live in `org.apache.flink.api.connector.sink2` (the core package). `DataSinkTranslator` imports the **streaming-package** versions, and nothing in the repo references the core-package ones → these two shims are dead code that helps neither 1.18 compilation nor runtime; the sink pre-write/pre-commit topology types are still missing under 1.18. Fix: move them to `org.apache.flink.streaming.api.connector.sink2`, or remove them and clarify the actual plan for 1.18 sink topology support.

#### Important (should fix)

- **[flink-cdc-flink1-compat .../SupportsCommitter.java] `createCommitter(CommitterInitContext)` references a type that does not exist in 1.18** (`CommitterInitContext` is absent from both flink-core and flink-streaming-java in 1.18 — verified), and this PR provides no shim for it. More critically: `DataSinkTranslator` compiles against 1.20.3, where the constructor is `CommitterOperatorFactory(SupportsCommitter, boolean, boolean)`, while 1.18 has `CommitterOperatorFactory(TwoPhaseCommittingSink, boolean, boolean)` — different method descriptors, so the committer path would very likely hit `NoSuchMethodError` at runtime on 1.18. Please clarify whether this PR only covers subtask-index access in runtime operators, or also commits to a working sink committer on 1.18; if the latter, a `CommitterInitContext` shim and a 1.18-compatible `CommitterOperatorFactory` layer are required.
- **[Tests and CI] The PR's central claim is never actually built or tested against 1.18.** The diff touches no `src/test` and no `pom.xml`. The root pom pins `flink.1.x.version=1.20.3`, `flink2-compat` uses 2.2.0, there is no 1.18 profile anywhere, and the CI workflow defaults to the `2.2.0` / `-Pflink2` matrix. So the compat modules are only ever compiled against 1.20.3 and 2.2.0. Suggestion: add a CI matrix or profile that compiles with `-Dflink.1.x.version=1.18.x`, plus minimal unit tests for both old and new paths — otherwise there is no evidence backing "1.18 compatibility".
- **[flink-cdc-flink1-compat duplicates real Flink classes] `flink1-compat` is packaged into the runtime jar with compile scope** (`flink-cdc-runtime/pom.xml`, no `<scope>`), shipping `org.apache.flink.api.connector.sink2.SupportsCommitter` while flink-core in 1.19/1.20 contains a class with the same FQCN. Although `org.apache.flink.*` is usually parent-first (real Flink classes win) so it is not immediately fatal, this is a fragile duplicate-class / potential shading conflict. Suggestion: relocate the compat artifacts or define an explicit scope/class-loading strategy, and verify on 1.19 (where `SupportsCommitter` already exists) that nothing is accidentally shadowed.

#### Minor (nice to have)

- **[All new files, class-level Javadoc] Copy-paste error:** the adapters under `flink-cdc-flink2-compat` (`RuntimeContextAdapter`, `SingleThreadFetcherManagerAdapter`, `SingleThreadMultiplexSourceReaderBaseAdapter`) are all documented as "Compatibility adapter for **Flink 1.18**" — they should say Flink 2.x to avoid misleading readers.
- **[flink2-compat SingleThreadFetcherManagerAdapter / SingleThreadMultiplexSourceReaderBaseAdapter] The constructors still accept a `FutureCompletingBlockingQueue<RecordsWithSplitIds<E>> elementsQueue` parameter that is silently discarded** (the flink2 `super` doesn't need it). Keeping both versions' constructor signatures aligned for uniform compilation is reasonable, but please add a comment stating the parameter is intentionally ignored on 2.x.
- **[FlussSink.java] The retained `createWriter(InitContext)` is now `@Deprecated` but lost its `@Override`.** It is an implementation of the `Sink` interface method — keep `@Override` so signature drift fails loudly. The `@Deprecated` semantics also seem contradictory: on 1.18 this is the only valid entry point.

### Verdict

**Ready to merge?** No

**Reasoning:** There are two defects that would crash at runtime on 1.18 (`DataSinkFunctionOperator` still calls `getTaskInfo()`, which does not exist in 1.18; the committer path's `CommitterOperatorFactory`/`CommitterInitContext` is incompatible with 1.18), and the `SupportsPreCommit/PreWriteTopology` shims are dead code in the wrong package. More fundamentally, nothing in the repo actually compiles or tests against 1.18, so the PR's core claim is unverified. Please complete the migration, fix the package placement, and add 1.18 compile/test evidence before merging.

sd4324530:
Hi, @leonardBang @loserwang1024 
I'm concerned this might confuse the community developers, as the documentation states that compatibility with Flink1.18 is no longer required.
If users do have this need, I think there are two solutions:
a. Keep this feature in a separate branch.
b. Update the documentation to explain that we need to ensure compatibility with Flink 1.18 in the future, and add CI to Flink 1.18 to avoid the risk of incompatibility.

I personally prefer solution A.
WDYT?

leonardBang:
+1 for a, 1.18 is pretty old version

PR #4408, opened 2026-05-22
[FLINK-39407][cdcup] Mount shared data volume on Flink containers for Paimon sink

jubins:

Fixes FLINK-39407 — the `cdcup` quickstart wizard configures the Paimon sink with warehouse path `/data/paimon-warehouse`, but the generated `docker-compose.yaml` never mounts the shared `cdc-data` volume on the Flink `jobmanager` and `taskmanager` services. Other service generators (`my_sql.rb`, `doris.rb`, `star_rocks.rb`) all mount `cdc-data:/data` on their own containers, but the Flink containers are missing the same mount. As a result, when `PaimonMetadataApplier.applyCreateTable` runs inside the JobManager and asks `FileSystemCatalog` to create `/data/<db>.db`, the path resolves to the container's ephemeral filesystem (which the user's host volume never sees), and MySQL → Paimon pipelines fail at first `CreateTableEvent` with either `DatabaseNotExistException` (CDC 3.2.1, which did not yet auto-create the database) or `RuntimeException: Create database location failed` (CDC 3.5.0). The user then inspects the host-side warehouse directory and finds it empty, confirming nothing reached the shared volume.

## Brief change log

- Added `volumes: ["cdc-data:/data"]` to the generated `jobmanager` service in `tools/cdcup/src/app.rb`, matching the volume key already declared at the top level of the generated compose file (`CDC_DATA_VOLUME`).
- Added the same `volumes: ["cdc-data:/data"]` entry to the generated `taskmanager` service, so the actual Paimon writes (which happen on the TM) also land on the shared volume.
- No changes to the Java connector code — `PaimonMetadataApplier` already auto-creates the database on master via `catalog.createDatabase(...)`; the bug was purely that the warehouse path was not actually backed by a shared mount.

## Verifying this change

This change is covered by existing tests:

- `PaimonMetadataApplierTest.testApplySchemaChange` — drops the database first, then applies a `CreateTableEvent` against a temp-dir warehouse, exercising the auto-database-creation path that the cdcup environment was failing to actually persist.
- `MySqlToPaimonE2eITCase` — full MySQL → Paimon pipeline with a real shared volume between JM and TM, demonstrating the expected working configuration that the cdcup-generated compose file now matches.

Manual verification: run `cdcup.sh`, choose MySQL → Paimon, inspect the generated `docker-compose.yaml` to confirm both Flink services carry `volumes: [cdc-data:/data]`, then run the pipeline and observe that `/data/paimon-warehouse/<db>.db/...` is created on the shared volume rather than disappearing into the container.

No new tests added — there is no Ruby test harness under `tools/cdcup/` (no `spec/`, no Gemfile, no test runner), and introducing one for a single-line config-generator fix would be disproportionate.

- **Dependencies** (does it add or upgrade a dependency): no
- **The public API**, i.e., is any changed class annotated with `@Public(Evolving)`: no — change is in a Ruby quickstart wizard under `tools/cdcup/`, not in any Java API.
- **The serializers**: no
- **The runtime per-record code paths** (performance sensitive): no
- **Anything that affects deployment or recovery**: no — this only changes a developer-experience tool that emits sample docker-compose files; production deployments compose their own YAML.
- **The S3 file system connector**: no

## Documentation

- **Does this pull request introduce a new feature?** no — it fixes the existing cdcup MySQL → Paimon scenario that was broken out of the box.
- **If yes, how is the feature documented?** n/a. The existing `docs/content/docs/get-started/quickstart-for-1.20/cdc-up-quickstart-guide.md` already describes the intended `/data` shared layout; this change just makes the generated compose match the documented behavior.

Yes — Claude was used as a pair-programming assistant for diagnosing the failure mode (matching the two stack traces against the cdcup-generated compose layout) and for drafting the fix. All code was written, understood, and verified by the author.

Generated-by: Claude Opus 4.7

### DIFF
--- a/tools/cdcup/src/app.rb
+++ b/tools/cdcup/src/app.rb
@@ -90,7 +90,6 @@ flink_cdc_version = @prompt.select('  ️ Which Flink CDC version would you like
   'hostname' => 'jobmanager',
   'ports' => ['8081'],
   'command' => 'jobmanager',
-  'volumes' => ["#{CDC_DATA_VOLUME}:/data"],
   'environment' => {
     'FLINK_PROPERTIES' => "jobmanager.rpc.address: jobmanager\nexecution.checkpointing.interval: 3000"
   }
@@ -100,7 +99,6 @@ flink_cdc_version = @prompt.select('  ️ Which Flink CDC version would you like
   'image' => "flink:#{flink_version}-scala_2.12",
   'hostname' => 'taskmanager',
   'command' => 'taskmanager',
-  'volumes' => ["#{CDC_DATA_VOLUME}:/data"],
   'environment' => {
     'FLINK_PROPERTIES' => "jobmanager.rpc.address: jobmanager\ntaskmanager.numberOfTaskSlots: 4\nexecution.checkpointing.interval: 3000"
   }
--- a/tools/cdcup/src/app.rb
+++ b/tools/cdcup/src/app.rb
@@ -90,6 +90,7 @@ flink_cdc_version = @prompt.select('  ️ Which Flink CDC version would you like
   'hostname' => 'jobmanager',
   'ports' => ['8081'],
   'command' => 'jobmanager',
+  'volumes' => ["#{CDC_DATA_VOLUME}:/data"],
   'environment' => {
     'FLINK_PROPERTIES' => "jobmanager.rpc.address: jobmanager\nexecution.checkpointing.interval: 3000"
   }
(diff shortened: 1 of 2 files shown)


tools/cdcup/src/app.rb
```diff
@@ -90,6 +90,7 @@
   'hostname' => 'jobmanager',
   'ports' => ['8081'],
   'command' => 'jobmanager',
+  'volumes' => ["#{CDC_DATA_VOLUME}:/data"],
```
yuxiqian: Will it be better if we append extra volume configs here?
https://github.com/apache/flink-cdc/blob/b6275e72ae6033a7cc95d18e8ebe28337b14f20e/tools/cdcup/src/sink/paimon.rb#L26

jubins: @yuxiqian good call, pushed an update that moves the volume mount into `Paimon.prepend_to_docker_compose_yaml` and reverts the unconditional mount in `app.rb`. Only Paimon needs `/data` on the Flink containers (since FileSystemCatalog writes happen inside JM/TM), so scoping it to just the sink generator

yuxiqian (approved):
Thanks for the quick fix!

PR #4409, opened 2026-05-22
[FLINK-39732] Introduce ObjectIdDiscoverer SPI for flexible table subscription (with default JdbcTableDiscoverer)

loserwang1024:
fix https://issues.apache.org/jira/browse/FLINK-39732

## Summary

This PR introduces a connector-agnostic `ObjectIdDiscoverer<T>` SPI to decouple object subscription discovery from individual CDC connectors.

Currently, each source connector tends to define its own way to decide which objects, especially tables, should be subscribed to, such as hard-coded table lists, regular expressions, or connector-specific include/exclude rules. This makes it hard to reuse discovery logic across connectors and hard for users to plug in dynamic discovery strategies such as metadata tables, custom SQL, REST services, or other external systems.

This PR also ships a default JDBC-backed table discoverer implementation for the common “subscription metadata table” pattern.

## Brief change log

- Added the generic `ObjectIdDiscoverer<T>` SPI for discovering arbitrary object identifier types.
- Added `ObjectIdDiscovererFactory<T>` with ServiceLoader-based discovery and matching by `type()` plus `objectIdClass()`.
- Kept `TableDiscoverer` and `TableDiscovererFactory` as `TableId`-specific convenience APIs on top of the generic SPI.
- Added backward-compatible SPI loading from both `ObjectIdDiscovererFactory` and legacy `TableDiscovererFactory` service files, with duplicate implementation classes de-duplicated.
- Added `JdbcTableDiscoverer` and `JdbcTableDiscovererFactory` as the default JDBC-backed implementation for table subscription discovery.
- Supported two JDBC discovery modes:
  - shared subscription table mode, filtered by `subscribe-id`;
  - custom query mode, where the first selected column is parsed as a fully-qualified `TableId`.
- Added tests for generic factory matching, duplicate/unknown factory handling, SPI loading, case-insensitive type matching, JDBC shared-table mode, custom-query mode, dynamic metadata changes, and required option validation.

可以，在 PR 描述里建议单独加一个 `## Public API` section，把 `ObjectIdDiscoverer`、`ObjectIdDiscovererFactory`、`TableDiscoverer` 的关系讲清楚。可以直接用下面这段：

## Public API

This PR introduces a new generic discovery SPI centered around `ObjectIdDiscoverer<T>`.

### `ObjectIdDiscoverer<T>`

`ObjectIdDiscoverer<T>` is the core abstraction for discovering a set of object identifiers from an external system or user-defined source.

```java
@PublicEvolving
public interface ObjectIdDiscoverer<T> extends Serializable, AutoCloseable {

    void open(Context context) throws Exception;

    Set<T> discover() throws Exception;

    @Override
    void close() throws Exception;

    interface Context {
        Configuration getConfiguration();

        ClassLoader getUserCodeClassLoader();
    }
}
```

The generic type `T` represents the identifier type returned by the discoverer. For example, a table discoverer returns `TableId`, while future implementations may return other object identifier types.

The lifecycle is:

- `open(Context)` initializes resources such as JDBC connections, clients, or caches.
- `discover()` returns the current set of discovered object identifiers.
- `close()` releases resources held by the discoverer.

### `ObjectIdDiscovererFactory<T>`

`ObjectIdDiscovererFactory<T>` is the SPI factory used to create discoverers and resolve implementations through Java `ServiceLoader`.

```java
@PublicEvolving
public interface ObjectIdDiscovererFactory<T> {

    String type();

    Class<T> objectIdClass();

    ObjectIdDiscoverer<T> createDiscoverer();

    static <T> ObjectIdDiscoverer<T> createDiscoverer(
            String type, Class<T> objectIdClass, ClassLoader classLoader);

    static <T> ObjectIdDiscovererFactory<T> discoverFactory(
            String type, Class<T> objectIdClass, ClassLoader classLoader);
}
```

Factories are selected by two dimensions:

- `type()` identifies the storage-side implementation, such as `jdbc`.
- `objectIdClass()` identifies the discovered object id type, such as `TableId.class`.

This allows the same storage type to support different object identifier types in the future without overloading a single string identifier.

Resolution rules:

- Factory `type()` matching is case-insensitive.
- `objectIdClass()` must match exactly.
- Unknown combinations fail with an `IllegalArgumentException` that lists available factories.
- Duplicate factories for the same `type()` and `objectIdClass()` fail with an `IllegalStateException`.

### `TableDiscoverer`

`TableDiscoverer` is a table-specific specialization of the generic SPI.

```java
@PublicEvolving
public interface TableDiscoverer extends ObjectIdDiscoverer<TableId> {}
```

It keeps the table discovery use case easy to use while allowing the underlying SPI to remain generic.

### `TableDiscovererFactory`

`TableDiscovererFactory` is the `TableId`-specific factory API built on top of `ObjectIdDiscovererFactory<TableId>`.

```java
@PublicEvolving
public interface TableDiscovererFactory extends ObjectIdDiscovererFactory<TableId> {

    default String identifier();

    @Override
    default Class<TableId> objectIdClass();

    @Override
    TableDiscoverer createDiscoverer();

    static TableDiscoverer createDiscoverer(String type, ClassLoader classLoader);
}
```

This keeps compatibility and convenience for table discovery users, while new generic discovery logic should be implemented against `ObjectIdDiscovererFactory<T>`.

## Built-in JDBC implementation

The JDBC discoverer reads the subscription list from any JDBC-compatible database. It supports two modes, chosen implicitly by the configuration:

### Mode A — Shared subscription table (default & recommended)

Designed for the common multi-tenant pattern where many CDC jobs share **one** metadata table and each job filters by a `subscribe-id`.

Effective query (built with a `PreparedStatement`, injection-safe on the parameter):

```sql
SELECT <column-name>
  FROM <table-name>
 WHERE <subscribe-id-column> = ?
```

| Key | Required | Default | Notes |
| --- | --- | --- | --- |
| `table.discoverer.jdbc.url` | yes | — | JDBC connection URL |
| `table.discoverer.jdbc.username` | yes | — | |
| `table.discoverer.jdbc.password` | yes | — | |
| `table.discoverer.jdbc.table-name` | yes | — | Shared metadata table |
| `table.discoverer.jdbc.subscribe-id` | **yes** | — | Identifies this subscription set; mandatory in Mode A |
| `table.discoverer.jdbc.column-name` | no | `subscribe_table_name` | Column holding fully-qualified table names |
| `table.discoverer.jdbc.subscribe-id-column` | no | `subscribe_id` | Column used in the `WHERE` clause |

Recommended schema:

```sql
CREATE TABLE cdc_subscriptions (
    subscribe_id         VARCHAR(64)  NOT NULL,
    subscribe_table_name VARCHAR(255) NOT NULL,
    PRIMARY KEY (subscribe_id, subscribe_table_name)
);
```

The legacy "unset `subscribe-id` ⇒ scan the whole metadata table" fallback has been intentionally removed: leaving `subscribe-id` blank is now an error, which prevents silent over-subscription in shared deployments.

### Mode B — Custom SELECT (escape hatch)

For uncommon layouts (JOINs, soft filters, multi-column metadata) users may set:

| Key | Required |
| --- | --- |
| `table.discoverer.jdbc.subscribe-query` | yes (when used) |

## Example

```yaml
# pipeline.yaml — Mode A (shared subscription table)
source:
  type: ...
  table.discoverer.type: jdbc
  table.discoverer.jdbc.url: jdbc:mysql://meta:3306/cdc_meta
  table.discoverer.jdbc.username: cdc
  table.discoverer.jdbc.password: ******
  table.discoverer.jdbc.table-name: cdc_subscriptions
  table.discoverer.jdbc.subscribe-id: orders-subscription
```

```yaml
# pipeline.yaml — Mode B (custom query)
source:
  type: ...
  table.discoverer.type: jdbc
  table.discoverer.jdbc.url: jdbc:mysql://meta:3306/cdc_meta
  table.discoverer.jdbc.username: cdc
  table.discoverer.jdbc.password: ******
  table.discoverer.jdbc.subscribe-query: |
    SELECT CONCAT(db, '.', tbl) FROM ops.cdc_meta
    WHERE enabled = 1 AND owner = 'team-orders'
```

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/JdbcTableDiscoverer.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/JdbcTableDiscoverer.java
@@ -1,0 +1,304 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.common.source.discover;
+
+import org.apache.flink.cdc.common.configuration.ConfigOption;
+import org.apache.flink.cdc.common.configuration.ConfigOptions;
+import org.apache.flink.cdc.common.configuration.Configuration;
+import org.apache.flink.cdc.common.event.TableId;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.sql.Connection;
+import java.sql.Driver;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.LinkedHashSet;
+import java.util.Properties;
+import java.util.ServiceLoader;
+import java.util.Set;
+
+/**
+ * A {@link TableDiscoverer} that reads the list of subscribed tables from a JDBC database.
+ *
+ * <p>This implementation connects to any JDBC-compatible database (e.g., MySQL, PostgreSQL) and
+ * returns fully-qualified table names parsed as {@link TableId} objects.
+ *
+ * <h3>Default mode — shared subscription table (recommended)</h3>
+ *
+ * <p>By default, the discoverer assumes that subscriptions for many CDC jobs live in a single
+ * shared database table, and that each subscription set is identified by a {@code subscribe-id}.
+ * The discoverer issues a parameterized
+ *
+ * <pre>{@code
+ * SELECT column-name FROM table-name WHERE subscribe-id-column = ?
+ * }</pre>
+ *
+ * <p>using a {@link PreparedStatement} (injection-safe), and only the rows whose subscribe-id
+ * matches the configured value are returned.
+ *
+ * <p><b>Required keys:</b> {@code table.discoverer.jdbc.url}, {@code
+ * table.discoverer.jdbc.username}, {@code table.discoverer.jdbc.password}, {@code
+ * table.discoverer.jdbc.table-name}, {@code table.discoverer.jdbc.subscribe-id}.
+ *
+ * <p><b>Optional keys:</b> {@code table.discoverer.jdbc.column-name} (defaults to {@code
+ * "subscribe_table_name"}), {@code table.discoverer.jdbc.subscribe-id-column} (defaults to {@code
+ * "subscribe_id"}).
+ *
+ * <p><b>Recommended schema:</b>
+ *
+ * <pre>{@code
+ * CREATE TABLE cdc_subscriptions (
+ *     subscribe_id         VARCHAR(64)  NOT NULL,
+ *     subscribe_table_name VARCHAR(255) NOT NULL,
+ *     PRIMARY KEY (subscribe_id, subscribe_table_name)
+ * );
+ * INSERT INTO cdc_subscriptions VALUES
+ *   ('orders-subscription',    'source_db.orders'),
+ *   ('orders-subscription',    'source_db.order_items'),
+ *   ('analytics-subscription', 'analytics_db.user_events');
+ * }</pre>
+ *
+ * <h3>Advanced escape hatch — custom query (overrides the default mode)</h3>
+ *
+ * <p>For uncommon layouts (e.g., needing JOINs or extra filters), users may set {@code
+ * table.discoverer.jdbc.subscribe-query} to any {@code SELECT} statement; column #1 of each row is
+ * treated as a fully-qualified table name. <b>When this option is set it takes priority over the
+ * default mode and all of {@code table-name}, {@code column-name}, {@code subscribe-id-column} and
+ * {@code subscribe-id} are ignored.</b> Use this only when the default schema cannot model your
+ * subscriptions.
+ *
+ * <p>Null values and rows that cannot be parsed into a valid {@link TableId} are silently skipped.
+ */
+public class JdbcTableDiscoverer implements TableDiscoverer {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcTableDiscoverer.class);
+
+    public static final ConfigOption<String> JDBC_URL =
+            ConfigOptions.key("table.discoverer.jdbc.url")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription("The JDBC connection URL for the table discovery database.");
+
+    public static final ConfigOption<String> USERNAME =
+            ConfigOptions.key("table.discoverer.jdbc.username")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription("The JDBC username for the table discovery database.");
+
+    public static final ConfigOption<String> PASSWORD =
+            ConfigOptions.key("table.discoverer.jdbc.password")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription("The JDBC password for the table discovery database.");
+
+    public static final ConfigOption<String> SUBSCRIBE_QUERY =
+            ConfigOptions.key("table.discoverer.jdbc.subscribe-query")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Custom SELECT statement used to discover subscribed tables. When set, "
+                                    + "this takes priority over the shared-table options. Column #1 of "
+                                    + "every row must be a fully-qualified table name.");
+
+    public static final ConfigOption<String> TABLE_NAME =
+            ConfigOptions.key("table.discoverer.jdbc.table-name")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The shared subscription table that stores subscription entries for "
+                                    + "one or more CDC jobs. Required in shared-table mode.");
+
+    public static final ConfigOption<String> COLUMN_NAME =
+            ConfigOptions.key("table.discoverer.jdbc.column-name")
+                    .stringType()
+                    .defaultValue("subscribe_table_name")
+                    .withDescription(
+                            "The column name in the subscription table that contains the "
+                                    + "fully-qualified table names to subscribe to.");
+
+    public static final ConfigOption<String> SUBSCRIBE_ID_COLUMN =
+            ConfigOptions.key("table.discoverer.jdbc.subscribe-id-column")
+                    .stringType()
+                    .defaultValue("subscribe_id")
+                    .withDescription(
+                            "The column name in the subscription table that holds the "
+                                    + "subscription-set identifier used for filtering.");
+
+    public static final ConfigOption<String> SUBSCRIBE_ID =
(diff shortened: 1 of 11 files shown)


loserwang1024:
@leonardBang , CC

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/JdbcTableDiscoverer.java
```diff
@@ -0,0 +1,257 @@
...
+                    "JdbcTableDiscoverer running in shared-table mode. URL='{}', table='{}', "
+                            + "column='{}', subscribeIdColumn='{}', subscribeId='{}'.",
+                    jdbcUrl,
+                    tableName,
+                    columnName,
+                    subscribeIdColumn,
+                    subscribeId);
+        }
+
+        connection = DriverManager.getConnection(jdbcUrl, username, password);
```
leonardBang: Context now exposes a user code class loader, but it is not used here when opening the JDBC connection. In deployments where the driver only exists in the user or connector classpath, this can still fail with `No suitable driver`. Please explicitly load or register the driver with that class loader before calling `DriverManager.getConnection(...)`.

docs/themes/book
leonardBang, line 1: This submodule deletion looks unrelated to the TableDiscoverer change and leaves the repo in an inconsistent state: `.gitmodules` still declares `docs/themes/book`, `docs/README.md` still asks contributors to initialize submodules, and `.github/workflows/docs.sh` still does so before building the docs. Please revert this deletion unless the full submodule cleanup is part of this PR.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/TableDiscovererFactory.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * SPI factory that creates a {@link TableDiscoverer}. Implementations are discovered at runtime via
+ * Java's {@link ServiceLoader} mechanism. To register a new discoverer type, add a line with the
+ * fully-qualified factory class name to:
+ *
+ * <pre>{@code
+ * META-INF/services/org.apache.flink.cdc.common.source.discover.TableDiscovererFactory
+ * }</pre>
+ *
+ * <p>The discoverer type is selected by the user via the {@code table.discoverer.type} option,
```
leonardBang: I could not find a production integration point in this PR that actually consumes `table.discoverer.type` or calls `TableDiscovererFactory.createDiscoverer(...)`; the usages I found are only in tests. As written, this documentation implies the option is already wired into the real source or data source creation flow. Could you either add that integration here or narrow the documentation so it does not overstate the current functionality?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/JdbcTableDiscovererFactory.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.common.source.discover;
+
+/**
+ * {@link TableDiscovererFactory} for {@link JdbcTableDiscoverer}. Activated when {@code
```
leonardBang: I grepped the PR branch and the only `TableDiscovererFactory.createDiscoverer(...)` call sites I could find are in `JdbcTableDiscovererITCase`; I could not find any `src/main` caller yet. Because of that, saying this is already activated by `table.discoverer.type = 'jdbc'` reads stronger than the current implementation. Could you either wire this option into the production source creation path in this PR or soften this Javadoc for now?

loserwang1024: >  I could not find any src/main caller yet.

I will use it in fluss yaml source later (poc code is here:https://github.com/loserwang1024/flink-cdc-connectors/tree/fluss-yaml-source).

 > Could you either wire this option into the production source creation path in this PR or soften this Javadoc for now?

I will soften this Javadoc for now.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/TableDiscoverer.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    void open(Context context) throws Exception;
+
+    /**
+     * Discovers and returns the set of tables to subscribe to.
+     *
+     * @return A set of {@link TableId} representing the tables to read.
+     * @throws Exception if the discovery fails.
+     */
+    Set<TableId> discover() throws Exception;
```
yuxiqian: Can we make `TableId` generic here? Table ID might not be a suitable abstraction for multimodal data sources.

yuxiqian: Maybe rename `TableDiscoverer` => `ObjectIdDiscoverer`, too?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/ObjectIdDiscoverer.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.flink.cdc.common.annotation.PublicEvolving;
+import org.apache.flink.cdc.common.configuration.Configuration;
+
+import java.io.Serializable;
+import java.util.Set;
+
+/** Pluggable abstraction for discovering a set of object identifiers. */
+@PublicEvolving
+public interface ObjectIdDiscoverer<ObjectIdentifier> extends Serializable, AutoCloseable {
```
yuxiqian: Just call it `<T>`? `ObjectIdentifier` looks like a specific class.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/ObjectIdDiscovererFactory.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+
+/** SPI factory that creates an {@link ObjectIdDiscoverer}. */
+@PublicEvolving
+public interface ObjectIdDiscovererFactory<ObjectIdentifier> {
```
yuxiqian: Ditto

loserwang1024:
@yuxiqian , just modified it.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/ObjectIdDiscovererFactory.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+
+/** SPI factory that creates an {@link ObjectIdDiscoverer}. */
+@PublicEvolving
+public interface ObjectIdDiscovererFactory<T> {
```
yuxiqian, line 33: minor: shall we implement `ObjectIdDiscovererFactory<T>` by inheriting `org.apache.flink.cdc.common.factories.Factory`? Thus we don't need to re-implement config validation, factory discovering here.

flink-cdc-common/src/main/resources/META-INF/services/org.apache.flink.cdc.common.source.discover.TableDiscovererFactory
```diff
@@ -0,0 +1,16 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
yuxiqian, line 1: By reusing standard `org.apache.flink.cdc.common.factories.Factory`,  we don't need two META-INF service definition here.

yuxiqian (approved):
Thanks Hongshun for the nice work, just left some minor comments.

leonardBang (approved):
Thanks @loserwang1024 for the update and @yuxiqian for the review, LGTM.

One minor suggestion, we'd better update the commit message and PR description according the PR discussion to help track and reflect the merged code.

PR #4411, opened 2026-05-25
Fix CdcAsFlussRow.java file location to match package declaration

haruki-830:
The file CdcAsFlussRow.java was incorrectly located in sink/row/row/ directory while its package declaration is org.apache.flink.cdc.connectors.fluss.sink.row. This mismatch violates Java's requirement that package declaration must match the file system path.


lvyanquan (approved):
+1.

PR #4413, opened 2026-05-25
[FLINK-39749][mysql-cdc] Support configurable string chunk key comparison mode to align with MySQL collation

ziyanTOP:

Fix chunk splitting and binlog event routing issues when MySQL collation differs from Java's natural String ordering.

Java \`String.compareTo()\` is case-sensitive (Unicode code point order), while MySQL collations like \`utf8mb4_general_ci\` are case-insensitive. This mismatch causes chunk boundaries computed by Java to diverge from actual MySQL row ordering, leading to premature unbounded chunks, overlapping splits, or lost binlog events.

See [FLINK-39749](https://issues.apache.org/jira/browse/FLINK-39749) for details.

## Brief change log

- Introduce \`ChunkKeyCompareMode\` enum: \`DEFAULT\`, \`CASE_INSENSITIVE\`, \`BINARY\`
- \`DEFAULT\`: preserves existing behavior (\`String.compareTo()\`)
- \`CASE_INSENSITIVE\`: uses \`String.compareToIgnoreCase()\` for Java-side comparisons
- \`BINARY\`: injects \`BINARY\` keyword in SQL predicates and uses byte-level comparison in Java
- Cover all three API layers: DataStream API, Flink SQL, Pipeline YAML
- Update documentation (EN + ZH) and add test coverage

## Verifying this change

This change is already covered by existing tests (\`MySqlTableSourceFactoryTest\`) and has been verified in a production Pipeline job (MySQL -> Paimon) with \`CHAR(36)\` UUID primary key and \`utf8mb4_general_ci\` collation.

- The public API, i.e., is any changed class annotated with \`@Public(Evolving)\`: no
- The serializers: no
- The runtime per-record code paths (performance sensitive): yes (affects chunk key comparison in snapshot/binlog phases)
- The S3 file system connector: no

## Documentation

### DIFF
--- a/docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md
+++ b/docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md
@@ -441,6 +441,19 @@ Flink SQL> SELECT * FROM orders;
         如果跳过 backfill ，快照阶段捕获表的更改将在稍后的 binlog 读取阶段被回放，而不是合并到快照中。<br>
         警告：跳过 backfill 可能会导致数据不一致，因为快照阶段发生的某些 binlog 事件可能会被重放（仅保证 at-least-once ）。
         例如，更新快照阶段已更新的值，或删除快照阶段已删除的数据。这些重放的 binlog 事件应进行特殊处理。
+      </td>
+    </tr>
+    <tr>
+      <td>scan.incremental.snapshot.string-key.compare-mode</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">default</td>
+      <td>String</td>
+      <td>
+        增量快照阶段字符串类型 chunk key 的比较模式。可选值包括：
+        <li><code>default</code>：使用 Java 默认字符串比较（区分大小写，Unicode 码点顺序）。可能与 MySQL 的 <code>utf8mb4_general_ci</code> 等大小写不敏感排序规则不一致。</li>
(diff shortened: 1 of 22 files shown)


docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md
```diff
@@ -441,6 +441,19 @@ Flink SQL> SELECT * FROM orders;
         如果跳过 backfill ，快照阶段捕获表的更改将在稍后的 binlog 读取阶段被回放，而不是合并到快照中。<br>
         警告：跳过 backfill 可能会导致数据不一致，因为快照阶段发生的某些 binlog 事件可能会被重放（仅保证 at-least-once ）。
         例如，更新快照阶段已更新的值，或删除快照阶段已删除的数据。这些重放的 binlog 事件应进行特殊处理。
+      </td>
+    </tr>
+    <tr>
+      <td>scan.incremental.snapshot.string-key.compare-mode</td>
```
lvyanquan, line 447: We can use `SHOW VARIABLES LIKE "collation_server"` to query the character set in use from MySQL, avoiding the introduction of a new configuration parameter.

lvyanquan (dismissed):

docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md
ziyanTOP: Good point, but `collation_server` is just the server default — the actual table or column can override it. To auto-detect properly we'd have to query `information_schema.COLUMNS` per table at startup, and it gets messy with composite PKs where each column might have a different collation.
So I went with explicit config for now — `default` keeps backward compat and won't break anyone. We can add an `auto` mode later: detect each table's chunk-key collation during snapshot, store it per-table in the split state, and handle them individually

PR #4418, opened 2026-05-29
[FLINK-39775][mysql] Release snapshot split metadata after entering the stream phase

spoorthibasu:
**TL;DR:** Fixes a JobManager OOM on large tables by releasing the snapshot split metadata once the source enters the binlog phase. The release is tied to the current assignment, so a reader failover can't drop metadata the new reader still needs. Opt-in, default off.

## Root Cause

After the snapshot phase finishes and the binlog split is assigned, the coordinator keeps the full snapshot split metadata in memory indefinitely:

- `MySqlSnapshotSplitAssigner` holds `assignedSplits`, `splitFinishedOffsets`, and `tableSchemas`.
- `snapshotState()` rebuilds the complete `SnapshotPendingSplitsState` on every checkpoint, and `notifyCheckpointComplete()` never drops it.
- `MySqlSourceEnumerator` keeps a second copy in its `binlogSplitMeta` cache.

With many finished splits (around 300K for a 2.5B-row table at the default chunk size of 8096), every post-snapshot checkpoint re-serializes the entire state. That holds JobManager heap high, and since the state is copied during checkpoint serialization and transfer, it also drives up direct memory and can OOM-kill the container even when heap stays under its limit.

## Fix

Release the metadata once it is no longer needed for correctness: after the binlog split is assigned, its metadata assembled by the reader, and that covered by a completed checkpoint.

### Reader-side (completion signal)

When a complete binlog split enters reading (metadata arrived inline or was assembled over meta-group requests), `MySqlSourceReader` sends a `BinlogSplitMetaAssembledEvent`. It is re-sent every time a complete split re-enters reading, including after a restore, so it stays idempotent.

### Coordinator-side (gated release)

- Opt-in behind `scan.incremental.snapshot.metadata.release.enabled` (default `false`), so existing jobs are unchanged.
- On the event, `MySqlHybridSplitAssigner` records that the reader assembled the metadata. `snapshotState()` schedules the release at the current checkpoint id, and the clearing runs in `notifyCheckpointComplete()` once that checkpoint completes.
- The clearing drops `assignedSplits`, `splitFinishedOffsets`, and `tableSchemas`, and the enumerator drops its `binlogSplitMeta` cache. `alreadyProcessedTables`, the assigner status, and the chunk splitter state are kept so a restore does not re-discover tables.
- Enabling this with `scan.newly-added-table.enabled` fails fast at job submission, since that flow needs the metadata this option releases. `scan.binlog.newly-added-table.enabled` reads from the binlog and is unaffected.

## Why the release is safe

- Clearing runs only in `notifyCheckpointComplete()`, never `snapshotState()`. Flink re-delivers (`addSplitsBack`) only assignments not covered by the last completed checkpoint. The binlog split was assigned before the release-scheduling checkpoint, so by the time release runs it is checkpoint-covered and can't be added back into an emptied assigner. `createBinlogSplit()` also asserts the metadata was not released, turning any future regression into a fast failure rather than silent data loss.
- If the binlog reader resets before release, the scheduled release is reset and the metadata is intact, so the split is recreated normally. This holds whether or not the split is handed back: a reader whose assignment was already checkpoint-covered restores it from its own state and hands nothing back.
- On restore from a released ("light") state, a persisted flag tells the assigner the metadata was released, so it does not re-discover tables or recreate the binlog split. The reader restores its own complete split, requests no metadata, and re-sends the assembled event.
- The format change is scoped to the option. `PendingSplitsStateSerializer` writes version 6 with the flag only when metadata release is enabled, so a job that leaves the option off keeps writing version 5 and stays restorable by an older connector build. Restoring a version-6 checkpoint whose flag is set with the option turned off fails fast rather than dropping the marker.

## Tying the assembled event to its assignment

The assembled event must belong to a specific assignment. Flink does not drop source events from a failed reader attempt (the non-speculative `SourceCoordinator` hands the event to the enumerator with no attempt check), and the event carried no identity, so on a large table this interleaving was possible:

1. The reader assembles the split over many meta-group requests and sends the assembled event.
2. The reader fails; the split is added back and re-created with empty metadata.
3. A fresh reader re-requests meta groups, spanning several checkpoints on a large table.
4. The stale event from the dead attempt arrives late and re-arms the release.
5. The release fires while the fresh reader still needs the metadata.

To close this, each assignment carries a generation:

- The coordinator bumps the generation whenever the binlog reader resets, including a failover that hands nothing back because its assignment was checkpoint-covered.
- The generation is stamped on the meta groups and echoed back in the assembled event. The coordinator releases only when the reported generation matches the current one, so a stale event with an older generation is ignored.
- A split small enough to ship inline never requests meta groups, so its reader reports a sentinel the coordinator always accepts. That is safe because an inline reader never needs the coordinator's metadata.
- The coordinator learns which subtask holds the binlog split from the assignment itself and ignores a stale `BinlogSplitAssignedEvent` from a failed attempt, so reset detection can't be pointed at the wrong subtask.
- The generation lives entirely on the transient source events.

## Parallelism

The release logic lives only in the singleton split enumerator, and one binlog split is assigned to one reader, so there are no per-subtask copies to reconcile. The one parallelism-relevant case is failover of the binlog-holding reader, already covered by the generation check and the reset handling above.

## Tests

- **MySqlSnapshotMetadataReleaseTest** (new): release only after the assembled event and a completed checkpoint; no release without the event, before the checkpoint, or when the option is off; both-options, released-restore-with-newly-added-table, and released-restore-with-the-option-off fail fast; add-back and empty add-back invalidate the assembly; stale old-generation events and a stale `BinlogSplitAssignedEvent` are ignored; inline splits still release after a failover; and a released-state restore does not re-discover tables.
- **PendingSplitsStateSerializerTest**: round-trips the released ("light") state and confirms the persisted flag, empty maps, and `isBinlogSplitAssigned` survive under version 6. It also checks that a default-off serializer keeps writing the byte-exact version-5 format and that a version-5 checkpoint restores with the flag false.
- **MySqlSourceReaderTest** / **MySqlSourceEnumeratorTest**: the reader emits the assembled event with the right generation (or inline sentinel), and the event is a safe no-op for a binlog-only assigner.
- Existing **MySqlSourceITCase**: TaskManager and JobManager failover in the binlog phase with release enabled, plus multi-table reads at higher parallelism, still pass with no data loss.

## Notes

The common path is unchanged until the source is well into the binlog phase, and the option is off by default, so existing jobs behave exactly as before.

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
@@ -21,15 +21,19 @@ import org.apache.flink.api.connector.source.SplitEnumeratorContext;
 import org.apache.flink.cdc.connectors.mysql.source.assigners.state.HybridPendingSplitsState;
 import org.apache.flink.cdc.connectors.mysql.source.assigners.state.PendingSplitsState;
 import org.apache.flink.cdc.connectors.mysql.source.config.MySqlSourceConfig;
+import org.apache.flink.cdc.connectors.mysql.source.events.BinlogSplitMetaAssembledEvent;
 import org.apache.flink.cdc.connectors.mysql.source.offset.BinlogOffset;
 import org.apache.flink.cdc.connectors.mysql.source.split.FinishedSnapshotSplitInfo;
 import org.apache.flink.cdc.connectors.mysql.source.split.MySqlBinlogSplit;
 import org.apache.flink.cdc.connectors.mysql.source.split.MySqlSchemalessSnapshotSplit;
 import org.apache.flink.cdc.connectors.mysql.source.split.MySqlSplit;
+import org.apache.flink.util.Preconditions;
 
 import io.debezium.relational.TableId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -51,6 +55,20 @@ public class MySqlHybridSplitAssigner implements MySqlSplitAssigner {
     private final MySqlSourceConfig sourceConfig;
 
     private boolean isBinlogSplitAssigned;
+
+    /** Whether the reader has reported holding the complete binlog split. */
+    private boolean binlogSplitMetaAssembled;
+
+    /** Checkpoint at which the metadata release was scheduled; released once it completes. */
+    @Nullable private Long checkpointIdToReleaseMeta;
+
+    /**
+     * Generation of the current binlog split assignment, bumped on every add-back. It rides on the
+     * meta groups served to the reader and comes back in the assembled event, so a stale event from
+     * a failed attempt (older generation) can't arm the release for a fresh reader. Not
+     * checkpointed: it resets to 0 on restore, which is fine because the reader re-learns it.
+     */
+    private long binlogAssignmentGeneration;
 
     private final MySqlSnapshotSplitAssigner snapshotSplitAssigner;
 
@@ -161,6 +179,12 @@ public class MySqlHybridSplitAssigner implements MySqlSplitAssigner {
             } else {
                 // we don't store the split, but will re-create binlog split later
                 isBinlogSplitAssigned = false;
+                // re-creating the binlog split: the reader must re-assemble and re-report
+                // before the snapshot metadata can be released again. Bumping the generation
+                // invalidates any assembled event still in flight from the failed attempt.
+                binlogSplitMetaAssembled = false;
+                checkpointIdToReleaseMeta = null;
+                binlogAssignmentGeneration++;
             }
         }
         snapshotSplitAssigner.addSplits(snapshotSplits);
@@ -168,6 +192,16 @@ public class MySqlHybridSplitAssigner implements MySqlSplitAssigner {
 
     @Override
     public PendingSplitsState snapshotState(long checkpointId) {
+        // Schedule releasing the snapshot metadata once the binlog split is assigned and the reader
+        // has assembled it; the release happens in notifyCheckpointComplete. Skipped when
+        // newly-added-table scan is on, since that flow may still need the metadata.
+        if (isBinlogSplitAssigned
+                && binlogSplitMetaAssembled
+                && checkpointIdToReleaseMeta == null
+                && !snapshotSplitAssigner.isSnapshotMetaReleased()
+                && !sourceConfig.isScanNewlyAddedTableEnabled()) {
+            checkpointIdToReleaseMeta = checkpointId;
+        }
         return new HybridPendingSplitsState(
                 snapshotSplitAssigner.snapshotState(checkpointId), isBinlogSplitAssigned);
     }
@@ -175,6 +209,45 @@ public class MySqlHybridSplitAssigner implements MySqlSplitAssigner {
     @Override
     public void notifyCheckpointComplete(long checkpointId) {
         snapshotSplitAssigner.notifyCheckpointComplete(checkpointId);
+        // Release the snapshot metadata only after the checkpoint covering the binlog split
+        // assignment completes. Doing it here (not in snapshotState) keeps the assignment
+        // checkpoint-covered, so addSplitsBack can never return the split to an emptied assigner.
+        if (checkpointIdToReleaseMeta != null
+                && checkpointId >= checkpointIdToReleaseMeta
+                && !snapshotSplitAssigner.isSnapshotMetaReleased()) {
+            snapshotSplitAssigner.releaseSnapshotMetadata();
+        }
+    }
+
+    /**
+     * Marks that the reader holds the complete binlog split, arming the metadata release. A
+     * reported generation that is neither the current one nor the inline sentinel is ignored, which
+     * drops a stale event from a reader attempt that an add-back has already superseded.
+     */
+    public void onBinlogSplitMetaAssembled(long reportedGeneration) {
+        // An inline reader requests no meta groups, so it reports COMPLETE_WITHOUT_META_GENERATION;
+        // honor that regardless, since it never needs the coordinator's metadata. Otherwise require
+        // an exact match, which drops a stale event from a failed attempt superseded by an
+        // add-back.
+        if (reportedGeneration != BinlogSplitMetaAssembledEvent.COMPLETE_WITHOUT_META_GENERATION
+                && reportedGeneration != binlogAssignmentGeneration) {
+            LOG.info(
+                    "Ignoring binlog split assembled event for generation {}; current generation is {}.",
+                    reportedGeneration,
+                    binlogAssignmentGeneration);
+            return;
+        }
+        this.binlogSplitMetaAssembled = true;
+    }
+
+    /** Returns the current binlog split assignment generation. */
+    public long getBinlogAssignmentGeneration() {
+        return binlogAssignmentGeneration;
+    }
+
+    /** Returns whether the snapshot split metadata has been released. */
+    public boolean isSnapshotMetaReleased() {
+        return snapshotSplitAssigner.isSnapshotMetaReleased();
     }
 
     @Override
@@ -205,6 +278,10 @@ public class MySqlHybridSplitAssigner implements MySqlSplitAssigner {
     // --------------------------------------------------------------------------------------------
 
     private MySqlBinlogSplit createBinlogSplit() {
+        Preconditions.checkState(
+                !snapshotSplitAssigner.isSnapshotMetaReleased(),
+                "Snapshot metadata was already released; the binlog split must not be re-created "
+                        + "after that.");
         final List<MySqlSchemalessSnapshotSplit> assignedSnapshotSplit =
                 new ArrayList<>(snapshotSplitAssigner.getAssignedSplits().values());
 
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
@@ -115,6 +115,21 @@ public class MySqlHybridSplitAssigner implements MySqlSplitAssigner {
         this.snapshotSplitAssigner = snapshotSplitAssigner;
         this.isBinlogSplitAssigned = isBinlogSplitAssigned;
         this.splitMetaGroupSize = splitMetaGroupSize;
+        // Fail fast on an incompatible restore. A job that released its snapshot metadata cannot
+        // later enable scan.newly-added-table, because the splits and schemas that flow needs are
+        // gone. Reject it loudly instead of silently building an inconsistent binlog split.
+        if (sourceConfig.isScanNewlyAddedTableEnabled()
+                && isBinlogSplitAssigned
+                && snapshotSplitAssigner.isSnapshotMetaReleased()) {
+            throw new IllegalStateException(
+                    "scan.newly-added-table.enabled cannot be turned on for a job that previously "
+                            + "released its snapshot split metadata "
+                            + "(scan.incremental.snapshot.metadata.release.enabled=true). The "
+                            + "assigned splits, finished offsets and table schemas needed to capture "
+                            + "newly added tables are no longer in state. Start the job from a fresh "
+                            + "state, or keep metadata release disabled if newly-added-table scanning "
+                            + "is required.");
+        }
     }
 
     @Override
@@ -192,13 +207,14 @@ public class MySqlHybridSplitAssigner implements MySqlSplitAssigner {
 
     @Override
     public PendingSplitsState snapshotState(long checkpointId) {
-        // Schedule releasing the snapshot metadata once the binlog split is assigned and the reader
-        // has assembled it; the release happens in notifyCheckpointComplete. Skipped when
-        // newly-added-table scan is on, since that flow may still need the metadata.
+        // Schedule the metadata release once the binlog split is assigned and the reader has
+        // assembled it; it happens in notifyCheckpointComplete. Gated behind the opt-in, and
+        // skipped when newly-added-table scan is on, since that flow may still need the metadata.
         if (isBinlogSplitAssigned
                 && binlogSplitMetaAssembled
                 && checkpointIdToReleaseMeta == null
                 && !snapshotSplitAssigner.isSnapshotMetaReleased()
+                && sourceConfig.isReleaseSnapshotMetadataEnabled()
                 && !sourceConfig.isScanNewlyAddedTableEnabled()) {
             checkpointIdToReleaseMeta = checkpointId;
         }
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
@@ -68,6 +68,7 @@ public class MySqlSourceConfigFactory implements Serializable {
     private boolean includeHeartbeatEvents = false;
     private boolean includeTransactionMetadataEvents = false;
     private boolean scanNewlyAddedTableEnabled = false;
+    private boolean releaseSnapshotMetadataEnabled = false;
     private boolean closeIdleReaders = false;
     private Properties jdbcProperties;
     private Duration heartbeatInterval = MySqlSourceOptions.HEARTBEAT_INTERVAL.defaultValue();
@@ -258,6 +259,17 @@ public class MySqlSourceConfigFactory implements Serializable {
         return this;
     }
 
+    /**
+     * Whether the source coordinator should release the snapshot split metadata after entering the
+     * binlog phase to reduce JobManager memory. Incompatible with {@link
+     * #scanNewlyAddedTableEnabled(boolean)}; disabled by default.
+     */
+    public MySqlSourceConfigFactory releaseSnapshotMetadataEnabled(
+            boolean releaseSnapshotMetadataEnabled) {
+        this.releaseSnapshotMetadataEnabled = releaseSnapshotMetadataEnabled;
+        return this;
+    }
+
     /** Custom properties that will overwrite the default JDBC connection URL. */
     public MySqlSourceConfigFactory jdbcProperties(Properties jdbcProperties) {
         this.jdbcProperties = jdbcProperties;
@@ -444,6 +456,7 @@ public class MySqlSourceConfigFactory implements Serializable {
                 parseOnLineSchemaChanges,
                 treatTinyInt1AsBoolean,
                 useLegacyJsonFormat,
-                assignUnboundedChunkFirst);
+                assignUnboundedChunkFirst,
+                releaseSnapshotMetadataEnabled);
     }
 }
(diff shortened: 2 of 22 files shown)


spoorthibasu:
Hi @lvyanquan , this PR is ready for review when you have time. It fixes a JobManager OOM on large tables by releasing the snapshot split metadata once the source is in the binlog phase. The description walks through the reasoning and the failover cases. CI is waiting on approval, could you please trigger it when you get a chance? Thanks!

spoorthibasu:
Hi @lvyanquan , could you please approve the workflow run on this PR? Happy to address anything that comes back.

spoorthibasu:
Hi @lvyanquan , the only failing check is OceanBaseE2eITCase, and it errored in setup because its OceanBase test container timed out on startup ("Timed out waiting for ... boot success!", container exited with code 1). This PR doesn't touch any OceanBase code, and the same E2E suite passed on the other Flink version and parallelism combinations, so it looks like an unrelated container flake. Could you please re-run the failed job when you get a chance?

spoorthibasu:
Hi @lvyanquan , just checking in. This PR is ready for review when you have time. The only red check is Pipeline E2E Tests 2.x (4-Parallelism): its OceanBaseE2eITCase test errored out on a container boot timeout unrelated to this PR, and the same suite passed on the other version and parallelism combinations.

spoorthibasu:
Hi @ruanhang1993, @yuxiqian, would either of you have time to review this one when you get a chance? It's a MySQL source change that releases the snapshot split metadata once the source enters the binlog phase, to fix a JobManager OOM on large tables (around 300K splits for a 2.5B-row table). The description covers the failover cases. CI is green except for the Pipeline E2E Tests 2.x (4-Parallelism) run, where the OceanBaseE2eITCase test errored out on an unrelated container boot timeout. Happy to rebase or walk through anything. Thanks!

spoorthibasu:
Hi @leonardBang, you assigned me FLINK-39775 a while back, and PR #4418 has been green and ready for a few weeks now (the only red is an unrelated OceanBase container boot timeout). It fixes the JobManager OOM on large MySQL tables by releasing the snapshot-split metadata from the coordinator once the source enters the binlog phase. Would you have time to take a look when you get a chance, or point me to the right reviewer? Happy to rebase if that helps. Thanks!

cc @loserwang1024, in case you have time to weigh in, given your work on the incremental-source split handling.

spoorthibasu:
@ruanhang1993 gentle reminder on this one, it's ready for review when you have time. The only red check is an unrelated OceanBase container flake. Happy to answer questions or rebase. cc @leonardBang, thanks again for routing it.

lvyanquan:
Hi @spoorthibasu,

Thanks for your continued follow-up. I was quite busy for a while, but I have some time now and can start reviewing this PR.

lvyanquan:
One thing I would like to clarify is the exact OOM scenario this PR is targeting.

My understanding is that the coordinator state mainly keeps growing during the snapshot phase, as more snapshot splits are assigned and finished. Once the job has successfully entered the binlog/incremental phase, the number of finished snapshot splits should be stable. If a checkpoint containing the full snapshot metadata has already completed successfully, then the coordinator state size of later checkpoints should be roughly the same.

So I wonder whether the reported OOM happened while the snapshot metadata was still growing, or after the job had already entered the binlog phase. In the latter case, I agree that keeping and repeatedly serializing such a large state is still wasteful and can keep memory pressure high, but it would be helpful to clarify why it still leads to OOM after the state size has become stable, e.g. checkpoint overlap, delayed GC, direct memory/container RSS pressure, etc.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
```diff
@@ -161,20 +179,75 @@ public void addSplits(Collection<MySqlSplit> splits) {
...
     @Override
     public PendingSplitsState snapshotState(long checkpointId) {
+        // Schedule releasing the snapshot metadata once the binlog split is assigned and the reader
+        // has assembled it; the release happens in notifyCheckpointComplete. Skipped when
+        // newly-added-table scan is on, since that flow may still need the metadata.
+        if (isBinlogSplitAssigned
+                && binlogSplitMetaAssembled
+                && checkpointIdToReleaseMeta == null
+                && !snapshotSplitAssigner.isSnapshotMetaReleased()
+                && !sourceConfig.isScanNewlyAddedTableEnabled()) {
```
lvyanquan, line 218: This flag is not part of the checkpointed enumerator state. For a job that has already started with `scan.newly-added-table.enabled = false`, once this PR releases the snapshot metadata, enabling `scan.newly-added-table.enabled` later for the same job becomes incompatible: the old assignedSplits, splitFinishedOffsets, and tableSchemas are already gone, while the newly-added-table flow may still need them to update the binlog split.

This changes the previous behavior, where the metadata was always kept in state. Now requiring jobs that need to discover newly added tables to configure `scan.newly-added-table.enabled = true` from the very beginning is a functional regression. This would be hard for users to accept.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/enumerator/MySqlSourceEnumerator.java
lvyanquan, line 332: I think there is another unsafe case around stale `BinlogSplitMetaRequestEvent`.

This PR handles stale `BinlogSplitMetaAssembledEvent` by using the assignment generation, but `BinlogSplitMetaRequestEvent` is still not tied to any reader attempt. A request sent by a failed reader attempt may arrive after the current reader has already assembled the metadata and the checkpoint completing the release has finished.

After the release, `releaseSnapshotMetadata()` clears `assignedSplits`, `splitFinishedOffsets`, and `tableSchemas`, and `MySqlSourceEnumerator#notifyCheckpointComplete()` also sets `binlogSplitMeta = null`. If a stale `BinlogSplitMetaRequestEvent` arrives after that, `sendBinlogMeta()` will try to rebuild `binlogSplitMeta` from `splitAssigner.getFinishedSplitInfos()`. Since the assigner metadata has already been cleared, this can return an empty list and throw `FlinkRuntimeException`.

This means a stale metadata request from a failed reader attempt can fail the enumerator/job after the metadata has been released. Could we explicitly handle this case by ignoring/logging metadata requests after the snapshot metadata has been released, instead of trying to rebuild the metadata and failing the job?

spoorthibasu:
> So I wonder whether the reported OOM happened while the snapshot metadata was still growing, or after the job had already entered the binlog phase. In the latter case, I agree that keeping and repeatedly serializing such a large state is still wasteful and can keep memory pressure high, but it would be helpful to clarify why it still leads to OOM after the state size has become stable, e.g. checkpoint overlap, delayed GC, direct memory/container RSS pressure, etc.

It's the second case you described, after the job has entered the binlog phase rather than while the snapshot state is still growing. The finished-split count is stable by then, so later checkpoints carry a roughly constant state size.

It's the direct memory / container RSS pressure you listed. Even with the state size stable, every post-snapshot checkpoint still re-serializes and transfers the full metadata, around 300K finished splits for a 2.5B-row table at the default chunk size. That can create several transient copies in a short window, the serialized state bytes plus the off-heap buffers used during transfer. Since Flink doesn't set a direct-memory limit for the JobManager by default (`jobmanager.memory.enable-jvm-direct-memory-limit=false`), that off-heap pressure isn't bounded early by a JVM-level direct-memory error and can instead build up in process RSS until the container is OOMKilled, even while the heap looks fine. This was reported on CDC 3.5.0 / Flink 1.20.x with a Paimon sink.

So it's exactly the repeated serialization of the large stable state you flagged as wasteful. Happy to add a note on this to the PR description if that would help.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
spoorthibasu: Agreed, this shouldn't change the default behavior for existing jobs. I've made the whole optimization opt-in behind a new option `scan.incremental.snapshot.metadata.release.enabled` (default false). With it off, nothing is released, so the metadata stays in state and `scan.newly-added-table.enabled` can be turned on later just like today. Only a job that explicitly enables the option releases the metadata.

For the cross-restart case you're pointing at, a job that released and then restarts with `scan.newly-added-table.enabled=true`, I added a fail-fast on restore. If a released state is restored together with newly-added-table enabled, the assigner throws a clear error instead of silently building an inconsistent binlog split. That released state is detected from the existing checkpoint fields (finished snapshot, emptied assigned-splits map, tables already processed), so there is no checkpoint or serializer format change and old checkpoints still restore. Both paths are tested, `testNoReleaseWhenReleaseMetadataDisabled` for the default-off behavior and `testFailFastWhenNewlyAddedEnabledOnReleasedRestore` for the cross-restart guard.

Since this is opt-in, a user hitting the OOM might not realize the flag exists. If you think it's worth it, I can add a warning log when the coordinator is still holding a large snapshot state after the binlog phase, so it surfaces in the JobManager logs where they'd be looking. Happy to add it or leave it out.

Does this direction and the option name look right to you?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/enumerator/MySqlSourceEnumerator.java
spoorthibasu: Agreed, a stale request shouldn't be able to fail the job. I added a guard in `sendBinlogMeta`. Once the metadata has been released, a `BinlogSplitMetaRequestEvent` can only come from a failed reader attempt, since a live reader that assembled the split recovers it from its own checkpointed state and never re-requests. The enumerator now logs and ignores the request instead of rebuilding from the emptied assigner and throwing. I also made the released state survive a restart, reconstructed from the checkpointed fields, so the guard still holds after restore, not only within the run that released. Both cases are tested, `testStaleBinlogMetaRequestAfterReleaseIsIgnored` within the run and `testStaleBinlogMetaRequestAfterLightRestoreIsIgnored` after restore.

spoorthibasu:
Thanks for the review @lvyanquan! I've replied to your comments above, could you take another look when you have time? CI just needs your approval to run on the fork.

spoorthibasu:
@lvyanquan these failures look unrelated to this change. mysql-source unit tests and Source E2E are green. The pipeline E2E reds are both at 4-Parallelism, which is currently flaky on master and passes here at 1-Parallelism (the 2.x run failed on `MySqlToDorisE2eITCase`, which also fails on master). The other is `SqlServerConnectorITCase.testAllTypes`, in the SqlServer connector this PR doesn't touch. Happy to rebase, or a re-run may be simplest.

spoorthibasu:
@lvyanquan could you please re-run the failed jobs when you have time? They're unrelated flakes, as noted above. Thanks!

spoorthibasu:
Thanks for the re-run @lvyanquan! CI is all green now, so this is ready for another look whenever you have time.

spoorthibasu:
Hi @leonardBang @lvyanquan, could you help review this when you have a chance? The two points from the July review are addressed and it's green and conflict-free. Thanks a lot!

lvyanquan (commented):
This was accidentally generated by the agent. I will review it again and leave a new comment.

~~Thanks, the two correctness concerns from the previous review are addressed. The assignment-generation check, checkpoint-covered release, and stale meta-request guard look reasonable.~~

~~I found one remaining documentation gap before approval:~~

~~1. The new user-facing option is only documented in the English pages. Please add the corresponding entries to:~~
   ~~- `docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md`~~
   ~~- `docs/content.zh/docs/connectors/pipeline-connectors/mysql.md`~~

~~2. Please update the PR description. It still says **“No new configuration”** and describes the release as if it happens automatically, but the latest commit makes it opt-in via `scan.incremental.snapshot.metadata.release.enabled` with default `false`.~~

~~3. It would also be useful for the option docs to state that metadata is released only after a successful checkpoint covering the assembled binlog split. If checkpointing is disabled or no checkpoint completes, the coordinator must retain the metadata.~~

spoorthibasu:
Hi @lvyanquan, I've added the Chinese doc entries (mysql-cdc.md and mysql.md) and a note that the metadata is only released after a successful checkpoint. The Chinese is a best-effort translation since I don't read it myself, so please correct the wording where needed. This commit is docs-only on top of the previous green run. Could you please take another look and approve the CI run when you have a chance? cc @leonardBang

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
lvyanquan, line 428: Could we fail fast when both options are enabled instead of silently disabling metadata release here?
```
Preconditions.checkState(
        !(scanNewlyAddedTableEnabled && releaseSnapshotMetadataEnabled),
        "scan.incremental.snapshot.metadata.release.enabled and "
                + "scan.newly-added-table.enabled cannot both be enabled.");
```

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
```diff
@@ -161,20 +194,76 @@ public void addSplits(Collection<MySqlSplit> splits) {
             } else {
                 // we don't store the split, but will re-create binlog split later
                 isBinlogSplitAssigned = false;
+                // re-creating the binlog split: the reader must re-assemble and re-report
+                // before the snapshot metadata can be released again. Bumping the generation
+                // invalidates any assembled event still in flight from the failed attempt.
+                binlogSplitMetaAssembled = false;
+                checkpointIdToReleaseMeta = null;
+                binlogAssignmentGeneration++;
```
lvyanquan, line 202: This only handles cases where the binlog split is actually added back. If its assignment was already checkpointed, a reader reset may call addSplitsBack() with an empty list, so the generation is not incremented even though the reader
may restore an incomplete metadata state. A delayed assembled event from the failed attempt could then trigger an unsafe release.

Could we also reset the assembled state and increment the generation whenever the binlog reader is reset, including empty add-back cases, and add a test for this scenario?

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/config/MySqlSourceConfigFactory.java
spoorthibasu: `createConfig` now fails fast when both options are enabled, rather than silently dropping the release. Test: `testFailFastWhenBothOptionsEnabled`.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/MySqlHybridSplitAssigner.java
spoorthibasu: The invalidation now runs whenever the binlog reader resets (via `onBinlogReaderReset` in `addSplitsBack`), so it also covers the empty add-back that happens when the assignment is already checkpointed. Test: `testEmptyAddBackFromBinlogReaderInvalidatesStaleAssembledEvent`.

spoorthibasu:
Both comments are addressed inline. Two related changes in the same area:

- The coordinator's record of which subtask holds the binlog split (used by the reset detection) is now only set from `BinlogSplitAssignedEvent` when it is unknown, and ignores an event naming a different subtask, so a stale event from a failed attempt can't point it at the wrong reader. Test: `testStaleBinlogSplitAssignedEventDoesNotCorruptHolder`.
- The released flag is now persisted in the checkpoint (serializer v6, older checkpoints default to not-released) rather than reconstructed on restore from the other state fields, which matches how the serializer already versions its other fields. Happy to keep it reconstructed if you'd rather avoid the version bump.

Docs (EN and ZH) and the PR description are updated. The Chinese is a best-effort translation of the English, so please feel free to correct the wording.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializer.java
```diff
@@ -45,8 +45,7 @@
  */
 public class PendingSplitsStateSerializer implements SimpleVersionedSerializer<PendingSplitsState> {
 
-    // TODO: need proper implementation of the new version
-    private static final int VERSION = 5;
+    private static final int VERSION = 6;
```
lvyanquan: This unconditional version bump changes the checkpoint format for every MySQL source job, even when
`scan.incremental.snapshot.metadata.release.enabled`
 is false. As a result, a default-off job cannot roll back to an older connector after taking a checkpoint
with the new version.

Since this optimization is opt-in, jobs that do not enable it should keep writing the existing v5 format without the released flag. Could we make the serializer version and flag serialization conditional on the option, and only write v6 when metadata release is enabled?

If a checkpoint already contains `snapshotMetaReleased=true`, restoring it with
the option disabled should fail fast or otherwise continue writing v6, since downgrading that state to v5 would lose the released-state marker.

lvyanquan (commented):
Hi @spoorthibasu. I have no further concerns about the flow when `scan.incremental.snapshot.metadata.release.enabled=true`. However, I found a major version compatibility issue: the state format of jobs that do not enable this option should not be changed.

lvyanquan:
I also apologize that this PR has gone through several rounds of review. Since the MySQL connector is very widely used, we need to review changes carefully to avoid introducing any potential risks when upgrading jobs that were previously running well. That said, I believe this PR is now well prepared and very close to being ready for merge.

flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/org/apache/flink/cdc/connectors/mysql/source/assigners/state/PendingSplitsStateSerializer.java
spoorthibasu: The serializer now writes v6 only when the option is enabled, so a default-off job keeps writing v5 and stays restorable by an older build. Restoring a released state with the option off fails fast instead of dropping the marker. Tests: `testReleaseDisabledJobKeepsWritingV5Format` and `testFailFastWhenReleaseDisabledOnReleasedRestore`.

spoorthibasu:
Thanks for the careful review @lvyanquan, I appreciate you taking the time on this one, and I've addressed it in the latest commit.

lvyanquan (approved):
LGTM.

lvyanquan:
@leonardBang @ruanhang1993 CC’ing you to see if you have any further feedback. If there are no additional comments, I’ll merge this in a few days.

leonardBang (approved):
Thanks @spoorthibasu for the great work and @lvyanquan for the detailed review!

+1 for the PR overall.

After carefully reviewing the state handling, checkpoint/restore behavior, and compatibility implications, I think the current design and implementation are safe.  

Also sorry that the review took a bit longer than usual @spoorthibasu. I have to explain to that MySQL CDC is arguably the most important connector in Flink CDC, and changes involving checkpoint state and compatibility deserve extra caution. Thanks for the solid work and for iterating on the details.

spoorthibasu:
Thanks @leonardBang, I appreciate the kind words and the time you took to review this carefully.

PR #4423, opened 2026-06-03
[FLINK-39777] Support configurable HashFunction strategies in PrePartitionOperator

haruki-830:
#### Summary

This commit adds configurable partitioning strategy support to Flink CDC pipelines, enabling users to switch to table-id hashing for small tables via YAML configuration, reducing unnecessary primary-key hashing overhead.

#### Key Changes

1. New HashFunctionStrategy Enum
- Introduced HashFunctionStrategy enum with two options: PRIMARY_KEY (hash by TableId + primary keys) and TABLE_ID (hash by TableId only).
- Designed with @PublicEvolving annotation, allowing future strategies like ROUND_ROBIN or COLUMNS.

2. New TableIdHashFunctionProvider
- Added TableIdHashFunctionProvider that computes hash based solely on TableId.
- Uses singleton pattern for HashFunction since it is stateless.
- Suitable for small tables, or tables with no/changing primary keys.

3. Pipeline Configuration Option
- Added pipeline option "partitioning.strategy" to allow switching strategies via YAML.
- When unset, behavior is identical to current, ensuring full backward compatibility.

4. Comprehensive Testing
- Added TableIdHashFunctionProviderTest with 7 test.
- Added PrePartitionOperatorTest case verifying TABLE_ID strategy.
- Added FlinkPipelineComposerTest configurations for PRIMARY_KEY and TABLE_ID strategy validation.

##### Route same-table events to a single subtask
```
pipeline:
  name: my-cdc-job
  partitioning.strategy: TABLE_ID
```

##### Force load-balanced distribution by primary keys
```
pipeline:
  name: my-cdc-job
  partitioning.strategy: PRIMARY_KEY
```

  #### JIRA Reference

  [https://issues.apache.org/jira/browse/FLINK-39777](url)


haruki-830:
@lvyanquan could you please help review this PR?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/pipeline/PipelineOptions.java
```diff
@@ -136,5 +136,24 @@ public class PipelineOptions {
                     .withDescription(
                             "The timeout time for SchemaOperator to wait downstream SchemaChangeEvent applying finished, the default value is 3 minutes.");
 
+    public static final ConfigOption<HashFunctionStrategy> PIPELINE_PARTITIONING_STRATEGY =
+            ConfigOptions.key("partitioning.strategy")
```
lvyanquan: Please update the [document](https://nightlies.apache.org/flink/flink-cdc-docs-release-3.6/docs/core-concept/data-pipeline/#pipeline-configurations) to guide users on how to use it.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/pipeline/PipelineOptions.java
```diff
@@ -136,5 +136,24 @@ public class PipelineOptions {
...
+                    .enumType(HashFunctionStrategy.class)
+                    .noDefaultValue()
```
lvyanquan: The current option uses `null` (via `.noDefaultValue()`) to mean "use the sink's own provider", which is an implicit convention. I'd suggest adding a third enum value `SINK_DEFINED` in `HashFunctionStrategy` and making it the default.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/TableIdHashFunctionProvider.java
```diff
@@ -0,0 +1,68 @@
...
+        @Override
+        public int hashcode(DataChangeEvent event) {
+            List<Object> objectsToHash = new ArrayList<>();
+            TableId tableId = event.tableId();
+            Optional.ofNullable(tableId.getNamespace()).ifPresent(objectsToHash::add);
+            Optional.ofNullable(tableId.getSchemaName()).ifPresent(objectsToHash::add);
+            objectsToHash.add(tableId.getTableName());
+
+            // Calculate hash
+            return (Objects.hash(objectsToHash.toArray()) * 31) & 0x7FFFFFFF;
```
lvyanquan: TableIdHashFunctionProvider.java:55-65 allocates an ArrayList and calls `Objects.hash(...)` on **every event**, but the result depends only on `event.tableId()` — which is already passed to `getHashFunction(tableId, schema)`.

It would be better to drop the `INSTANCE` singleton and have the provider new a `TableIdHashFunction(tableId)` per call, computing the hash once in the constructor and just returning the cached int from `hashcode(event)`.

flink-cdc-common/src/test/java/org/apache/flink/cdc/common/sink/TableIdHashFunctionProviderTest.java
```diff
@@ -0,0 +1,172 @@
...
+        HashFunction<DataChangeEvent> hashFunction2 = provider.getHashFunction(TABLE_B, SCHEMA);
+
+        // Both should return the same singleton instance
+        assertThat(hashFunction1).isSameAs(hashFunction2);
+    }
+
+    private DataChangeEvent createInsertEvent(TableId tableId, int id, String name) {
+        return DataChangeEvent.insertEvent(
+                tableId,
+                org.apache.flink.cdc.common.data.GenericRecordData.of(
```
lvyanquan: Add an import instead of fully-qualified
  name for consistency with the rest of the codebase.

lvyanquan (commented):
Thanks @haruki-830 for the feature, left some comments.

lvyanquan:
Please also update https://github.com/apache/flink-cdc/blob/master/flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full.yaml#L56 / https://github.com/apache/flink-cdc/blob/master/flink-cdc-cli/src/test/resources/definitions/pipeline-definition-full-with-repsym.yaml#L59 and related tests.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/TableIdHashFunctionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+        private static final long serialVersionUID = 1L;
+
+        private final int cachedHash;
+
+        TableIdHashFunction(TableId tableId) {
+            List<Object> objectsToHash = new ArrayList<>();
+            Optional.ofNullable(tableId.getNamespace()).ifPresent(objectsToHash::add);
+            Optional.ofNullable(tableId.getSchemaName()).ifPresent(objectsToHash::add);
+            objectsToHash.add(tableId.getTableName());
+            this.cachedHash = (Objects.hash(objectsToHash.toArray()) * 31) & 0x7FFFFFFF;
```
lvyanquan: Using TableId#hashCode directly instead of Objects.hash(objectsToHash.toArray()) would be simpler.

lvyanquan (approved):
+1.

PR #4424, opened 2026-06-03
[FLINK-39832] Fix Oracle NUMBER(p,0) with p>=19 incorrectly mapped to BIGINT

ironman-star:
Jira: https://issues.apache.org/jira/browse/FLINK-39832

  OracleTypeUtils.fromDbzColumn mapped NUMBER(p,0) to BIGINT for all precisions, but Debezium encodes p>=19 as DECIMAL(BYTES). The mismatch caused the sink to read a BinaryRecordData pointer as a long, producing a constant garbage PK (171798691841) and collapsing all rows on upsert.

  Fix: return DECIMAL(p,0) when p>=19, aligned with Debezium runtime behavior.

  Tested with Oracle XE 11.2, NUMBER(19,0) PK, including 19-digit large IDs (9000000000000000001). Snapshot and incremental (INSERT/UPDATE/DELETE) all verified correct.


ironman-star:
@lvyanquan Would you please review this code?

lvyanquan:

lvyanquan:
It would be better if you could add related tests.

ironman-star:
@lvyanquan Already added the tests.

ironman-star:
@lvyanquan I've already fix the OracleMetadataAccessorITCase tests and ask for rerun the CI test again.

ironman-star:
@lvyanquan I only modified the code related to the Oracle connector, but I don't know why the PostgreSQL tests are failing.

lvyanquan (approved):
+1.

PR #4425, opened 2026-06-03
[FLINK-39833][pipeline][starrocks] Escape special characters in column comment and default value when building DDL

linguoxuan:
When the StarRocks pipeline connector builds DDL statements (CREATE TABLE / ALTER TABLE ADD COLUMN), special characters in column comments and default values (such as backslash `\`, double quote `"`, newline `\n`, carriage return `\r`) are not properly escaped.

This causes the generated DDL to be syntactically invalid, leading to schema synchronization failures at runtime.

### Root Cause

The column comment and default value strings are directly interpolated into the DDL template `DEFAULT "..."` and `COMMENT "..."` without any escaping.

### Fix

Introduce a utility method `StarRocksUtils.escapeSqlStringLiteral()` that escapes:
- `\` → `\\`
- `"` → `""`
- `\n` / `\r` → space

Apply this escaping in:
- `StarRocksEnrichedCatalog` (CREATE TABLE DDL building)
- `StarRocksMetadataApplier` (ALTER TABLE ADD COLUMN)
- `StarRocksUtils` (initial schema conversion and table comment)

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
@@ -91,10 +91,12 @@ public class StarRocksUtils {
                     new StarRocksColumn.Builder()
                             .setColumnName(column.getName())
                             .setOrdinalPosition(i)
-                            .setColumnComment(column.getComment())
+                            .setColumnComment(escapeSqlStringLiteral(column.getComment()))
                             .setDefaultValue(
-                                    convertInvalidTimestampDefaultValue(
-                                            column.getDefaultValueExpression(), column.getType()));
+                                    escapeSqlStringLiteral(
+                                            convertInvalidTimestampDefaultValue(
+                                                    column.getDefaultValueExpression(),
+                                                    column.getType())));
             toStarRocksDataType(
                     column,
                     i < primaryKeyCount,
@@ -529,5 +531,32 @@ public class StarRocksUtils {
         }
 
         return defaultValue;
+    }
+
+    public static String escapeSqlStringLiteral(String value) {
+        if (value == null) {
+            return null;
+        }
+        StringBuilder sb = new StringBuilder(value.length() + 8);
+        for (int i = 0; i < value.length(); i++) {
+            char c = value.charAt(i);
+            switch (c) {
+                case '\\':
+                    sb.append("\\\\");
+                    break;
+                case '"':
+                    sb.append("\"\"");
+                    break;
+                case '\n':
+                    sb.append("\\n");
+                    break;
+                case '\r':
+                    sb.append("\\r");
+                    break;
+                default:
+                    sb.append(c);
+            }
+        }
+        return sb.toString();
     }
 }
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/test/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksMetadataApplierITCase.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/test/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksMetadataApplierITCase.java
@@ -617,6 +617,145 @@ class StarRocksMetadataApplierITCase extends StarRocksSinkTestBase {
         assertEqualsInOrder(expected, actual);
     }
 
+    @Test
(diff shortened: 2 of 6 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
```diff
@@ -492,4 +494,29 @@ public static String convertInvalidTimestampDefaultValue(
...
+            switch (c) {
+                case '\\':
+                    sb.append("\\\\");
+                    break;
+                case '"':
+                    sb.append("\"\"");
+                    break;
+                case '\n':
+                case '\r':
+                    sb.append(' ');
```
haruki-830: I think this may change the semantics of DEFAULT values. This helper is used for both COMMENT and DEFAULT, but replacing LF/CR with spaces means a source default like `line1\nline2` will become `line1 line2` in StarRocks.
StarRocks string literal parsing already supports backslash escape sequences.So for DEFAULT values, it seems better to emit escaped SQL literals such as `\\n` / `\\r` rather than replacing them with spaces.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
```diff
@@ -108,7 +110,7 @@ public static StarRocksTable toStarRocksTable(
                         .setTableKeys(schema.primaryKeys())
                         // use primary keys as distribution keys by default
                         .setDistributionKeys(schema.primaryKeys())
-                        .setComment(schema.comment());
+                        .setComment(escapeSqlStringLiteral(schema.comment()));
```
haruki-830: It looks like this escapes table comments, but the current StarRocksCatalog create-table path does not seem to emit table comments in the generated DDL.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/test/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksMetadataApplierITCase.java
```diff
@@ -583,6 +583,145 @@ void testMysqlDefaultTimestampValueConversionInAddColumn() throws Exception {
...
+        runJobWithEvents(Collections.singletonList(new CreateTableEvent(tableId, schema)));
+
+        List<String> actual = inspectTableSchema(tableId);
+
+        List<String> expected =
+                Arrays.asList(
+                        "id | int | NO | true | null",
+                        "col_backslash | varchar(300) | YES | false | default\\value",
+                        "col_quote | varchar(300) | YES | false | default\"value",
+                        "col_newline | varchar(300) | YES | false | default value",
```
haruki-830: The new coverage is helpful. I wonder if the expected value for newline defaults should preserve the original DEFAULT semantics instead of expecting it to become a space. Otherwise the test may lock in behavior where escaping changes the actual default value.

haruki-830 (commented):
It looks like this PR has conflicts with the current master, so it may need a rebase.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
linguoxuan: just revert.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/test/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksMetadataApplierITCase.java
linguoxuan: fix

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-starrocks/src/main/java/org/apache/flink/cdc/connectors/starrocks/sink/StarRocksUtils.java
linguoxuan: fix

PR #4427, opened 2026-06-03
[hotfix] Use explicit UTF-8 charset in coerceToBytes and guard Optional.get() in Kafka serializers

dangzitou:
## What this PR does

This PR fixes two minor but important issues:

### 1. Platform-default charset in `SchemaMergingUtils.coerceToBytes()`

**File:** `flink-cdc-common/.../SchemaMergingUtils.java:617`

`originalField.toString().getBytes()` uses the platform-default charset, which can vary across JVM configurations and operating systems. This causes inconsistent behavior when converting fields to bytes during schema evolution.

**Fix:** Use `StandardCharsets.UTF_8` explicitly.

### 2. Unchecked `Optional.get()` in Kafka serialization schemas

**Files:**
- `CsvSerializationSchema.java:109`
- `JsonSerializationSchema.java:132`

Both files call `schema.getColumn(...).get()` without checking if the Optional is present. If a primary key column is missing from the schema (e.g., after schema evolution), this throws a bare `NoSuchElementException` with no context.

**Fix:** Replace with `orElseThrow()` that includes the missing column name in the error message.

## Testing

These are straightforward defensive improvements:
- The charset fix ensures consistent UTF-8 encoding regardless of JVM locale
- The Optional guard improves error diagnostics without changing happy-path behavior

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
@@ -74,6 +74,7 @@ import org.apache.flink.shaded.guava31.com.google.common.io.BaseEncoding;
 import javax.annotation.Nullable;
 
 import java.math.BigDecimal;
+import java.nio.charset.StandardCharsets;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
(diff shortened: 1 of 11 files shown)


yuxiqian:
Hi @dangzitou, thank you for your contribution!

It looks like this PR addresses 3 independent issues (the UTF-8 fix, Doris, and Kafka). Could you please split them into separate JIRA tickets with one PR each? That makes review easier and keeps the git history clean.

Thanks again!

PR #4428, opened 2026-06-04
[hotfix] Use explicit UTF-8 charset in getBytes() calls across connectors

dangzitou:
## What this PR does

Replaces platform-default `String.getBytes()` with `String.getBytes(StandardCharsets.UTF_8)` across multiple modules to ensure consistent encoding behavior regardless of JVM locale or OS configuration.

### Affected files

- `SchemaMergingUtils.java` + test — core schema coercion
- `DebeziumJsonSerializationSchema.java` — Kafka Debezium JSON default value handling
- `RowDataTiKVEventDeserializationSchemaBase.java` — TiDB source connector
- `BinaryTypeReturningClass.java` / `VarBinaryTypeReturningClass.java` — UDF examples

### Why

`String.getBytes()` without an explicit charset uses the JVM's default charset, which varies across environments (e.g., `US-ASCII` on some minimal Docker images, `GBK` on Chinese Windows). This causes silent data corruption when non-ASCII characters are involved. Using `UTF-8` explicitly makes the behavior deterministic.

## Testing

These are straightforward defensive improvements — the fix ensures consistent UTF-8 encoding regardless of JVM locale.

---

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
@@ -74,6 +74,7 @@ import org.apache.flink.shaded.guava31.com.google.common.io.BaseEncoding;
 import javax.annotation.Nullable;
 
 import java.math.BigDecimal;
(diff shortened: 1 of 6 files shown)


dangzitou:

PR #4429, opened 2026-06-04
[hotfix] Guard Optional.get() in Doris connector schema lookups

dangzitou:
## What this PR does

Replaces bare `Optional.get()` calls with `Optional.orElseThrow()` in the Doris pipeline connector to provide meaningful error messages when columns are missing from the schema.

### Affected files

- `DorisMetadataApplier.java` — column lookup during schema evolution
- `DorisSchemaUtils.java` — partition key column lookup

### Why

`Optional.get()` throws a bare `NoSuchElementException` with no context when the Optional is empty. This makes debugging difficult when a primary key or partition key column is unexpectedly missing (e.g., after schema evolution). Using `orElseThrow()` with a descriptive message including the column name makes failures immediately diagnosable.

## Testing

Behavioral change only in the error path — happy-path behavior is unchanged.

---


dangzitou:

PR #4430, opened 2026-06-04
[hotfix] Guard Optional.get() in Kafka and StarRocks connector schema lookups

dangzitou:
## What this PR does

Replaces bare `Optional.get()` calls with `Optional.orElseThrow()` in the Kafka and StarRocks pipeline connectors to provide meaningful error messages when primary key columns are missing from the schema.

### Affected files

- `CsvSerializationSchema.java` — CSV serialization primary key lookup
- `JsonSerializationSchema.java` — JSON serialization primary key lookup
- `StarRocksUtils.java` — StarRocks primary key column ordering

### Why

`Optional.get()` throws a bare `NoSuchElementException` with no context when the Optional is empty. This makes debugging difficult when a primary key column is unexpectedly missing (e.g., after schema evolution). Using `orElseThrow()` with a descriptive message including the column name makes failures immediately diagnosable.

## Testing

Behavioral change only in the error path — happy-path behavior is unchanged.

---


haruki-830 (commented):
The StarRocks change looks good.

One issue on the Kafka side: in both `CsvSerializationSchema` and `JsonSerializationSchema`, the `orElseThrow` lambda references the loop counter i, which isn't effectively final — so the code won't compile.

We may extract the PK name into a local variable before the lambda (e.g. `String primaryKey = schema.primaryKeys().get(i)`) to fix it.

dangzitou:
Thanks for catching this. I updated both Kafka loops to capture `primaryKey` before the `orElseThrow` lambda, so the lambda no longer references the mutable loop counter.

Validation: `mvn -pl flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-kafka -am -DskipTests compile` completed successfully (including Spotless and Checkstyle).

dangzitou:

PR #4435, opened 2026-06-04
[FLINK-39844][postgres] Filter snapshot table columns by schema name to fix LIKE-wildcard cross-schema column leakage

JNSimba:
https://issues.apache.org/jira/browse/FLINK-39844

During the snapshot phase, the PostgreSQL connector reads column structure via JDBC `DatabaseMetaData#getColumns(catalog, schemaPattern, tableNamePattern, columnNamePattern)`. Per the JDBC spec, **both** `schemaPattern` and `tableNamePattern` are LIKE patterns, where `_` matches any single character and `%` matches any sequence of characters. Both are legal identifier characters in PostgreSQL, so `getColumns` can return columns from other schemas/tables that were never meant to match.

An exact filter on the **table name** already exists, but the **schema name was never validated**. When two schemas have names that are wildcard matches of each other (e.g. `sch_test` and `schxtest`, where `_` matches `x`) and both contain a same-named table, capturing `sch_test.<table>` also pulls in the look-alike schema's columns. The table-name filter cannot tell them apart, so the snapshot fails with `IllegalStateException: Duplicate key Optional.empty` (or, when columns differ, silently merges columns from the wrong schema).

## Brief change log

- `PostgresConnection#doReadTableColumn`: also compare the result-set `TABLE_SCHEM` (column 2) against `TableId.schema()`, in addition to the existing `TABLE_NAME` check. The schema check is skipped when the requested schema is `null`, so columns are not dropped when no schema is specified. This is a pure after-the-fact filter; it does not change the metadata query and does not affect normal (non-wildcard) schemas/tables.
- `SimilarTableNamesITCase` / `similar_names.sql`: add a cross-schema case with two schemas (`sch_test` / `schxtest`) that are wildcard matches of each other, each holding a same-named table, verifying that only the target schema's snapshot and incremental data are captured.

## Verifying this change

- Added `SimilarTableNamesITCase#testReadTableWithSimilarSchemaNameUnderscore`, which fails (snapshot `IllegalStateException: Duplicate key Optional.empty`) without the fix and passes with it.

## Does this pull request potentially affect one of the following parts:

- The public API, i.e., is any changed class annotated with `@Public(Evolving)`: no
- The serializers: no
- The runtime per-record code paths (performance sensitive): no
- The connector code base: yes (postgres-cdc snapshot column reading)

## Documentation

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-postgres-cdc/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
@@ -709,8 +709,13 @@ public class PostgresConnection extends JdbcConnection {
         // - When querying 'user_sink', the pattern may also match 'userbsink' (due to '_')
         // - When querying 'user%data' (where % is literal), it may match 'user_test_data' (due to
         // '%')
+        // Both the schema name and table name are passed to DatabaseMetaData#getColumns as LIKE
+        // patterns, so wildcards in either may produce columns from other schemas/tables. Verify
+        // both against the requested TableId (skipping the schema check when no schema is given).
+        final String resultSchemaName = columnMetadata.getString(2);
         final String resultTableName = columnMetadata.getString(3);
-        if (!tableId.table().equals(resultTableName)) {
+        if (!tableId.table().equals(resultTableName)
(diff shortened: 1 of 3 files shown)


Mrart (commented):
I think we need add ut in ‘PostgresPipelineITCase’ to cover this change

JNSimba:
> I think we need add ut in ‘PostgresPipelineITCase’ to cover this change

Should SimilarTableNamesITCase.java be able to override this? This is a source-level change, so perhaps we don't need to add a case to the pipeline?

PR #4437, opened 2026-06-04
[FLINK-38647] Support changelog-mode=upsert + null-tolerance

cansakiroglu:
FLINK-38647 reports a NullPointerException from the Postgres Pipeline source when a captured table has REPLICA IDENTITY DEFAULT and an UPDATE arrives. Root cause: PostgresDataSource hardcodes DebeziumChangelogMode.ALL, which makes the deserializer extract a before-image that is null under DEFAULT replication, hence NPE'ing.

This change fixes the issue from two angles:

1. Connector side: expose 'changelog-mode' YAML option on the Postgres Pipeline connector. Accepts 'all' (default, current behaviour) or 'upsert'. In upsert mode the source emits UPDATE events with before == null and only after populated, so the pipeline runs cleanly under REPLICA IDENTITY DEFAULT without requiring FULL (which increases WAL volume).

2. Runtime side: make serializer null-tolerant. Three changes:

- copy() null-guards each recordDataSerializer.copy() call so the chained CopyingChainingOutput.pushToOperator path stops NPE'ing on null before/after.
- serialize() writes 2 leading boolean presence flags before conditionally writing each record. Required because the previous serialize() already skipped writing null fields, but deserialize() always tried to read two records back.
- deserialize() reads the 2 flag bytes and skips the corresponding read when absent.

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/factory/PostgresDataSourceFactory.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/factory/PostgresDataSourceFactory.java
@@ -24,6 +24,7 @@ import org.apache.flink.cdc.common.event.TableId;
 import org.apache.flink.cdc.common.factories.DataSourceFactory;
 import org.apache.flink.cdc.common.factories.Factory;
 import org.apache.flink.cdc.common.factories.FactoryHelper;
+import org.apache.flink.cdc.common.schema.Schema;
 import org.apache.flink.cdc.common.schema.Selectors;
 import org.apache.flink.cdc.common.source.DataSource;
 import org.apache.flink.cdc.common.utils.StringUtils;
@@ -31,6 +32,7 @@ import org.apache.flink.cdc.connectors.base.options.SourceOptions;
 import org.apache.flink.cdc.connectors.base.options.StartupOptions;
 import org.apache.flink.cdc.connectors.postgres.source.PostgresDataSource;
 import org.apache.flink.cdc.connectors.postgres.source.PostgresSourceBuilder;
+import org.apache.flink.cdc.connectors.postgres.source.config.PostgresSourceConfig;
 import org.apache.flink.cdc.connectors.postgres.source.config.PostgresSourceConfigFactory;
 import org.apache.flink.cdc.connectors.postgres.table.PostgreSQLReadableMetadata;
 import org.apache.flink.cdc.connectors.postgres.utils.PostgresSchemaUtils;
@@ -38,6 +40,7 @@ import org.apache.flink.cdc.debezium.table.DebeziumChangelogMode;
 import org.apache.flink.table.api.ValidationException;
 import org.apache.flink.table.data.RowData;
 
+import io.debezium.connector.postgresql.connection.PostgresConnection;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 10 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/PostgresDataSourceOptions.java
```diff
@@ -281,4 +281,17 @@ public class PostgresDataSourceOptions {
                     .defaultValue(false)
                     .withDescription(
                             "Whether to infer CDC column types when processing pgoutput Relation messages.");
+
+    @Experimental
+    public static final ConfigOption<String> CHANGELOG_MODE =
+            ConfigOptions.key("changelog-mode")
+                    .stringType()
+                    .defaultValue("all")
```
haruki-830: I noticed the existing SQL connector (flink-connector-postgres-cdc) defines this option using enumType(DebeziumChangelogMode.class). It might be worth aligning with that approach — it would give us compile-time safety and also avoid the null handling concern in the factory. Something like:
```
public static final ConfigOption<DebeziumChangelogMode> CHANGELOG_MODE =
        ConfigOptions.key("changelog-mode")
                .enumType(DebeziumChangelogMode.class)
                .defaultValue(DebeziumChangelogMode.ALL);
```

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/factory/PostgresDataSourceFactory.java
```diff
@@ -203,8 +205,22 @@ public DataSource createDataSource(Context context) {
         String metadataList = config.get(METADATA_LIST);
         List<PostgreSQLReadableMetadata> readableMetadataList = listReadableMetadata(metadataList);
 
-        // Create a custom PostgresDataSource that passes the includeDatabaseInTableId flag
-        return new PostgresDataSource(configFactory, readableMetadataList);
+        String changelogModeRaw = config.get(CHANGELOG_MODE);
+        DebeziumChangelogMode changelogMode;
+        switch (changelogModeRaw.toLowerCase()) {
```
haruki-830: If enumType is adopted for CHANGELOG_MODE, this switch block could be simplified to a single config.get(CHANGELOG_MODE) call.
Also, it might be good to add a primary key validation for upsert mode, similar to what the SQL connector does in `PostgreSQLTableFactory`.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/event/DataChangeEventSerializer.java
```diff
@@ -62,10 +62,17 @@ public void serialize(DataChangeEvent event, DataOutputView target) throws IOExc
...
         tableIdSerializer.serialize(event.tableId(), target);
 
-        if (event.before() != null) {
+        // Write explicit presence flags so deserialize can recover null before/after.
+        // Required when changelog-mode=upsert (FLINK-38647) emits UPDATE events with
+        // before == null. Adds 2 bytes per event but makes the wire format symmetric.
+        boolean beforePresent = event.before() != null;
+        boolean afterPresent = event.after() != null;
+        target.writeBoolean(beforePresent);
+        target.writeBoolean(afterPresent);
```
haruki-830: This looks like a wire format breaking change — old deserializers would interpret these 2 boolean bytes as the start of the before RecordData, which could cause stream corruption on checkpoint restore.Would it be possible to bump DataChangeEventSerializerSnapshot.CURRENT_VERSION to 2 in this PR and add a version-aware fallback? 
I think this would be important to include before merging, since checkpoint restoration is a production-critical path.

haruki-830:
Nice PR! I've left a few inline comments with suggestions on the implementation. It would also be great to have some test coverage — mainly roundtrip tests for the serializer with null before/after, and a basic unit test. Looking forward to the updates!

dev-donghwan:
@cansakiroglu

Thank you for this PR.

This PR resolves the source part, but the sink part is still broken.

Case:

1. flink-cdc pipeline -> Paimon sink (`changelog-mode: upsert`)
2. the Paimon sink calls `PaimonWriterHelper.convertEventToFullGenericRows()`
3. the UPDATE branch requires `before` -> NPE, and the job fails permanently

```
SinkWrapperException: Failed to handle event ... DataChangeEvent{tableId=..., before=null, ..., op=UPDATE}
Caused by: java.lang.NullPointerException:
  Cannot invoke "org.apache.flink.cdc.common.data.RecordData.getArity()" because "recordData" is null
  at PaimonWriterHelper.convertRecordDataToGenericRow(...)
  at PaimonWriterHelper.convertEventToFullGenericRows(...)   // UPDATE branch reads before() unconditionally
```

The reason is that a PostgreSQL upsert event does not have `before`, only `after` — the same failure pattern FLINK-39684 fixed for REPLACE in the same method.

We verified a one-line guard end-to-end (PG -> Paimon, DML + schema evolution):

```java
case UPDATE:
    if (hasPrimaryKey && dataChangeEvent.before() != null) {   // skip UPDATE_BEFORE when there is nothing to build it from
        ...
```

For primary-key tables this is safe: `+U`-only and `-U`+`+U` converge to the same merged state, and PK-changing updates never arrive as UPDATE (Debezium decomposes them into DELETE + CREATE). Repro test: `DataChangeEvent.updateEvent(tableId, null, afterData)` in `PaimonWriterHelperTest` — every existing test passes a non-null `before`.

Note: the Kafka sink has the same pattern — `CanalJsonSerializationSchema` and `DebeziumJsonSerializationSchema` read `before()` for UPDATE (and REPLACE) too.

Would you prefer to include the sink-side null-tolerance in this PR, or should we file a separate JIRA as a follow-up to FLINK-39684? 

cc. @haruki-830 @leonardBang @yuxiqian

yuxiqian:
Thanks Donghwan for the kind review. I believe it's incorrect for the PostgreSQL source to emit partial UpdateEvents that leave the before field `null`, as it may break many assumptions in runtime operators / sinks.

I wonder if we can emit the `REPLACE` event in UPSERT changelog mode as it carries +U only? Thus we don't have to modify sinks to be "null-tolerance".

dev-donghwan:
Thanks @yuxiqian, that makes sense — emitting `REPLACE` (+U only) in upsert changelog mode, instead of an `UPDATE` event with a null `before`.

Agreed this is the cleaner approach: `REPLACE` is already handled by the event serializer and all pipeline sinks (e.g. FLINK-39684 already covers it in the Paimon writer), so no sink needs to become null-tolerant, and the null-tolerant `DataChangeEventSerializer` changes in this PR would no longer be needed either.

We implemented this and verified it with a real pipeline run on Docker (Flink 2.2.1, PostgreSQL 15 with `REPLICA IDENTITY DEFAULT`, pgoutput) — the UPDATE arrives at the sink as `op=REPLACE` carrying the after image only.

@cansakiroglu I've opened a PR against your branch with this change as a single commit: https://github.com/cansakiroglu/flink-cdc/pull/1

dev-donghwan (approved):

PR #4441, opened 2026-06-12
[FLINK-39916][mysql] Fix MySQL batch startup option validation error

llphxd:
This PR fixes an incorrect placeholder used in the MySQL Pipeline Connector batch startup option validation error.

## JIRA
[FLINK-39916](https://issues.apache.org/jira/browse/FLINK-39916)

## Change log
- Replace the SLF4J-style `{}` placeholder with the Java `String.format` placeholder `%s` in `MySqlDataSourceFactory.java`, so the actual startup option can be rendered correctly in the thrown `IllegalArgumentException`.

- Additional Cleanup: Remove an unused `DOT_PLACEHOLDER` constant from `MySqlDataSourceFactory.java`.

## Verifying this change
This is a small validation error output fix with no test coverage added.

## Why are the changes needed?
- This avoids showing an unresolved `{}` placeholder when an unsupported MySQL startup option is used in BATCH pipeline mode.

## How was this patch tested?
- Ran the related unit test locally.


haruki-830 (approved):
+1.

lvyanquan (approved):
+1.

PR #4446, opened 2026-06-18
[minor][pipeline-connector/sqlserver] Fix SQL Server schema restore without per-record JDBC

leonardBang:
Restore CreateTableEvent state from source splits during reader initialization so savepoint recovery does not fall back to per-record JDBC lookups.


leonardBang:
@Daishuyuan Would you like to review this PR when you have time?

leonardBang:
@ruanhang1993 Would you like to help review this PR when you have time ?

ruanhang1993 (approved):
LGTM.

PR #3643, opened 2024-10-14
[FLINK-35360] support Flink cdc pipeline Yarn application mode

MOBIN-F:
hi @Mrart ,I merged your [PR](https://github.com/apache/flink-cdc/pull/3599) and tested it, but found some problem,i fixed it and added some tests.

flink-cdc-pipeline is a great feature, but currently only supports yarn-session in yarn. I think the cdc-community should support application mode as soon as possible so that users can promote the use of pipeline features within the company.

main changes of this PR are:
1. Added CliExecutor#main method as the program entry of application mode, because in application mode, the main method runs in the cluster
2. org.apache.flink.core.fs.Path replaces java.nio.file.Path, because additionalJars may be on the remote hdfs or oss
3. MysqlE2eWithYarnApplicationITCase is the flink yarn application test class, which will start yarnminicluster to submit flink jobs
@yuxiqian @leonardBang @Mrart Can you help review this PR? Thank you～

PR #4451, opened 2026-06-29
[FLINK-40005][cdc-cli] CliExecutor application mode resolves pipeline definition from local path (k8s), remote path or inline content

Zhile:

Fix [FLINK-40005]: in **application mode**, `CliExecutor.main` failed to start a YAML pipeline with `Missing required field "source" in top-level configuration`.

`args[0]` was passed straight to `YamlPipelineDefinitionParser.parse(String, ...)`, which treats its `String` argument as YAML **content**. When `args[0]` is a file **path**, the path itself is parsed as YAML, producing a document without `source`/`sink` and failing validation.

The two native application deployment executors share `CliExecutor.main` but pass `args[0]` with different semantics — the root of the problem:
- `K8SApplicationDeploymentExecutor` sets `APPLICATION_ARGS = commandLine.getArgList()`,  so `args[0]` is the pipeline definition **file path** (shipped into the JobManager container, e.g. mounted via a ConfigMap by the Flink Kubernetes Operator).
- `YarnApplicationDeploymentExecutor` reads the file on the client side and sets `APPLICATION_ARGS` to the pipeline definition **content**.

So a fix that only reads `args[0]` as a path would fix K8S but **regress YARN** (whose `args[0]` is already content, not a path).

Regression introduced by #3643 (FLINK-35360, *Support Yarn application mode for yaml job*).

## Brief change log

- `CliExecutor.main` now resolves `args[0]` via `resolvePipelineDef(...)`, in three cases:
  1. Explicit-scheme path (`s3://`/`hdfs://`/`oss://`/`file://`) — read through Flink's `FileSystem` so the matching plugin resolves it.
  2. Bare local file path — read with the local JVM file API (`Files.readAllBytes`) rather than Flink's `FileSystem`, whose cluster default may be S3/HDFS and would not resolve a local path shipped next to the JobManager.
  3. Otherwise the value is already the pipeline definition content (YARN) — parsed verbatim.
- Fixes K8S application mode without regressing YARN, and additionally supports placing the pipeline definition file on remote storage (S3/HDFS/OSS).
- Update Kubernetes deployment docs (EN + ZH): the `FlinkDeployment` example now uses `entryClass: org.apache.flink.cdc.cli.CliExecutor` without `--use-mini-cluster`, and the "native application mode is not supported" note is removed.

## Verifying this change

- `CliExecutorTest` covers all three resolution paths (local path / scheme path via `file://`, same code path as `s3://` / inline content) plus the original regression.
- Manually verified on a Flink 1.20 Kubernetes Application-mode deployment (Flink Kubernetes Operator): the pipeline starts and runs; previously it failed with `Missing required field "source"`.

- Dependencies: no
- The public API: no
- The serializers / checkpointing: no

### DIFF
--- a/docs/content.zh/docs/deployment/kubernetes.md
+++ b/docs/content.zh/docs/deployment/kubernetes.md
@@ -241,10 +241,9 @@ spec:
   imagePullPolicy: Always
   job:
     args:
-      - '--use-mini-cluster'
       - /opt/flink/flink-cdc-{{< param Version >}}/conf/mysql-to-doris.yaml
-    entryClass: org.apache.flink.cdc.cli.CliFrontend
-    jarURI: 'local:///opt/flink/flink-cdc-{{< param Version >}}/lib/flink-cdc-dist-{{< param Version >}}.jar'
+    entryClass: org.apache.flink.cdc.cli.CliExecutor
+    jarURI: 'local:///opt/flink/lib/flink-cdc-dist-{{< param Version >}}.jar'
     parallelism: 1
     state: running
     upgradeMode: savepoint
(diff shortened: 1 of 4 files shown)


Zhile:
@MOBIN-F  can you help to review? I'm not sure if this applies the original design.

Zhile:
@yuxiqian @Mrart @leonardBang can you guys help to review this PR? If there's any issue, I can modify it.

yuxiqian:
@MOBIN-F Could you please take a look at this PR?

PR #4457, opened 2026-07-01
[FLINK-40043] fix the paimon pipeline job couldn’t increase job parallelism.

sd4324530:
Fix a NoSuchElementException thrown by PaimonSink.restoreWriter when a CDC pipeline job using the Paimon sink is restored from a savepoint with a higher pipeline.parallelism than the original run.

## Bug Description
Rescaling a pipeline job up via savepoint (stop --withSavepoint → modify pipeline.parallelism → run --fromSavepoint) causes the new sink subtasks to fail to initialize, taking the entire job down at startup. The same operation with the same or lower parallelism works as expected.

## Stack trace:

```
java.util.NoSuchElementException
    at java.util.ArrayList$Itr.next(ArrayList.java:1000)
    at org.apache.flink.cdc.connectors.paimon.sink.v2.PaimonSink.restoreWriter(PaimonSink.java:101)
    at org.apache.flink.streaming.runtime.operators.sink.StatefulSinkWriterStateHandler.createWriter(StatefulSinkWriterStateHandler.java:120)
    at org.apache.flink.streaming.runtime.operators.sink.SinkWriterOperator.initializeState(SinkWriterOperator.java:153)
    at org.apache.flink.cdc.runtime.operators.sink.DataSinkWriterOperator.initializeState(DataSinkWriterOperator.java:132)
    ...
```

## Root Cause Analysis
PaimonSink implements SupportsWriterState<InputT, PaimonWriterState>. Each PaimonWriter snapshots one PaimonWriterState (containing the randomly generated commitUser) in PaimonWriter.snapshotState():

```java
@Override
public List<PaimonWriterState> snapshotState(long checkpointId) {
    return Collections.singletonList(stateCache);
}
```

When restoring, Flink's StatefulSinkWriterStateHandler.createWriter() (a Flink framework class) takes the union of all previously snapshotted states, and re-distributes them across the new parallelism based on its internal channel computer. With P_old < P_new, some of the new subtasks will receive zero PaimonWriterState entries — this is expected behavior of the SupportsWriterState contract.

However, PaimonSink.restoreWriter does not handle the empty-collection case:

```java
@Override
public StatefulSinkWriter<InputT, PaimonWriterState> restoreWriter(
        WriterInitContext context, Collection<PaimonWriterState> paimonWriterStates) {
    long lastCheckpointId = ...;
    Preconditions.checkNotNull(paimonWriterStates);                       // only null check
    String storedCommitUser = paimonWriterStates.iterator().next().getCommitUser();  // <-- NPE on empty
    return new PaimonWriter<>(catalogOptions, context.metricGroup(), storedCommitUser, ...);
}
```

Calling iterator().next() on an empty collection throws NoSuchElementException. This is a contract violation of the SupportsWriterState interface, which explicitly allows empty collections (see Flink's StreamingFileSink and FileSink for correct implementations).

The issue was masked before FLINK-38478 introduced per-writer random commitUsers stored in state — the previous commitUser was a static field on the sink and didn't need to be restored per-subtask.

## Fix
When the restored state collection is empty (i.e., this subtask did not receive any historical state during rescale-out), fall back to the commitUser field on the PaimonSink instance. This is safe because all subtasks of the same job share the same PaimonSink instance (and therefore the same commitUser), and the commitUser is re-generated only on a fresh job startup — not on a state restore.

```java
@Override
public StatefulSinkWriter<InputT, PaimonWriterState> restoreWriter(
        WriterInitContext context, Collection<PaimonWriterState> paimonWriterStates) {
    long lastCheckpointId =
            context.getRestoredCheckpointId()
                    .orElse(CheckpointIDCounter.INITIAL_CHECKPOINT_ID - 1);
    Preconditions.checkNotNull(paimonWriterStates);
    String storedCommitUser =
            paimonWriterStates.isEmpty()
                    ? commitUser
                    : paimonWriterStates.iterator().next().getCommitUser();
    return new PaimonWriter<>(
            catalogOptions,
            context.metricGroup(),
            storedCommitUser,
            serializer,
            lastCheckpointId);
}
```

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonSink.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-paimon/src/main/java/org/apache/flink/cdc/connectors/paimon/sink/v2/PaimonSink.java
@@ -98,7 +98,10 @@ public class PaimonSink<InputT>
                 context.getRestoredCheckpointId()
                         .orElse(CheckpointIDCounter.INITIAL_CHECKPOINT_ID - 1);
         Preconditions.checkNotNull(paimonWriterStates);
-        String storedCommitUser = paimonWriterStates.iterator().next().getCommitUser();
+        String storedCommitUser =
+                paimonWriterStates.isEmpty()
+                        ? this.commitUser
+                        : paimonWriterStates.iterator().next().getCommitUser();
         return new PaimonWriter<>(
                 catalogOptions,
                 context.metricGroup(),
--- a/flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/migration/MySqlToPaimonMigrationITCase.java
+++ b/flink-cdc-e2e-tests/flink-cdc-pipeline-e2e-tests/src/test/java/org/apache/flink/cdc/pipeline/tests/migration/MySqlToPaimonMigrationITCase.java
@@ -262,6 +262,97 @@ class MySqlToPaimonMigrationITCase extends PipelineTestEnvironment {
 
     @ParameterizedTest(name = "{0} -> SNAPSHOT")
     @EnumSource(names = {"SNAPSHOT"})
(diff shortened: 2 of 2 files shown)


sd4324530:
@lvyanquan Would you like to help review this PR when you have time?

lvyanquan (approved):
+1.

PR #4459, opened 2026-07-02
[FLINK-40052][mysql] Roll back assigner status in addSplits to prevent NPE in createBinlogSplit

linguoxuan:
Rollback assigner status when failed splits are re-added, to prevent the hybrid assigner from prematurely creating the binlog split before all re-added splits are finished again.


haruki-830 (approved):
+1

PR #4463, opened 2026-07-07
[hotfix] Bump Mockito for JDK 17 tests

yuxiqian:
Bump Mockito version to fix JDK 17 tests, as older version could not handle JDK 17 bytecodes properly.


lvyanquan (approved):
+1.

PR #4464, opened 2026-07-07
[FLINK-40087][mysql] Fix MySqlDefaultValueConverter fails to parse charset-introducer default values wrapped in double quotes and/or containing padding spaces (e.g. `_UTF8MB4" 0 "`)

gong:
FLINK-40087 Fix MySqlDefaultValueConverter fails to parse charset-introducer default values wrapped in double quotes and/or containing padding spaces (e.g. `_UTF8MB4" 0 "`)

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlDefaultValueConverter.java
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mysql-cdc/src/main/java/io/debezium/connector/mysql/MySqlDefaultValueConverter.java
@@ -60,7 +60,7 @@ public class MySqlDefaultValueConverter implements DefaultValueConverter {
             Pattern.compile("([0-9]*-[0-9]*-[0-9]*) ([0-9]*:[0-9]*:[0-9]*(\\.([0-9]*))?)");
 
     private static final Pattern CHARSET_INTRODUCER_PATTERN =
-            Pattern.compile("^_[A-Za-z0-9]+'(.*)'$");
+            Pattern.compile("^_[A-Za-z0-9]+(?:'(.*)'|\"(.*)\")$");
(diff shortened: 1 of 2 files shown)


gong:
@leonardBang @yuxiqian PTAL

yuxiqian (commented):
One minor comment: can we add tests for snapshotting tables with double quoted default value fields?

IIRC default values are parsed differently in snapshotting (by SHOW CREATE TABLES) and incremental stages (by parsing DDL).

gong:
> One minor comment: can we add tests for snapshotting tables with double quoted default value fields?
> 
> IIRC default values are parsed differently in snapshotting (by SHOW CREATE TABLES) and incremental stages (by parsing DDL).

@yuxiqian Thanks for the review! I added testSnapshotWithCharsetIntroducerDefaultValue which creates a table with both single-quoted (_utf8mb4' 0 ') and double-quoted (_UTF8MB4" 0 ") character-set-introducer defaults (covering TINYINT / INT / BIGINT / DECIMAL / DATE / DATETIME / TINYINT(1)) before the Flink source starts, so the schema is discovered via SHOW CREATE TABLE during the snapshot phase.
While writing this I probed the actual output of SHOW CREATE TABLE on MySQL and found that the server normalizes those default expressions on the write path — it strips the introducer, strips the padding inside the quotes, and rewrites double quotes to single quotes. So the snapshot path never actually observes the raw shape and cannot serve as a reverse reproduction for the bug. The new case therefore acts as a positive verification that the fix does not introduce any regression on the snapshot schema discovery path (source reaches RUNNING, snapshot rows arrive, and a subsequent binlog INSERT relying on server-side defaults also flows through). The reverse reproduction remains covered by testAlterWithCharsetIntroducerDefaultValue which feeds the raw ALTER TABLE DDL through the binlog. I documented this in the Javadoc of the new test.

gong:
git rebase master

PR #4467, opened 2026-07-09
[FLINK-39760] Fix missing JTS core dependency of PostgreSQL YAML connector

haruki-830:
### Summary
This commit adds the missing JTS core dependency to the PostgreSQL YAML connector, ensuring the connector can be built independently when geometry handling code uses JTS classes.
### Key Changes
**Explicit JTS Core Dependency**

- Added org.locationtech.jts:jts-core to flink-cdc-pipeline-connector-postgres.
- Uses jts.version set to 1.19.0, matching the JTS version already used transitively in the project dependency tree.

**Fix Standalone Connector Build**

- PostgresEventDeserializer imports and uses JTS classes such as Coordinate and WKBReader for geometry value handling.
- Before this change, the PostgreSQL YAML connector did not declare jts-core in its own module dependencies, which could break isolated module builds or packaging workflows.

### JIRA Reference
[https://issues.apache.org/jira/browse/FLINK-39760](url)

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/pom.xml
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/pom.xml
@@ -188,6 +188,7 @@ limitations under the License.
                                     <include>com.google.guava:*</include>
                                     <include>org.apache.kafka:*</include>
                                     <include>org.postgresql:postgresql</include>
+                                    <include>org.locationtech.jts:jts-core</include>
                                     <include>com.fasterxml.*:*</include>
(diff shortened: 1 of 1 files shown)


flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/pom.xml
lvyanquan, line 195: Please also add org.locationtech.jts:jts-core to the shade includes

haruki-830: Addressed, thanks.

lvyanquan (approved):
+1.

PR #4471, opened 2026-07-13
[FLINK-40130][connect/paimon] Fix missing primary key type compatibility check of Paimon sink

haruki-830:
#### Summary

This commit adds primary key type compatibility validation to the Flink CDC pipeline Paimon sink. When writing to an existing Paimon table, the sink now rejects incompatible upstream primary key types before entering the record coercion path, preventing unexpected conversion failures or incorrect behavior.

#### Key Changes

1. Primary Key Type Compatibility Validation
- Added validation in `BucketAssignOperator` for existing Paimon tables.
- Checks whether each existing Paimon primary key column is type-compatible with the corresponding upstream primary key column before record coercion.
- Reports a clear error when the primary key type is incompatible.

2. Reuse Existing Type Compatibility Logic
- Made `SchemaMergingUtils.isDataTypeCompatible(...)` public and reused it directly in the Paimon sink for primary key type checks.
- Ignores nullability during this check, so validation focuses on type compatibility while nullability remains enforced by the normal write/table validation path.

3. Test Coverage
- Added coverage for incompatible primary key types.
- Added coverage for compatible primary key type changes to ensure valid cases continue to work.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40130](url)

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/utils/SchemaMergingUtils.java
@@ -327,8 +327,12 @@ public class SchemaMergingUtils {
         return coercedRow;
     }
 
-    @VisibleForTesting
-    static boolean isDataTypeCompatible(@Nullable DataType currentType, DataType upcomingType) {
+    /**
+     * Checks whether the upcoming data type can fit into the current data type. A missing current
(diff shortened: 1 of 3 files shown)


haruki-830:
Hi @lvyanquan, could you please help review this PR when you have time? Thanks!

lvyanquan (approved):
+1.

PR #4473, opened 2026-07-14
[FLINK-40148][common] Fix TableIdRouter producing wrong sink table for multi-digit source suffix

sd4324530:

Fix `TableIdRouter.resolveReplacement` so it routes tables correctly when the source-table regex contains a multi-character capturing group (e.g. a multi-digit suffix like `13`, `14`, `16`) regardless of whether the sink-table uses a `$N` back-reference.

Previously, with a rule such as:

    route:
      - source-table: db_6.table_([1-9]|1[0-6])
        sink-table:   new_db_6.table_merged

`db_6.table_13` was wrongly routed to `new_db_6.table_merged3` (and similarly `_14` → `...merged4`, `_16` → `...merged6`). The bug was caused by `Matcher.find()` + `Matcher.replaceAll` in `resolveReplacement`: `find()` only matched the prefix `db_6.table_1` (because the alternation `[1-9]` wins over `1[0-6]` for input `13`), and `replaceAll` left the trailing `3` appended to the configured sink-table. Tables 1-9 were unaffected because the regex consumed the entire single-digit suffix.

## Brief change log

- Replace `Matcher.find()` + `Matcher.replaceAll(sinkTable)` with `Matcher.matches()` + `Matcher.appendReplacement` / `Matcher.appendTail` in `TableIdRouter.resolveReplacement`. This guarantees the whole source table id is consumed before the substitution is applied, so trailing characters can no longer leak into the sink-table name.
- Add `testRouteWithoutBackReferenceWithMultiDigitSuffix` and `testRouteWithBackReferenceAndMultiDigitCapture` to `TableIdRouterTest` as regression coverage for this bug.

---

## Verifying this change

- Added unit tests in `flink-cdc-runtime/src/test/java/org/apache/flink/cdc/common/route/TableIdRouterTest.java`:
  - `testRouteWithoutBackReferenceWithMultiDigitSuffix` — reproduces the originally reported scenario (sink-table without `$1`, multi-digit source suffix); fails on master, passes with the fix.
  - `testRouteWithBackReferenceAndMultiDigitCapture` — guards the same root cause for the `$1` back-reference case to prevent future regressions.
- All 12 tests in `TableIdRouterTest` pass, together with the 7 tests in `TableIdRouterMatchModeTest` and the 8 tests in `SchemaDerivatorTest` (27 tests total, no regressions).
- Manually verified by running the failing test against unmodified `master`: the new test reproduces wrong sink-table names such as `new_db_6.table_merged0`, `...merged1`, `...merged3`, `...merged4`, `...merged5`, `...merged6`, which match the issue's report.

---

- [x] Yes

Generated-by: cluade code with minimax m3

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/route/TableIdRouter.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/route/TableIdRouter.java
@@ -173,9 +173,20 @@ public class TableIdRouter {
         if (route.f2 != null) {
             return TableId.parse(route.f1.replace(route.f2, originalTable.getTableName()));
         } else {
+            // Use matches() so that the entire source table id is consumed. After matches() the
+            // matcher holds the full match, and we feed it through appendReplacement /
+            // appendTail to perform the `$N` substitution only on the matched region. We cannot
+            // use Matcher.replaceAll here because it internally calls reset() + find(), which
+            // would only match a prefix of the source table id (when the regex alternation
+            // order matters, e.g. `([1-9]|1[0-6])` matches `1` from `13` and leaves `3` as a
+            // leftover that gets appended to the sink table name, routing
+            // `table_13` to `table_3`).
             Matcher matcher = route.f0.matcher(originalTable.toString());
-            if (matcher.find()) {
-                return TableId.parse(matcher.replaceAll(route.f1));
+            if (matcher.matches()) {
+                StringBuilder sb = new StringBuilder();
+                matcher.appendReplacement(sb, route.f1);
+                matcher.appendTail(sb);
+                return TableId.parse(sb.toString());
             }
         }
         return TableId.parse(route.f1);
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/route/TableIdRouter.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/route/TableIdRouter.java
@@ -178,9 +178,8 @@ public class TableIdRouter {
             // appendTail to perform the `$N` substitution only on the matched region. We cannot
             // use Matcher.replaceAll here because it internally calls reset() + find(), which
             // would only match a prefix of the source table id (when the regex alternation
-            // order matters, e.g. `([1-9]|1[0-6])` matches `1` from `13` and leaves `3` as a
-            // leftover that gets appended to the sink table name, routing
-            // `table_13` to `table_3`).
+            // order matters, e.g. `([1-9]|1[0-6])` matching `1` in `db_6.table_13`), leaving `3` as
+            // a leftover that gets appended to the configured sink-table name (e.g. `...merged3`).
             Matcher matcher = route.f0.matcher(originalTable.toString());
             if (matcher.matches()) {
                 StringBuilder sb = new StringBuilder();
--- a/flink-cdc-runtime/src/test/java/org/apache/flink/cdc/common/route/TableIdRouterTest.java
+++ b/flink-cdc-runtime/src/test/java/org/apache/flink/cdc/common/route/TableIdRouterTest.java
@@ -178,6 +178,60 @@ public class TableIdRouterTest extends SchemaTestBase {
     }
 
     @Test
+    void testRouteWithoutBackReferenceWithMultiDigitSuffix() {
+        // Regression test: when the source-table regex matches multi-digit suffixes (e.g. 10-16)
+        // but the sink-table does NOT contain a `$1` back-reference, every matched source table
+        // should still be routed to the exact sink-table declared in the routing rule. Previously
+        // TableIdRouter.resolveReplacement used Matcher.find(), which only matched a prefix of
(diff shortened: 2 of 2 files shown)


sd4324530:
Please take a look when you have time.

sd4324530:
I couldn't reproduce the CI error in my local environment, so it's probably unrelated to this PR.

lvyanquan:
The fix itself is correct, but it introduces a **behavioral compatibility change**. Existing jobs may produce different source-to-sink table mappings after upgrading. From a user’s perspective, data may silently stop appearing in the previously mapped sink tables, which could look like silent data loss. We should carefully assess the affected scenarios and compatibility impact before merging this change.

CC @yuxiqian, WDYT?

yuxiqian:
Thanks Yanquan for pointing out this, it's incorrect behavior indeed. As the buggy behavior only occurs in one minor version (3.6) and there's no clean workaround here, I would suggest this to be merged in the next major release.

lvyanquan:
Hi, @sd4324530.

I suggest adding a section to the route documentation that explains the behavior in Flink CDC 3.6.0, the corrected behavior introduced by this change, and the potential compatibility impact on existing jobs.

It would also be helpful to add a test that genuinely verifies multi-character capture groups. For example, use a sink pattern such as `new_db_6.table_$1_suffix`, where the old implementation produces `table_1_suffix3` for `table_13`, while the corrected implementation produces `table_13_suffix`.

sd4324530:
> Hi, @sd4324530.
> 
> I suggest adding a section to the route documentation that explains the behavior in Flink CDC 3.6.0, the corrected behavior introduced by this change, and the potential compatibility impact on existing jobs.
> 
> It would also be helpful to add a test that genuinely verifies multi-character capture groups. For example, use a sink pattern such as `new_db_6.table_$1_suffix`, where the old implementation produces `table_1_suffix3` for `table_13`, while the corrected implementation produces `table_13_suffix`.

@lvyanquan Thank you very much for your review and suggestions. I have added compatibility information to the document, and add a test that genuinely verifies multi-character capture groups. Please take a look when you have time.

docs/content.zh/docs/core-concept/route.md
```diff
@@ -130,3 +130,66 @@ route:
...
+
+从 Flink CDC 3.7.0 起，source-table 正则会完整匹配整个 source 表 id，所有捕获组（包括跨越多字符的
+捕获组）都会被完整捕获。下表展示了同样的 source 表 id、同样的配置，在 3.6.0 与 3.7.0 中分别会被
+路由到哪里，便于直接对比。
+
+| Source 表 id         | source-table                              | sink-table                          | 3.6.0 中的路由结果                            | 3.7.0 中的路由结果                       |
+|--------------------|-------------------------------------------|-------------------------------------|--------------------------------------------|-----------------------------------|
+| `new_db_6.table_1` | `new_db_6.table_([1-9]|1[0-6])`           | `new_db_6.table_$1_suffix`          | `new_db_6.table_1_suffix`                  | `new_db_6.table_1_suffix`         |
+| `new_db_6.table_13`| `new_db_6.table_([1-9]|1[0-6])`           | `new_db_6.table_$1_suffix`          | `new_db_6.table_1_suffix3`                 | `new_db_6.table_13_suffix`        |
+| `new_db_6.table_17`| `new_db_6.table_([1-9]|1[0-6])`           | `new_db_6.table_$1_suffix`          | `new_db_6.table_1_suffix7`                 | `new_db_6.table_17`（不匹配）         |
```
lvyanquan: The `table_17` example in this compatibility table is not accurate. Although the old `resolveReplacement()` implementation would match the `table_1` prefix with `find()`, `calculateRoute()` first checks the rule with `Matcher.matches()`. Since `table_17` does not match either `[1-9]` or `1[0-6]` completely, `resolveReplacement()` is never invoked. Therefore, both 3.6.0 and the fixed version fall back to the original source table, `new_db_6.table_17`.

The incorrect mapping `new_db_6.table_1_suffix7` would only be observed if `find()`/`replaceAll()` were invoked directly, not through `TableIdRouter.route()`. Please update this row and the related explanation about “partially matched” source tables accordingly.

docs/content/docs/core-concept/route.md
```diff
@@ -131,3 +131,76 @@ As a result, such table will be routed to downstream table `sink_db_foo.sink_tab
 Standard RegExp capturing could not be used with `replace-symbol` options.
 
 {{< /hint >}}
+
+# Compatibility Notes — Multi-Character Capturing Groups
+
+When a `source-table` rule uses a capturing group that can match more than one character — for
```
lvyanquan: This compatibility change affects route rules whose regular expressions can match a source table ID completely while also matching a shorter prefix with Matcher.find(). This commonly occurs when alternatives are ordered from shorter to longer, such as table_1|table_10 or ([1-9]|1[0-6]). The issue is not limited to regular expressions that contain capturing groups.

lvyanquan (approved):
+1.

PR #4474, opened 2026-07-15
[FLINK-40151][runtime] Support missing YAML Transform predicates and nullable BOOLEAN logical evaluation

haruki-830:
### Summary
This commit improves YAML Transform predicate support and nullable BOOLEAN logical evaluation. It adds support for several previously unsupported SQL predicates, and fixes runtime failures when logical expressions evaluate nullable BOOLEAN operands.

### Key Changes
**Additional Predicate Support**

- Added support for `IS UNKNOWN` / `IS NOT UNKNOWN`.
- Added support for `IS DISTINCT FROM` / `IS NOT DISTINCT FROM`.
- Added support for `LIKE ... ESCAPE`.
- Added support for `SIMILAR TO` / `NOT SIMILAR TO`.

**Nullable BOOLEAN Logical Evaluation**

- Added null-safe handling for `AND`, `OR`, and `NOT` with nullable BOOLEAN operands.
- Added null-safe handling for `IS TRUE` / `IS NOT TRUE`.
- Added null-safe handling for `IS FALSE` / `IS NOT FALSE`.
- Kept boolean condition consumers null-safe so `UNKNOWN` does not cause Java unboxing failures.

**Test Coverage**

- Added unit tests for distinct predicates, nullable BOOLEAN logical functions, legacy comparison behavior, legacy two-argument `LIKE`, `LIKE ... ESCAPE`, and `SIMILAR TO`.
- Updated parser tests for generated Janino expressions.
- Updated YAML transform specs for nullable BOOLEAN logic and the newly supported predicates.

**JIRA Reference**  
[https://issues.apache.org/jira/browse/FLINK-40151](https://issues.apache.org/jira/browse/FLINK-40151)

### DIFF
--- a/docs/content.zh/docs/core-concept/transform.md
+++ b/docs/content.zh/docs/core-concept/transform.md
@@ -122,21 +122,29 @@ Flink CDC 使用 [Calcite](https://calcite.apache.org/) 来解析表达式并且
 | value1 <= value2                     | lessThanOrEqual(value1, value2)              | 如果 value1 小于等于 value2，返回 TRUE；如果 value1 或 value2 为 NULL，返回 FALSE。 |
 | value IS NULL                        | isNull(value)                                | 如果 value 为 NULL，返回 TRUE。                                      |
 | value IS NOT NULL                    | isNotNull(value)                             | 如果 value 不为 NULL，返回 TRUE。                                     |
+| value1 IS DISTINCT FROM value2       | isDistinctFrom(value1, value2)               | 如果 value1 和 value2 是不同的值，返回 TRUE。NULL 会作为普通值参与比较，且不会返回 NULL。 |
+| value1 IS NOT DISTINCT FROM value2   | isNotDistinctFrom(value1, value2)            | 如果 value1 和 value2 不是不同的值，返回 TRUE。NULL 会作为普通值参与比较，且不会返回 NULL。 |
 | value1 BETWEEN value2 AND value3     | betweenAsymmetric(value1, value2, value3)    | 如果 value1 大于等于 value2 且小于等于 value3，返回 TRUE。                    |
 | value1 NOT BETWEEN value2 AND value3 | notBetweenAsymmetric(value1, value2, value3) | 如果 value1 小于 value2 或大于 value3，返回 TRUE。                        |
-| string1 LIKE string2                 | like(string1, string2)                       | 如果 string1 匹配模式 string2，返回 TRUE。                              |
-| string1 NOT LIKE string2             | notLike(string1, string2)                    | 如果 string1 不匹配模式 string2，返回 TRUE。                             |
+| string1 LIKE string2                 | like(string1, string2)                       | 如果 string1 按 Java 正则表达式和子串匹配语义匹配 string2，返回 TRUE。           |
+| string1 NOT LIKE string2             | notLike(string1, string2)                    | 如果 string1 按 Java 正则表达式和子串匹配语义不匹配 string2，返回 TRUE。          |
+| string1 LIKE string2 ESCAPE string3  | like(string1, string2, string3)              | 如果 string1 匹配 SQL LIKE 模式 string2，返回 TRUE。`%` 匹配任意长度字符串，`_` 匹配单个字符，string3 用于转义通配符。如果任一参数为 NULL，返回 NULL。 |
+| string1 NOT LIKE string2 ESCAPE string3 | notLike(string1, string2, string3)        | 如果 string1 不匹配 SQL LIKE 模式 string2，返回 TRUE。如果任一参数为 NULL，返回 NULL。 |
+| string1 SIMILAR TO string2           | similarTo(string1, string2)                  | 如果 string1 匹配 SQL SIMILAR TO 模式 string2，返回 TRUE。如果任一参数为 NULL，返回 NULL。 |
+| string1 NOT SIMILAR TO string2       | notSimilarTo(string1, string2)               | 如果 string1 不匹配 SQL SIMILAR TO 模式 string2，返回 TRUE。如果任一参数为 NULL，返回 NULL。 |
+| string1 SIMILAR TO string2 ESCAPE string3 | similarTo(string1, string2, string3)    | 如果 string1 使用 string3 作为转义字符匹配 SQL SIMILAR TO 模式 string2，返回 TRUE。如果任一参数为 NULL，返回 NULL。 |
+| string1 NOT SIMILAR TO string2 ESCAPE string3 | notSimilarTo(string1, string2, string3) | 如果 string1 使用 string3 作为转义字符不匹配 SQL SIMILAR TO 模式 string2，返回 TRUE。如果任一参数为 NULL，返回 NULL。 |
 | value1 IN (value2 [, value3]* )      | in(value1, value2 [, value3]*)               | 如果 value1 存在于给定列表 (value2, value3, …) 中，返回 TRUE。              |
 | value1 NOT IN (value2 [, value3]* )  | notIn(value1, value2 [, value3]*)            | 如果 value1 不存在于给定列表 (value2, value3, …) 中，返回 TRUE。             |
 
 ## 逻辑函数
 
-逻辑函数遵循 SQL 的三值逻辑来处理可为 NULL 的 BOOLEAN 值。`AND` 和 `OR` 会在左操作数已经决定结果时短路，不再计算右操作数。
+逻辑函数遵循 SQL 的三值逻辑来处理可为 NULL 的 BOOLEAN 值。`AND` 和 `OR` 会在左操作数已经决定结果时短路，不再计算右操作数。生成的 Janino 代码会根据操作数可空性使用原生运算符、条件表达式或惰性函数调用。
 
 | 函数                     | Janino 代码            | 描述                                                                 |
 |------------------------|----------------------|--------------------------------------------------------------------|
-| boolean1 OR boolean2   | or(boolean1, boolean2)  | 如果任一值为 TRUE，返回 TRUE；如果没有 TRUE 且至少一个值为 NULL，返回 NULL。               |
-| boolean1 AND boolean2  | and(boolean1, boolean2) | 如果任一值为 FALSE，返回 FALSE；如果没有 FALSE 且至少一个值为 NULL，返回 NULL。            |
+| boolean1 OR boolean2   | 短路 OR 表达式  | 如果任一值为 TRUE，返回 TRUE；如果没有 TRUE 且至少一个值为 NULL，返回 NULL。               |
+| boolean1 AND boolean2  | 短路 AND 表达式 | 如果任一值为 FALSE，返回 FALSE；如果没有 FALSE 且至少一个值为 NULL，返回 NULL。            |
 | NOT boolean            | not(boolean)           | 如果 boolean 为 FALSE，返回 TRUE；如果 boolean 为 TRUE，返回 FALSE；如果为 NULL，返回 NULL。 |
 | boolean IS FALSE       | isFalse(boolean)       | 如果 boolean 为 FALSE，返回 TRUE；如果 boolean 为 TRUE 或 NULL，返回 FALSE。          |
 | boolean IS NOT FALSE   | isNotFalse(boolean)    | 如果 boolean 为 TRUE 或 NULL，返回 TRUE；如果 boolean 为 FALSE，返回 FALSE。          |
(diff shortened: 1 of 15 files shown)


haruki-830:
@lvyanquan Could u PTAL when u have time? Thanks!

lvyanquan:
Thanks for the work here. The changes split into three buckets with very different compatibility impact. I'd suggest applying buckets 1–2 unconditionally and gating bucket 3 behind an explicit option.

**1. New operators (currently unparseable) — keep always-on**

SQL: `IS [NOT] DISTINCT FROM`, `IS [NOT] UNKNOWN`, `SIMILAR TO` / `NOT SIMILAR TO`, `LIKE ... ESCAPE '<char>'`.

These fail validation on the current release, so no existing pipeline can depend on them. Safe to enable unconditionally.

**2. Crash → 3-valued logic, non-NULL behavior unchanged — keep always-on**

SQL: `AND`, `OR`, `NOT`, `IS [NOT] TRUE`, `IS [NOT] FALSE` where the operand is a NULL boolean (nullable `BOOLEAN` column or `CAST(NULL AS BOOLEAN)`).

These previously threw an NPE on NULL; results for non-NULL operands are identical. Pure crash fix, safe to enable unconditionally.

**3. Silent result changes — must be gated behind an explicit option**

SQL:
- Comparison / set predicates: `=`, `<>`, `>`, `>=`, `<`, `<=`, `IN`, `NOT IN`, `BETWEEN`, `NOT BETWEEN`. On a NULL operand the result changes from `false`/`true` to `NULL`. For negated forms (`<>`, `NOT IN`, `NOT BETWEEN`) this is `true → NULL`, which **drops rows in `WHERE`** and **flips `CASE` / `IF` branches**.
- `LIKE` / `NOT LIKE`: pattern engine changes from Java regex (`Pattern.find()`, substring match) to SQL wildcards (`%`, `_`, whole-string match). This changes results for **every row, not just NULL** — e.g. `'xabcy' LIKE 'abc'` goes `true → false`, `col LIKE 'A%'` goes `false → true`.

This is a backward-incompatible change to user-facing semantics, and the silent nature makes it dangerous: after the upgrade the YAML is unchanged, the job stays healthy, and no error or warning is emitted, yet the output data changes. For a CDC sync, altered filter results change which `INSERT` / `UPDATE` / `DELETE` events reach the sink, so the sink silently diverges from the source with no signal to detect or repair it. It therefore must not be the default on a minor/patch release.

**Recommendation**

Introduce `pipeline.transform.expression.semantics` (enum `LEGACY` | `FLINK_SQL`, default `LEGACY`). Keep the current behavior under `LEGACY`; apply bucket 3 only under `FLINK_SQL`. Buckets 1 and 2 remain always-on.

Also, to ease review, consider splitting this so buckets 1–2 and bucket 3 land as separate commits (or separate PRs) — the crash/parse fixes can then be merged quickly, while the semantic change is reviewed on its own.

haruki-830:
> Also, to ease review, consider splitting this so buckets 1–2 and bucket 3 land as separate commits (or separate PRs) — the crash/parse fixes can then be merged quickly, while the semantic change is reviewed on its own.

Thanks for the detailed review and the compatibility analysis. I agree with the split.

I will separate this work into two PRs:
1. The first PR will only include the always-on changes with no silent behavior change:
   - newly supported operators that are currently unparseable, such as `IS [NOT] DISTINCT FROM`, `IS [NOT] UNKNOWN`, `SIMILAR TO` / `NOT SIMILAR TO`, and `LIKE ... ESCAPE`
   - NULL boolean crash fixes for `AND`, `OR`, `NOT`, `IS [NOT] TRUE`, and `IS [NOT] FALSE`
2. The second PR will handle the backward-incompatible semantic changes behind an explicit option:
   - comparison predicates, `IN`, `BETWEEN`, and legacy `LIKE` / `NOT LIKE`
   - add `pipeline.transform.expression.semantics` with `LEGACY` as the default and `FLINK_SQL` as the opt-in SQL-compatible behavior

This should allow the parse/crash fixes to move forward independently, while keeping the silent result-changing semantics isolated and explicitly gated.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -322,6 +334,21 @@ private static Java.Rvalue generateUnaryOperation(
...
+        return new Java.MethodInvocation(Location.NOWHERE, null, functionName, atoms);
+    }
+
+    private static Java.Rvalue generateLazyBinaryFunctionOperation(
+            Context context, SqlBasicCall sqlBasicCall, String functionName, Java.Rvalue[] atoms) {
+        if (atoms.length != 2) {
+            throw new ParseException("Unrecognized expression: " + sqlBasicCall.toString());
+        }
+        Java.Rvalue rightOperandSupplier =
+                new Java.AmbiguousName(Location.NOWHERE, new String[] {"() -> " + atoms[1]});
```
lvyanquan: Could we add an end-to-end compilation test to `JaninoCompilerTest`?

  The existing tests only verify the generated expression string or invoke
  `LogicalFunctions` directly. They do not compile the generated expression with
  Janino, so they cannot detect that Janino 3.1.10 does not support lambda
  expressions.

lvyanquan:
Please update both the English and Chinese transform documentation to reflect these changes

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
haruki-830: Addressed in the latest commit.

lvyanquan:
Also cc’ing @yuxiqian.

docs/content.zh/docs/core-concept/transform.md
```diff
@@ -120,8 +120,8 @@ Flink CDC 使用 [Calcite](https://calcite.apache.org/) 来解析表达式并且
 | value1 >= value2                     | greaterThanOrEqual(value1, value2)           | 如果 value1 大于等于 value2，返回 TRUE；如果 value1 或 value2 为 NULL，返回 FALSE。 |
 | value1 < value2                      | lessThan(value1, value2)                     | 如果 value1 小于 value2，返回 TRUE；如果 value1 或 value2 为 NULL，返回 FALSE。   |
 | value1 <= value2                     | lessThanOrEqual(value1, value2)              | 如果 value1 小于等于 value2，返回 TRUE；如果 value1 或 value2 为 NULL，返回 FALSE。 |
-| value IS NULL                        | null == value                                | 如果 value 为 NULL，返回 TRUE。                                      |
-| value IS NOT NULL                    | null != value                                | 如果 value 不为 NULL，返回 TRUE。                                     |
+| value IS NULL                        | isNull(value)                                | 如果 value 为 NULL，返回 TRUE。                                      |
+| value IS NOT NULL                    | isNotNull(value)                             | 如果 value 不为 NULL，返回 TRUE。                                     |
```
lvyanquan, line 124: The following functions are not documented:

  - IS DISTINCT FROM
  - IS NOT DISTINCT FROM
  - LIKE ... ESCAPE
  - NOT LIKE ... ESCAPE
  - SIMILAR TO
  - NOT SIMILAR TO
  - SIMILAR TO ... ESCAPE
  - NOT SIMILAR TO ... ESCAPE

Moreover, the two forms of LIKE are handled differently:

  - Two-argument LIKE: uses Java regular expression semantics.
  - LIKE with ESCAPE: uses SQL % and _ wildcard semantics.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -322,6 +334,27 @@ private static Java.Rvalue generateUnaryOperation(
...
+    private static Java.Rvalue generateLazyBinaryFunctionOperation(
+            Context context, SqlBasicCall sqlBasicCall, String functionName, Java.Rvalue[] atoms) {
+        if (atoms.length != 2) {
+            throw new ParseException("Unrecognized expression: " + sqlBasicCall.toString());
+        }
+        Java.Rvalue rightOperandSupplier =
+                new Java.AmbiguousName(
+                        Location.NOWHERE,
+                        new String[] {
+                            "new java.util.function.Supplier<Boolean>() { public Boolean get() { return "
```
lvyanquan, line 351: A possible optimization is to generate different code based on operand nullability:

- If both operands are non-null, use native &&/||.
- If only the left operand is non-null, use a conditional expression:
      - AND: left ? right : Boolean.FALSE
      - OR: left ? Boolean.TRUE : right
- If the left operand is nullable, keep the current Supplier fallback to preserve three-valued logic and short-circuit evaluation.

This avoids unnecessary per-record Supplier allocations in common non-null cases.

flink-cdc-composer/src/test/resources/specs/comparison.yaml
```diff
@@ -165,6 +165,26 @@
...
+- do: Distinct From Op
+  projection: id_, int_ IS DISTINCT FROM 4 AS comp_1, int_ IS NOT DISTINCT FROM 4 AS comp_2, CAST(NULL AS INT) IS DISTINCT FROM int_ AS comp_3, CAST(NULL AS INT) IS NOT DISTINCT FROM int_ AS comp_4
+  primary-key: id_
+  expect: |-
+    CreateTableEvent{tableId=foo.bar.baz, schema=columns={`id_` BIGINT NOT NULL 'Identifier',`comp_1` BOOLEAN,`comp_2` BOOLEAN,`comp_3` BOOLEAN,`comp_4` BOOLEAN}, primaryKeys=id_, options=()}
+    DataChangeEvent{tableId=foo.bar.baz, before=[], after=[1, false, true, true, false], op=INSERT, meta=()}
+    DataChangeEvent{tableId=foo.bar.baz, before=[1, false, true, true, false], after=[-1, true, false, true, false], op=UPDATE, meta=()}
+    DataChangeEvent{tableId=foo.bar.baz, before=[-1, true, false, true, false], after=[], op=DELETE, meta=()}
+    DataChangeEvent{tableId=foo.bar.baz, before=[], after=[0, true, false, false, true], op=INSERT, meta=()}
+    DataChangeEvent{tableId=foo.bar.baz, before=[0, true, false, false, true], after=[], op=DELETE, meta=()}
```
lvyanquan, line 187: Lacks end-to-end tests for:
  - NOT LIKE ... ESCAPE
  - SIMILAR TO ... ESCAPE
  - NOT SIMILAR TO ... ESCAPE
  - NULL inputs for these expressions

haruki-830:
@lvyanquan Thanks for the detailed feedback.

I split the follow-up work into separate JIRA issues:

- https://issues.apache.org/jira/browse/FLINK-40203
- https://issues.apache.org/jira/browse/FLINK-40204

For this PR, I kept the scope limited to the always-on changes that do not introduce silent behavior changes. The backward-incompatible semantic alignment work, such as comparison / IN / BETWEEN / two-argument LIKE behavior, will be handled under the explicit semantics option in a follow-up PR.

For nullable AND / OR, I applied the localized optimization suggested here: native `&&` / `||` for non-null operands, conditional expressions when only the left operand is non-null, and the existing lazy fallback when the left operand is nullable. Avoiding that fallback completely would require statement-level codegen, so I tracked it separately instead of expanding this PR into an expression translator refactor.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
yuxiqian: We may add a test spec like `true OR 1/0`, `false AND 1/0` to ensure the short-circuit logic is effective.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -322,6 +335,131 @@ private static Java.Rvalue generateUnaryOperation(
...
+            return isIdentifierNullable(context, (SqlIdentifier) sqlNode);
+        }
+        if (sqlNode instanceof SqlLiteral) {
+            return ((SqlLiteral) sqlNode).getValue() == null;
+        }
+        if (sqlNode instanceof SqlBasicCall) {
+            return isBasicCallNullable(context, (SqlBasicCall) sqlNode);
+        }
+        return true;
+    }
```
yuxiqian, line 406: Instead of deducing `isIdentifierNullable` at the Janino compiler level, can we check the corresponding `RelNode` nullability during Calcite validation?

Current implementation is OK though, as the compilation happens just once.

yuxiqian (commented):
Thanks Haruki for the contribution, just left some minor comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
haruki-830: Thanks for the suggestion. I agree that deriving nullability from Calcite/RelNode would be cleaner.

Since the current check only runs once during expression compilation and is conservative for unknown expressions, I kept it local in this PR to avoid expanding the scope. I think it would be better to revisit this in the follow-up PR for statement-level Transform expression codegen.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
haruki-830: Added them in the latest commit. Thanks!

lvyanquan (approved):
+1.

PR #4478, opened 2026-07-17
[FLINK-40161] Log original MySQL DDL events before parsing

haruki-830:
#### Summary

This commit adds logging for original MySQL DDL events before they are parsed by the Flink CDC MySQL source. When DDL parsing fails, users can now find the original DDL statement together with its binlog coordinates in the logs, making troubleshooting easier.

#### Key Changes

1. Log Original MySQL DDL Events
- Added info-level logging in `MySqlStreamingChangeEventSource` before calling `parseStreamingDdl(...)`.
- Logs the database name and original DDL SQL.
- Logs binlog file name and event start/end positions when the event header provides them.

2. Keep ROLLBACK Handling Separate
- Keeps the existing dedicated warning path for `ROLLBACK`.
- Avoids logging `ROLLBACK` as a DDL event because it is not a schema change.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40161](url)


haruki-830:
@lvyanquan PTAL.THANKS!

lvyanquan (approved):
+1.

PR #4480, opened 2026-07-21
[FLINK-40202] Fix ConcurrentModificationException when initializing enumerator metrics

naivedogger:
This closes FLINK-40202.

`HybridSplitAssigner#open()` initializes enumerator metrics *after* calling `snapshotSplitAssigner.open()`, which has already started an asynchronous splitting thread that mutates `remainingSplits`. `initEnumeratorMetrics()` then iterates that list without synchronization, causing the following exception:

image

The fix initializes the enumerator metrics before opening the snapshot assigner, so metrics are seeded before the background splitting thread starts.


loserwang1024 (approved):
LGTM! @leonardBang  ,please help check it again.

leonardBang:
+1 from my side

leonardBang (approved):

PR #4482, opened 2026-07-22
[FLINK-40204][runtime] Introduce statement-level codegen for YAML Transform expressions

haruki-830:
#### Summary

This commit introduces statement-level code generation for Flink CDC YAML Transform expressions. Transform expressions are now represented by a `GeneratedExpression` intermediate structure and compiled through Janino `ScriptEvaluator`, allowing nullable `AND` / `OR` short-circuit logic to be generated as explicit `if` / `else` statements instead of per-record `Supplier` wrappers.

#### Key Changes

1. Introduce Statement-Level Transform Codegen
- Added `GeneratedExpression` to carry generated statements, result term, and result Java type.
- Added `TransformExpressionEvaluator` as a small wrapper around Janino `ScriptEvaluator`.
- Updated projection and filter processors to compile and evaluate generated scripts.

2. Support Statement-Level Nullable AND / OR
- Added `translateSqlNodeToGeneratedExpression(...)` in `JaninoCompiler`.
- Generates explicit short-circuit `if` / `else` blocks for nullable `AND` and `OR`.
- Ensures the right operand is evaluated only when required.

3. Improve Generated Script Diagnostics
- Updated compile/evaluation errors and debug output to print the generated script instead of only the result term such as `result$0`.
- Added validation at the script compilation entry to reject empty result terms before passing generated scripts to Janino.

4. Test Coverage
- Added coverage for nullable `AND` / `OR` SQL three-valued logic.
- Added short-circuit tests for:
  - `false AND (1 / 0 > 0)`
  - `true OR (1 / 0 > 0)`
- Added projection and filter coverage.
- Added assertions that generated code no longer contains `new Supplier`.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40204](url)

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/GeneratedExpression.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/GeneratedExpression.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.cdc.runtime.parser;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/** Statement-level Java code generated from a transform expression. */
+public class GeneratedExpression implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private final String code;
+    private final String resultTerm;
+    private final Class<?> resultClass;
+
+    private GeneratedExpression(String code, String resultTerm, Class<?> resultClass) {
+        this.code = Objects.requireNonNull(code);
+        this.resultTerm = Objects.requireNonNull(resultTerm);
+        this.resultClass = Objects.requireNonNull(resultClass);
+    }
+
+    public String getCode() {
+        return code;
+    }
+
+    public String getResultTerm() {
+        return resultTerm;
+    }
+
+    public Class<?> getResultClass() {
+        return resultClass;
+    }
+
+    public String asScript() {
+        StringBuilder script = new StringBuilder();
+        appendCode(script, code);
+        script.append("return ").append(resultTerm).append(";");
+        return script.toString();
+    }
+
+    public static GeneratedExpression of(String code, String resultTerm, Class<?> resultClass) {
+        return new GeneratedExpression(code, resultTerm, resultClass);
+    }
+
+    public static GeneratedExpression fromExpression(String resultTerm, Class<?> resultClass) {
+        return of("", resultTerm, resultClass);
+    }
+
+    private static void appendCode(StringBuilder builder, String code) {
+        if (code.isEmpty()) {
+            return;
+        }
+        builder.append(code);
+        if (code.charAt(code.length() - 1) != '\n') {
+            builder.append('\n');
+        }
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        GeneratedExpression that = (GeneratedExpression) o;
+        return code.equals(that.code)
+                && resultTerm.equals(that.resultTerm)
+                && resultClass.equals(that.resultClass);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(code, resultTerm, resultClass);
+    }
+
+    @Override
+    public String toString() {
+        return "GeneratedExpression{"
+                + "code='"
+                + code
+                + '\''
+                + ", resultTerm='"
+                + resultTerm
+                + '\''
+                + ", resultClass="
+                + resultClass
+                + '}';
+    }
+}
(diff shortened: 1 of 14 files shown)


haruki-830:
@lvyanquan PTAL when you have time, Thanks!

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -804,6 +904,265 @@ private static String generateInvokeExpression(UserDefinedFunctionDescriptor udf
...
+                for (SqlNode node : (SqlNodeList) sqlNode) {
+                    translateSqlNodeToGeneratedAtoms(node, atoms);
+                }
+            } else if (sqlNode instanceof SqlIdentifier
+                    || sqlNode instanceof SqlLiteral
+                    || sqlNode instanceof SqlBasicCall
+                    || sqlNode instanceof SqlCase) {
+                atoms.add(translate(sqlNode, deduceGeneratedExpressionClass(context, sqlNode)));
+            }
+        }
```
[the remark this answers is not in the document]
haruki-830: Addressed in the latest commit.

lvyanquan:
This change offers very limited benefits but has a relatively large impact surface. I think we need to evaluate it carefully.

CC @yuxiqian

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/GeneratedExpression.java
```diff
@@ -0,0 +1,106 @@
...
+    public Class<?> getResultClass() {
+        return resultClass;
+    }
+
+    public String asScript() {
+        StringBuilder script = new StringBuilder();
+        appendCode(script, code);
+        script.append("return ").append(resultTerm).append(";");
+        return script.toString();
+    }
```
yuxiqian, line 54: Better avoid string manipulations on dumped Java codes. Is `Java.ReturnStatement` useful here?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/GeneratedExpression.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private static void appendCode(StringBuilder builder, String code) {
+        if (code.isEmpty()) {
+            return;
+        }
+        builder.append(code);
+        if (code.charAt(code.length() - 1) != '\n') {
+            builder.append('\n');
+        }
+    }
```
yuxiqian, line 72: Ditto, avoid string ops

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -804,6 +916,418 @@ private static String generateInvokeExpression(UserDefinedFunctionDescriptor udf
...
+                return translateSqlBasicCall((SqlBasicCall) sqlNode, resultClass);
+            }
+            if (sqlNode instanceof SqlCase) {
+                return translateSqlCase((SqlCase) sqlNode, resultClass);
+            }
+            Java.Rvalue rvalue = translateSqlNodeToJaninoRvalue(context, sqlNode);
+            if (rvalue == null) {
+                throw new ParseException("Unrecognized expression: " + sqlNode);
+            }
+            return GeneratedExpression.fromExpression(rvalue.toString(), resultClass);
```
yuxiqian, line 938: `rvalue.toString()` is for debugging purposes only and may not generate Java code with correct parentheses.

For example, SQL statement `(a - b) / (c + 1)` will be converted to `a - b / c + 1`.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -496,6 +511,103 @@ private static Java.Rvalue generateEqualsOperation(
...
+            }
+            return Double.class;
+        }
+        if (sqlLiteral.getTypeName() == SqlTypeName.BOOLEAN) {
+            return Boolean.class;
+        }
+        return Object.class;
+    }
+
+    private static boolean isBooleanResultCall(SqlBasicCall sqlBasicCall) {
```
yuxiqian, line 580: One may write a UDF returning `BOOLEAN`?

yuxiqian (commented):
Thanks Haruki for the contribution!

I also doubt the necessity of such changes. Deducing types from SqlNode and manipulating generated code strings looks fragile. Is there any alternative solution to implement "short-circuit call"?

haruki-830:
Thanks @lvyanquan and @yuxiqian for the feedback. 

I ran a JMH benchmark comparing the Supplier path against an equivalent if/else script and found no statistically significant difference — all configurations fall in the 330k-372k ops/ms range with overlapping confidence intervals, that means the JIT escape analysis reliably eliminates the anonymous `Supplier` allocation, so the per-record allocation concern is not a real bottleneck in practice.

Given the lack of measurable benefit and the correctness risks @yuxiqian identified, I agree this change is not warranted. I'll go ahead and close this PR later.

yuxiqian:
> Thanks @lvyanquan and @yuxiqian for the feedback.
> 
> I ran a JMH benchmark comparing the Supplier path against an equivalent if/else script and found no statistically significant difference — all configurations fall in the 330k-372k ops/ms range with overlapping confidence intervals, that means the JIT escape analysis reliably eliminates the anonymous `Supplier` allocation, so the per-record allocation concern is not a real bottleneck in practice.
> 
> Given the lack of measurable benefit and the correctness risks @yuxiqian identified, I agree this change is not warranted. I'll go ahead and close this PR later.

Thanks for the detailed benchmark report.

> rvalue.toString() is for debugging purposes only and may not generate Java code with correct parentheses.
>
> For example, SQL statement (a - b) / (c + 1) will be converted to a - b / c + 1.

Seems this is an existing bug, not introduced by changes in this PR. Could you please investigate this later?

haruki-830:
> 
> > rvalue.toString() is for debugging purposes only and may not generate Java code with correct parentheses.
> > For example, SQL statement (a - b) / (c + 1) will be converted to a - b / c + 1.
> 
> Seems this is an existing bug, not introduced by changes in this PR. Could you please investigate this later?

Thanks for pointing out this bug! I confirmed that `BinaryOperation.toString()` does lose parentheses when a sub-expression has lower precedence than its parent operator. This affects the existing `translateSqlNodeToJaninoExpression` path as well. 

I'll open a separate PR to fix it.

PR #4484, opened 2026-07-23
[FLINK-40203][runtime] Add an explicit option for Flink SQL-compatible YAML Transform expression semantics

haruki-830:
#### Summary

This commit adds an explicit expression semantics option to YAML Transform. The default `LEGACY` mode preserves existing pipeline behavior, while the opt-in `FLINK_SQL` mode aligns supported predicates with Flink SQL semantics.

#### Key Changes

1. Configurable Transform Expression Semantics
- Added the `pipeline.transform.expression.semantics` option.
- Added `LEGACY` as the default value to preserve backward compatibility.
- Added `FLINK_SQL` as an opt-in mode for Flink SQL-compatible expression evaluation.
- Propagated the configured semantics through the composer and Transform runtime.

2. Flink SQL-Compatible Predicate Semantics
- Added SQL three-valued logic for `=`, `<>`, `<`, `<=`, `>`, and `>=`.
- Added SQL-compatible `BETWEEN` / `NOT BETWEEN` and `IN` / `NOT IN` behavior.
- Represents SQL `UNKNOWN` as a nullable `Boolean`.
- Changed two-argument `LIKE` / `NOT LIKE` to use SQL `%` and `_` wildcards with whole-string matching.
- Keeps rows in Transform filters only when the result is `TRUE`, while projections preserve `UNKNOWN` as `NULL`.

- Added runtime unit tests for both legacy and Flink SQL semantics.
- Added YAML Transform specs covering comparison, `BETWEEN`, `IN`, `LIKE`, and row-level filtering with `UNKNOWN`.
- Added YAML configuration parsing coverage for the new option and its default value.
- Documented the configuration and behavioral differences in both English and Chinese.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40203](url)


docs/content.zh/docs/core-concept/data-pipeline.md
```diff
@@ -119,10 +119,27 @@ under the License.
...
 | `route-mode`                  | [路由]({{< ref "docs/core-concept/route" >}}#路由模式)规则的匹配模式。可选值：`ALL_MATCH`（默认，应用所有匹配的规则）或 `FIRST_MATCH`（只应用第一个匹配的规则）。                                                                                                                                                                                                                                                                                                                    | optional          |
+| `transform.expression.semantics` | Transform 表达式的求值语义。可选值：`LEGACY`（默认，保持现有行为）或 `FLINK_SQL`（对支持的谓词使用与 Flink SQL 兼容的语义）。                                                                                                                                                                                                                                                                                                                                                 | optional          |
```
yuxiqian: Call it DEFAULT or something similar? Legacy sounds like it will be removed soon.

haruki-830:
@lvyanquan @yuxiqian Would you take a look at this? Thanks!

docs/content.zh/docs/core-concept/data-pipeline.md
haruki-830: That makes sense. I’ll rename it to DEFAULT and update the documentation to clarify that it preserves the current behavior.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/impl/ComparisonFunctions.java
```diff
@@ -61,27 +72,80 @@ public static boolean greaterThan(Object lhs, Object rhs) {
         return universalCompares(lhs, rhs) > 0;
     }
 
+    public static Boolean sqlGreaterThan(Object lhs, Object rhs) {
+        if (lhs == null || rhs == null) {
+            return null;
+        }
+        return universalCompares(lhs, rhs) > 0;
```
lvyanquan: The existing `universalCompares` should remain unchanged to preserve `DEFAULT` behavior. However, the new `FLINK_SQL` path should use a SQL-specific numeric comparator to avoid `double` precision loss, cross-type equality failures such as `Long.equals(Integer)`, and scale-sensitive `BigDecimal.equals`. This comparator should be shared by equality, ordering, `BETWEEN`, and `IN`.

haruki-830: Addressed in the latest commit. Also rebased onto master while at it.

lvyanquan:
Hi @haruki-830 would you like to rebase to master to solve the conflicts.

leonardBang:
@haruki-830 Could you rebase this PR to latest master branch ?

PR #4488, opened 2026-07-24
[FLINK-40223][runtime] Support additional Flink SQL string functions in YAML Transform

haruki-830:
#### Summary

This commit expands the string function coverage of Flink CDC YAML Transform by adding commonly used Flink SQL string manipulation, search, predicate, and encoding functions. These functions can now be used in row-level transform projections and filters.

#### Key Changes

1. Additional String Function Support
- Added `LTRIM`, `RTRIM`, `BTRIM`, `OVERLAY`, `POSITION`, `LOCATE`, and `INSTR`.
- Added `CONCAT_WS`, `LPAD`, `RPAD`, `REPLACE`, `REPEAT`, `LEFT`, and `RIGHT`.
- Added `STARTSWITH`, `ENDSWITH`, `TO_BASE64`, and `FROM_BASE64`.
- Handles null inputs, Unicode code points, and function-specific boundary cases consistently with Flink SQL semantics.

2. Parser and Runtime Integration
- Registered the new functions with their operand and return type definitions in the transform operator table.
- Added parser handling for SQL-specific `OVERLAY` and `POSITION` syntax.
- Integrated the functions with Janino expression compilation for runtime evaluation.

- Added unit tests for function behavior, null handling, Unicode strings, and boundary cases.
- Added parser and transform runtime tests for valid and invalid expressions.
- Added YAML specification coverage for projection results, including matched, unmatched, and null inputs.
- Documented the new functions in both the English and Chinese transform documentation.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40223](url)


haruki-830:

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/impl/StringFunctions.java
```diff
@@ -176,6 +197,193 @@ public static String lower(String str) {
...
+        return Base64.getEncoder().encodeToString(str.getBytes(StandardCharsets.UTF_8));
+    }
+
+    public static String fromBase64(String str) {
+        if (str == null) {
+            return null;
+        }
+        return new String(
+                Base64.getDecoder().decode(str.getBytes(StandardCharsets.UTF_8)),
+                StandardCharsets.UTF_8);
```
yuxiqian, line 512: If decoded bytes are not valid UTF-8 string, it will be replaced by `U+FFFD`. Maybe we should provide a variant function for base64 string <-> bytes.

docs/content.zh/docs/core-concept/transform.md
```diff
@@ -177,10 +177,28 @@ Flink CDC 使用 [Calcite](https://calcite.apache.org/) 来解析表达式并且
...
+| RIGHT(string, integer)                                     | right(string, integer)                        | 返回 STRING 最右侧 integer 个字符。                                                                                      |
+| STARTSWITH(string1, string2)                               | startswith(string1, string2)                  | 返回 STRING1 是否以 STRING2 开头，支持字符字符串和二进制字符串。                                                                     |
+| ENDSWITH(string1, string2)                                 | endswith(string1, string2)                    | 返回 STRING1 是否以 STRING2 结尾，支持字符字符串和二进制字符串。                                                                     |
+| TO_BASE64(string)                                          | toBase64(string)                              | 将 STRING 编码为 base64 字符串。                                                                                      |
+| FROM_BASE64(string)                                        | fromBase64(string)                            | 解码 base64 字符串。                                                                                                 |
```
yuxiqian: Ditto, supporting FROM/TO_BASE64 with generic binary would be useful.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
```diff
@@ -196,12 +196,33 @@ public static SqlSelect parseSelect(String statement) {
...
     }
 
+    private static void replaceParserBoundStringOperators(SqlNode sqlNode) {
+        if (sqlNode instanceof SqlBasicCall) {
+            SqlBasicCall call = (SqlBasicCall) sqlNode;
+            if ("OVERLAY".equalsIgnoreCase(call.getOperator().getName())) {
+                call.setOperator(TransformSqlOperatorTable.OVERLAY);
+            } else if (call.getKind() == SqlKind.POSITION) {
+                call.setOperator(TransformSqlOperatorTable.POSITION);
+            }
```
yuxiqian: This would be unnecessary if we use Calcite standard `SqlStdOperatorTable.OVERLAY` and `.POSITION`.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/metadata/TransformSqlOperatorTable.java
```diff
@@ -174,6 +242,38 @@ public void lookupOperatorOverloads(
     public static final SqlFunction UPPER = SqlStdOperatorTable.UPPER;
     public static final SqlFunction LOWER = SqlStdOperatorTable.LOWER;
     public static final SqlFunction TRIM = SqlStdOperatorTable.TRIM;
+    public static final SqlFunction LTRIM =
```
yuxiqian, line 178: We may use definitions in SqlStdOperatorTable if possible.

yuxiqian (commented):
Thanks Haruki for the nice work, just left some minor comments.

haruki-830:
@yuxiqian Thanks for the review! I addressed the comments in the lastest commit.

flink-cdc-composer/src/test/resources/specs/string.yaml
```diff
@@ -147,3 +147,23 @@
     DataChangeEvent{tableId=foo.bar.baz, before=[-1, 天地玄黄宇宙洪荒, {{天地玄黄宇宙洪荒}}], after=[], op=DELETE, meta=()}
     DataChangeEvent{tableId=foo.bar.baz, before=[], after=[0, null, {{null}}], op=INSERT, meta=()}
     DataChangeEvent{tableId=foo.bar.baz, before=[0, null, {{null}}], after=[], op=DELETE, meta=()}
+- do: Additional String Search Functions
```
yuxiqian, line 150: Could you please ensure all new functions (and overloads) are covered here?

Splitting multiple spec blocks is favored for readability.

yuxiqian (commented):
+1 from my side (except one minor comment on test specs). Need another eye from Kunni.

docs/content/docs/core-concept/transform.md
```diff
@@ -178,10 +178,29 @@ Logical functions follow SQL three-valued logic for nullable BOOLEAN values. `AN
...
+| ENDSWITH(string1, string2)                                  | endswith(string1, string2)                       | Returns whether STRING1 ends with STRING2. Character and binary strings are supported.                                                                                                            |
+| TO_BASE64(string \| binary)                                 | toBase64(string \| binary)                       | Encodes a character or binary string to a base64 string.                                                                                                                                          |
+| FROM_BASE64(string)                                         | fromBase64(string)                               | Decodes a base64 string to a UTF-8 character string.                                                                                                                                               |
```
lvyanquan: Flink 2.2's current FROM_BASE64 runtime implementation preserves the decoded byte sequence in BinaryStringData, including bytes that are not valid UTF-8. For example, TO_BASE64(FROM_BASE64('wyg=')) returns 'wyg='.
This implementation converts the decoded bytes to java.lang.String, so malformed UTF-8 sequences are replaced with U+FFFD. Consequently, the same round trip returns '77+9KA=='.

Given that STRING values currently flow through java.lang.String in the transform expression engine, matching Flink's byte-preserving behavior would require broader runtime/type-conversion changes. Could we document that FROM_BASE64 assumes valid UTF-8 input and recommend FROM_BASE64_BINARY for arbitrary binary content? It would also be helpful to mention this intentional divergence from Flink 2.2 in the PR description and add a test covering non-UTF-8 decoded bytes.

lvyanquan (approved):
LGTM. Only left one minor suggestion.

docs/content/docs/core-concept/transform.md
haruki-830: Thanks for the suggestion! Addressed in the latest commit.

yuxiqian (approved):

PR #4491, opened 2026-07-27
[FLINK-40239][runtime] Support additional Flink SQL regular expression functions in YAML Transform

haruki-830:
#### Summary

This commit adds support for five additional Flink SQL regular expression functions in YAML Transform: `REGEXP_EXTRACT`, `REGEXP_EXTRACT_ALL`, `REGEXP_COUNT`, `REGEXP_INSTR`, and `REGEXP_SUBSTR`.

The functions use Java regular expression syntax and align with Flink SQL semantics for NULL values, invalid patterns, capture groups, and no-match cases.

#### Key Changes

1. Additional Regular Expression Functions
- Added implementations for the five missing regular expression functions.
- Supports capture group indexes in `REGEXP_EXTRACT` and `REGEXP_EXTRACT_ALL`.
- Uses the expected no-match behavior:
  - `REGEXP_EXTRACT` and `REGEXP_SUBSTR` return NULL.
  - `REGEXP_EXTRACT_ALL` returns an empty array.
  - `REGEXP_COUNT` and `REGEXP_INSTR` return 0.
- Invalid regular expressions and invalid capture group indexes return NULL.

2. YAML Transform Integration
- Registered the new functions in the Transform SQL operator table.
- Added return type inference for `ARRAY<STRING>` from `REGEXP_EXTRACT_ALL`.
- Verified that `REGEXP_EXTRACT_ALL` works correctly in YAML Transform projections.
- Existing `REGEXP_REPLACE` behavior remains unchanged.

3. Test Coverage
- Added unit tests covering NULL arguments, invalid patterns, no matches, capture groups, and invalid group indexes.
- Added parser tests for function registration and expression translation.
- Added projection tests covering the runtime conversion of `REGEXP_EXTRACT_ALL` results.

4. Documentation
- Documented the five new functions in both English and Chinese Transform documentation.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40239](url)


haruki-830:

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/impl/StringFunctions.java
```diff
@@ -62,6 +65,72 @@ public static String regexpReplace(String str, String regex, String replacement)
...
+    /** Returns a string extracted with a specified regular expression and capture group index. */
+    public static String regexpExtract(String str, String regex) {
+        return regexpExtract(str, regex, 0);
+    }
+
+    public static String regexpExtract(String str, String regex, Number extractIndex) {
+        if (extractIndex == null || extractIndex.longValue() < 0) {
+            return null;
+        }
+        Matcher matcher = getRegexpMatcher(str, regex);
```
lvyanquan, line 80: `getRegexpMatcher()` calls `Pattern.compile(regex)` for every record. Since regexes in YAML transforms are usually constant, this adds unnecessary CPU overhead to the per-record hot path. Please use a bounded `ThreadLocalCache<String, Pattern>` while preserving the current invalid-regex behavior.

haruki-830: Addressed in the latest commit.

lvyanquan (approved):
+1.

PR #4492, opened 2026-07-28
[FLINK-40240][runtime] Support TRY_CAST, IFNULL and NULLIF in YAML Transform

haruki-830:
#### Summary

This commit adds `TRY_CAST`, `IFNULL`, and `NULLIF` support to YAML Transform projections and filters. It aligns their behavior with Flink SQL while preserving the existing `CAST` failure semantics.

#### Key Changes

1. Function and Syntax Support
- Added operator and type metadata for `TRY_CAST`, `IFNULL`, and `NULLIF`.
- Added preprocessing for `TRY_CAST(expr AS type)` because Calcite 1.32 does not support this syntax directly.
- Handles nested expressions, different letter cases, quoted text, identifiers, and SQL comments.

2. Code Generation and Runtime Semantics
- Added Janino code generation for all three functions.
- `TRY_CAST` supports the same target types as the existing `CAST` implementation and returns `NULL` only for data conversion failures.
- Unsupported conversion paths fail during pipeline validation, while invalid configuration, UDF, and internal errors continue to fail the task.
- Added null-safe `IFNULL` and numeric-aware/deep-equality `NULLIF` evaluation without changing existing `CAST` behavior.

- Added parser, code generation, runtime, and YAML specification coverage for projections and filters.
- Covered successful and failed casts, nested expressions, nullability, invalid syntax, and function combinations.
- Verified the changes with both default Flink 1.20 and the Flink 2 profile.
- Updated the English and Chinese Transform documentation.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40240](https://issues.apache.org/jira/browse/FLINK-40240)

### DIFF
--- a/docs/content.zh/docs/core-concept/transform.md
+++ b/docs/content.zh/docs/core-concept/transform.md
@@ -244,10 +244,14 @@ Flink CDC 使用 [Calcite](https://calcite.apache.org/) 来解析表达式并且
 | CASE WHEN condition1 THEN result1 (WHEN condition2 THEN result2)* (ELSE result_z) END                                 | 嵌套三元表达式                          | 当第一个 conditionX 满足时，返回 resultX。如果没有条件满足，如果提供了 result_z 则返回 result_z，否则返回 NULL。                              |
 | COALESCE(value1 [, value2]*)                                                                                          | coalesce(Object... objects)      | 返回第一个不为 NULL 的参数。如果所有参数都为 NULL，则返回 NULL。返回类型是所有参数中限制最少的公共类型。如果所有参数都可为空，则返回类型也可为空。                          |
 | IF(condition, true_value, false_value)                                                                                | condition ? true_value : false_value | 如果条件满足，返回 true_value，否则返回 false_value。例如，IF(5 > 3, 5, 3) 返回 5。                                              |
+| IFNULL(value, replacement)                                                                                           | ifNull(value, replacement)       | 当 value 为 NULL 时返回 replacement，否则返回 value。两个参数必须存在公共类型。仅当 replacement 可为 NULL 时，结果才可为 NULL。                  |
+| NULLIF(value1, value2)                                                                                               | nullIf(value1, value2)           | 当 value1 与 value2 相等时返回 NULL，否则返回 value1。返回类型为 value1 对应的可空类型。                                         |
 
 ## 转换函数
 
 你可以使用 `CAST( <EXPR> AS <T> )` 语法将任何有效的表达式 `<EXPR>` 转换为特定类型 `<T>`。可能的转换路径如下：
+
+`TRY_CAST( <EXPR> AS <T> )` 支持与 `CAST` 相同的转换路径。当某个值无法转换时返回 NULL；不支持的源类型到目标类型转换路径会在 pipeline 校验或初始化阶段被拒绝。配置错误、UDF 异常和内部程序错误不会被转换为 NULL。
 
 | 源类型                                 | 目标类型      | 说明                                               |
 |-------------------------------------|-----------|--------------------------------------------------|
--- a/docs/content.zh/docs/core-concept/transform.md
+++ b/docs/content.zh/docs/core-concept/transform.md
@@ -247,6 +247,8 @@ Flink CDC 使用 [Calcite](https://calcite.apache.org/) 来解析表达式并且
 | IFNULL(value, replacement)                                                                                           | ifNull(value, replacement)       | 当 value 为 NULL 时返回 replacement，否则返回 value。两个参数必须存在公共类型。仅当 replacement 可为 NULL 时，结果才可为 NULL。                  |
 | NULLIF(value1, value2)                                                                                               | nullIf(value1, value2)           | 当 value1 与 value2 相等时返回 NULL，否则返回 value1。返回类型为 value1 对应的可空类型。                                         |
 
+`NULLIF` 会对数值参数进行跨数值类型比较。例如，`NULLIF(CAST(1 AS INT), CAST(1 AS BIGINT))` 返回 NULL。该行为与当前 Transform `=` 运算符的类型敏感相等比较不同；对于上述混合类型比较，`=` 返回 FALSE。
+
 ## 转换函数
 
 你可以使用 `CAST( <EXPR> AS <T> )` 语法将任何有效的表达式 `<EXPR>` 转换为特定类型 `<T>`。可能的转换路径如下：
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/impl/CastingFunctions.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/impl/CastingFunctions.java
@@ -264,7 +264,51 @@ public class CastingFunctions {
     }
 
     public static LocalDateTime castToTimestamp(Object object, String timezone) {
-        ZoneId zoneId = ZoneId.of(timezone);
+        return castToTimestamp(object, ZoneId.of(timezone), false);
+    }
+
+    public static String tryCastToString(Object object) {
+        return castToString(object);
+    }
+
+    public static Boolean tryCastToBoolean(Object object) {
+        return castToBoolean(object);
+    }
+
+    public static Byte tryCastToByte(Object object) {
+        return castToByte(object);
+    }
+
+    public static Short tryCastToShort(Object object) {
+        return castToShort(object);
+    }
+
+    public static Integer tryCastToInteger(Object object) {
+        return castToInteger(object);
+    }
+
+    public static Long tryCastToLong(Object object) {
+        return castToLong(object);
+    }
+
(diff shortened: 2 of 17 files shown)


haruki-830:

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/impl/CastingFunctions.java
```diff
@@ -264,7 +264,51 @@ public static BigDecimal castToBigDecimal(Object object, int precision, int scal
...
+
+    public static Float tryCastToFloat(Object object) {
+        return castToFloat(object);
+    }
+
+    public static Double tryCastToDouble(Object object) {
+        return castToDouble(object);
+    }
+
+    public static BigDecimal tryCastToBigDecimal(Object object, int precision, int scale) {
```
lvyanquan, line 363: Except for TIMESTAMP, all newly added tryCastToXxx methods simply delegate to the existing castToXxx implementations. Therefore, TRY_CAST cannot reliably detect conversion failures or distinguish its behavior from CAST.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/impl/LogicalFunctions.java
```diff
@@ -102,4 +104,45 @@ public static Object coalesce(Object... objects) {
...
+        if (left instanceof Double || right instanceof Double) {
+            double leftValue = left.doubleValue();
+            double rightValue = right.doubleValue();
+            if (Double.isFinite(leftValue) && Double.isFinite(rightValue)) {
+                return BigDecimal.valueOf(leftValue).compareTo(BigDecimal.valueOf(rightValue)) == 0;
+            }
+            return Double.compare(leftValue, rightValue) == 0;
+        }
+        if (left instanceof Float || right instanceof Float) {
+            return Float.compare(left.floatValue(), right.floatValue()) == 0;
```
lvyanquan: Float.compare and Double.compare do not match SQL numeric equality semantics. In particular, they treat -0.0 and 0.0 as different and treat two NaN values as equal. As a result, NULLIF(-0.0, 0.0) incorrectly returns the first value, while NULLIF(NaN, NaN) incorrectly returns NULL.

Please compare floating-point operands using primitive numeric equality after coercing them to the common numeric type, rather than using Float.compare or Double.compare.

Tests covering signed zero, NaN, and infinity should also be added.

lvyanquan: The Float branch in `valuesEqualForNullIf` downgrades the other side via `floatValue()`, so `nullIf(16_777_217L, 16_777_217f)` wrongly returns null — beyond float's 24-bit mantissa (2²⁴), values get rounded and compared equal, contradicting both SQL semantics (which promotes to DOUBLE) and the `=` operator here. Since `float → double` is lossless, please promote to double like the Double branch above does. A regression test for `NULLIF(BIGINT, FLOAT)` with values beyond 2²⁴ would be appreciated.

lvyanquan (changes requested):

haruki-830:
@lvyanquan Thanks for the careful review! I’ve addressed the requested changes in the latest commit. The branch has also been rebased onto the latest master. Please take another look when you have time.

yuxiqian:
Could you please rebase with latest master and resolve conflicts if any?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -795,6 +839,10 @@ private static Java.Rvalue generateOtherFunctionOperation(
             } else {
                 throw new ParseException("Unrecognized expression: " + sqlBasicCall);
             }
+        } else if (operationName.equals("IFNULL")) {
+            return generateIfNullOperation(context, sqlBasicCall, atoms);
+        } else if (operationName.equals("NULLIF")) {
+            return generateNullIfOperation(context, sqlBasicCall, atoms);
```
lvyanquan, line 859: The new IFNULL/NULLIF/TRY_CAST branches are checked before the UDF lookup, which breaks the existing UDF-first convention (see the orElseGet fallback below — that's why UPPER/CONCAT never collide with same-named UDFs). 

I verified with a probe: a UDF named ifnull/try_cast that compiled fine before this PR now fails with Unrecognized expression, and a UDF named nullif is silently replaced by the built-in. 

Please guard the three dispatch points (the IFNULL/NULLIF branches here, the TRY_CAST check in translateSqlNodeToJaninoRvalue, and case `NULLIF` in the SqlKind switch) so registered UDFs take precedence, which also matches Flink SQL's function resolution order.

docs/content.zh/docs/core-concept/transform.md
```diff
@@ -244,11 +244,15 @@ Flink CDC 使用 [Calcite](https://calcite.apache.org/) 来解析表达式并且
...
 | COALESCE(value1 [, value2]*)                                                                                          | coalesce(Object... objects)      | 返回第一个不为 NULL 的参数。如果所有参数都为 NULL，则返回 NULL。返回类型是所有参数中限制最少的公共类型。如果所有参数都可为空，则返回类型也可为空。                          |
 | IF(condition, true_value, false_value)                                                                                | condition ? true_value : false_value | 如果条件满足，返回 true_value，否则返回 false_value。例如，IF(5 > 3, 5, 3) 返回 5。                                              |
+| IFNULL(value, replacement)                                                                                           | ifNull(value, replacement)       | 当 value 为 NULL 时返回 replacement，否则返回 value。两个参数必须存在公共类型。仅当 replacement 可为 NULL 时，结果才可为 NULL。                  |
+| NULLIF(value1, value2)                                                                                               | nullIf(value1, value2)           | 当 value1 与 value2 相等时返回 NULL，否则返回 value1。返回类型为 value1 对应的可空类型。                                         |
```
lvyanquan, line 248: NULLIF's numeric cross-type equality matches Flink SQL (1 = 1L → TRUE), but the existing `=` operator in this engine uses strictequals(`valueEquals`), so 1 = 1L is false here — users may observe WHERE a = b false while NULLIF(a, b) IS NULL true. Since fixing `=` is out of scope for this PR, could this divergence be documented in transform.md, and ideally tracked in a follow-up JIRA for aligning valueEquals with Flink SQL?

lvyanquan (changes requested):

docs/content.zh/docs/core-concept/transform.md
haruki-830: Thanks for pointing this out! [FLINK-40203](https://issues.apache.org/jira/browse/FLINK-40203) / PR [#4484](https://github.com/apache/flink-cdc/pull/4484) is introducing an opt-in `FLINK_SQL` expression semantics mode, under which `=` compares numeric operands across numeric types and is therefore consistent with `NULLIF`. The default mode intentionally preserves the existing type-sensitive `=` behavior for backward compatibility. In the latest commit, I have documented this difference rather than duplicating the comparison-semantics changes from FLINK-40203.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
```diff
@@ -564,7 +569,9 @@ private static boolean isIdentifierNullable(Context context, SqlIdentifier sqlId
     }
 
     private static boolean isBasicCallNullable(Context context, SqlBasicCall sqlBasicCall) {
-        if (sqlBasicCall.getOperator().getName().equalsIgnoreCase("IFNULL")) {
+        if (sqlBasicCall.getOperator().getName().equalsIgnoreCase("IFNULL")
```
lvyanquan: Thanks for the fix — the precedence behavior is now correct and well tested. 

One structural concern though: the current approach guards each built-in branch with a negative check (`!findUserDefinedFunction(...).isPresent()`), scattered across four places (the `TRY_CAST` entry, `case NULLIF:` in the SqlKind switch, the IFNULL/NULLIF branches here, and `isBasicCallNullable`). This inverts the established convention in this method — the `orElseGet` fallback below is structurally "UDF first, built-in fallback", whereas the fix is "built-in first, exclude UDFs case by case".

Would you consider consolidating this into a single positive dispatch point at the top of function-call translation? Something like:

```java
Optional<UserDefinedFunctionDescriptor> udf = findUserDefinedFunction(context, functionName);
if (udf.isPresent()) {
    return generateUdfInvocation(context, udf.get(), atoms);
}
// built-in special cases (TRY_CAST / IF / IFNULL / NULLIF / ...) below
```

lvyanquan (approved):
Overall LGTM, but I'd suggest staying consistent with the existing implementation style rather than special-casing the newly added functions.

lvyanquan:
Thanks @haruki-830 for the quick response!

PR #4493, opened 2026-08-01
[FLINK-40274][cdc-cli] Remove unsupported kubernetes-session from target option help message

jx2lee:

This PR corrects the Flink CDC CLI help text for the `--target` option.

The help message listed `kubernetes-session` as a supported deployment target, but
Flink CDC currently supports `kubernetes-application` only for Kubernetes-specific
submission. Kubernetes session clusters can still be used by starting the Flink
session cluster separately and submitting through the existing `remote` target.

## Brief change log

- Removed the unsupported `kubernetes-session` value from the `--target` help text.
- Kept the help message aligned with the deployment targets defined by `ComposeDeployment`.

---

## Verifying this change

This change can be verified as follows:

- Run `mvn --batch-mode --no-transfer-progress -pl flink-cdc-cli -am test`
- Check `flink-cdc.sh --help` and confirm that the `--target` option no longer lists `kubernetes-session`.

## Documentation

---

- [X] Yes (OpenAI Codex)

Generated-by: OpenAI Codex

### DIFF
--- a/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontendOptions.java
+++ b/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontendOptions.java
@@ -51,8 +51,12 @@ public class CliFrontendOptions {
                     .longOpt("target")
                     .hasArg()
                     .desc(
-                            "The deployment target for the execution. This can take one of the following values "
-                                    + "local/remote/yarn-session/yarn-application/kubernetes-application")
+                            "The deployment target for the execution. This can take one of the following values:\n"
+                                    + "- local\n"
+                                    + "- remote\n"
+                                    + "- yarn-session\n"
+                                    + "- yarn-application\n"
+                                    + "- kubernetes-application")
                     .build();
 
     public static final Option USE_MINI_CLUSTER =
--- a/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontendOptions.java
+++ b/flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontendOptions.java
@@ -52,8 +52,7 @@ public class CliFrontendOptions {
                     .hasArg()
                     .desc(
                             "The deployment target for the execution. This can take one of the following values "
-                                    + "local/remote/yarn-session/yarn-application/kubernetes-session/kubernetes"
-                                    + "-application")
+                                    + "local/remote/yarn-session/yarn-application/kubernetes-application")
                     .build();
 
     public static final Option USE_MINI_CLUSTER =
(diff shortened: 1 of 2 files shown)


featzhang (approved):

flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontendOptions.java
```diff
@@ -52,8 +52,7 @@ public class CliFrontendOptions {
                     .hasArg()
                     .desc(
                             "The deployment target for the execution. This can take one of the following values "
```
yuxiqian: Shall we re-organize the description message with manual line breaking and ListElements?

jx2lee: @yuxiqian  Did you mean ListElements format, like example below ?
```
                    .desc(
                            "The deployment target for the execution. This can take one of the following values:\n"
                                    + "- local\n"
                                    + "- remote\n"
                                    + "- yarn-session\n"
                                    + "- yarn-application\n"
                                    + "- kubernetes-application")
                    .build();
 ```

Comment (yuxiqian) [reply flink-cdc-cli/src/main/java/org/apache/flink/cdc/cli/CliFrontendOptions.java]: Yes! It could also be constructed like this:

```java
    public static final ConfigOption<SchemaChangeBehavior> PIPELINE_SCHEMA_CHANGE_BEHAVIOR =
            ConfigOptions.key("schema.change.behavior")
                    .enumType(SchemaChangeBehavior.class)
                    .defaultValue(SchemaChangeBehavior.LENIENT)
                    .withDescription(
                            Description.builder()
                                    .text("Behavior for handling schema change events. ")
                                    .linebreak()
                                    .add(
                                            ListElement.list(
                                                    text("IGNORE: Drop all schema change events."),
                                                    text(
                                                            "LENIENT: Apply schema changes to downstream tolerantly, and keeps executing if applying fails."),
                                                    text(
                                                            "TRY_EVOLVE: Apply schema changes to downstream, but keeps executing if applying fails."),
                                                    text(
                                                            "EVOLVE: Apply schema changes to downstream. This requires sink to support handling schema changes."),
                                                    text(
                                                            "EXCEPTION: Throw an exception to terminate the sync pipeline.")))
                                    .build());
```

jx2lee: @yuxiqian Thanks for the suggestion. Since Commons CLI's Option.desc(...) only accepts a String, we need to format Description into plain text before passing it to the option.

Would you prefer adding a reusable PlainTextFormatter next to HtmlFormatter under flink-cdc-common/src/main/java/org/apache/flink/cdc/common/configuration/description/, or keeping this scoped by adding a small private/static formatter inside CliFrontendOptions?

yuxiqian: Oh I see. As the description formatter is not available as we're using commons CLI library, your previous code (manual newline and list items) is more appropriate.

jx2lee: @yuxiqian I fixed it. Thanks for reviewing!

yuxiqian (approved):
Thanks for your contribution!

PR #4494, opened 2026-08-03
[FLINK-39722][pipeline-connector][fluss] Support Fluss CDC Yaml Source.

loserwang1024:

## Brief change log

Support Fluss yaml source.

---

## Verifying this change

- *Manually tested by ...*

## Documentation

---

### DIFF
--- a/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/TableDiscovererFactory.java
+++ b/flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/TableDiscovererFactory.java
@@ -26,10 +26,6 @@ import org.apache.flink.cdc.common.event.TableId;
 public interface TableDiscovererFactory extends ObjectIdDiscovererFactory<TableId> {
 
     /** Compatibility alias for the old single identifier, now meaning storage-side type. */
-    default String identifier() {
-        return type();
-    }
-
     @Override
     default Class<TableId> objectIdClass() {
         return TableId.class;
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/pom.xml
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/pom.xml
@@ -101,7 +101,7 @@ limitations under the License.
         </dependency>
         <dependency>
             <groupId>org.apache.fluss</groupId>
-            <artifactId>fluss-flink-1.20</artifactId>
+            <artifactId>fluss-flink-${fluss.flink.major.version}</artifactId>
             <version>${fluss.version}</version>
             <scope>test</scope>
         </dependency>
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/pom.xml
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/pom.xml
@@ -33,7 +33,7 @@ limitations under the License.
 
 
     <properties>
-        <fluss.version>0.9.0-incubating</fluss.version>
+        <fluss.version>1.0-SNAPSHOT</fluss.version>
     </properties>
 
     <dependencies>
@@ -43,6 +43,12 @@ limitations under the License.
             <version>${fluss.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-connector-base</artifactId>
+            <version>${flink.version}</version>
+            <scope>provided</scope>
+        </dependency>
 
         <!-- Test dependencies -->
         <dependency>
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/pom.xml
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/pom.xml
@@ -33,6 +33,7 @@ limitations under the License.
 
 
     <properties>
+        <!-- Target the upcoming Fluss 1.0 release and use its public Apache snapshot until then. -->
         <fluss.version>1.0-SNAPSHOT</fluss.version>
     </properties>
 
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/v2/FlussSinkWriter.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/v2/FlussSinkWriter.java
@@ -70,7 +70,7 @@ public class FlussSinkWriter<InputT> implements SinkWriter<InputT> {
     }
 
     public void initialize(SinkWriterMetricGroup metricGroup) throws IOException {
-        LOG.info("Opening Fluss with config {}", flussConfig);
+        LOG.info("Opening Fluss sink writer.");
         this.metricGroup = metricGroup;
         flinkMetricRegistry =
                 new WrapperFlussMetricRegistry(
@@ -109,7 +109,6 @@ public class FlussSinkWriter<InputT> implements SinkWriter<InputT> {
                     continue;
                 }
                 MultiTableWriteRecord writeRecord = toWriteRecord(opType, tablePath, row, schemaId);
-                LOG.info("------writeRecord  " + writeRecord);
                 CompletableFuture<?> writeFuture = multiTableWriter.write(writeRecord);
                 writeFuture.whenComplete(
                         (ignored, throwable) -> {
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/v2/FlussSinkWriter.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/v2/FlussSinkWriter.java
@@ -25,10 +25,8 @@ import org.apache.flink.metrics.groups.SinkWriterMetricGroup;
 
 import org.apache.fluss.client.Connection;
 import org.apache.fluss.client.ConnectionFactory;
-import org.apache.fluss.client.table.Table;
-import org.apache.fluss.client.table.writer.AppendWriter;
-import org.apache.fluss.client.table.writer.TableWriter;
-import org.apache.fluss.client.table.writer.UpsertWriter;
+import org.apache.fluss.client.table.writer.MultiTableWriteRecord;
+import org.apache.fluss.client.table.writer.MultiTableWriter;
 import org.apache.fluss.config.Configuration;
 import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.metrics.Gauge;
@@ -40,9 +38,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
 /** Base class for Flink {@link SinkWriter} implementations in Fluss. */
@@ -63,8 +58,7 @@ public class FlussSinkWriter<InputT> implements SinkWriter<InputT> {
     private transient Counter numRecordsOutErrorsCounter;
     private volatile Throwable asyncWriterException;
 
-    private final Map<TablePath, TableWriter> writerMap;
-    private final Map<TablePath, Table> tableMap;
+    private transient MultiTableWriter multiTableWriter;
 
     public FlussSinkWriter(
             Configuration flussConfig,
@@ -73,8 +67,6 @@ public class FlussSinkWriter<InputT> implements SinkWriter<InputT> {
         this.flussConfig = flussConfig;
         this.mailboxExecutor = mailboxExecutor;
         this.flussRecordSerializer = flussRecordSerializer;
-        this.writerMap = new HashMap<>();
-        this.tableMap = new HashMap<>();
     }
 
     public void initialize(SinkWriterMetricGroup metricGroup) throws IOException {
@@ -85,6 +77,7 @@ public class FlussSinkWriter<InputT> implements SinkWriter<InputT> {
                         metricGroup, Collections.singleton(MetricNames.WRITER_SEND_LATENCY_MS));
         connection = ConnectionFactory.createConnection(flussConfig, flinkMetricRegistry);
         flussRecordSerializer.open(connection);
+        multiTableWriter = connection.getMultiTable().newMultiTableWrite().createWriter();
 
         initMetrics();
     }
@@ -101,41 +94,22 @@ public class FlussSinkWriter<InputT> implements SinkWriter<InputT> {
 
         try {
             FlussEvent flussEvent = flussRecordSerializer.serialize(inputValue);
-
-            TablePath tablePath = flussEvent.getTablePath();
-
-            if (flussEvent.isShouldRefreshSchema() || !writerMap.containsKey(tablePath)) {
-                // refresh table schema
-                if (tableMap.containsKey(tablePath)) {
-                    Table table = tableMap.remove(tablePath);
-                    writerMap.remove(tablePath);
-                    table.close();
-                }
-
-                Table table = connection.getTable(tablePath);
-                TableWriter writer;
-                if (table.getTableInfo().hasPrimaryKey()) {
-                    writer = table.newUpsert().createWriter();
-                } else {
-                    writer = table.newAppend().createWriter();
-                }
-                tableMap.put(tablePath, table);
-                writerMap.put(tablePath, writer);
-            }
-
-            List<FlussRowWithOp> rowWithOps = flussEvent.getRowWithOps();
-            if (rowWithOps == null) {
+            if (flussEvent == null || flussEvent.getRowWithOps() == null) {
                 return;
             }
-            for (FlussRowWithOp rowWithOp : rowWithOps) {
+
+            TablePath tablePath = flussEvent.getTablePath();
+            int schemaId = flussEvent.getSchemaId();
+
+            for (FlussRowWithOp rowWithOp : flussEvent.getRowWithOps()) {
                 FlussOperationType opType = rowWithOp.getOperationType();
                 InternalRow row = rowWithOp.getRow();
                 if (opType == FlussOperationType.IGNORE) {
                     // skip writing the row
-                    return;
+                    continue;
                 }
                 CompletableFuture<?> writeFuture =
-                        write(writerMap.get(tablePath), opType, row, tablePath);
+                        multiTableWriter.write(toWriteRecord(opType, tablePath, row, schemaId));
                 writeFuture.whenComplete(
                         (ignored, throwable) -> {
                             if (throwable != null) {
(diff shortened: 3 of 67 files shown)


fxbing:
Ran this PR (5a7ce328) end-to-end as a Fluss→Fluss pipeline, built against apache/fluss main (22cb1969). Build is fine and the PR's own tests are green (74 UT + 43 IT). Two issues showed up in E2E:

  1. Source deserializer doesn't handle ARRAY / MAP / ROW

  FlussRecordDeserializer.convertFlussField covers the 16 scalar types and throws for everything else:

  java.lang.UnsupportedOperationException: Unsupported Fluss data type for deserialization: ArrayType
    at ...FlussRecordDeserializer.convertFlussField(FlussRecordDeserializer.java:396)
    at ...FlussRecordDeserializer.convertFlussRowToCdcRecord(FlussRecordDeserializer.java:201)

  Repro: a source table with tags ARRAY<STRING> (same for MAP / ROW) — the job fails on the first record and goes into a restart loop. The sink side already supports these types via CdcAsFlussArray /
  CdcAsFlussMap / CdcAsFlussRow, so the gap looks source-only.

  2. CreateTableEvent isn't reconciled against an existing sink table

  If the sink table already exists with fewer columns than the source, FlussMetaDataApplier treats CreateTableEvent as create-if-not-exists and leaves the physical schema as is. Upstream-only columns are then
  silently dropped while the job stays RUNNING with no error.

  Repro: sink (id, payload), source (id, payload, extra_col) → the row lands as (1, 'cold-start') and extra_col is lost. The coordinator does pass the full schema down:

  Step 3.5 - Corresponding schema changes are: [CreateTableEvent{... `extra_col` STRING ...}]

  A worse variant: when the sink is missing a middle column (source (id, name, sale_quantity) vs sink (id, sale_quantity)), the write still succeeds with name projected away rather than failing fast.

  Would it make sense for the metadata applier to reload the real TableInfo on CreateTableEvent, append missing nullable trailing columns, and fail fast when the difference can't be reconciled that way?

loserwang1024:
@fxbing 

> 1. Source deserializer doesn't handle ARRAY / MAP / ROW

support it now.

> 2. CreateTableEvent isn't reconciled against an existing sink table

This is a sink behavior, thus is not included in this PR. I have raise a new issue to solve it: https://issues.apache.org/jira/browse/FLINK-40349

fxbing (approved):

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/FlussSource.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public SimpleVersionedSerializer<FlussSourceEnumState> getEnumeratorCheckpointSerializer() {
+        return new FlussSourceEnumStateSerializer();
+    }
+
+    @Override
+    public SourceReader<T, FlussSplitBase> createReader(SourceReaderContext readerContext) {
+        FlussSourceReaderMetrics sourceReaderMetrics =
+                new FlussSourceReaderMetrics(readerContext.metricGroup());
+        WrapperFlussMetricRegistry metricRegistry =
```
fxbing, line 156: `WrapperFlussMetricRegistry` wraps Fluss `METER`s as `WrapperFlussMeter`, but it never registers a `MetricViewUpdater` (or otherwise calls `update()` on the underlying `MeterView`). `WrapperFlussMeter#getRate()` therefore only exposes the cached value.

The new source creates this registry in `FlussSource#createReader`, so scanner rate metrics such as `fetchRequestsPerSecond` and `remoteFetch*` will remain at their initial `0` even while records are being fetched. Could we register and close a `MetricViewUpdater` for this registry, as the Fluss metric integration requires?

https://github.com/apache/flink-cdc/blob/ed5fc73911f32ed7acb308052b62dcb11f27c71f/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/v2/metrics/WrapperFlussMetricRegistry.java#L116

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/pom.xml
```diff
@@ -33,7 +33,7 @@ limitations under the License.
 
 
     <properties>
-        <fluss.version>0.9.0-incubating</fluss.version>
+        <fluss.version>1.0-SNAPSHOT</fluss.version>
```
leonardBang, line 37: Could we clarify which publicly available Fluss artifact this PR is expected to build against?
As of August 20, 2026, the Apache snapshot repository still resolves the relevant Fluss artifacts to `1.0-20260802.061439-1`; it does not currently show a newer publication. A locally installed artifact or private mirror would not be reproducible for community builds.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/v2/FlussSinkWriter.java
```diff
@@ -101,41 +94,23 @@ public void write(InputT inputValue, Context context) throws IOException {
...
                 InternalRow row = rowWithOp.getRow();
                 if (opType == FlussOperationType.IGNORE) {
                     // skip writing the row
-                    return;
+                    continue;
                 }
-                CompletableFuture<?> writeFuture =
-                        write(writerMap.get(tablePath), opType, row, tablePath);
+                MultiTableWriteRecord writeRecord = toWriteRecord(opType, tablePath, row, schemaId);
+                LOG.info("------writeRecord  " + writeRecord);
```
leonardBang: Could we remove this per-record INFO log?

This is on the writer's hot path and eagerly converts every `MultiTableWriteRecord` to a string. Under normal production throughput it may generate a large amount of logging, add per-record overhead, and expose business row contents.

If diagnostic logging is needed, a guarded DEBUG log that avoids printing the complete record would be safer.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/metrics/FlussSourceReaderMetrics.java
```diff
@@ -0,0 +1,128 @@
...
+
+    // -------- Helper functions --------
+    private void registerOffsetMetricsForTableBucket(TableBucket tableBucket) {
+        final MetricGroup metricGroup =
+                tableBucket.getPartitionId() == null
+                        ? this.flussSourceReaderMetricGroup
+                        : this.flussSourceReaderMetricGroup.addGroup(
+                                PARTITION_GROUP, String.valueOf(tableBucket.getPartitionId()));
+        final MetricGroup bucketGroup =
+                metricGroup.addGroup(BUCKET_GROUP, String.valueOf(tableBucket.getBucket()));
```
leonardBang: Should the metric group include the table identity?

A CDC source reader may consume multiple tables, and different tables can have the same partition and bucket IDs. In that case they register the same metric identifier, so the gauges may collide or become indistinguishable.

Could we add the database/table path or table ID to the metric group and cover two tables with the same bucket number in a test?

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/enumerator/FlussSourceEnumerator.java
```diff
@@ -0,0 +1,564 @@
...
+    //  Phase 4: Handle new splits — mark assigned and assign to readers (callback)
+    // -------------------------------------------------------------------------
+
+    /**
+     * Receives newly created splits, records their {@link PhysicalTablePath}s as assigned, and
+     * distributes the splits to registered readers.
+     */
+    private void handleTableBucketChanges(List<FlussSplitBase> newSplits, Throwable error) {
+        if (error != null) {
+            LOG.error("Error creating splits for new table-buckets", error);
```
leonardBang: Could this split-initialization failure be propagated rather than returning after logging it?
Failures while resolving startup offsets or creating snapshot/log splits currently leave the discovered buckets unassigned. In one-shot discovery mode the operation is never retried, so the job may remain RUNNING without producing data.
A test that injects an offset-initialization failure and verifies that the source fails, rather than silently becoming idle, would help cover this behavior.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/enumerator/FlussSourceEnumerator.java
```diff
@@ -0,0 +1,564 @@
...
+    //  Phase 2: Check for table-bucket changes (callback)
+    // -------------------------------------------------------------------------
+
+    /**
+     * Compares the discovered table-buckets against already-assigned {@link PhysicalTablePath}s and
+     * triggers split creation for newly discovered table-buckets.
+     */
+    private void checkTableBucketChanges(List<TableBucketInfo> allBuckets, Throwable error) {
+        if (error != null) {
+            LOG.error("Error discovering subscribed table-buckets", error);
```
leonardBang: Should the initial discovery failure be propagated to the coordinator instead of only being logged?
When periodic discovery is disabled, this callback is invoked only once. If table discovery or a metadata request fails, returning here leaves the job RUNNING without assigning any splits, and there is no subsequent retry.
Could we fail the job for initial/one-shot discovery failures? For periodic discovery, an explicit bounded retry policy may be more appropriate.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/deserializer/FlussRecordDeserializer.java
```diff
@@ -0,0 +1,504 @@
...
+                events.add(
+                        new CreateTableEvent(
+                                tableId,
+                                buildCdcSchema(
+                                        restoredRowType,
+                                        record.getPrimaryKeyNames(),
+                                        record.getPartitionKeyNames())));
+            }
+
+            Integer cachedSchemaId = latestSchemaIdCache.get(tablePath);
```
leonardBang: Could we restore the schema with the highest schema ID here instead of keeping whichever split is processed first?

Different splits of the same table may checkpoint different schema IDs, and the order in which restored splits are applied is not guaranteed to be newest-first. Keeping the first schema can therefore initialize the table cache with an older schema and may emit an outdated CreateTableEvent.

It may be helpful to add restore tests for both split orders {schema 1, schema 2} and {schema 2, schema 1}.

leonardBang (commented):
Thanks @loserwang1024 for the contribution, I left some comments

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/deserializer/FlussRecordDeserializer.java
```diff
@@ -0,0 +1,504 @@
...
+    /**
+     * Restores the internal schema caches from a recovered split. This seeds the
+     * latestSchemaIdCache, latestRowTypeCache, and latestRecordDataGeneratorCache so that schema
+     * changes occurring after the last checkpoint can still be detected.
+     */
+    @Override
+    public List<Event> restoreState(TablePath tablePath, int schemaId, RowType rowType) {
+        ensureCacheInitialized();
+        // Multiple splits may read log with different schemaIds; only reserve the first one.
+        if (!latestSchemaIdCache.containsKey(tablePath)) {
```
leonardBang: Could this split-initialization failure be propagated rather than returning after logging it?
Failures while resolving startup offsets or creating snapshot/log splits currently leave the discovered buckets unassigned. In one-shot discovery mode the operation is never retried, so the job may remain RUNNING without producing data.
A test that injects an offset-initialization failure and verifies that the source fails, rather than silently becoming idle, would help cover this behavior.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/deserializer/FlussRecordDeserializer.java
```diff
@@ -0,0 +1,504 @@
...
+            TablePath tablePath,
+            InternalRow initialRow,
+            RowType initialRowType,
+            boolean isSchemaChangeEvent) {
+        RowType latestRowType = latestRowTypeCache.get(tablePath);
+        InternalRow row = initialRow;
+
+        // A reader maybe subscribe multiple split, thus only inferred by the latest schema(also the
+        // widest)
+        if (isSchemaChangeEvent) {
```
leonardBang: Would it make sense to project the row whenever initialRowType differs from latestRowType, rather than only when the current record advances the cached schema ID?

The Fluss MultiTableLogScanner uses dynamic schema resolution and preserves each record's original schema ID. With multiple buckets, records may therefore arrive in the order schema 1 → schema 2 → schema 1. After schema 2 updates the cache, the final schema 1 row is read using schema 2's field converters and currently fails with:
ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2

Could we project every older-schema row to the cached latest schema?  It would also be helpful to add a regression test covering schema 1 → schema 2 → schema 1.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/pom.xml
loserwang1024: done

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/metrics/FlussSourceReaderMetrics.java
loserwang1024: Remove all the bucket level metrics because https://github.com/apache/fluss/pull/3828. Add it later if needed.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/reader/FlussSplitReader.java
```diff
@@ -0,0 +1,421 @@
...
+
+                        maxRecordTimestamp = Math.max(maxRecordTimestamp, record.timestamp());
+                    }
+                }
+            }
+        }
+
+        // Report event time lag
+        if (maxRecordTimestamp > 0) {
+            sourceReaderMetrics.reportRecordEventTime(fetchTimestamp - maxRecordTimestamp);
```
leonardBang: Could we calculate the fetch event-time lag using the current time after `poll()` completes?

`fetchTimestamp` is captured before the blocking poll. A record appended while the poll is waiting may therefore have a timestamp later than `fetchTimestamp`, causing `currentFetchEventTimeLag` to report a negative value.

It may also be helpful to clamp the result to zero to tolerate clock skew and add a test covering a record timestamp later than the poll start time.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/reader/FlussSplitReader.java
```diff
@@ -0,0 +1,421 @@
...
+    private void closeCurrentBoundedSplit() throws IOException {
+        try {
+            if (currentBatchScanner != null) {
+                currentBatchScanner.close();
+            }
+        } catch (Exception e) {
+            throw new IOException("Failed to close batch scanner", e);
+        }
+
+        // todo: 可以封装为一个对象
```
leonardBang: Could we remove this TODO or rewrite it in English? Comments in the project should be written in English. For example: `// TODO: Encapsulate the snapshot reader state in a dedicated object.`

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/deserializer/FlussRecordDeserializer.java
```diff
@@ -0,0 +1,503 @@
...
+
+    /**
+     * Restores the internal schema caches from a recovered split. This seeds the
+     * latestSchemaIdCache, latestRowTypeCache, and latestRecordDataGeneratorCache so that schema
+     * changes occurring after the last checkpoint can still be detected.
+     */
+    @Override
+    public List<Event> restoreState(TablePath tablePath, int schemaId, RowType rowType) {
+        ensureCacheInitialized();
+        // Multiple splits may read log with different schemaIds; only reserve the first one.
```
leonardBang: Could we retain the restored schema with the highest schema ID here instead of whichever split is initialized first?

Different splits of the same table may checkpoint different schema IDs, and `initializedState()` applies them independently during recovery. If the older split is processed first, `restoredCreateTableRowTypeCache` is initialized with an outdated schema and the first post-recovery record emits an outdated `CreateTableEvent`.

Could we replace the cached schema whenever `schemaId` is greater than the currently restored schema ID? It would also be helpful to cover both restoration orders: `{schema 1, schema 2}` and `{schema 2, schema 1}`.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/TableDiscovererFactory.java
```diff
@@ -26,10 +26,6 @@
 public interface TableDiscovererFactory extends ObjectIdDiscovererFactory<TableId> {
 
     /** Compatibility alias for the old single identifier, now meaning storage-side type. */
```
leonardBang, line 28: Could we keep the `identifier()` compatibility alias here? `TableDiscovererFactory` is `@PublicEvolving`, and the current master API intentionally retains this default method as the legacy name for `type()`. Removing it in an unrelated Fluss source PR can break third-party discoverer code compiled against the current SPI and also leaves the compatibility-alias Javadoc attached to `objectIdClass()`. Unless this removal is part of a separately reviewed SPI migration, restoring the alias would avoid an unnecessary compatibility regression.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/enumerator/FlussSourceEnumerator.java
```diff
@@ -0,0 +1,595 @@
...
+            long scanDiscoveryIntervalMs,
+            FlussSourceEnumState restoredState) {
+        this(
+                context,
+                discoverer,
+                flussConfig,
+                sourceConfig,
+                offsetsInitializer,
+                scanDiscoveryIntervalMs,
+                restoredState.getAssignedPhysicalTablePaths());
```
leonardBang: Could we restore `getRemainingSplits()` into the pending assignments here? `snapshotState()` persists splits whose startup offsets or snapshot IDs have already been initialized but not yet assigned, while this constructor only restores `assignedPhysicalTablePaths`. I reproduced this by checkpointing an unassigned latest-mode split at offset 1, appending another record, and restoring the enumerator: the restored split was rediscovered at offset 2 instead of retaining offset 1. Restoring `remainingSplits` made the test pass. The restore path should also recalculate owners for the current parallelism and prevent these restored paths from being initialized again by discovery.

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/source/enumerator/FlussSourceEnumerator.java
```diff
@@ -0,0 +1,595 @@
...
+    private List<FlussSplitBase> initHybridSnapshotLogSplits(
+            TablePath tablePath,
+            @Nullable String partitionName,
+            List<TableBucketInfo> bucketInfos,
+            OffsetsInitializer.BucketOffsetsRetriever retriever)
+            throws Exception {
+        List<FlussSplitBase> splits = new ArrayList<>();
+
+        // Get KV snapshots for this table/partition
+        KvSnapshots kvSnapshots =
```
leonardBang, line 358: Could we acquire and checkpoint a KV snapshot lease before creating these hybrid splits? The split stores a concrete snapshot ID, but the Fluss server retains only a bounded number of completed snapshots, so an unleased snapshot may be pruned while a slow full scan is still running or before the job restores from a checkpoint/savepoint. I reproduced this with `kv.snapshot.num-retained=1`: after the enumerator checkpointed snapshot 0 and two newer snapshots were created, `createBatchScanner(tableBucket, snapshotId)` failed with `KvSnapshotNotExistException`; acquiring a lease made the same test pass. The native Fluss source persists its lease ID and releases snapshots only after their consumption is checkpointed. Could we implement the corresponding lease lifecycle here and add this recovery test?

leonardBang (commented):
Thanks for the update. I found three additional issues during the full review.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/source/discover/TableDiscovererFactory.java
loserwang1024: Since https://issues.apache.org/jira/browse/FLINK-39732 is for the same version of the Fluss YAML source, semantic simplicity takes priority.

leonardBang (approved):
Thanks @loserwang1024 for the update and @fxbing for the review, LGTM now

leonardBang:
Will merge once CI green....

PR #4496, opened 2026-08-04
[FLINK-40313][runtime] Support nested collection functions in Transforms

haruki-830:
#### Summary

This commit extends Flink CDC Transform expressions with Flink SQL collection constructors and functions. It adds support for constructing and operating on nested ARRAY, MAP, and ROW values in projection and filter expressions.

#### Key Changes

1. Collection Constructors
- Added support for `ARRAY[...]`, `MAP[...]`, and `ROW(...)`.
- Added recursive type coercion for nested collection constructors.
- Added ROW type conversion with generated field names such as `f0` and `f1`.

2. Collection Functions
- Added support for `CARDINALITY`, `ARRAY_CONTAINS`, `ARRAY_POSITION`, and `ELEMENT`.
- Supports arrays and maps where applicable, including nested values and NULL handling.
- Reports an error when `ELEMENT` receives an array containing multiple elements.

3. Parser and Expression Compilation
- Enabled explicit `ROW(...)` constructors while preserving existing MySQL conformance behavior.
- Added Calcite operator registration and Janino expression generation for the new constructors and functions.
- Supports the new expressions in both projections and filters.

4. Test Coverage and Documentation
- Added unit tests for constructors, functions, NULL behavior, nested values, and error cases.
- Added Transform specification tests for projections, filters, and nested collection types.
- Updated the English and Chinese Transform documentation.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40313](https://issues.apache.org/jira/browse/FLINK-40313)

### DIFF
--- a/docs/content/docs/core-concept/transform.md
+++ b/docs/content/docs/core-concept/transform.md
@@ -282,6 +282,13 @@ Struct functions are used to access elements in ARRAY, MAP, ROW, and VARIANT typ
 
 | Function | Janino Code | Description |
 | -------- | ----------- | ----------- |
+| ARRAY[value, ...] | array(value, ...) | Creates an array from the given values. The values must have a common type. |
+| MAP[key, value, ...] | map(key, value, ...) | Creates a map from key-value pairs. Keys and values must have common types respectively. |
+| ROW(value, ...) | row(value, ...) | Creates a row from the given values. Field names are generated as `f0`, `f1`, and so on. |
+| CARDINALITY(arrayOrMap) | cardinality(arrayOrMap) | Returns the number of elements in an array or map. Returns NULL if the input is NULL. |
+| ARRAY_CONTAINS(array, value) | arrayContains(array, value) | Returns whether the array contains the given value. |
+| ARRAY_POSITION(array, value) | arrayPosition(array, value) | Returns the 1-based position of the first matching value, or 0 if it is not found. |
+| ELEMENT(array) | element(array) | Returns the element of a single-element array. Returns NULL for an empty array and throws an exception if the array has more than one element. |
 | array[index] | itemAccess(array, index) | Returns the element at position `index` in the array. Index is 1-based (SQL standard). Returns NULL if the index is out of bounds or if the array is NULL. |
 | map[key] | itemAccess(map, key) | Returns the value associated with `key` in the map. Returns NULL if the key does not exist or if the map is NULL. |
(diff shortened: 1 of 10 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/typeutils/CalciteDataTypeConverter.java
```diff
@@ -339,6 +339,10 @@ public static DataType convertCalciteRelDataTypeToDataType(RelDataType relDataTy
             case VARIANT:
                 return DataTypes.VARIANT();
             case ROW:
+                return DataTypes.ROW(
+                        relDataType.getFieldList().stream()
+                                .map(field -> convertCalciteRelDataTypeToDataType(field.getType()))
+                                .toArray(DataType[]::new));
```
lvyanquan: Could we preserve the original ROW field names here? `DataTypes.ROW(DataType...)` regenerates all field names as `f0`, `f1`, etc. For example, if `complex_row_` is `ROW<name STRING, length INT>`, `ELEMENT(ARRAY[complex_row_])` will incorrectly produce `ROW<f0 STRING, f1 INT>`. The same issue affects named ROW types nested in ARRAY or MAP and may change the downstream schema.  

Please construct the ROW type using Calcite’s field names, while normalizing Calcite-generated aliases for explicit `ROW(...)` constructors to `f0`, `f1`, etc. It would also be good to add a regression test covering a named ROW through `ELEMENT`, ARRAY, or MAP.

haruki-830: Fixed in the latest commit. Added a regression test as well.

docs/content/docs/core-concept/transform.md
```diff
@@ -282,6 +282,13 @@ Struct functions are used to access elements in ARRAY, MAP, ROW, and VARIANT typ
 
 | Function | Janino Code | Description |
 | -------- | ----------- | ----------- |
+| ARRAY[value, ...] | array(value, ...) | Creates an array from the given values. The values must have a common type. |
```
yuxiqian, line 285: Shall we add a bad-case test where element types differ?

yuxiqian (approved):
Looks great! Just left a trivial comment on test coverage, please take a look at your convenience.

docs/content/docs/core-concept/transform.md
haruki-830: ok^.^

PR #4497, opened 2026-08-04
[FLINK-40315][connect/mariadb] Add MariaDB CDC DataStream source

leosanqing:

Implements [FLINK-40315](https://issues.apache.org/jira/browse/FLINK-40315), a sub-task of
[FLINK-40314](https://issues.apache.org/jira/browse/FLINK-40314).

The MySQL CDC connector can already reach MariaDB over its MySQL-compatible wire protocol, but it
does not understand MariaDB GTIDs: it parses the MySQL `uuid:interval` form, while MariaDB uses
`domain-server-sequence`. Once a cluster fails over, scales out, or a job restarts and gets routed
to another server in the same GTID domain, that server has a different server id and the
checkpointed position can no longer be compared correctly — recovery either fails or silently
stops advancing.

Rather than adding MariaDB branches to the MySQL read path — the most critical path in Flink CDC —
this introduces a standalone `flink-connector-mariadb-cdc` module, following the `oceanbase-cdc`
precedent: depend on `flink-connector-mysql-cdc`, reuse its entire read pipeline, and inject
MariaDB behaviour.
**No line of `flink-connector-mysql-cdc` is modified.**

Design background: apache/flink-cdc#4468 validated MariaDB GTID support inside mysql-cdc under
FLINK-34807; that PR is not meant to be merged and serves as the reference for this split.

## Brief change log

**Module**

- Add `flink-connector-mariadb-cdc`, registered under `flink-cdc-source-connectors`, depending on
  `flink-connector-mysql-cdc` to reuse the snapshot and binlog read pipeline
- MySQL Connector/J is declared `provided`, so it is neither bundled nor exposed transitively;
  users supply the driver themselves
- Add `MariaDbTableSourceFactory` (extends `MySqlTableSourceFactory`, overriding only the
  identifier to `mariadb-cdc`) plus its `META-INF/services` registration. All MySQL CDC options
  are reused as-is.

**MariaDB binlog and GTID support** 

 three classes are added(MariaDB core):

- `MariaDBBinaryLogClient` — sets `@mariadb_slave_capability=4` so the server emits `MARIADB_GTID`
  events and the consumed GTID actually advances per transaction
- `MariaDbConnections` — reads `@@gtid_binlog_pos` (a lightweight helper, so `MySqlConnection` does
  not have to be overridden for a single query)
- `MariaDbGtidComparator` — domain+sequence containment, ignoring the server id

## Documentation

- Does this pull request introduce a new feature? (yes)
- If yes, how is the feature documented? (docs — tracked separately in
  [FLINK-40328](https://issues.apache.org/jira/browse/FLINK-40328))

---

- [x] Yes (Opus-4.8,GPT5.5-sol)

### DIFF
--- a/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mariadb-cdc/pom.xml
+++ b/flink-cdc-connect/flink-cdc-source-connectors/flink-connector-mariadb-cdc/pom.xml
@@ -67,6 +67,15 @@ limitations under the License.
             <scope>test</scope>
         </dependency>
 
+        <!-- Provides TestValuesTableFactory (the 'values' test sink) used by the recovery IT. -->
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-planner_${scala.binary.version}</artifactId>
+            <version>${flink.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
         <dependency>
             <groupId>org.apache.flink</groupId>
             <artifactId>flink-table-runtime</artifactId>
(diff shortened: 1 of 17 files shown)


leosanqing:
@leonardBang  @yuxiqian @haruki-830 hi, could you help me review this code pls?

PR #4498, opened 2026-08-05
[FLINK-40331][cdc-connect][fluss] Support forward shuffle in fluss sink.

loserwang1024:
https://issues.apache.org/jira/browse/FLINK-40331

## Brief change log

Add sink.partitioning.strategy in fluss sink to support FORWARD.

---

## Verifying this change

*(or)*

- *Manually tested by ...*

## Documentation

---

### DIFF
--- a/docs/content.zh/docs/connectors/pipeline-connectors/fluss.md
+++ b/docs/content.zh/docs/connectors/pipeline-connectors/fluss.md
@@ -100,6 +100,13 @@ Pipeline Connector Options
       <td>用于建立与 Fluss 集群初始连接的主机/端口对列表。 </td>
     </tr>
     <tr>
+      <td>sink.partitioning.strategy</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">DEFAULT</td>
+      <td>String</td>
+      <td>DataChangeEvent 路由使用的分区策略。可选值为 <code>DEFAULT</code> 和 <code>FORWARD</code>。<code>DEFAULT</code> 对主键表按主键 hash，对 log 表采用 round-robin。<code>FORWARD</code> 将数据事件发送到与上游 subtask 索引相同的下游 subtask，用于 Fluss 到 Fluss 的数据同步。上游数据分布必须与 Fluss 表的分桶策略保持一致，否则可能出现数据正确性问题。</td>
+    </tr>
+    <tr>
       <td>bucket.key</td>
       <td>optional</td>
       <td style="word-wrap: break-word;">(none)</td>
--- a/docs/content.zh/docs/connectors/pipeline-connectors/fluss.md
+++ b/docs/content.zh/docs/connectors/pipeline-connectors/fluss.md
@@ -104,7 +104,7 @@ Pipeline Connector Options
       <td>optional</td>
       <td style="word-wrap: break-word;">DEFAULT</td>
       <td>String</td>
-      <td>DataChangeEvent 路由使用的分区策略。可选值为 <code>DEFAULT</code> 和 <code>FORWARD</code>。<code>DEFAULT</code> 对主键表按主键 hash，对 log 表采用 round-robin。<code>FORWARD</code> 将数据事件发送到与上游 subtask 索引相同的下游 subtask，用于 Fluss 到 Fluss 的数据同步。上游数据分布必须与 Fluss 表的分桶策略保持一致，否则可能出现数据正确性问题。</td>
+      <td>DataChangeEvent 路由使用的分区策略。可选值为 <code>DEFAULT</code> 和 <code>FORWARD</code>。<code>DEFAULT</code> 对主键表按主键进行哈希分区，对日志表采用随机分发，将事件分散到各下游 subtask 以实现负载均衡。<code>FORWARD</code> 将数据事件发送到与上游 subtask 索引相同的下游 subtask，用于 Fluss 到 Fluss 的数据同步。上游数据分布必须与 Fluss 表的分桶策略保持一致，否则可能出现数据正确性问题。</td>
     </tr>
     <tr>
       <td>bucket.key</td>
--- a/docs/content/docs/connectors/pipeline-connectors/fluss.md
+++ b/docs/content/docs/connectors/pipeline-connectors/fluss.md
@@ -101,6 +101,13 @@ Pipeline Connector Options
       <td>The bootstrap servers for the Fluss sink connection. </td>
     </tr>
     <tr>
+      <td>sink.partitioning.strategy</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">DEFAULT</td>
+      <td>String</td>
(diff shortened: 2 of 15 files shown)


flink-cdc-common/src/main/java/org/apache/flink/cdc/common/function/HashFunction.java
```diff
@@ -28,4 +28,14 @@
...
 
     int hashcode(T event);
+
+    /**
+     * Calculates the hash code with the upstream source subtask index.
+     *
+     * <p>Implementations that do not depend on the source subtask index can continue implementing
+     * {@link #hashcode(Object)} only.
+     */
+    default int hashcode(int sourceIndex, T event) {
```
lvyanquan: Hi @loserwang1024, thanks for the PR! One suggestion on the `HashFunction` change: instead of a raw `int sourceIndex`, consider passing a small immutable `HashContext` object:

```java
@Internal
public class HashContext {

    private final int sourceSubtaskIndex;
    private final int downstreamParallelism;

    public HashContext(int sourceSubtaskIndex, int downstreamParallelism) { ... }

    public int getSourceSubtaskIndex() { ... }

    public int getDownstreamParallelism() { ... }
}
```

```java
default int hashcode(HashContext context, T event) {
    return hashcode(event);
}
```

Reasons:

1. **Self-documenting call sites** — `hashcode(subTaskId, event)` gives no hint what the bare `int` means; `hashcode(hashContext, event)` does.
2. **Extensible without signature churn** — future strategies (bucket-aware mapping, fallback when `sourceIndex >= downstreamParallelism`) can read from the context instead of stacking more `int` overloads.
3. **Fails fast instead of silently skewing** — `ForwardHashFunction.hashcode(event)` can throw `UnsupportedOperationException` rather than returning `0` and routing everything to subtask 0 if someone calls the single-arg method by mistake.

The context can be created once in `open()` and reused, so there's no per-record allocation cost. Since `HashFunction` is `@Internal`, this is the cheapest time to make the change. WDYT?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/sink/ForwardHashFunctionProvider.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public HashFunction<DataChangeEvent> getHashFunction(@Nullable TableId tableId, Schema schema) {
+        return new ForwardHashFunction();
+    }
+
+    private static class ForwardHashFunction implements HashFunction<DataChangeEvent> {
+
+        @Override
+        public int hashcode(DataChangeEvent event) {
+            return 0;
```
naivedogger: I see that ForwardHashFunction overrides both hashcode(int sourceIndex, DataChangeEvent event) and hashcode(DataChangeEvent event). The single-arg version returns 0.
My understanding is that hashcode(int sourceIndex, T event) is a default method on the interface, so the only reason to implement the single-arg version is that it's the abstract contract. But this raises two questions:
1. Is the single-arg override actually needed? If every call site in the pipeline now uses the two-arg overload, can we just leave the single-arg version unimplemented (i.e., rely on the default)? Or does something in the framework force us to provide it?
2. If we do need to implement it, shouldn't it throw rather than silently return 0? Returning 0 means that if someone accidentally regresses a call site back to hashcode(event), every record from every source subtask gets routed to sink subtask 0.

docs/content.zh/docs/connectors/pipeline-connectors/fluss.md
```diff
@@ -99,6 +99,13 @@ Pipeline Connector Options
       <td>String</td>
       <td>用于建立与 Fluss 集群初始连接的主机/端口对列表。 </td>
     </tr>
+    <tr>
+      <td>sink.partitioning.strategy</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">DEFAULT</td>
+      <td>String</td>
+      <td>DataChangeEvent 路由使用的分区策略。可选值为 <code>DEFAULT</code> 和 <code>FORWARD</code>。<code>DEFAULT</code> 对主键表按主键 hash，对 log 表采用 round-robin。<code>FORWARD</code> 将数据事件发送到与上游 subtask 索引相同的下游 subtask，用于 Fluss 到 Fluss 的数据同步。上游数据分布必须与 Fluss 表的分桶策略保持一致，否则可能出现数据正确性问题。</td>
```
naivedogger: The docs and ConfigOption description say DEFAULT uses round-robin for log tables, but the actual implementation in FlussHashFunction uses ThreadLocalRandom.current().nextInt(), which is a random value, not round-robin. Could we align the docs here?

docs/content/docs/connectors/pipeline-connectors/fluss.md
```diff
@@ -100,6 +100,13 @@ Pipeline Connector Options
       <td>String</td>
       <td>The bootstrap servers for the Fluss sink connection. </td>
     </tr>
+    <tr>
+      <td>sink.partitioning.strategy</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">DEFAULT</td>
+      <td>String</td>
+      <td>The partitioning strategy for DataChangeEvent routing. Available values are <code>DEFAULT</code> and <code>FORWARD</code>. <code>DEFAULT</code> hashes primary key tables by primary keys and routes log tables in round-robin mode. <code>FORWARD</code> routes data events to downstream subtasks with the same indices as upstream and is intended for Fluss-to-Fluss data synchronization. The upstream data distribution must match the Fluss table bucket distribution; otherwise, data correctness issues may occur.</td>
```
naivedogger: ditto

flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-fluss/src/main/java/org/apache/flink/cdc/connectors/fluss/sink/FlussDataSinkOptions.java
```diff
@@ -33,6 +33,17 @@ public class FlussDataSinkOptions {
                     .noDefaultValue()
                     .withDescription("The bootstrap servers for the Fluss sink connection.");
 
+    public static final ConfigOption<SinkPartitioningStrategy> SINK_PARTITIONING_STRATEGY =
+            ConfigOptions.key("sink.partitioning.strategy")
+                    .enumType(SinkPartitioningStrategy.class)
+                    .defaultValue(SinkPartitioningStrategy.DEFAULT)
+                    .withDescription(
+                            "Partitioning strategy for DataChangeEvent routing. "
+                                    + "DEFAULT hashes primary key tables by primary keys and routes log tables in round-robin mode. "
```
naivedogger: ditto

naivedogger (approved):
Thanks @loserwang1024 for this pr. Overall, LGTM. Left some comments, PTAL.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/function/HashFunction.java
loserwang1024: done

docs/content/docs/connectors/pipeline-connectors/fluss.md
loserwang1024: done

leonardBang (approved):
LGTM

PR #4499, opened 2026-08-05
[FLINK-40337][runtime] Preserve operator precedence in generated Janino expressions

haruki-830:
#### Summary

This change preserves operator precedence when generating Janino expressions for Flink CDC pipeline transforms. It uses Janino's AST-aware unparser to retain the grouping and semantics of nested expressions.

#### Key Changes

1. Janino Expression Serialization
- Replaced `Java.Rvalue.toString()` with Janino's `Unparser`.
- Delegates precedence and parenthesis handling to Janino instead of maintaining operator rules manually.
- Supports binary and conditional expressions consistently.

2. Correct Nested Expression Generation
- Preserves expressions such as `(a + b) * c`, `(a OR b) AND c`, and `a / (b * c)`.
- Correctly handles conditional expressions used as arithmetic operands, such as `IF(a > b, a, b) + c`.
- Fixes transform expressions such as `weight / (height * height)` that previously changed semantics during code generation.

3. Test Coverage
- Added coverage for nested arithmetic, logical, and conditional expression generation.
- Added runtime evaluation tests to verify the semantics of generated Janino expressions.
- Verified expressions whose existing precedence already preserves their evaluation order.

#### JIRA Reference

[https://issues.apache.org/jira/browse/FLINK-40337](https://issues.apache.org/jira/browse/FLINK-40337)

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/JaninoCompiler.java
@@ -94,6 +94,16 @@ public class JaninoCompiler {
         "Arithmetic", "Casting", "Comparison", "Logical", "String", "Struct", "Temporal"
     };
 
+    private static final Map<String, Integer> BINARY_OPERATOR_PRECEDENCE =
+            Map.of(
+                    "||", 1,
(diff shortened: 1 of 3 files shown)


yuxiqian (approved):
+1, waiting for ci to pass.

PR #4505, opened 2026-08-11
[hotfix][runtime] Respect UDF precedence during type validation

haruki-830:

This is a follow-up to #4492 (FLINK-40240). It aligns Calcite type validation with Janino code generation by giving same-named UDFs precedence over built-in functions.

## Verifying this change

- Added a regression test for UDF return-type inference.

## Documentation

This PR does not introduce a new feature.

### DIFF
--- a/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
+++ b/flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
@@ -55,8 +55,10 @@ import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.SqlOperatorTable;
 import org.apache.calcite.sql.SqlSelect;
+import org.apache.calcite.sql.SqlSyntax;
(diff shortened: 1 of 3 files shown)


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
```diff
@@ -240,6 +248,67 @@ private static RelNode sqlToRel(
...
+    private static SqlNode resolveUserDefinedFunctions(
+            SqlNode sqlNode, List<SqlFunction> udfFunctions) {
+        return sqlNode.accept(
+                new SqlShuttle() {
+                    @Override
+                    public SqlNode visit(SqlCall call) {
+                        SqlNode visited = super.visit(call);
+                        if (visited instanceof SqlBasicCall) {
+                            SqlBasicCall basicCall = (SqlBasicCall) visited;
+                            udfFunctions.stream()
```
lvyanquan, line 263: resolveUserDefinedFunctions matches every SqlBasicCall solely by operator name, including structural operators. Since UDF names are not restricted, registering a UDF named as causes id AS alias to have its AS operator replaced by the UDF, so Calcite validates it as a function call rather than an alias expression.
Please restrict rebinding to actual function-call syntax and leave structural/special operators such as AS intact.

```
      if (basicCall.getOperator().getSyntax().family
              != SqlSyntax.FUNCTION) {
          return visited;
      }
```

lvyanquan (approved):
LGTM.

PR #4506, opened 2026-08-12
[FLINK-40385][postgre]Optimize efficiency issues of PostgreSQL during the full snapshot phase.

hql0312:

When fetching snapshot data, refreshSchema is executed to retrieve data of subscribed tables. This refresh operation has already been invoked via refresh within PostgresSourceFetchTaskContext.configure


hql0312:
https://issues.apache.org/jira/browse/FLINK-40385

PR #4508, opened 2026-08-17
[FLINK-40432][connect/tidb] Split key ranges by TiKV region via FLIP-27 enumerator

lifulong:

TiDB CDC used to split a table by evenly partitioning the entire `long` handle space (`Long.MIN_VALUE` ~ `Long.MAX_VALUE`).
For clustered `AUTO_INCREMENT` primary keys, real handles only occupy a tiny slice of that space. With parallelism 8, almost all rows land on a single subtask, which then OOMs during `KVClient.scan`.

This change assigns each reader a contiguous chunk of **real TiKV regions** covering the table, so snapshot and incremental CDC scale with region count instead of the theoretical `long` domain.

Region topology is discovered **once** by a FLIP-27 enumerator. Each parallel `SourceFunction.open()` used to query PD independently; if region split/merge happened while tasks started minutes apart, assigned key ranges could overlap (duplicate CDC) or leave gaps (lost data). The enumerator snapshots region boundaries, assigns one contiguous `KeyRange` per reader, and restore reuses checkpointed splits instead of rediscovering PD.

Runtime region split/merge is still handled by `CDCClient` **inside** the assigned key range. That only remaps physical regions onto a frozen key-space ownership and does not re-partition work across readers.

## Brief change log

- Add a thin FLIP-27 `TiKVSource` (`TiKVSourceEnumerator` + `TiKVSourceReader`)
- Enumerator queries PD once via `RangeSplitter.splitRangeByRegion`, clips/sorts/merges contiguous region chunks, and assigns one `TiKVKeyRangeSplit` per subtask
- Reader scans snapshot then streams CDC only for its assigned key range; `resolvedTs` is stored on the split
- Enumerator checkpoint stores unassigned splits; restore never rediscovers regions
- Fail the job if region meta cannot be fetched (no silent fallback to task0-owns-all-table)
- Table API uses `SourceProvider`; DataStream `TiDBSource.builder().build()` now returns `TiKVSource` (use `env.fromSource`)
- Deprecate `TiKVRichParallelSourceFunction` / `buildSourceFunction()`
- Keep the old handle-space helpers as deprecated APIs
- Update DataStream docs (en/zh)

## Verifying this change

- Added/Updated unit tests in `TableKeyRangeUtilsTest` and `TiKVKeyRangeSplitSerializerTest`
- Manually tested by our produce env job (region-based split)

## Documentation

- Does this pull request introduce a new feature? no (bugfix / correctness of parallel split)
- If yes, how is the feature documented? docs / JavaDocs (DataStream example switched to `fromSource`)

---

- [x] Yes (Cursor)

### DIFF
--- a/docs/content.zh/docs/connectors/flink-sources/tidb-cdc.md
+++ b/docs/content.zh/docs/connectors/flink-sources/tidb-cdc.md
@@ -237,15 +237,14 @@ The TiDB CDC source can work in parallel reading, because there is multiple task
 
 ### DataStream Source
 
-The TiDB CDC connector can also be a DataStream source. You can create a SourceFunction as the following shows:
-
-### DataStream Source
+The TiDB CDC connector can also be a DataStream source. Region-based splits are computed once by the source enumerator; restore reuses the checkpointed key ranges instead of querying PD again.
 
 ```java
+import org.apache.flink.api.common.eventtime.WatermarkStrategy;
 import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.connector.source.Source;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
-import org.apache.flink.streaming.api.functions.source.SourceFunction;
 import org.apache.flink.util.Collector;
 
 import org.apache.flink.cdc.connectors.tidb.TDBSourceOptions;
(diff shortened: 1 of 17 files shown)


lifulong:
@leonardBang @GOODBOY008 @lvyanquan  can you help review this change

lifulong:
@leonardBang @GOODBOY008 @lvyanquan can you help review this change?

Mrart:
This doesn't seem like the proper way to handle it. Usually, Flink CDC relies on JDBC for chunk splitting during the snapshot (full-load) phase, and the incremental phase stays in the same process. To prevent OOM errors, we should implement rate limiting on the gRPC client side.

PR #4509, opened 2026-08-17
[FLINK-40397][mysql] Add binlog position lag metrics for MySQL binlog reader

hadoopkandy:
                                                                                                                                                                                                                                 
  This PR adds binlog position lag metrics for the MySQL binlog reader, which measure the lag between the current consumed binlog offset and the latest master binlog offset.                                                      
                                                                                                                                                                                                                                 
  Unlike the existing `currentFetchEventTimeLag` metric (which only updates when there are events flowing), these metrics are meaningful even during idle periods — they reflect how far behind the reader is from the MySQL master
   at the position level.                                                                                                                                                                                                        
                                                                                                                                                                                                                                   
  The feature is **opt-in** via `scan.binlog.position-lag.interval.ms`. A value of `-1` (default) disables the feature entirely. When set to a positive value, the connector periodically executes `SHOW MASTER STATUS` at this    
  interval to calculate lag.                                   
                                                                                                                                                                                                                                   
  ### Brief change log                                                                                                                                                                                                             
                                                               
  - Added `BinlogLagCalculator` to compute binlog position lag, producing two independent results: transaction lag (GTID-based) and byte position lag (file+position based).                                                       
  - In `BinlogSplitReader`, conditionally fetch the master's current binlog offset via `SHOW MASTER STATUS` at the configured interval and store it in a shared `AtomicReference<BinlogOffset>`.                                 
  - In `MySqlRecordEmitter`, read the shared master offset and calculate lag against the current consumed offset, then report it via `MySqlSourceReaderMetrics`.                                                                   
  - Registered two new gauge metrics in `MySqlSourceReaderMetrics`:                                                                                                                                                                
    - `currentBinlogTransactionLag`: GTID-based transaction count lag (-1 when GTID is unavailable)                                                                                                                                
    - `currentBinlogBytePositionLag`: byte-level position lag (-1 when position info is unavailable)                                                                                                                               
  - Added configuration option `scan.binlog.position-lag.interval.ms` (default `-1`, disabled) to control the feature and polling frequency.                                                                                       
                                                                                                                                                                                                                                   
  ### Lag calculation strategy                                                                                                                                                                                                     
                                                                                                                                                                                                                                   
  | Mode | Metric | Lag meaning | Calculation |                                                                                                                                                                                    
  |------|--------|-------------|-------------|                                                                                                                                                                                  
  | GTID | `currentBinlogTransactionLag` | Transaction count difference | Sum of (master max txn ID - current max txn ID) per server UUID |                                                                                        
  | File-position (same file) | `currentBinlogBytePositionLag` | Byte offset difference | master position - current position |                                                                                                     
  | File-position (cross file) | `currentBinlogBytePositionLag` | Estimated byte difference | file sequence diff × 1,000,000 + master position - current position |                                                                
                                                                                                                                                                                                                                   
  ### Verifying this change                                                                                                                                                                                                        
                                                                                                                                                                                                                                   
  This change added tests:                                                                                                                                                                                                         
  - `BinlogLagCalculatorTest`: unit tests covering GTID mode (single/multiple UUIDs, disjoint intervals, starts-from-middle, same GTID fallback to position), file-position mode (same file, cross file, edge cases).
  - `MySqlRecordEmitterTest#testBinlogPositionLagMetricIsUpdated`: verifies that the byte position lag metric is updated correctly during record emission through the public API path.                                             
                                                                                                                                                                                                                                   
  ### Does this pull request potentially affect one of the following parts                                                                                                                                                         
                                                                                                                                                                                                                                   
  - Dependencies: no                                                                                                                                                                                                             
  - The public API: yes (new builder method `binlogPositionLagIntervalMs` and config option)
  - The serializers: no                                                                                                                                                                                                            
  - The runtime per-record code path: yes (lightweight gauge update during binlog phase only, disabled by default)                                                                                                                 
  - Anything that affects determine of shard: no                                                                                                                                                                                   
                                                                                                                                                                                                                                   
  ### Documentation                                                                                                                                                                                                                
                                                                                                                                                                                                                                   
  - [x] Does this pull request introduce a new feature? yes — new monitoring metrics and configuration option                                                                                                                      
  - [x] If yes, how is the feature documented? JavaDoc on metric constants; configuration option documented in both English and Chinese user-facing docs (`mysql-cdc.md`).

### DIFF
--- a/docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md
+++ b/docs/content.zh/docs/connectors/flink-sources/mysql-cdc.md
@@ -432,6 +432,19 @@ Flink SQL> SELECT * FROM orders;
       </td>
     </tr>
     <tr>
+      <td>scan.binlog.position-lag.interval.ms</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">-1</td>
+      <td>Long</td>
+      <td>
+        获取 master binlog 状态以计算延迟指标的时间间隔（毫秒）。<br>
+        默认值 -1 表示关闭该功能。设置为正值（如 10000）时，连接器会按此间隔定期执行 SHOW MASTER STATUS 查询，并计算两个延迟指标：<br>
+        <li>currentBinlogTransactionLag：基于 GTID 的事务数延迟（非 GTID 模式下报告 -1）</li>
+        <li>currentBinlogBytePositionLag：基于 binlog 文件和位置的字节位置延迟</li>
+        注意：该指标仅在 binlog 增量读取阶段生效，snapshot 全量阶段不会产出此指标。开启此功能会对 MySQL 服务器产生额外的周期性查询。<br>
+      </td>
+    </tr>
+    <tr>
       <td>scan.incremental.snapshot.backfill.skip</td>
       <td>optional</td>
       <td style="word-wrap: break-word;">false</td>
--- a/docs/content/docs/connectors/flink-sources/mysql-cdc.md
+++ b/docs/content/docs/connectors/flink-sources/mysql-cdc.md
@@ -457,6 +457,19 @@ Only valid for cdc 1.x version. During a snapshot operation, the connector will 
       </td>
     </tr>
     <tr>
+      <td>scan.binlog.position-lag.interval.ms</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">-1</td>
+      <td>Long</td>
+      <td>
+        The interval in milliseconds to fetch master binlog status for lag metrics.<br>
+        A value of -1 (default) disables the feature. When set to a positive value (e.g. 10000), the connector periodically executes SHOW MASTER STATUS at this interval to calculate two lag metrics:<br>
+        <li>currentBinlogTransactionLag: GTID-based transaction count lag (reports -1 when GTID is unavailable)</li>
+        <li>currentBinlogBytePositionLag: byte position lag based on binlog file and position</li>
+        Note: this metric is only reported during the binlog (incremental) reading phase; it is not available during the snapshot (full) phase. Enabling this feature adds periodic queries to the MySQL server.<br>
+      </td>
+    </tr>
+    <tr>
       <td>scan.incremental.snapshot.backfill.skip</td>
       <td>optional</td>
       <td style="word-wrap: break-word;">false</td>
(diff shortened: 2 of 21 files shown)


lvyanquan:
Could we first clarify the unit and semantics of `currentBinlogPositionLag`?

Currently, the same gauge represents different quantities:

 - GTID sets differ: the value is an estimated transaction count.
 - GTID sets are equal or GTID is disabled: the value is a byte-position difference.
 - Binlog files differ: the value is a synthetic estimate based on `fileSequenceDiff * 1_000_000`.

As a result, the metric may switch units at runtime and cannot be interpreted consistently or used with a stable alert threshold. The cross-file value also has no physical unit.

I think we should define whether this metric represents transaction count, byte lag, or an abstract position distance before finalizing the calculation. If both transaction and byte lag are useful, exposing separate metrics may be clearer.

lvyanquan:
And Could we make this metric opt-in through a source configuration instead of enabling it by default?

This is not a purely local metric. It executes `SHOW MASTER STATUS` against the MySQL server every 10 seconds. Enabling it by default means that all existing CDC jobs will start issuing additional queries after upgrading, even when users do not need this metric. When many CDC jobs connect to the same MySQL instance, these recurring queries may accumulate and introduce unexpected server-side overhead.

I suggest disabling this feature by default and providing a configuration option to enable it explicitly. When disabled, the connector should skip both metric registration and the periodic master-status query.

hadoopkandy:
> And Could we make this metric opt-in through a source configuration instead of enabling it by default?
> 
> This is not a purely local metric. It executes `SHOW MASTER STATUS` against the MySQL server every 10 seconds. Enabling it by default means that all existing CDC jobs will start issuing additional queries after upgrading, even when users do not need this metric. When many CDC jobs connect to the same MySQL instance, these recurring queries may accumulate and introduce unexpected server-side overhead.
> 
> I suggest disabling this feature by default and providing a configuration option to enable it explicitly. When disabled, the connector should skip both metric registration and the periodic master-status query.

 Great feedback. Addressed both concerns:                                                                                                                                                                                         
                                                                                                                                                                                                                                   
  1. Opt-in via configuration: Added scan.binlog.position-lag.interval.ms (default -1, meaning disabled). A positive value enables the feature and controls the polling interval. When disabled, both the periodic SHOW MASTER     
  STATUS query and metric registration are skipped entirely.                                                                                                                                                                       
  2. Metric semantics split: Replaced the single currentBinlogPositionLag with two independent metrics:                                                                                                                          
    - currentBinlogTransactionLag: GTID-based transaction count lag (available only in GTID mode)                                                                                                                                  
    - currentBinlogBytePositionLag: byte-level lag based on binlog file + position (always available when enabled)                                                                                                                 
                                                                                                                                                                                                                                   
  This avoids the ambiguity of mixing different units in one metric and gives users clear, actionable signals.

hadoopkandy:
@lvyanquan PTAL

PR #4513, opened 2026-08-21
[FLINK-40453][postgres] Not load table schema when initializing record emitter

naivedogger:

Fixes [FLINK-40453](https://issues.apache.org/jira/browse/FLINK-40453).

`PostgresPipelineRecordEmitter` currently builds all `CreateTableEvent`s eagerly in its constructor (and again in the `applySplit()` override), which opens a JDBC connection and fetches every captured table's schema up front — before any split is processed. This adds startup overhead that scales with the number of captured tables. A table's schema is only needed when its first record / low-watermark / schema-change event is emitted, so loading everything at initialization is wasteful.

## Brief change log

- Remove the eager `generateCreateTableEvent()` call from the constructor and drop the `applySplit()` override that pre-populates the cache.
- Resolve and cache each `CreateTableEvent` lazily on demand, falling back to a JDBC lookup only when the split state does not already carry the table schema.
- Behavior for snapshot / incremental / schema-change paths is unchanged.

---

## Verifying this change

- Added unit tests in `PostgresPipelineRecordEmitterTest`, asserting that constructing the emitter opens no JDBC connection and that assigning a snapshot split does not trigger loading of all table schemas.

## Documentation

---


loserwang1024 (approved):
LGTM

PR #4514, opened 2026-08-26
[FLINK-40473] Fix yarn-application mode failing to locate flink-cdc-dist jar with Flink version suffix

llphxd:

Submitting a YAML pipeline with `-t yarn-application` using the official
`flink-cdc-3.6.0-1.20` binary distribution always fails with:

java.io.FileNotFoundException: Failed to fetch Flink CDC dist jar from path: /path/to/flink-cdc-3.6.0-1.20/lib

The `lib` directory does contain the dist jar. `YarnApplicationDeploymentExecutor`
locates it by matching file names against
`^flink-cdc-dist-(\d+(\.\d+)*)(-SNAPSHOT)?\.jar$`, which only accepts `-SNAPSHOT`
or `.jar` after the version number.

Since 3.6.0 the binary distribution is released once per Flink minor version, and
the release script sets `revision` to `<release-version>-<flink-version>`.  The `-1.20` segment never matches the
pattern.

## JIRA
Fixes [FLINK-40473](https://issues.apache.org/jira/browse/FLINK-40473).

## Brief change log

- Match the dist jar by its `flink-cdc-dist-` prefix and `.jar` suffix instead of a
  version pattern, so the lookup no longer depends on how the version is formatted.
- Extract the lib directory scan into `getFlinkCDCDistJar(Path)` so it can be unit
  tested without mutating the `FLINK_CDC_HOME` environment variable.
- Add `YarnApplicationDeploymentExecutorTest`.

The set of names accepted by the old pattern is a strict subset of what the new
predicate accepts, so no previously working setup is affected.

---

## Verifying this change

- Added unit tests in
  `flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/deployment/YarnApplicationDeploymentExecutorTest.java`:
  - `testFindDistJarWhateverVersionItCarries` covers `flink-cdc-dist-3.5.0.jar`,
    `flink-cdc-dist-3.7-SNAPSHOT.jar`, `flink-cdc-dist-3.6.0-1.20.jar` and
    `flink-cdc-dist-3.7.0-2.0.jar`. The last two fail without this fix. A connector
    jar and a directory named like a dist jar are placed in the same lib directory
    to verify neither is picked up.
  - `testDistJarNotFound` verifies that a lib directory holding only connector jars
    still raises `FileNotFoundException`, i.e. the prefix match does not over-match.

The failure happens in the first statement of `deploy()`, before any interaction
with YARN, so the unit test covers the full code path that was broken.

## Documentation

### DIFF
--- a/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/deployment/YarnApplicationDeploymentExecutor.java
+++ b/flink-cdc-composer/src/main/java/org/apache/flink/cdc/composer/flink/deployment/YarnApplicationDeploymentExecutor.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.cdc.composer.flink.deployment;
 
+import org.apache.flink.cdc.common.annotation.VisibleForTesting;
 import org.apache.flink.cdc.common.utils.Preconditions;
 import org.apache.flink.cdc.composer.PipelineDeploymentExecutor;
 import org.apache.flink.cdc.composer.PipelineExecution;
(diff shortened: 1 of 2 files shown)


lvyanquan (approved):
+1.

PR #4519, opened 2026-08-31
[FLINK-40512][postgres] Cache typeRegistry to reduce the busy work on inferring schema change event

loserwang1024:

## Brief change log

List the key changes in this PR. 

---

## Verifying this change

*(or)*

- *Manually tested by ...*

## Documentation

---

### DIFF
--- a/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresPipelineRecordEmitter.java
+++ b/flink-cdc-connect/flink-cdc-pipeline-connectors/flink-cdc-pipeline-connector-postgres/src/main/java/org/apache/flink/cdc/connectors/postgres/source/reader/PostgresPipelineRecordEmitter.java
@@ -32,6 +32,7 @@ import org.apache.flink.cdc.debezium.DebeziumDeserializationSchema;
 import org.apache.flink.cdc.debezium.event.DebeziumEventDeserializationSchema;
 import org.apache.flink.connector.base.source.reader.RecordEmitter;
 
+import io.debezium.connector.postgresql.TypeRegistry;
 import io.debezium.connector.postgresql.connection.PostgresConnection;
 import io.debezium.data.Envelope;
 import io.debezium.relational.Table;
@@ -65,6 +66,8 @@ public class PostgresPipelineRecordEmitter<T> extends PostgresSourceRecordEmitte
     private final Set<TableId> alreadySendCreateTableTables;
     private final boolean includeDatabaseInTableId;
     private final Map<TableId, CreateTableEvent> createTableEventCache;
+
+    private TypeRegistry typeRegistry;
 
     public PostgresPipelineRecordEmitter(
             DebeziumDeserializationSchema<T> debeziumDeserializationSchema,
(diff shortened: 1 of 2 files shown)


flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceStreamFetcher.java
```diff
@@ -221,9 +221,10 @@ protected boolean shouldEmit(SourceRecord sourceRecord) {
             }
             // not in the monitored splits scope, do not emit
             return false;
+        } else if (taskContext.isSchemaChangeEvent(sourceRecord)) {
```
leonardBang: Would it make sense to first verify that an excluded relation-schema record can reach this branch through the production path? `RelationAwarePostgresSchema.applySchemaChangesForTable()` already checks `!isFilteredOut(table.id())` before dispatching, and Debezium's parent implementation applies the same table filter. The new test directly calls `CDCPostgresDispatcher.dispatch(excludedTable)`, so it bypasses that gate and only covers an artificially queued record. Could we add a test that produces an actual excluded-table relation/DDL event (or point to the path that bypasses this filter)? If none exists, this base fetcher API extension may be unnecessary.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/FetchTask.java
```diff
@@ -66,6 +66,14 @@ interface Context {
 
         boolean isDataChangeRecord(SourceRecord record);
 
+        /** Returns whether the given record is a schema change event. */
+        boolean isSchemaChangeEvent(SourceRecord record);
```
leonardBang: Could `isSchemaChangeEvent` be a default method returning `false`? `FetchTask.Context` is public, and adding an abstract method forces every existing implementation to recompile or implement it. Returning `false` retains the previous behavior here—records that are not data changes continue to fall through to the existing emit path—while Postgres can override it. This would also remove the need for a no-op MongoDB override.

yuxiqian: I'm a little dubious about this, as the `isSchemaChangeEvent` check should be deterministic, and providing a possibly wrong default implementation is misleading.

Do we really need this method? Can we just check it in the `shouldEmitSchemaChangeEvent` method? The invocation cost should be the same.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/external/IncrementalSourceStreamFetcher.java
loserwang1024: you're right, I have found the real cause is typeRegistry loading.

yuxiqian (approved):
Thanks Hongshun for investigating this, changes looks reasonable to me.

PR #4520, opened 2026-09-01
[FLINK-40513][runtime] Introduce multimodal AI functions for image understanding

haruki-830:
This PR adds multimodal AI function support to Flink CDC Pipeline transforms.

`AI_IMAGE_COMPLETE` uses the standard OpenAI-compatible vision chat protocol. `AI_IMAGE_EMBED` and `SupportsImageEmbedding` are added as extension points, but image embedding is not yet supported by any built-in client. DashScope-specific embedding is intentionally out of scope.

## Brief change log

- Added `SupportsImageTextGeneration` and `SupportsImageEmbedding`.
- Added `AI_IMAGE_COMPLETE` and `AI_IMAGE_EMBED`.
- Added runtime parsing, type inference, capability validation, and UDF precedence handling.
- Implemented OpenAI-compatible vision chat using MIME detection and Base64 data URLs.
- Added tests for supported image formats, Base64 encoding, null input, and invalid input.
- Updated the English and Chinese documentation.

## Verifying this change

- Updated `AiFunctionsTest` and `AiFunctionParserTest`.
- Updated `OpenAiCompatibleModelClientTest` and `OpenAiCompatibleModelClientFactoryTest`.
- Updated `FlinkPipelineAiFunctionITCase`.
- Verified the tests with both the default Flink profile and `-Pflink2`.
- Verified the complete repository with `mvn spotless:check`.

## Documentation

---

- [X] Yes (OpenAI Codex)

Generated-by: OpenAI Codex

### DIFF
--- a/docs/content.zh/docs/core-concept/ai-model.md
+++ b/docs/content.zh/docs/core-concept/ai-model.md
@@ -52,7 +52,7 @@ AI 模型可用于 transform 表达式中的文本生成、文本分析、embedd
 
 OpenAI-compatible 模型客户端通过标准 vision chat 支持 `AI_IMAGE_COMPLETE`。客户端会识别 PNG、JPEG、GIF 和 WebP 图片，并将图片编码为 Base64 data URL。图片为 `NULL` 时直接返回 `NULL`，且不会调用模型；图片为空或格式无法识别时，会在发送请求前报错。
 
-`AI_IMAGE_EMBED` 当前只提供框架函数和 provider capability。OpenAI-compatible 模型客户端不实现图片 embedding，社区发行包目前也没有可用于生产的图片 embedding provider。需要图片向量化的用户需要等待后续 provider 实现。
+`AI_IMAGE_EMBED` 当前只提供框架函数和 provider capability。OpenAI API 目前没有定义标准的图片向量化协议，因此图片 embedding 需要由具体 provider 单独实现。OpenAI-compatible 模型客户端不实现图片 embedding，社区发行包目前也没有可用于生产的图片 embedding provider。需要图片向量化的用户需要等待后续 provider 实现。
 
 六个专用文本函数使用内置英文 prompt 模板，但输入文本可以是任意语言。输入为 `NULL` 时直接返回 `NULL`，且不会调用模型；模型返回 `NULL` 时也返回 `NULL`。非空文本响应必须是语法合法的 JSON，否则当前记录处理失败，错误信息会标明具体 AI 函数。运行时只校验 JSON 语法，不校验响应字段是否存在或字段类型是否匹配。
 
--- a/docs/content.zh/docs/core-concept/ai-model.md
+++ b/docs/content.zh/docs/core-concept/ai-model.md
@@ -24,11 +24,11 @@ under the License.
 
 # AI 模型
 
-AI 模型可用于 transform 表达式中的文本生成、文本分析和 embedding。
+AI 模型可用于 transform 表达式中的文本生成、文本分析、embedding 和图片理解。
 
 ## AI Functions
 
-模型名称必须是字符串常量，并引用 `pipeline.model` 中声明的模型。文本函数要求模型客户端支持文本生成，`AI_EMBED` 要求模型客户端支持 embedding；Pipeline 会在执行前校验引用模型的 capability 是否匹配。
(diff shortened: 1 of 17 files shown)


docs/content.zh/docs/core-concept/ai-model.md
```diff
@@ -43,13 +43,24 @@ AI 模型可用于 transform 表达式中的文本生成、文本分析和 embed
...
+以下多模态函数从 `BYTES` 字段读取图片数据：
+
+| 函数 | 说明 | 返回类型 |
+|------|------|----------|
+| `AI_IMAGE_COMPLETE(model, image, prompt)` | 根据图片和自然语言 prompt 生成文本。 | `STRING` |
+| `AI_IMAGE_EMBED(model, image)` | 将图片转换为 embedding 向量。 | `ARRAY<FLOAT>` |
+
+OpenAI-compatible 模型客户端通过标准 vision chat 支持 `AI_IMAGE_COMPLETE`。客户端会识别 PNG、JPEG、GIF 和 WebP 图片，并将图片编码为 Base64 data URL。图片为 `NULL` 时直接返回 `NULL`，且不会调用模型；图片为空或格式无法识别时，会在发送请求前报错。
+
+`AI_IMAGE_EMBED` 当前只提供框架函数和 provider capability。OpenAI-compatible 模型客户端不实现图片 embedding，社区发行包目前也没有可用于生产的图片 embedding provider。需要图片向量化的用户需要等待后续 provider 实现。
```
yuxiqian: Shall we mention that OpenAI API does not provide standard image embedding protocols here?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
yuxiqian, line 987: Seems this method requires model name string to be a compile-time literal string. Besides introducing more complexity in `JaninoCompiler`, It forbids the following use case:

```sql
AI_COMPLETE(
   IF(important_job, 'ultimate_model', 'cheap_model'), ...
)
```

Do you think it's a good idea to postpone the capability check from SQL compile time (the first we met `CreateTableEvent`) to evaluation time (first `DataChangeEvent`)?

yuxiqian (commented):
Thanks Haruki for the nice work, just left some minor comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
haruki-830: Thanks for the suggestion! I agree that this direction makes sense. Deferring model resolution to evaluation time would remove the literal-only restriction in `JaninoCompiler` and enable dynamic model selection such as:
>
```sql
AI_COMPLETE(
    IF(important_job, 'ultimate_model', 'cheap_model'),
    content,
    prompt
)
```
>
One possible implementation would be to introduce an internal `AiModelClientResolver` that wraps the existing `Map<String, AiModelClient>` held by the transform operators and inject it into the evaluation scope as a single parameter. `JaninoCompiler` would preserve the model argument as a regular string expression and pass both the resolver and the evaluated model name to `AiFunctions`. The function could then preserve the existing `NULL` short-circuit behavior before resolving the client and validating its capability at evaluation time.
>
For literal model names, we could retain the existing validation as a fail-fast pre-check, while still using the same runtime resolution path during evaluation.
>
Since this changes the shared evaluation pipeline for the entire AI function family introduced in #4512, rather than only the image functions added by this PR, would you prefer this to be handled in a separate JIRA and PR, or should I include it in this PR?

docs/content.zh/docs/core-concept/ai-model.md
haruki-830: OK. Addressed in the latest commit.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/parser/TransformParser.java
yuxiqian: Thank you for the writeup! I was hoping we could removing special cases for AI functions and delaying checks to `impl.AiFunctions`, but that could be discussed in another PR.

yuxiqian (approved):

PR #4525, opened 2026-09-07
[FLINK-40572][runtime] Support dynamic AI model selection

haruki-830:

This PR fixes [FLINK-40572](https://issues.apache.org/jira/browse/FLINK-40572) and follows up on the remaining work discussed in [#4520](https://github.com/apache/flink-cdc/pull/4520).

Previously, the model argument of an AI function had to be a string literal. The parser extracted the model name before execution, and `JaninoCompiler` rewrote it to a pre-bound model client variable. As a result, expressions such as columns, `IF`, and `CASE` could not be used to select an AI model dynamically.

This PR evaluates the model argument for each record and resolves the selected model at runtime. It allows a pipeline to choose among the models declared in `pipeline.model` based on record contents.

## Brief change log

- Pass the declared model client map to generated expressions for runtime model lookup.
- Evaluate the model argument as a regular `STRING` expression for every record.
- Support dynamic model selection through columns, `IF`, and `CASE` expressions.
- Remove the static model extraction and capability validation from the parser and composer.
- Remove the special model-literal rewriting logic from `JaninoCompiler`.
- Validate null, undeclared, and capability-incompatible models when the AI function is invoked.
- Preserve null-input short-circuit behavior without resolving the model.
- Preserve the precedence of user-defined functions with the same name.
- Update the English and Chinese AI model documentation.

---

## Verifying this change

This change added and updated tests and can be verified as follows:

- Updated `AiFunctionsTest` to cover runtime model resolution, missing models, null model names, incompatible capabilities, and null-input short-circuiting.
- Updated `AiFunctionParserTest` to cover dynamic model expressions and user-defined function precedence.
- Updated `FlinkPipelineAiFunctionITCase` to verify dynamic model selection in a running pipeline.
- Verified the runtime and composer tests with Flink 1.20.
- Verified the runtime and composer tests with Flink 2.x using the `flink2` profile.
- Verified code formatting with `mvn spotless:check`.

## Documentation

- If yes, how is the feature documented? English and Chinese documentation

---

- [X] Yes (OpenAI Codex)

Generated-by: OpenAI Codex (GPT-5)
```

### CODE REVIEW

File: flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/ai/AiModelClientResolver.java [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.flink.cdc.common.model.AiModelClient;
+import org.apache.flink.cdc.common.utils.Preconditions;
+
+import javax.annotation.Nullable;
+
+import java.util.Map;
+
+/** Resolves AI model clients by their logical names during expression evaluation. */
+@Internal
+public class AiModelClientResolver {
```
yuxiqian: Seems this resolver is just wrapping the Map internally. Can we keep it as-is for simplicity?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/functions/impl/AiFunctions.java
```diff
@@ -101,43 +142,88 @@ private static BinaryVariant generateText(
...
+        float[] embedding = model.embedImage(image);
+        return embedding == null ? null : Floats.asList(embedding);
+    }
+
+    private static <T> T resolveModel(
+            AiModelClientResolver modelClientResolver,
+            String modelName,
+            String functionName,
+            Class<T> requiredCapability,
+            String capabilityName) {
```
yuxiqian: Stating the required interface explicitly could be useful:

```suggestion
    private static <T> T resolveModel(
            AiModelClientResolver modelClientResolver,
            String modelName,
            String functionName,
            Class<T> requiredCapability) {
```

and

```java
throw new UnsupportedOperationException(
    "Model '"
            + modelName
            + "' could not be used in "
            + functionName
            + " because it does not implement "
            + requiredCapability.getSimpleName()
            + " interface.");
```

yuxiqian (commented):
Thanks Haruki for the cleaning-up work, just left some minor comments.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/TransformExpressionCompiler.java
```diff
@@ -85,10 +70,8 @@ public static ExpressionEvaluator compileExpression(
                             argumentClasses.add(Class.forName(udfFunction.getClasspath()));
                         }
 
-                        for (String paramName : modelClients.keySet()) {
-                            argumentNames.add(paramName);
-                            argumentClasses.add(AiModelClient.class);
-                        }
+                        argumentNames.add(JaninoCompiler.DEFAULT_AI_MODEL_CLIENTS);
```
yuxiqian, line 73: Not relevant to this PR, but the previous argument declaration & binding implementation is rather weird:

- Projection column/time parameters are declared and bound in `ProjectionColumnProcessor`.
- Filter column/time parameters are declared and bound in `TransformFilterProcessor`.
- UDF and AI model parameters are declared in `TransformExpressionCompiler`.
- Their values are bound separately in `ProjectionColumnProcessor` and `TransformFilterProcessor`.

The order is consistent, but the logic is scattered. Shall we build the complete parameter layout in one shared place and let `TransformExpressionCompiler` only consume it? This could be done later in a following ticket.

yuxiqian (approved):
LGTM, just left a follow up note.

PR #4526, opened 2026-09-07
[FLINK-40552][runtime] Support asynchronous transform execution

haruki-830:

JIRA: https://issues.apache.org/jira/browse/FLINK-40552

This PR introduces opt-in ordered asynchronous execution for post-transform processing. It improves the throughput of I/O-bound transform expressions, such as AI model calls, while preserving CDC event ordering and schema consistency.

## Brief change log

- Extract the common post-transform logic into `PostTransformProcessor` for both synchronous and asynchronous execution.
- Add asynchronous execution options for enablement, timeout, capacity, and worker threads.
- Use ordered asynchronous waiting so that `DataChangeEvent`s can execute concurrently while outputs remain ordered.
- Treat `SchemaChangeEvent`s as barriers to preserve schema/data ordering.
- Preserve schema state across checkpoints and same-parallelism savepoints.
- Emit the restored `CreateTableEvent` only once before subsequent data events.
- Add compatibility adapters for both Flink 1.x and Flink 2.x.
- Document asynchronous execution options and thread-safety requirements.

## Verifying this change

This change added unit tests covering:

- Concurrent execution and ordered output of data change events.
- Schema change barriers and schema/data ordering.
- Timeout and transform exception propagation.
- Checkpoint restoration.
- Same-parallelism savepoint restoration.
- Exactly-once emission of the restored `CreateTableEvent`.
- Selection of the asynchronous post-transform operator by the composer.

## Documentation

---

- [X] Yes — OpenAI Codex (GPT-5)

Generated-by: OpenAI Codex (GPT-5)


flink-cdc-runtime/src/test/java/org/apache/flink/cdc/runtime/operators/transform/async/AsyncPostTransformFunctionTest.java
```diff
@@ -0,0 +1,526 @@
...
+    private static void drainMailbox(OneInputStreamOperatorTestHarness<Event, Event> harness)
+            throws Exception {
+        while (true) {
+            Mail mail = harness.getTaskMailbox().tryTake(TaskMailbox.MIN_PRIORITY).orElse(null);
+            if (mail == null) {
+                return;
+            }
+            mail.run();
+        }
+    }
```
yuxiqian, line 291: This test case seems too low-level and artificial. Replacing it with some IT cases might work better?

yuxiqian (commented):
Thanks Haruki for the contribution!

Could you please add IT / E2e cases to cover this feature? Currently there are merely 2 unit tests. For simplicity we may make current transform e2e cases parameterized.

yuxiqian:
CI is failing. Mind you take a look?

haruki-830:
The CI failure was caused by an outdated validation call that had already been removed by FLINK-40572. The branch has now been rebased and the stale call has been removed.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/async/AsyncPostTransformProcessor.java
yuxiqian, line 1: Though it's nice to extract common parts between sync and async operators to eliminate code duplication, it might be potentially risky as it affects existing code, and makes it pain if we decide to modify async implementation later.

Shall we create a parallel AsyncPostTransformFunction in dedicated `org.apache.flink.cdc.runtime.operators.transform.async` package, without modifying PostTransformOperator at all?

docs/content.zh/docs/core-concept/data-pipeline.md
```diff
@@ -125,5 +125,11 @@ under the License.
...
+| `transform.async-execution.worker-threads` | 每个异步 PostTransform 任务使用的工作线程数，默认值为 16。                                                                                                                                                                                                                                                                                                                                                                                                               | optional          |
+
+异步 PostTransform 适用于 AI 模型调用等 I/O 密集型表达式。DataChangeEvent 可能并发调用 UDF 和 AI 模型客户端，但输出和所有 SchemaChangeEvent 仍保持有序，因此 UDF 和 AI 模型客户端实现必须是线程安全的。为保证 schema 状态一致，checkpoint 或 savepoint 前会等待尚未完成的异步请求，长时间运行的请求可能会延长 checkpoint 时间。savepoint 只支持在并行度不变时恢复，并且不能在从已有 savepoint 恢复时开启或关闭此选项。
```
yuxiqian, line 135: Maybe we should mention that async transform feature is experimental and unstable now.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/async/AsyncPostTransformProcessor.java
haruki-830: OK, addressed in the latest commit!

docs/content.zh/docs/core-concept/data-pipeline.md
haruki-830: The documentation now marks async transform as experimental and unstable.

flink-cdc-flink2-compat/src/main/java/org/apache/flink/cdc/runtime/operators/AsyncWaitOperatorAdapter.java
```diff
@@ -0,0 +1,61 @@
...
+                processingTimeService,
+                mailboxExecutor);
+    }
+
+    @Override
+    public void prepareSnapshotPreBarrier(long checkpointId) throws Exception {
+        // The async function owns schema state, so it must not advance beyond records retained in
+        // AsyncWaitOperator's recovery queue.
+        endInput();
+        super.prepareSnapshotPreBarrier(checkpointId);
```
yuxiqian: Does this effectively forbids performing checkpoint with in-flight events?

IIUC this would make checkpoint duration much longer, but makes state recovery simpler. @lvyanquan Do you think it's acceptable?

flink-cdc-flink1-compat/src/main/java/org/apache/flink/cdc/runtime/operators/AsyncWaitOperatorAdapter.java
```diff
@@ -0,0 +1,61 @@
...
+        setup(parameters.getContainingTask(), parameters.getStreamConfig(), parameters.getOutput());
+    }
+
+    @Override
+    public void prepareSnapshotPreBarrier(long checkpointId) throws Exception {
+        // The async function owns schema state, so it must not advance beyond records retained in
+        // AsyncWaitOperator's recovery queue.
+        endInput();
+        super.prepareSnapshotPreBarrier(checkpointId);
+    }
```
yuxiqian, line 60: IIUC this adapter will ensure that checkpoints are completed after all inflight got handled. This may cause longer checkpoint time but eliminates some state-recovery issues.

@lvyanquan Do you think it's an acceptable trade-off?

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/async/AsyncPostTransformFunction.java
```diff
@@ -0,0 +1,958 @@
...
+ */
+public class AsyncPostTransformFunction extends RichAsyncFunction<Event, Event>
+        implements CheckpointedFunction, Serializable {
+
+    private static final long serialVersionUID = 1L;
+    private static final String TABLE_STATE_NAME = "async-post-transform-table-state";
+    private static final long EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS = 30L;
+    private static final Logger LOG = LoggerFactory.getLogger(AsyncPostTransformFunction.class);
+
+    private static final int TABLE_STATE_VERSION = 2;
```
lvyanquan: `TABLE_STATE_VERSION` starts at `2` but version 1 never existed in the codebase. Either start at `1` or add a brief comment explaining why, so future readers don't wonder what v1 was.

yuxiqian:
Kindly ping @lvyanquan and @haruki-830, just resolved some conflicts.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/async/AsyncPostTransformFunction.java
yuxiqian: Done.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/transform/async/AsyncPostTransformFunction.java
```diff
@@ -0,0 +1,958 @@
...
+    @Override
+    public void asyncInvoke(Event event, ResultFuture<Event> resultFuture) {
+        if (event instanceof DataChangeEvent) {
+            // Resolve a restored CreateTableEvent on the mailbox thread so ORDERED output
+            // deterministically prepends it to the first data event for the table.
+            TableId tableId = ((DataChangeEvent) event).tableId();
+            List<Event> prependedEvents = prependCreateTableEventIfNeeded(tableId);
+            asyncInvokeDataChangeEvent(event, resultFuture, prependedEvents);
+        } else {
+            asyncInvokeBarrierEvent(event, resultFuture);
```
lvyanquan, line 287: **`CreateTableEvent` can be emitted after a schema change (with a stale schema) on restore**

The restored `CreateTableEvent` is only prepended when the first post-restore event is a `DataChangeEvent` (`asyncInvoke`, line 279). The `SchemaChangeEvent` branch of `process()` (line 407) never marks the table as emitted.

Reproduced locally with `restore → AddColumnEvent → DataChangeEvent`:

```text
AddColumnEvent{A, addedColumns=[region AFTER blk]}
CreateTableEvent{A, schema=(id, name, blk)}   // stale — missing `region`
DataChangeEvent{A, ...}                        // transformed against the new schema
```

So the `CreateTableEvent` arrives after the schema change, and carries the pre-change schema while the following record uses the post-change one. Downstream `registerNewSchema` bumps the version for any incoming `CreateTableEvent`, so the latest evolved schema rolls back to a version without `region` — while `MetadataApplier` has already applied the `ALTER TABLE` and records are laid out for the new schema. The sink then fails on a column mismatch or silently drops the new column.

`ORDERED` mode can't help here, since the `CreateTableEvent` is packed inside the data event's own result.

Could you add a test for `restore → SchemaChangeEvent → DataChangeEvent`?

lvyanquan (approved):
+1.

PR #4412, opened 2026-05-25
[FLINK-39748][postgres] Fix snapshot timestamp drift for historical TIMESTAMP/DATE columns

JNSimba:
This closes FLINK-39748.

The Postgres CDC snapshot path reads column values via a bare `rs.getObject(i + 1)` in `PostgresScanFetchTask`. For TIMESTAMP / TIMESTAMPTZ / DATE columns, the PG JDBC driver constructs the returned `java.sql.Timestamp` / `java.sql.Date` through `GregorianCalendar` (default Julian/Gregorian cutover at 1582-10-15) using the JVM default time zone. This makes pre-cutover dates drift by N days (e.g. `0001-01-01` by 2 days), and also adds an LMT delta on JVMs whose default zone has an LMT segment (e.g. `Asia/Shanghai` is `+08:05:43` until 1901, vs `+08:00:00` after).

The Postgres logical decoding (streaming) path does not pass through `GregorianCalendar`, so the same row produces different Debezium records on snapshot vs streaming, breaking idempotent UPSERT semantics for downstream sinks.

This patch:

1. Routes the snapshot path through `PostgresConnection.getColumnValue`, which already does per-type dispatch for `MONEY` / `BIT` / `NUMERIC` / `TIME` / `TIMETZ`, by replacing the bare `rs.getObject(i + 1)` in `PostgresScanFetchTask.createDataEventsForTable` with `jdbcConnection.getColumnValue(rs, i + 1, column, table, databaseSchema)`. This mirrors how Debezium's own `RelationalSnapshotChangeEventSource` reads rows.

2. Extends the switch in `PostgresConnection.getColumnValue` with three new cases for `PgOid.TIMESTAMP` / `TIMESTAMPTZ` / `DATE`, reading the columns as `java.time.LocalDateTime` / `OffsetDateTime` / `LocalDate` via `rs.getObject(columnIndex, ...class)`. This bypasses `GregorianCalendar`. PG `+/-infinity` sentinels are preserved as `Timestamp(Long.MAX/MIN_VALUE)` to keep the existing downstream contract.

A regression test in `PostgresScanFetchTaskTest` snapshots boundary dates (`0001-01-01`, `1582-10-04`, `1582-10-15`, `1900-12-31`, `1901-01-02`, and a microsecond-precision value) for TIMESTAMP and DATE columns and asserts the produced Debezium record values match the proleptic-UTC expectation.

PR #4530, opened 2026-09-08
[FLINK-40594][cdc-common] Preserve TIME precision in the pipeline runtime

tchivs:

Fixes [FLINK-40594](https://issues.apache.org/jira/browse/FLINK-40594): the pipeline runtime stores TIME as millisecond-of-day, so every `TIME(p)` with `p > 3` is silently truncated even though `TimeType` accepts a precision of up to 9 and `DebeziumEventDeserializationSchema` already builds values through `fromMicroOfDay` / `fromNanoOfDay`.

Beyond the lost digits, `TimeData.equals`, `hashCode` and `compareTo` compared `millisOfDay`, so two `TIME(6)` values differing only in microseconds were indistinguishable and collapsed during deduplication, ordering and key comparison.

## Brief change log

- `TimeData` stores nanosecond-of-day and exposes `toMicroOfDay()` / `toNanoOfDay()`; equality, hash and comparison now use that value. `toMillisOfDay()` keeps its signature and truncating behaviour, so existing callers stay source and binary compatible.
- `TimeDataSerializer` becomes precision-aware, following the existing shape of `TimestampDataSerializer` / `LocalZonedTimestampDataSerializer` instead of introducing a new pattern:
  - `precision <= 3` keeps the historical four-byte millisecond encoding and resolves as `isCompatibleAsIs`, so existing state keeps working;
  - `precision > 3` uses an eight-byte nanosecond-of-day encoding and resolves as `isCompatibleAfterMigration`, through a versioned `TypeSerializerSnapshot` that still reads the legacy snapshot envelope and four-byte payload.
- The binary readers/writers (`BinaryRecordData`, `BinaryArrayData`, `AbstractBinaryWriter`, `BinaryArrayWriter`), the converters and `InternalSerializers` thread the declared precision through.
- Two existing expectations that pinned the truncation are corrected rather than relaxed:
  - the `TIME(6)` / `TIME(9)` transform expectation now asserts the retained fraction instead of `21:48:25.123`;
  - the temporal-function check compares `LOCALTIME` at whole-second granularity, because `LOCALTIME` and `CURRENT_TIME` are `TIME(0)` and previously carried a fraction their declared type does not have.

---

## Verifying this change

- *Added unit tests in `flink-cdc-common/src/test/java/org/apache/flink/cdc/common/data/TimeDataTest.java`* covering nanosecond retention and the equality/ordering behaviour.
- *Updated unit tests in `flink-cdc-runtime/.../serializer/data/TimeDataSerializerTest.java`*: it becomes an abstract base with concrete `TimeDataSerializer3Test` / `6Test` / `9Test`, plus a `TimeDataSerializerCompatibilityTest` asserting `isCompatibleAsIs` for the millisecond encoding, `isCompatibleAfterMigration` for the widened one, and successful reading of the legacy snapshot envelope and four-byte payload.
- *Updated unit tests in `flink-cdc-common/.../converter/InternalObjectConverterTest.java` and `JavaObjectConverterTest.java`.*
- *Updated integration tests in `flink-cdc-composer/.../FlinkPipelineTransformITCase.java` and `flink-cdc-pipeline-connector-postgres/.../PostgresFullTypesITCase.java`.* The Postgres case gains a direct microsecond-of-day assertion, because the existing `TIME(6)` expectation compared two equally truncated values and therefore passed vacuously.

Local results on top of `9f23c0356`, built and run on the JDK 11 baseline (`openjdk 11.0.24`; verified the emitted bytecode is class file major 55):

| Module | Result |
|---|---|
| `flink-cdc-common` | 96 tests, 0 failures |
| `flink-cdc-runtime` | 992 tests, 0 failures, 0 errors, 1 skipped |
| `flink-cdc-composer` (`FlinkPipelineTransformITCase`) | 71 tests, 0 failures |
| Spotless | passes |

The same modules were also green on JDK 17.

Two notes for reviewers:

- Existing `TimeDataSerializerTest` data included `TimeData.fromNanoOfDay(102400)`, `204800` and `409600`. Under millisecond storage all three collapsed to zero at construction, so the previous round-trip assertion passed while the intended distinct values never existed; the precision-parameterised tests replace it.
- The precision-parameterised serializer tests cover `TIME(3)` on the legacy four-byte encoding, `TIME(6)` and `TIME(9)` on the widened one, and the snapshot compatibility matrix, so a regression in either encoding path fails a specific test rather than a shared one.

## Documentation

---

## Update: follow-up commit for the CI failures

`GenericRecordDataSerializer` still encoded TIME as millisecond-of-day, so a `GenericRecordData` round trip kept truncating `TIME(p > 3)`. It now writes nanosecond-of-day under a new tag and still reads the legacy tag, covered by `RecordDataSerializerTest#testGenericRecordDataKeepsSubMillisecondTime`.

Behaviour note for the change log: because the internal converters normalise a value to its declared precision, `LOCALTIME` and `CURRENT_TIME` — both `TIME(0)` — now emit whole seconds instead of carrying a millisecond fraction their declared type does not have. `FlinkPipelineTransformITCase` and `TransformE2eITCase` assert that.

`PostgresFullTypesITCase` expected `18:00:22.123456` from the snapshot phase. That is not reachable today: `PostgresScanFetchTask` reads snapshot rows with `ResultSet#getObject`, and pgjdbc materialises `time` as `java.sql.Time`, which only carries milliseconds — the value is already truncated before it reaches the runtime. Measured on this branch: the fixture row `time_6_c = 18:00:22.123456` arrives from the snapshot as `nanoOfDay = 64822123000000` (`18:00:22.123`), while a row inserted afterwards with `time_6_c = 19:00:22.123456` arrives from the change stream as `nanoOfDay = 68422123456000` (`19:00:22.123456`). The snapshot expectation now records the millisecond resolution, and microsecond retention is asserted on the change stream. The snapshot read itself looks like a separate `[postgres]` defect and is left for its own JIRA.

---

Generated-by: Claude Code (Anthropic)

The change was prepared, verified and described with AI assistance; I have reviewed every hunk, ran the suites listed above, and take responsibility for the correctness of the change.


yuxiqian:
Thanks Lanny for the contribution, please check the CI failing cases as they seems relevant.

@leonardBang IIRC the TIME(3) precision has been discussed earlier, and the conclusion is we'd better keep the same memory layout with Flink SQL, truncating to 3-digits. As apache/flink#26954 has extended precision up to 9, should this be reconsidered now?

leonardBang:
> Thanks Lanny for the contribution, please check the CI failing cases as they seems relevant.
> 
> @leonardBang IIRC the TIME(3) precision has been discussed earlier, and the conclusion is we'd better keep the same memory layout with Flink SQL, truncating to 3-digits. As [apache/flink#26954](https://github.com/apache/flink/pull/26954) has extended precision up to 9, should this be reconsidered now?

Thanks @tchivs for the contribution and @yuxiqian for the investigation, I think it's time to support  TIME(p) with p > 3

tchivs:
Thanks @yuxiqian and @leonardBang. I pushed 2e37478 which fixes the CI failures; all six red jobs came down to three causes, and only two of them were failures of this change.

**1. `TransformE2eITCase#testTemporalFunctions` (the four Pipeline E2E jobs) — my oversight.**
`verifyDataRecord` carried the same `LOCALTIME` expectation that I had already corrected in `FlinkPipelineTransformITCase`, and I missed the E2E copy. Since the converters normalise a value to its declared precision, and `LOCALTIME` / `CURRENT_TIME` are `TIME(0)`, the pipeline now emits whole seconds there; both tests assert that consistently. I have added this to the change log in the description, because it is user-visible.

**2. `GenericRecordDataSerializer` — an incompleteness of the fix that CI does not cover.**
It still encoded TIME as millisecond-of-day, so a `GenericRecordData` round trip truncated `TIME(p > 3)` even after the binary path was fixed. It now writes nanosecond-of-day under a new tag and keeps reading the legacy tag, so old state stays readable. `RecordDataSerializerTest#testGenericRecordDataKeepsSubMillisecondTime` fails without that change.

**3. `PostgresFullTypesITCase#testTimeTypesWithTemporalMode{Adaptive,MicroSeconds}` (Pipeline Unit Tests 1.x and 2.x) — a pre-existing PostgreSQL snapshot defect that this change made visible.**
The `18:00:22.123456` expectation is not reachable in the snapshot phase: `PostgresScanFetchTask` reads snapshot rows with `ResultSet#getObject`, and pgjdbc materialises `time` as `java.sql.Time`, which only carries milliseconds, so the microseconds are gone before the value reaches the runtime. Under millisecond storage both sides of that assertion were equally truncated, which is why it used to pass. Measured on this branch:

| phase | `time_6_c` in PostgreSQL | emitted `nanoOfDay` |
|---|---|---|
| snapshot | `18:00:22.123456` | `64822123000000` (`18:00:22.123`) |
| change stream | `19:00:22.123456` | `68422123456000` (`19:00:22.123456`) |

So the runtime change does preserve microseconds where the source actually delivers them. The snapshot expectation now states the millisecond resolution with a comment, and microsecond retention is asserted on the change stream instead of the vacuous snapshot comparison. Fixing the snapshot read looks like a separate `[postgres]` change to me (reading `Types.TIME` as `LocalTime` rather than `java.sql.Time`); I would rather file that as its own JIRA than widen this PR — happy to do so if you agree.

On the memory-layout question: the serializer keeps the historical four-byte millisecond encoding and `isCompatibleAsIs` for `precision <= 3`, and only uses the eight-byte nanosecond-of-day encoding for `precision > 3`, reached through a versioned snapshot that still reads the legacy envelope. So `TIME(3)` state and layout are unchanged, and only the precisions that Flink SQL has now extended in apache/flink#26954 use the wider slot.

Verified locally on the JDK 11 target (built and run with JDK 17, Flink `1.20.3` profile): `flink-cdc-common` and `flink-cdc-runtime` 993 tests green, `PostgresFullTypesITCase` time-type and full-type cases green, `FlinkPipelineTransformITCase#testDateAndTimeCastingFunctions` and `#testTransformWithTemporalFunction` green, `flink-cdc-pipeline-e2e-tests` compiles, Spotless passes on the touched modules. I do not have a local environment to run the pipeline E2E suite itself, so I am relying on CI for that job.

tchivs:
CI came back on 2e37478: **18 of 19 checks green**, and every case that this PR actually broke now passes.

- `Pipeline Unit Tests` and `Pipeline Connectors Unit Tests 2.x` are green, so `PostgresFullTypesITCase` is fixed on both Flink lines.
- `Pipeline E2E Tests (1-Parallelism)`, `Pipeline E2E Tests 2.x (1-Parallelism)` and `Pipeline E2E Tests 2.x (4-Parallelism)` are green.
- Only `Pipeline E2E Tests (4-Parallelism) / test (11, 1.20.3, pipeline_e2e)` is red, and `TransformE2eITCase` reports `Tests run: 24, Failures: 0, Errors: 1` there — no assertion failure, so `testTemporalFunctions` passed in that job too.

The two remaining failures look unrelated to TIME:

1. `TransformE2eITCase#testAssortedSchemaTransform(boolean)[1]` — `TimeoutException: failed to get specific event: CreateTableEvent{tableId=transform_test_duaf6z.terminus, schema=columns={ID INT NOT NULL, VERSION STRING, NAME STRING}, ...}`. No temporal column is involved; the job simply did not reach that event inside the timeout.
2. `MySqlToIcebergE2eITCase#testSyncWholeDatabase` — the collected rows still had 10 fields while the expectation has 16 (`..., 1, 1, 1, 1, 1, 1, 1, 1, 1, 1`), i.e. the assertion ran before the added columns had propagated. Again no TIME column in the diff.

Both are in the 4-parallelism `1.20.3` job only, and that job flakes on `master` as well, independently of this PR: nightly run `34072345338` (2026-09-07) failed the same job on `MySqlToHudiE2eITCase#testSyncWholeDatabase`, and run `33935167454` (2026-09-05) failed the 1- and 4-parallelism E2E jobs.

I cannot re-run it myself (`Must have admin rights to Repository`). @yuxiqian could you kick off a re-run of that single job when you get a chance? If you would rather see a fresh full run, I can push a no-op commit instead — just say which you prefer.

tchivs:
Two findings while waiting, one of which changes what I said earlier about needing a new JIRA.

**The snapshot TIME truncation is already covered by FLINK-39748 / #4412.**
`PostgresConnection#getColumnValue` deliberately reads `PgOid.TIME` and `TIMETZ` as a string, with the comment "In order to guarantee that we resolve TIMETZ columns with proper microsecond precision". The microseconds are lost only because `PostgresScanFetchTask.createDataEventsForTable` bypasses that method with a bare `rs.getObject(i + 1)` — which is exactly the bypass @JNSimba removes in #4412 for TIMESTAMP / TIMESTAMPTZ / DATE. So no separate JIRA is needed; TIME comes along for free there.

That does create an ordering interaction worth flagging: once #4412 lands, `time_6_c` will arrive from the snapshot with microseconds, so the expectation I just corrected has to go back to `18:00:22.123456`. I pushed f178753, which says so in the test comment and names FLINK-39748, so whoever merges second sees why the line moves. #4412 does not currently touch `PostgresFullTypesITCase`, so if it merges after this PR its CI will flag that one line. The change-stream assertion stays valid either way.

**On the remaining red job:** `Pipeline E2E Tests (4-Parallelism)` fails across unrelated branches, not just here. Of the last 20 completed `Flink CDC CI` runs: `34074617215` (#4526's branch FLINK-40572) failed the 1- and 4-parallelism E2E jobs and its next run `34308394179` was green with no code change to those tests; `33623144908` on `master` failed the 1-parallelism jobs; `34179458569` on `master` failed `Source Unit Tests 2.x`; and nightly `34072345338` failed the same 4-parallelism job on `MySqlToHudiE2eITCase#testSyncWholeDatabase` — the same `testSyncWholeDatabase` shape as the `MySqlToIcebergE2eITCase` failure here.

A re-run should be enough; I still cannot trigger one myself.

tchivs:
One practical note: pushing f178753 replaced the check rollup, and its two workflow runs (`34480001818` Flink CDC CI, `34480001311` Labeler-Trigger) are sitting at `action_required`, so this PR currently shows no checks at all. Approving that run also re-runs the flaky job, so it covers the re-run request from my previous comment — sorry for the extra click, I did not expect a push from this fork to need approval again.

For the record, the green result I reported was on 2e37478: 18 of 19 checks green, with only `Pipeline E2E Tests (4-Parallelism) / test (11, 1.20.3, pipeline_e2e)` red on two failures unrelated to TIME. f178753 only touches comments in `PostgresFullTypesITCase` (naming FLINK-39748 as the change that will restore the `18:00:22.123456` snapshot expectation), so the code under test is identical to 2e37478.

yuxiqian:
@tchivs Retriggered.

Kindly reminder: please keep the reply concise and avoid pasting LLM generated outputs here to make our conversation more effective.

tchivs:
Thanks for the retrigger — and fair point on the length. Apologies, I'll keep replies short here.

That job's only failures were two flakes with no TIME column involved: `MySqlToIcebergE2eITCase#testSyncWholeDatabase` and a `CreateTableEvent` timeout in `TransformE2eITCase#testAssortedSchemaTransform`. `testTemporalFunctions` passed.

Heads-up for whoever merges second: the snapshot-side `time_6_c` truncation is what #4412 (FLINK-39748) fixes, so the expectation I patched to `18:00:22.123` needs to go back to `18:00:22.123456` when that lands.

tchivs:
Still red, but a different job and test this time: `Pipeline E2E Tests 2.x (4-Parallelism)` → `SchemaEvolvingTransformE2eITCase#testFineGrainedSchemaEvolution`, a `TimeoutException` waiting for an `AlterColumnTypeEvent` after 199s. Last run it was `MySqlToIcebergE2eITCase#testSyncWholeDatabase` plus a timeout in `TransformE2eITCase#testAssortedSchemaTransform`, on the 1.20.3 job. No TIME column in any of them.

Otherwise 18/19 green, `testTemporalFunctions` and the postgres time-type cases included. Three runs, three different E2E timeouts — shall we treat that suite as pre-existing flake and track it separately? I can file the JIRA if you'd like.

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/RecordData.java
```diff
@@ -170,6 +170,16 @@ public interface RecordData {
...
 
+    /**
+     * Returns the Time data at the given position using its declared precision.
+     *
+     * <p>The default implementation preserves compatibility with record implementations whose
+     * representation is independent of precision.
+     */
+    default TimeData getTime(int pos, int precision) {
+        return getTime(pos);
+    }
```
yuxiqian, line 181: It's concerning to modify public API. Shall we use the same memory layout for all precisions, so this would be unnecessary?

flink-cdc-common/src/main/java/org/apache/flink/cdc/common/data/binary/BinaryArrayData.java
```diff
@@ -226,6 +229,21 @@ public int getInt(int pos) {
         return BinarySegmentUtils.getInt(segments, getElementOffset(pos, 4));
     }
 
+    @Override
+    public TimeData getTime(int pos, int precision) {
+        assertIndexIsValid(pos);
+        if (precision <= 3) {
```
yuxiqian, line 235: It is possible not to use different memory layout? Seems fragile if precision were lost in the pipeline.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/serializer/data/writer/AbstractBinaryWriter.java
```diff
@@ -236,7 +236,11 @@ public void writeDate(int pos, DateData value) {
 
     @Override
     public void writeTime(int pos, TimeData value, int precision) {
-        writeInt(pos, value.toMillisOfDay());
+        if (precision <= 3) {
+            writeInt(pos, value.toMillisOfDay());
+        } else {
+            writeLong(pos, Long.MIN_VALUE | value.toNanoOfDay());
+        }
```
yuxiqian, line 243: If you refer to the writeTimestamp method, you may see that the precision for temporal types are not enforced, and we won't truncate over-precision digits or use different memory layouts depending on precisions. I would prefer follow the same way and avoid modifying too much public API.

flink-cdc-composer/src/test/java/org/apache/flink/cdc/composer/flink/FlinkPipelineTransformITCase.java
```diff
@@ -3072,7 +3072,7 @@ void testDateAndTimeCastingFunctions() throws Exception {
         assertThat(outputEvents)
                 .containsExactlyInAnyOrder(
                         "CreateTableEvent{tableId=default_namespace.default_schema.my_table, schema=columns={`id` INT NOT NULL,`date_0` DATE,`time_0` TIME(0),`time_3` TIME(3),`time_6` TIME(6),`time_9` TIME(9),`date_0_str` STRING,`time_0_str` STRING,`time_3_str` STRING,`time_6_str` STRING,`time_9_str` STRING}, primaryKeys=id, options=()}",
-                        "DataChangeEvent{tableId=default_namespace.default_schema.my_table, before=[], after=[1, 1999-12-31, 21:48:25, 21:48:25.123, 21:48:25.123, 21:48:25.123, 1999-12-31, 21:48:25, 21:48:25.123, 21:48:25.123, 21:48:25.123], op=INSERT, meta=()}",
+                        "DataChangeEvent{tableId=default_namespace.default_schema.my_table, before=[], after=[1, 1999-12-31, 21:48:25, 21:48:25.123, 21:48:25.123456, 21:48:25.123456789, 1999-12-31, 21:48:25, 21:48:25.123, 21:48:25.123456, 21:48:25.123456789], op=INSERT, meta=()}",
```
yuxiqian, line 3075: Please also add tests to verify how transform / routing operators handle TIME with different precisions.

yuxiqian (commented):
@tchivs I just left some comments, PTAL when available.

Plus, please resolve conflicts with latest master branch.

yuxiqian:
After carefully reviewed this PR, I believe it's not possible to extend TIME precision without breaking backwards compatibility.

As AsyncTransformOperator may serialize DataChangeEvents into states, it's not safe to change the underlying type from INT to LONG. TIME(>3) fields could not be deserialized correctly if we choose a higher underlying INT type, as current serializer isn't versioned.

The original Flink PR https://github.com/apache/flink/pull/26954#discussion_r2316303536 didn't modify it, too. Thus, I suggest postponing this PR for now.

cc @leonardBang @ruanhang1993 WDYT?

leonardBang:
Backwards compatibility is important for users, +1 to postpone this PR.

PR #4538, opened 2026-09-14
[FLINK-40646][Postgres] Pooled JDBC connections use the default application name.

loserwang1024:
AI-Contributed/Feature: 0/12
AI-Contributed/UT: 0/4

## Brief change log

List the key changes in this PR. 

---

## Verifying this change

*(or)*

- *Manually tested by ...*

## Documentation

---


leonardBang (commented):
## Review: [FLINK-40646][Postgres] Pooled JDBC connections use the default application name

### Strengths

- **The fix hits the root cause.** Although `PostgresDialect.openJdbcConnection` passes `CONNECTION_NAME` as `connectionUsage` to `PostgresConnection` (where shaded Debezium's `addDefaultSettings` injects `ApplicationName`), the custom `JdbcConnectionFactory` goes through the HikariCP pooled path: `JdbcConnectionPoolFactory.createPooledDataSource` builds the URL only from `getJdbcUrl(sourceConfig)`, so the `ApplicationName` in `JdbcConfiguration` is entirely discarded — this is exactly why pooled connections show up in `pg_stat_activity` as the pgjdbc default `PostgreSQL JDBC Driver`. Encoding `ApplicationName` directly into the JDBC URL in `PostgresConnectionPoolFactory.getJdbcUrl` is the minimal, on-target fix.
- **The parameter name is correct.** The URL parameter `ApplicationName` exactly matches pgjdbc's `PGProperty.APPLICATION_NAME` (verified against 42.5.5/42.3.10 bytecode; the default is indeed `PostgreSQL JDBC Driver`).
- **Naming consistency.** Reusing `PostgresDialect.CONNECTION_NAME` rather than inventing a new string makes pooled connections, the non-pooled `PostgresConnection`, and the replication connection (walsender side already carries `ApplicationName` via `addDefaultSettings`) all appear as a unified `postgres-cdc-connector` in `pg_stat_activity` — achieving the observability goal.
- **The regression test is genuinely effective.** `testJdbcConnectionApplicationName` obtains a *pooled* connection via `dialect.openJdbcConnection()`, and the `SHOW application_name` assertion would return `PostgreSQL JDBC Driver` and fail before the fix — a true behavior-level regression test, not merely a check of URL string concatenation; the connection is correctly returned/closed via try-with-resources.
- **Test infrastructure is updated in sync.** `PostgresTestBase.getJdbcConnection` adapts to the new 4-parameter `JDBC_URL_PATTERN`, with no compilation gaps (repo-wide grep confirms no other `new PostgresConnectionPoolFactory()` call sites).

### Issues

#### Critical (must fix)

- None. The fix path is correct, the change surface is small, and tests cover the fixed behavior; no data-consistency or resource-leak issues found.

#### Important (should fix)

- **[PostgresDialectTest.java:114] The test hardcodes the expected value as the literal `"postgres-cdc-connector"`, while the production single source of truth is `PostgresDialect.CONNECTION_NAME` (private).** The two strings could drift apart silently in the future; a drift would currently make the test fail (acceptable), but a better approach is to promote `CONNECTION_NAME` to a `@VisibleForTesting` package-level/public constant and reference it from the test. Suggestion: change the constant in `PostgresDialect` to `static final String CONNECTION_NAME` (drop `private`) or add `@VisibleForTesting`, and reference it in the test. If the PMC considers hardcoding acceptable (as a lock on the default value), this can be downgraded.

#### Minor (nice to have)

- **[PostgresConnectionPoolFactory.java:32-34] The implicit no-arg constructor was removed, and `JDBC_URL_PATTERN` (a public constant) changed semantics from 3 to 4 parameters** → a source-incompatible change for downstream code that directly instantiates this public class or references the constant. The class carries no `@Public`/`@PublicEvolving` annotation and is connector-internal, so risk is low. Suggestion: optionally keep a no-arg constructor delegating to the default name, or add `@Internal` semantics on the class; add Javadoc to `JDBC_URL_PATTERN` and the constructor (explaining the `%s` order and the purpose of `applicationName`).
- **[PostgresConnectionPoolFactory.java:42] `String.format` concatenates `applicationName` without URL-encoding.** The current value `postgres-cdc-connector` is safe, but if this value ever becomes configurable (containing spaces, `&`, `=`, or non-ASCII), it would break the URL or inject extra parameters. Suggestion: validate with `Preconditions.checkNotNull` in the constructor, or comment that only URL-safe values are accepted; for longer-term evolution, consider injecting via Hikari's `addDataSourceProperty`.
- **[PostgresConnectionPoolFactory.java:29 + JdbcConnectionPools reuse] `applicationName` becomes factory instance state, but `getPoolId`'s `ConnectionPoolId` does not include it.** If factories are ever constructed with different applicationNames in the same JVM, `JdbcConnectionPools` (a static singleton cached by host/port/user/database/identifier) would reuse the first-built pool and silently keep the old application_name. Currently only `CONNECTION_NAME` is used repo-wide, so no actual bug. Suggestion: a comment documenting this constraint suffices.
- **[PostgresTestBase.java:129] The test infra passes the literal `"test-driver"` as the application name.** No functional impact, but extracting it to a constant (e.g. `TEST_APPLICATION_NAME`) would improve readability.
- **[Overall design] The PR introduces no new `ConfigOption`; application_name is fixed to `postgres-cdc-connector` and not user-customizable.** As a bugfix (correcting the default value) the scope is reasonable and the default appropriate. If the community wants it configurable (e.g. appending a job name to distinguish multiple Flink jobs in `pg_stat_activity`), suggest opening a follow-up JIRA rather than expanding this PR.

### Verdict

**Ready to merge?** Yes

**Reasoning:** The fix precisely targets the root cause (the pooled URL dropping `ApplicationName`), the implementation is minimal, the naming is consistent with the non-pooled path, and it comes with a real regression test that fails before the fix. All findings are non-blocking Minor-level improvements (constant reuse, Javadoc, URL-encoding defense) that can be handled in this PR or followed up later.

loserwang1024:
@leonardBang  fix the "Important (should fix)" one.

leonardBang (approved):

PR #4541, opened 2026-09-14
[FLINK-40656][runtime] Reset schema coordinator state on checkpoint restore

fxbing:

Fix repeated `Unexpected evolving status: EVOLVING` failures after a failed schema update and same-instance coordinator recovery. Restoring only the schema manager leaves the previous attempt's transient protocol state behind, preventing schema requests from being replayed. See [FLINK-40656](https://issues.apache.org/jira/browse/FLINK-40656).

## Brief change log

- Fence queued and incoming work during reset, and wait for the previous schema-change worker to terminate before installing restored state.
- Reinitialize transient state in both schema coordinators, restore an empty schema manager for null checkpoint state, and preserve schemas restored before startup.
- Complete cancelled request/checkpoint futures exceptionally and suppress failure callbacks from the resetting attempt.

---

## Verifying this change

- Added coordinator regressions for failed-apply recovery, null state, restore-before-start, late success/failure, bounded reset timeout/retry, and queued request/checkpoint rejection. The original recovery test reproduces `Unexpected evolving status: EVOLVING` without the fix.
- Verified `SchemaCoordinatorTest` and `SchemaEvolveTest` with Java 11 / Flink 1.20.3 and Java 17 / Flink 2.2.0 using clean builds; 20 tests passed in each configuration.
- Spotless and `git diff --check` passed.
- A temporary embedded Fluss recovery test passed with the three modified coordinator classes overlaid on the existing Java 11 / Flink 1.17 test environment: stop the Sink Coordinator, add a source column, restore the coordinator, and verify target schemas, full rows, and PK updates/deletes. Log replay duplicates are allowed; this was not a packaged-release E2E run.

## Documentation

- Does this pull request introduce a new feature? No.
- If yes, how is the feature documented? Not applicable.

---

- [X] Yes — Codex CLI; Claude assisted with design and review.

Generated-by: Codex CLI 0.153.4


flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/common/SchemaRegistry.java
```diff
@@ -155,6 +165,12 @@ public void close() throws Exception {
     /** Restore schema registry state from byte array. */
     protected abstract void restore(byte[] checkpointData) throws Exception;
 
+    /** Stops schema-change work from the previous coordinator generation. */
+    protected abstract void quiesceSchemaChangeExecutor() throws Exception;
+
+    /** Recreates transient state after a coordinator reset. */
+    protected abstract void reinitializeTransientState();
```
yuxiqian: Shall we call these methods `initialize` and `shutdown`? `initialize` may be invoked at the first time. And we may need to do more cleaning-up work besides shutting down executors.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
```diff
@@ -227,10 +252,9 @@ protected void handleUnrecoverableError(String taskDescription, Throwable t) {
         super.handleUnrecoverableError(taskDescription, t);
 
         // For each pending future, release it exceptionally before quitting
-        pendingRequests.forEach(
-                (index, tuple) -> {
-                    tuple.f1.completeExceptionally(t);
-                });
+        if (pendingRequests != null) {
```
yuxiqian, line 255: Could you please add some IT case in flink-cdc-composer, and some E2e cases to cover both regular and distributed topology?

yuxiqian (commented):
Thanks Xiaobing for the nice work! Just left some minor comments here.

flink-cdc-runtime/src/main/java/org/apache/flink/cdc/runtime/operators/schema/regular/SchemaCoordinator.java
```diff
@@ -119,8 +120,32 @@ public SchemaCoordinator(
     @Override
     public void start() throws Exception {
         super.start();
+        initialize();
```
yuxiqian: Call `initialize();` in `SchemaRegistry#start`?

yuxiqian:
Thanks for the rapid response, LGTM overall.

Shall we call `initialize` and log in base SchemaRegistry and remove overriding `start` in coordinators?

```java
// SchemaRegistry.java:
@Override
public void start() throws Exception {
    LOG.info("Starting SchemaRegistry - {}.", operatorName);
    initializeBaseRuntimeState();
    initialize();
    LOG.info("Started SchemaRegistry for {}. Parallelism: {}", operatorName, currentParallelism);
}
```

yuxiqian (approved):

PR #4545, opened 2026-09-16
[hotfix][route] Revert strict route rule regex matching case sensitivity

yuxiqian:

Prior to Flink CDC 3.6, route rules were matched case-insensitively.

In Flink CDC 3.6, [FLINK-38779](https://issues.apache.org/jira/browse/FLINK-38779) replaced `Selectors` with Java regular expressions in `TableIdRouter` to support standard RegExp replacement rules. However, the new `Pattern` was compiled without `Pattern.CASE_INSENSITIVE`, unintentionally changing the long-standing route matching behavior to case-sensitive.

As a result, existing route rules may no longer match when the database or table name casing differs from the emitted `TableId`. The source table then falls through without being routed, which is a backward-incompatible regression for jobs upgraded to Flink CDC 3.6.

This PR restores the case-insensitive route matching behavior used before Flink CDC 3.6 while retaining the RegExp replacement support introduced by FLINK-38779.

## Brief change log

Compile route rule patterns with `Pattern.CASE_INSENSITIVE` to restore the original behavior.

---

## Verifying this change

Unit tests added.

## Documentation

---

- [X] Yes (OpenAI Codex)

Generated-by: OpenAI Codex (gpt-5.6-sol)


lvyanquan (approved):
LGTM.

PR #4546, opened 2026-09-16
[ci][tests] Split pipeline e2e into stages

yuxiqian:
This PR splits pipeline e2e into 3 stages to avoid execution timeout.

As observed in #4541, we're almost hitting the 90 minutes time limit.


lvyanquan (approved):
+1.

PR #4549, opened 2026-09-17
[FLINK-40690][pipeline-connector/fluss] Preserve nested row field positions

fxbing:

Fix incorrect nested ROW values in the Fluss pipeline sink when the parent column mapping reorders columns or does not cover every child position. For example, mapping `[nested, id]` to `[id, nested]` must preserve `(7, 8)` inside the nested row. See FLINK-40690.

## Brief change log

- Wrap nested records with their own identity mapping using the existing factory, instead of reusing the parent mapping.
- Add regressions for reordered and sparse parent mappings, wider children, null fields, multiple ROW levels, and BinaryRecordData.
- Add a pipeline regression that writes to a pre-existing Fluss table with reordered top-level columns and reads back the complete result, asserting exactly `Row.of(42, Row.of(7, 8))` without a `LIMIT`.

## Verifying this change

- All six new regression tests fail with assertion failures before the fix.
- The new pipeline regression also fails against the old implementation after a real write/read round trip, returning `Row.of(42, Row.of(8, 7))`.
- Clean reactor builds passed all 28 selected tests with both Java 11 / Flink 1.20.3 and Java 17 / Flink 2.2.0: six `CdcAsFlussRowTest` cases, 19 `FlussConversionsTest` cases, two `FlussEventSerializationSchemaTest` cases, and the new `FlussSinkPipelineITCase#testNestedRowWithReorderedColumns`. No failures, errors, or skips.
- Spotless apply/check and `git diff --check` passed.
- Tests used an isolated local Maven repository with a consistent set of cached Apache Fluss `20260901.094454` SNAPSHOT artifacts, avoiding incompatible locally built SNAPSHOT jars. No project dependency or POM changes were required.

## Documentation

- Does this pull request introduce a new feature? No.
- If yes, how is the feature documented? Not applicable.

- [X] Yes — Codex CLI; Claude and GLM assisted with analysis, implementation, and review.


loserwang1024 (approved):
LGTM

PR #4550, opened 2026-09-17
[FLINK-40697][cdc-base] Release finished snapshot split metadata after entering the stream phase

spoorthibasu:

The incremental source framework (flink-cdc-base) keeps the full snapshot-split metadata (assigned splits, finished offsets, table schemas) in the split enumerator for the whole lifetime of the job and re-serializes it on every checkpoint. On a large multi-table job this keeps JobManager memory high and inflates checkpoint size.
 
FLINK-39775 fixed this for the MySQL connector. This PR generalizes the same mechanism into flink-cdc-base so the JDBC incremental dialects (Postgres, SqlServer, Oracle, Db2) get it, rather than each connector solving it separately.
 
The release is opt-in behind a new option `scan.incremental.snapshot.metadata.release.enabled` (default `false`), so existing jobs are unaffected. Once the stream split is assigned and the reader reports its metadata assembled, the enumerator schedules the release at the current checkpoint and clears it in `notifyCheckpointComplete`, so the assignment is always covered by a completed checkpoint. An assignment-generation guard rejects a stale assembled report from a reader that failed over. The checkpoint serializer is bumped to version 9 only for jobs that enable the option, so default-off jobs keep writing version 8 and stay restorable by an older build.

## Brief change log

- Add `scan.incremental.snapshot.metadata.release.enabled` (default `false`) to the base source options and config
- Add the release primitive to the snapshot assigner and thread it through the split and hybrid assigners
- Add the reader-to-enumerator assembled event and the enumerator release orchestration (arm in `snapshotState`, execute in `notifyCheckpointComplete`), with the assignment-generation guard
- Persist the released flag; `PendingSplitsStateSerializer` writes version 9 only when the option is enabled, else version 8
- Fail fast on release together with `scan.newly-added-table.enabled`, and on restoring an already-released state with the option off
- Expose the option through the DataStream builders and Flink SQL for Postgres, SqlServer, Oracle, and Db2
- Document the option (EN and ZH) for the four connectors
---

## Verifying this change

- Unit tests in flink-cdc-base for the release protocol, the generation guard, the conditional serializer versioning (version 8 vs 9 by the option), and the fail-fast paths
- Integration tests on real Postgres and SqlServer databases covering the group-fetch release path plus JobManager and TaskManager failover in the stream phase, verifying no snapshot or stream data is lost through the release and restore

## Documentation

The Chinese doc entries mirror the English and are a best-effort translation. Please correct the wording where needed.
---


flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/test/java/org/apache/flink/cdc/connectors/base/source/enumerator/IncrementalSourceEnumeratorReleaseTest.java
```diff
@@ -0,0 +1,458 @@
...
+            enumerator.notifyCheckpointComplete(100L);
+            assertThat(assigner.released).isFalse();
+            assertThat(finishedSnapshotSplitMeta(enumerator)).isNotNull();
+        }
+    }
+
+    private static Object finishedSnapshotSplitMeta(IncrementalSourceEnumerator enumerator)
+            throws Exception {
+        java.lang.reflect.Field field =
+                IncrementalSourceEnumerator.class.getDeclaredField("finishedSnapshotSplitMeta");
```
lvyanquan: **nit: use a package-private getter instead of reflection to access `finishedSnapshotSplitMeta`**

`IncrementalSourceEnumeratorReleaseTest` uses reflection to inspect `IncrementalSourceEnumerator.finishedSnapshotSplitMeta` to verify the cache is cleared after release. Reflection is brittle — renaming the field would silently break the test.

Consider adding a package-private getter on `IncrementalSourceEnumerator`:

```java
@VisibleForTesting
List<FinishedSnapshotSplitInfo> getFinishedSnapshotSplitMeta() {
    return finishedSnapshotSplitMeta;
}
```

Minor, not blocking.

lvyanquan (commented):
A high-quality PR overall. I only have one minor suggestion regarding the test code, which is not a blocker.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/test/java/org/apache/flink/cdc/connectors/base/source/enumerator/IncrementalSourceEnumeratorReleaseTest.java
spoorthibasu: Replaced the reflection with a package-private `@VisibleForTesting` `getFinishedSnapshotSplitMeta()` on the enumerator, and the test now calls it directly.

spoorthibasu:
Thanks for the review @lvyanquan. I've also prepared the same wiring for the MongoDB connector, which is the one remaining incremental dialect. Happy to open that as a follow-up once this is merged.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/IncrementalSourceReader.java
```diff
@@ -481,6 +501,14 @@ private void requestStreamSplitMetaIfNeeded(StreamSplit streamSplit) {
         } else {
             LOG.info("The meta of stream split {} has been collected success", splitId);
             this.addSplits(Collections.singletonList(streamSplit));
+            // Report the stream-split metadata assembled so the coordinator can release what it no
+            // longer needs to serve. Carries the assignment generation so a stale report from a
+            // failed attempt is ignored.
+            context.sendSourceEventToCoordinator(
```
lvyanquan: addSplits(streamSplit) enters the completed stream-split branch above, which already sends a StreamSplitMetaAssembledEvent. This therefore sends the event twice on the group-fetch path. Could we remove this duplicate notification and update the comment in the completed-split branch to also mention metadata assembled through group fetch? Non-blocking, since the coordinator handling is idempotent.

docs/content/docs/connectors/flink-sources/db2-cdc.md
```diff
@@ -263,6 +263,13 @@ Db2 server.
           If the flink version is greater than or equal to 1.15, the default value of 'execution.checkpointing.checkpoints-after-tasks-finish.enabled' has been changed to true,
           so it does not need to be explicitly configured 'execution.checkpointing.checkpoints-after-tasks-finish.enabled' = 'true'
       </td>
+    </tr>
+    <tr>
+      <td>scan.incremental.snapshot.metadata.release.enabled</td>
+      <td>optional</td>
+      <td style="word-wrap: break-word;">false</td>
+      <td>Boolean</td>
+      <td>Whether to release the snapshot split metadata (assigned splits, finished offsets and table schemas) held by the source coordinator once the source has entered the stream phase, to reduce JobManager memory on jobs with a very large number of snapshot splits. Disabled by default. Incompatible with scan.newly-added-table.enabled: enabling both fails at startup, and a job that has released the metadata cannot later enable newly-added-table scanning. Release happens only after a successful checkpoint; if checkpointing is disabled or no checkpoint completes, the metadata is retained, so this option has no effect without checkpointing.</td>
```
lvyanquan: Could we document the downgrade restriction for all connectors that expose this option, including MySQL? Once a checkpoint or savepoint is created with this option enabled, it cannot be used to restore the job after downgrading to Flink CDC 3.6.0 or an earlier version. MySQL has the same state compatibility restriction and exposes this option in both the Flink Source and Pipeline Connector APIs, so the corresponding English and Chinese MySQL documentation should be updated as well. This is non-blocking.

flink-cdc-connect/flink-cdc-source-connectors/flink-cdc-base/src/main/java/org/apache/flink/cdc/connectors/base/source/reader/IncrementalSourceReader.java
spoorthibasu: Removed the duplicate send, and the comment in the completed-split branch now mentions the group-fetch case along with the others that reach it.

docs/content/docs/connectors/flink-sources/db2-cdc.md
spoorthibasu: Documented for the four dialects and for MySQL, in both the Flink Source and Pipeline Connector docs, EN and ZH.

The Db2 docs also drop the `scan.newly-added-table.enabled` line, since Db2 doesn't expose that option.

spoorthibasu:
Thanks @lvyanquan, both are addressed inline.

Three other changes in this push:

- The release clears its armed checkpoint once it runs, so a stream-only job, which has no snapshot metadata to drop, doesn't re-enter that path on every later checkpoint.
- `IncrementalSourceReader` coverage for the assembled event: `testReaderSendsStreamSplitMetaAssembledEventForCompleteSplit` on the inline path and `testReaderEchoesAssignmentGenerationInAssembledEvent` on the group-fetch path.
- Ticket references follow the same split throughout, FLINK-40697 for what this ticket adds and FLINK-39775 where it points back at the original MySQL change.

lvyanquan (approved):
+1.

PR #4553, opened 2026-09-21
[FLINK-40732][mongodb] Release finished snapshot split metadata after entering the stream phase

spoorthibasu:

The MongoDB CDC connector keeps the snapshot split metadata (assigned splits, finished offsets, table schemas) in the source coordinator for the whole lifetime of the job and re-serializes it on every checkpoint. On a job with a very large number of snapshot splits this keeps JobManager memory high and inflates checkpoint size.

FLINK-40697 added the release to the incremental source framework (flink-cdc-base) behind `scan.incremental.snapshot.metadata.release.enabled`. The four JDBC dialects inherit the option from `BaseSourceConfig` and its validation from `JdbcSourceConfig`. MongoDB is built on the same framework but implements `SourceConfig` directly, so it inherits neither and the option cannot be enabled there. This PR adds it, which makes MongoDB the last connector built on flink-cdc-base to get the option.

The option stays opt-in and defaults to `false`, so existing jobs are unaffected. Unlike the JDBC dialects, MongoDB defaults `scan.incremental.snapshot.enabled` to `false`, so the release applies only when incremental snapshot is also enabled.

## Brief change log

- Add the `releaseSnapshotMetadataEnabled` field, getter and factory setter to `MongoDBSourceConfig` and `MongoDBSourceConfigFactory`
- Fail fast when the option is enabled together with `scan.newly-added-table.enabled`, which MongoDB supports; the check is implemented here because MongoDB does not extend `JdbcSourceConfig`
- Expose the option through `MongoDBSourceBuilder`, and through Flink SQL in `MongoDBTableSourceFactory` and `MongoDBTableSource`
- Document the option (EN and ZH), including the downgrade restriction

---

## Verifying this change

- Unit tests in `MongoDBSourceConfigFactoryTest` for the default, for the setter reaching the config, and for the fail-fast on both options
- A unit test in `MongoDBTableFactoryTest` that sets the option in the DDL and asserts it reaches `MongoDBTableSource`
- Integration tests on a real MongoDB container in `MongoDBParallelSourceITCase` covering the group-fetch release path plus JobManager and TaskManager failover in the stream phase, verifying no snapshot or stream data is lost through the release and restore

## Documentation

The Chinese doc entry mirrors the English and is a best-effort translation. Please correct the wording where needed.


lvyanquan (approved):
LGTM.

lvyanquan:
Merged. Thanks @spoorthibasu for this series of high-quality contributions.

